PR #262, opened 2020-03-27
WIP: Resolve "Spec pegged orders"

### LINKED ISSUE #216: Spec pegged orders
In GitLab by @barnabee on Feb 10, 2020, 15:39

Write a spec pegged orders

ashleyvega:
In GitLab by @barnabee on Feb 10, 2020, 15:39

Closes #216


ashleyvega:
In GitLab by @barnabee on Feb 10, 2020, 16:34

added 1 commit

<ul><li>4e17c58a - WIP first draft marker maker and pegged orders</li></ul>

[Compare with previous version](/vega-protocol/product/-/merge_requests/47/diffs?diff_id=74813012&start_sha=79188028b06bb19c8dca91cb672d40a94e6d5cf4)

tamlyn10:
Pete to let Barney know if it would be preferable for Barney to make further progress on this before handing over.

peterbarrow:
Questions still needing to be answered:
1. Do we want hidden orders?
2. Do we want to publish all order updates?
3. Do we want to include pegged orders in market depth/best-bid/best-ask?
4. Are FOK+IOC orders OK?

@tamlyn10 @davidsiska-vega Do you have any thoughts please?

barnabee:
Been chatting a bit to @peterbarrow on the "hidden" order thing (1 & 2) - obviously we can't actually hide orders but deciding what the core emits and what the APIs expose around pegs is interesting. I think he's going to propose some ideas but going to need a wider discussion to choose as it will be a question of trade-offs... (doesn't affect how the protocol works but how the APIs do and what someone sees through them and in Console)

barnabee:
> 4. Are FOK+IOC orders OK?

Unless @tamlyn10 disagrees I propose yes, we allow non-persistent orders to be pegged. Given latency it makes sense someone might want to do an IOC at an offset from the BBO/mid IMO.

0000-pegged-orders.md
```diff
@@ -0,0 +1,98 @@
+Feature name: Pegged and Market Maker Orders
+Start date: 2020-02-10
+Specification PR: https://github.com/vegaprotocol/product/pull/262
+
+# Pegged and Market Maker orders
+
+## Acceptance Criteria
+* Pegged orders can only be LIMIT orders, all other types are rejected
+* Pegged orders can only be GTT, GTC, IOC and FOK orders
+* A market must specifically support pegged orders
```
campbellssource: Clarification question: Does this mean that when proposing a market I'd have to say I want it to accept pegged orders? Or that there will be some types of markets that for some set of rules shouldn't be allowed to accept pegged orders?
If the former it might be worth updating market framework spec?

peterbarrow: From my understanding I think your first option is right, when creating a market you must specify that it can handle pegged orders during continuous trading. Once @barnabee confirms I will update the market framework spec.

campbellssource:
@peterbarrow fyi I sketch out some thoughts for what an order ticket might look like in console. https://github.com/vegaprotocol/console/issues/4035 
As I read the spec there is a notion of a group or orders, in the mock ups there is an option for each order in the group to share parameters: size, Reference, and TIF, or set them separately for each order. I'm not sure what will/should be should be supported.

campbellssource:
More considerations of pegged orders in console fyi: https://github.com/vegaprotocol/console/issues/4044

peterbarrow:
We had a meeting to discuss what events would be emitted from the core when handling pegged orders and how it would happen. These probably apply to the market depth spec more than pegged orders. We produced the following points:

Market Depth Data
* Market Depth should include pegged order volume
* We should separate the volume at each level into normal, pegged and market making order volumes to allow them to be drawn with different attributes in the console
* We should generate a best bid/ask pair for all orders and for all order excluding pegged.

Order Stream
* We will not distribute price level updates as they might be too frequent and overload the handlers

witgaw:
@peterbarrow (cc @barnabee) just pushed some minor edits, hope you don't mind - please let me know if you disagree with any on them.

The spec looks really good to me, just a few comments/questions from me:

1. I'd suggest removing the bit about Market Maker orders from this spec and moving it to a separate one. I don't think it contains enough detail for those orders and it's already fairly complex for pegged orders themselves thus I think it would make sense to split the two.

2. These things seem somewhat at odds with each other to me:

> Pegged orders can be GTC, GTC, IOC or FOK TIF orders. 

and 

> Pegged orders are restricted in what values can be used when they are created, these can be defined by a list of rules each order must abide with.
> 
> Buy orders must be pegged against best bid or mid and the offset must be negative (<0).
> Sell order must be pegged against best ask or mid and the offset must be positive (>0).

My understanding is that we never resolve orders off the book (we process them sequentially and an order either matches with another order(s) already on the book resulting in a trade or it gets added to the book). If that's the case and the restrictions on offset signs for buy/sell orders mentioned above apply then I don't see the IOC or FOK pegged orders ever having any effect.

Do you agree with the above interpretation or am I missing something? If so which of those assumption do we want to change? In my naive view a buy order with a positive offset (and sell order with a negative one) would still make sense if I'm interested in being a very active market maker that always posts the most competitive prices. However, there may be some practical reasons against that - please let me know if that's the case.

3. `PeggedOrder{
    PeggedType type
    OrderID    orderID
}`

Would `PeggedType` contain the `Offset` field or should it be added to the `PeggedOrder`?

I'm happy to make those modifications to the spec myself as long as you agree.

0000-pegged-orders.md
barnabee: Initially all markets should accept pegged orders — they are _required_ for market makers anyway.

barnabee:
@witgaw IOC and FOK pegged orders should be accepted IMO (though see the last part of this comment, too) but the below needs a slight tweak I think:

> Buy orders must be pegged against best bid or mid and the offset must be negative (<0).
> Sell order must be pegged against best ask or mid and the offset must be positive (>0).

Persistent pegged orders must not execute on entry or on being repriced because then you can have pegs interacting (trading) with other pegs, which means that as market maker orders are refreshed automatically they could get into a loop. We could do this by checking on each entry or reprice if they'd execute and cancelling them (like a true post only order) but that means that we couldn't guarnatee a market makers orders would always be valid. By restricting the peg offsets such that they cannot possibly trade on entry or reprice, this problem is avoided. (Imagine also the scenario of a low enough spread that the MM's orders are constantly price takers — they are then also supplying no useful liquidity under our model.) Similarly, a non-persistent pegged order must have a non-zero chance of trading on entry, otherwise it's pointless,

I think the above leads to the following set of rules (but someone please check!):

Type | Side | Bid Peg | Mid Peg | Offer Peg
------ | ---- | -------- | -------- | --------
Persistent (GTC, GTT, etc.) | Buy | `<= 0` | `< 0` | Not alllowed
Persistent (GTC, GTT, etc.) | Sell | `>= 0` | `> 0` | Not alllowed
Non persistent (IOC, FOK, etc.) | Buy | > 0 | > 0 | >= 0
Non persistent (IOC, FOK, etc.) | Sell | `<= 0` | `< 0` | `< 0`

> In my naive view a buy order with a positive offset (and sell order with a negative one) would still make sense if I'm interested in being a very active market maker that always posts the most competitive prices. 

Whilst that may be true, it causes problems as described above. One away from mid is pretty competitive — if it happens on both sides it'd lead to a spread of 2. 

In fact there is a futher question here of **what to do if mid is at a 1/2 tick size interval**. A suggestion is that when pegging, the mid is adjusted up or down depending on the side of the pegged order, i.e.

Best bid = 100, best offer = 105 => **mid = 102.5**
A buy pegged to `Mid - 1` should take the mid as 103, and thus be at `102`
A sell pegged to `Mid + 1` should take the mid as 102, and thus be at `103`

This looks good for peristent pegs, but I am not as sure for IOC/FOK.

To be honest it'd be OK to drop non-peristent (IOC/FOK) pegged orders from the first version of this spec. if it's make things simpler…? I think they're desirable because of our block latency, but maybe waiitng for someone to ask would save us some hassle? 🤷‍♂️

jeremyletang:
@peterbarrow @witgaw @barnabee what are the action to be done to get this spec merged now?

barnabee:
> @peterbarrow @witgaw @barnabee what are the action to be done to get this spec merged now?

I'm unsure. I made a comment above, can anyone tell me if either there are further questions that need answering or if it's updated and I need to review?

tamlyn10:
@peterbarrow to take the lead on finishing this.

peterbarrow:
I have integrated all the updates from above and I am happy for this to be merged if others are.

0000-pegged-orders.md
```diff
@@ -0,0 +1,107 @@
...
+
+When a party submits a new pegged order, only a LIMIT order is accepted. The party also specifies the reference price to which the order will be priced along with an offset to apply to this price. The reference price is looked up from the live market and the final price is calculated and used to insert the new order. If the price would result in a hit, the order is executed in the same way as a normal LIMIT order. If the price does not hit, the order is placed on the book at the back of the calculated price level.
+
+Whenever the reference price changes all the pegged orders that rely on it need to be repriced. We run through a time sorted list of all the pegged orders and remove each order from the book, recalculate it's price and then reinsert it into the orderbook at the back of the price queue. Following a price move margin checks take place on the positions of the parties. If a p[egged order is to be inserted at a price level that does not currently exist, that price level is created. Likewise if a pegged order is the only order at a price level and it removed, the price level is removed as well.
+
```
tamlyn10, line 35: >Following a price move margin checks take place on the positions of the parties. 

Is this a generic statement or specific to this spec? In future, margins may be calculated under different triggers than a price move, trade etc.

0000-pegged-orders.md
```diff
@@ -0,0 +1,107 @@
...
+Pegged orders can be GTC or GTC TIF orders with IOC and FOK being added in the second phase of pegged orders. This means they might never land on the book or they can hit the book and be cancelled at any time and in the case of GTT they can expire and be removed from the book in the same way that normal GTT orders can.
+
+If the reference point moves to such a value that it would create an invalid order once the offset was applied, the pegged order is parked. As the reference price moves, any orders on the parked list will be evaluated to see if they can come back into the order book.
+
+When a pegged order is removed from the book due to cancelling, expiring or filling. The order details is removed from the pegged/parked orders list.
+
```
tamlyn10, line 46: minor thing - is this meant to be one sentence?

0000-pegged-orders.md
```diff
@@ -0,0 +1,107 @@
...
+Pegged orders are limit orders where the price is specified of the form `REFERENCE +/- OFFSET`, therefore 'pegged' is a _price type_, and can be used for any limit order that is valid during continuous trading. A pegged order's price is calculated from the value of the reference price on entry to the order book. Pegged orders that are persistent will be repriced, losing time priority, _after processing any event_ which causes the `REFERENCE` price to change. Pegged orders are not permitted in some trading period types, most notably auctions, and pegged orders that are on the book at the start of such a period will be parked (moved to a separate off-book area) in time priority until they are cancelled or expire, or the market enters a period that allows pegs, in which case they are re-priced and added back to the order book. Pegged orders entered during a period that does not accept them will be added to the parked area. Pegged orders submitted to a market with a main trading mode that does not support pegged orders will be rejected.
+
```
tamlyn10, line 25: We just wrote this into the market making mechanics spec...

>As pegged orders are parked during an auction are parked and not placed on the book, margin checks will not occur. This includes checking the orders margin when checking the validity of the transaction.

>Engineering notes:
>- check that other pegged orders are treated the same

0000-pegged-orders.md
peterbarrow: This is a generic comment as all orders entering the book must have their margin level checked before it will be allowed to be inserted. We currently do that with amends in that as we are removing the order from the book and re-inserting, we perform a margin check for the new amended order details. Do you think this is incorrect going forward?

0000-pegged-orders.md
peterbarrow: Thanks, fixed.

0000-pegged-orders.md
peterbarrow: I added a section to the spec to cover this, thanks Tamlyn.

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+When a pegged order is removed from the book due to cancelling, expiring or filling, the order details are removed from the pegged/parked orders list.
+
+# Reference-level explanation
+
+Pegged orders are restricted in what values can be used when they are created, these can be defined by a list of rules each order must abide with.
+
+| Type	                          | Side  | Bid Peg | Mid Peg |  Offer Peg  |
+|---------------------------------|-------|---------|---------|-------------|
+| Persistent (GTC, GTT, etc.)	  | Buy	  | <= 0    | < 0     | Not allowed |
+| Persistent (GTC, GTT, etc.)	  | Sell  | >= 0    | > 0     | Not allowed |
```
barnabee: For persistent sells the bid peg is not allowed and you may peg to mid with > 0 or the best offer with >= 0

barnabee (changes requested):

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+
+Pegged orders are limit orders where the price is specified of the form `REFERENCE +/- OFFSET`, therefore 'pegged' is a _price type_, and can be used for any limit order that is valid during continuous trading. A pegged order's price is calculated from the value of the reference price on entry to the order book. Pegged orders that are persistent will be repriced, losing time priority, _after processing any event_ which causes the `REFERENCE` price to change. Pegged orders are not permitted in some trading period types, most notably auctions, and pegged orders that are on the book at the start of such a period will be parked (moved to a separate off-book area) in time priority until they are cancelled or expire, or the market enters a period that allows pegs, in which case they are re-priced and added back to the order book. Pegged orders entered during a period that does not accept them will be added to the parked area. Pegged orders submitted to a market with a main trading mode that does not support pegged orders will be rejected.
+
```
tamlyn10, line 25: When pegged orders are returned to the continuous trading limit order book after auction, I'm assuming they have lost time priority relative to the uncrossed limit orders that migrate back to a limit book after auction? If so, could we write here just for abundant clarity?

barnabee: I don’t understand this comment, but may be missing something? Limit orders (that are not parked and don’t expire) stay on the book through the auction and into continuous trading. No migration of orders occurs going into or out of auction or continuous trading, they are both trading modes on the same order book. We may want to note that when pegs are unparked they don’t receive special treatment and are added to the back of the price level (lowest time priority).

**Some descriptions generally about peg pricing and time priority that may be useful to check the spec against and add in if anything is missing:**

Pegged orders being added back into the book after being parked (either due to an auction or their reference price not being available) are treated the same as any other order being added to the book and therefore come after all persistent orders at that price level that are already on the book (i.e. the pegs are added to the end/back of the price level as if new incoming orders). This is also true if a pegged order is re-priced _and its price changes_, as price amendments are equivalent to a cancel/replace so the order enters at the back of its new price level. 

When there are multiple pegged orders needing reprice, they must be repriced in order of entry (note: certain types of amend are considered amend in place and others as cancel/replace, for the cancel replace type, the entry time becomes the time of the amend, i.e. a pegged order loses its reprice ordering priority). Generally the way I’ve seen this is to maintain an ordered list of pegged orders for use in re-pricing (and parking/unparking), with new pegged orders added to the end of this list and cancel/replace amends causing the order to be removed from the list and re-added at the end.

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
+Feature name: Pegged and Market Maker Orders
+Start date: 2020-02-10
+Specification PR: https://github.com/vegaprotocol/product/pull/262
+
+# Pegged and Market Maker orders
+
+## Acceptance Criteria
+- [ ] Pegged orders can only be LIMIT orders, all other types are rejected.
+- [ ] Pegged orders can only be GTT and GTC orders. IOC and FOK will be added in the second phase of pegged orders.
+- [ ] A markets ability to handle pegged orders is set at creation time and should be turned on by default.
```
barnabee: Currently all markets must support pegged orders as they are required for liquidity provision (market making). This could be revisited if/when we have other liquidity sources.

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+- [ ] Pegged orders are removed from the order book when going into an auction and are parked.
+- [ ] Parked orders are returned to the order book once continuous trading is resumed.
```
barnabee: "returned to the order book in entry order"

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+- [ ] Pegged orders are repriced when their reference price moves.
```
barnabee, line 13: Add: Pegged orders are not repriced and do not lose time priority when their specific reference price is unchanged, even if other peg reference prices move.

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+- [ ] Cancelling a pegged order removes it from the book and the pegged/parked slice.
+- [ ] An expired pegged order is removed from the book and the pegged/parked slice.
```
barnabee, line 15: Add case for removal when they are filled?

0000-pegged-orders.md
barnabee: After re-reading the spec we say "parked in time priority". I do think this should be tightened up with something from the above because time priority is not defined at book level, only per price level (it can't necessarily be ascertained by looking at creation or update timestamp, for example after an amend), so we should be clear about how the ordering of peg reprices happens across the whole book.

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Guide-level explanation
+
+**Reference Price:** This is the price against which the final order priced is calculated. Possible options are best bid/ask and mid price.
+
+**Offset:** This is a value added to the reference price. It can be negative.
```
barnabee: Add that this must be a multiple of the tick size.

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+
+When a party submits a new pegged order, only a LIMIT order is accepted. The party also specifies the reference price to which the order will be priced along with an offset to apply to this price. The reference price is looked up from the live market and the final price is calculated and used to insert the new order. If the price would result in a hit, the order is executed in the same way as a normal LIMIT order. If the price does not hit, the order is placed on the book at the back of the calculated price level.
```
barnabee: Note that the allowable peg reference + offset combinations should prevent a persistent (GTT/GTC) peg from hitting another order on reprice so this should not need to be checked for pegged limit orders.

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+
+Whenever the reference price changes all the pegged orders that rely on it need to be repriced. We run through a time sorted list of all the pegged orders and remove each order from the book, recalculate it's price and then reinsert it into the orderbook at the back of the price queue. Following a price move margin checks take place on the positions of the parties. If a pegged order is to be inserted at a price level that does not currently exist, that price level is created. Likewise if a pegged order is the only order at a price level and it is removed, the price level is removed as well.
```
barnabee: When running through the sorted list of pegs we need to ensure we do not remove and replace any pegged orders that are pegged to an unmoved reference, so that they don't lost time priority. 

The sorted list of pegged orders should be sorted by entry time (i.e. when they were submitted OR last underwent a cancel/replace amend).

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+
+Pegged orders can be GTC or GTC TIF orders with IOC and FOK being added in the second phase of pegged orders. This means they might never land on the book or they can hit the book and be cancelled at any time and in the case of GTT they can expire and be removed from the book in the same way that normal GTT orders can.
+
+If the reference point moves to such a value that it would create an invalid order once the offset was applied, the pegged order is parked. As the reference price moves, any orders on the parked list will be evaluated to see if they can come back into the order book.
```
barnabee: I guess this covers if the reference value does not exist (i.e. there is no best bid)?

0000-pegged-orders.md
```diff
@@ -0,0 +1,109 @@
...
+| Non persistent (IOC, FOK, etc.) |	Buy   | > 0     | > 0     | >= 0        |
+| Non persistent (IOC, FOK, etc.) |	Sell  | <= 0    | < 0	  | < 0         |
+
+As the calculation of mid price can result in an invalid result due to precision, the value will be rounded depending on which side of the book is being handled.
+
+For persistent pegged orders:
+
+Best bid = 100, best offer = 105 => mid = **102.5**<br>
+A buy pegged to Mid - 1 should take the mid as 103, and thus be at 102<br>
+A sell pegged to Mid + 1 should take the mid as 102, and thus be at 103
```
barnabee, line 64: I think we need acceptance criteria for this nuanced case

barnabee (changes requested):
This is pretty close I think! Few quite small comments/additions and clarifications but nothing major

0000-pegged-orders.md
peterbarrow: Updated, thanks.

0000-pegged-orders.md
peterbarrow: I liked your description above and have added it to the spec, thanks.

0000-pegged-orders.md
peterbarrow: Updated, thanks.

0000-pegged-orders.md
peterbarrow: Updated, thanks.

0000-pegged-orders.md
peterbarrow: Added.

0000-pegged-orders.md
peterbarrow: Yes, added.

0000-pegged-orders.md
peterbarrow: Done, thanks,

0000-pegged-orders.md
peterbarrow: Removed old details.

0000-pegged-orders.md
peterbarrow: Added no best bid example.

0000-pegged-orders.md
peterbarrow: Done

0000-pegged-orders.md
peterbarrow: Added and for tests too.

barnabee (approved):
Looks great! Thanks @peterbarrow

PR #267, opened 2020-03-27
assets proposal

ashleyvega:
In GitLab by @jeremyletang on Mar 26, 2020, 17:40

### DIFF
--- a/specs/0027-asset-proposal.md
+++ b/specs/0027-asset-proposal.md
@@ -1,0 +1,90 @@
+Feature name: assets proposals
+Start date: 2020-03-25
+
+# Summary
+This spec cover the creation, discoverability of a settlement asset in the vega network.
+
+All markets require a settlement asset, this settlement asset will be hosted in a blockchain and all informations about them will be relayed from a vega bridge on the given asset blockchain.
+This spec cover how a vega user can propose a new asset, what's the mechanism to make this asset available inside vega, and usable by any market.
+As of now, we will only cover implementation of ERC-20 tokens, following will specs will cover implementation of other asset (first ETH, then BTC, etc, etc...), so implementation needs to keep in mind these details.
+
+# Reference-level explanation
+
+## Proposing a new asset
+
+The proposal of a new asset must be done through the governance system.
+This mean the introduction of a new proposal type in order to propose a new asset to be added to the network.
+This proposal could be done by anyone with stake in vega token, altho we expect market makers to do it.
+
+The proposal vote for an asset is done in two step, first by the validator nodes firsts, then by the token holders.
+
+First, When a new asset is proposed to vega, the asset validity is verified with their origin blockchain, this allow the vega network to recover information about the asset class (e.g: symbol, name, decimal place, etc, etc).
+If the asset is accepted by the node, the node will then send his own vote as a transaction to the chain, so the other validators can keep track of whom is accepting the new asset.
+This first phase may be configured through network parameters (e.g: duration of the phase, how much validator needs to succeed to validate the asset).
+In a first version we could hard code these value (e.g: 1 hours duration for the node to validate the asset, only 2+1/3 of the node needs to succeed.
+
+Once this first step is done, if enough nodes were able to validate the asset, the network will proceed with accepting token holders votes, but if not enough node were able to validate the asset, then the new asset is rejected.
+
+The second part of the vote is the normal governance flow.
+
+## Enabling a new asset on the bridge
+
+Once the proposal is accepted, the original submitter of the proposal, can reach out to the nodes, and get confirmation (as a signature, e.g: of the contract address of the asset to enable) of the acceptance of the new asset.
+The original submitter must them agregate 2+1/3 signature from the node, and send them to the bridge. The bridge will then whitelist this new asset, and start forwarding event with it.
+(this is the mutli signature scheme, which allow us to validate decisions in between the bridges and the vega network and is discribed in the bridge specs).
+
+## Enabling a new asset on vega
+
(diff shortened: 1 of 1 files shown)


ashleyvega:
In GitLab by @jeremyletang on Mar 26, 2020, 18:05

changed title from **{-fist shot at -}assets proposal** to **assets proposal**

Vegaklaus:
What is the format of the proposal ? Is there a proposal ID, or is every proposal its own ID ? In the former case, is it assured every validator  sees the same proposal ?

General: why is 'node' often used in singular ? Is it a typo or does that mean something specific ?


Line 19: Probably not a big issue, but can we be spammed with proposals ? Should verifiers only allow, say, 1 proposal per tokenholder per day ?

Line 21: Maybe more specify: "the validators verify" rather than "is verified". Is this functionality of the bridge to allow this verification or
coded elsewhere ?
Can the asset be invalidated in mid-vote, e.g., by spending the bitcoin after the first validators verified it ?

Line 22: Just to be sure, this means there's no no vote, just yes and abstain and wait for timeout ?

Line 24: One hour in blocktime ? Different validators might see it at different times, so one hour is not a clearly defined term. Maybe "x blocks after the first vote for that proposal on the blockchain" 
	
2+1/3 can be 2/3 rounded down + 1 (I suppose it means traitors+1, where the number of parties is 3*traitors+1)

Line 28: Is there a link to that ?
Line 32: More a question for the governance flow - the multisig is issues by the validators who confirm that the governance vote was yes ?

As all votes require only 1/3+2, it is possible to have two votes on the same asset that would succeed on independend subsets of the validators. 
Can that cause an issue (the same asset being used in two markets, the same asset being accepted twice for the same market,...) ?

jeremyletang:
> What is the format of the proposal ? Is there a proposal ID, or is every proposal its own ID ? In the former case, is it assured every validator sees the same proposal ?

The proposal get an ID assigned and sequence just after going through consensus (as we are doing with he orders).

> General: why is 'node' often used in singular ? Is it a typo or does that mean something specific ?

A typo.

> Line 19: Probably not a big issue, but can we be spammed with proposals ? Should verifiers only allow, say, 1 proposal per tokenholder per day ?

Yes any rule required to reduce / stop spam can be implemented, but as the other parts of the system were spam is possible I just did not spend any though on that so far, and thought it could be done as part of some future work.

> Line 21: Maybe more specify: "the validators verify" rather than "is verified". Is this functionality of the bridge to allow this verification or
> coded elsewhere ?
> Can the asset be invalidated in mid-vote, e.g., by spending the bitcoin after the first validators verified it ?

We do not verify and asset as been deposit in here, but we verify the asset exist on the chain (meaning in case of an erc 20 token, that the contract for the token exisrts, and we can get information about it, and can use it).

> Line 22: Just to be sure, this means there's no no vote, just yes and abstain and wait for timeout ?

yes so far. but this can be changed if you require to.

> Line 24: One hour in blocktime ? Different validators might see it at different times, so one hour is not a clearly defined term. Maybe "x blocks after the first vote for that proposal on the blockchain"

Yes this is in blocktime speaking. it could be in term of block or anything, the existing proposal already use time not blocks for acceptance, so I reused that.

> 2+1/3 can be 2/3 rounded down + 1 (I suppose it means traitors+1, where the number of parties is 3*traitors+1)
> 
> Line 28: Is there a link to that ?

The specs as already been merged yes: https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md

> Line 32: More a question for the governance flow - the multisig is issues by the validators who confirm that the governance vote was yes ?

Well each node will issue a signature once the network governance vote is yes, then this can be agregated to be sent to the mutli sig verification on the bridge.

> As all votes require only 1/3+2, it is possible to have two votes on the same asset that would succeed on independend subsets of the validators.
> Can that cause an issue (the same asset being used in two markets, the same asset being accepted twice for the same market,...) ?

I'm not sure about that, I would imagine ideally at least 2+1 / 3 of the validator needs to accept the asset for the governance to work. to start with all nodes will need to validated it successfully.
A market can so far have 1 asset,  the market itself cannot accept twice something, the network should probably not  accept twice the same asset

edd:
Transaction spam thoughts go here: https://github.com/vegaprotocol/product/issues/217

Vegaklaus:
The lack of no-votes is not an issue, though the current voting allows two votes to succeed on independent sets of validators at the same time (each getting their own set of 2+t/3 voters). It's not clear to me if that is a problem (if the same asset is in two markets or two assets voted into one market), and how the governance flow would handle this.

One more question, the initial proposal is also submitted as a blockchain transaction ?

edd:
@EVODelavega can probably answer this best (cc @jeremyletang)

jeremyletang:
> The lack of no-votes is not an issue, though the current voting allows two votes to succeed on independent sets of validators at the same time (each getting their own set of 2+t/3 voters). It's not clear to me if that is a problem (if the same asset is in two markets or two assets voted into one market), and how the governance flow would handle this.

I'm not sure I understand this, so far we'll require all node to vote on favor of the asset, if even one node cannot validate the asset, and then not voting for it, the asset proposal will be stopped, and won't fo further (through the rest of the governance).

So far an asset could be use by multiple market for sure, also the definition of a market (market framework) do not allow two assets.

> One more question, the initial proposal is also submitted as a blockchain transaction ?

Yes, the proposal for a new asset is first submitted to the chain, so it get gossiped through the network, once the node have seen the transaction they then validate the asset exist on the foreign chainm and vote to accept it through the vega network.

tamlyn10:
@C0deMunk33 have added you to this spec PR as an owner..

tamlyn10:
What is status update on this spec, Jeremy, Danny?

tamlyn10:
Implemented already. @jeremyletang to review the spec and merge etc.

PR #271, opened 2020-04-09
Create Event Queue Spec

C0deMunk33:


C0deMunk33:
Now open for comment

specs/0000-event-queue.md
```diff
@@ -0,0 +1,94 @@
...
+# Acceptance Criteria
+## Event Queue
+* Event Queue calls `Get_Subscribed_Event_Sources` and successfully parses the response
+* Event Queue gathers and filters applicable events from Ethereum node
+* Event Queue calls `Event_Acceptance_Status` and marks accepted events internally as complete
+* Event Queue propagates unaccepted events to Vega node via `Propagate_Event` 
+
+## Vega Nodes
+* Vega nodes responds to `Get_Subscribed_Event_Sources` with a list of valid smart contract events
+* Vega nodes accept events submitted to `Propagate_Event` and verifies them against valid Ethereum node
```
edd: If a transaction is not verified (I.e. the data received via the event does not match the data received when looking up the transaction) we should explicitly error. `Event_Acceptance_Status` only provides a Boolean. Perhaps it could be an enum and contain an entry for events rejected for failing verification? I am not married to that solution, so anything that properly lets us represent that status is fine by me.

edd (changes requested):
One change in the acceptance criteria, which is just around clarity. It might also be good to add some details on how what verification with Ethereum looks like. I assume it is:

- Talk to a local Ethereum API
- Talk to a remotely hosted Ethereum API
- Fetch transaction ID
- Check that balances and accounts involved match the event

What would cause it to fail this check? That’s what we need to capture. Otherwise this looks good to me.

specs/0000-event-queue.md
```diff
@@ -0,0 +1,94 @@
...
+# Summary
+Latency and throughput is of paramount importance to Vega validator nodes, so monitoring various 3rd party (and notably low transaction speed) blockchains and crypto-assets isn’t possible. 
+To counter this problem we have created an event queue management system. 
+This event queue allows the standardization and propagation of transactions from 3rd party asset chains into the Vega event message format. 
+
+# Guide-level explanation
+Events and transactions are gathered and processed by the Event Queue from a given 3rd party blockchain, and only the ones subscribed to by Vega nodes will be propagated through consensus and then validated by the validator nodes. 
+This makes the event queue works as a buffer between the slow/complicated world of various blockchains, and the high throughput, low latency of Vega Core.
+This message queue will use grpc to communicate with the Vega network via 3 main functions: `Get_Subscribed_Event_Sources`, `Propagate_Event`, and `Event_Acceptance_Status`.
+The first, `Get_Subscribed_Event_Sources`, returns a list of smart contract addresses that consensus has deemed valid as sources of events
```
barnabee: Do we also need to be able to specify the specific event that will be subscribed to? (I think this will likely be useful for Ethereum based data sources aka oracles, for instance)

specs/0000-event-queue.md
```diff
@@ -0,0 +1,94 @@
...
+    rpc Get_Subscribed_Event_Sources() returns (Get_Subscribed_Event_Sources_Response);
+    rpc Propagate_Event(Event_Propagation_Request) returns (Event_Propagation_Response);
+    rpc Event_Acceptance_Status(Event_Acceptance_Status_Request) returns (Event_Acceptance_Status_Response);    
+}
+
+message Get_Subscribed_Event_Sources_Response {
+    repeated string subscirbed_event_source=1; 
+}
+
+enum Event_Types {
```
barnabee: How will this scale when we add events for use as general data sourcing and other uses (control Vega accounts/position from Eth smart contract for instance)? Is this something we need to think about now?

specs/0000-event-queue.md
```diff
@@ -0,0 +1,94 @@
...
+    Deposit_Minimum_Set=5;
+}
+
+message Event_Propagation_Request {
+        string event_source = 1; //address of bridge
+        string asset_source = 2; //address of asset
+        string asset_id = 3; //ID of asset specific to that asset_source
+        Event_Types event_type = 4; // enumerated event type
+        string source_party_id = 5; // source ethereum address 20 bytes hex preceded by 0x or other party ID of user/contract/system this pertains to
+        string target_party_id = 6; // provided Vega public key of party to target the event to
```
barnabee: This seems like it would be part of the event detail rather than the propagation request? (Some events in future won't have a target party)

specs/0000-event-queue.md
```diff
@@ -0,0 +1,94 @@
...
+enum Event_Types {
+    UNKNOWN = 0;
+    Asset_Deposited=1;
+    Asset_Withdrawn=2;
+    Asset_Listed=3;
+    Asset_Delisted=4;
+    Deposit_Minimum_Set=5;
+}
+
+message Event_Propagation_Request {
```
barnabee: Where is the event content sent?

specs/0000-event-queue.md
```diff
@@ -0,0 +1,94 @@
...
+To counter this problem we have created an event queue management system. 
+This event queue allows the standardization and propagation of transactions from 3rd party asset chains into the Vega event message format. 
+
+# Guide-level explanation
+Events and transactions are gathered and processed by the Event Queue from a given 3rd party blockchain, and only the ones subscribed to by Vega nodes will be propagated through consensus and then validated by the validator nodes. 
+This makes the event queue works as a buffer between the slow/complicated world of various blockchains, and the high throughput, low latency of Vega Core.
+This message queue will use grpc to communicate with the Vega network via 3 main functions: `Get_Subscribed_Event_Sources`, `Propagate_Event`, and `Event_Acceptance_Status`.
+The first, `Get_Subscribed_Event_Sources`, returns a list of smart contract addresses that consensus has deemed valid as sources of events   
+The second function, `Propagate_Event`, allows an event queue to send events raised on 3rd party blockchains (deposits, withdrawals, etc) through Vega consensus to ensure an event has been seen by the network.
```
barnabee: Should clarify that Vega transactions and APIs must support:
- multiple blockchains as sources
- multiple source on any given blockchain (i.e. if we had two bridge contracts on the same Ethereum network for different asset types or multiple contracts being used as data sources/oracles)

ashleyvega (commented):
I've made some cosmetic changes, and have no further comments on this high-level spec, though I'd like to see a tech spec with some implementation details at some stage.

barnabee (changes requested):
Looking for an update on commetns I requested

specs/0000-event-queue.md
C0deMunk33: indeed. I've updated this to say:
"returns a list of smart contract addresses and events that consensus has deemed as a valid source."

and the Vega node sim spits out the events as well as the contract addresses.

specs/0000-event-queue.md
C0deMunk33: It's not. The node verifies the contents from the source. this request simply lets the node know that there is an event it should care about. Since there's no point in trusting the event queue for info like that, there's no real reason to send it along.

specs/0000-event-queue.md
C0deMunk33: indeed. and something that would be verified by the node... I will remove this

specs/0000-event-queue.md
C0deMunk33: I see this enum as an expandable list and, as each of these Event_Types are individually parsed, adding a new event type would be pretty straight forward.

specs/0000-event-queue.md
C0deMunk33: noted and updated

specs/0000-event-queue.md
C0deMunk33: I agree. Having an explicit "rejected" status makes sense

C0deMunk33: Added enum for this

C0deMunk33 (commented):
updated doc based on comments:
https://github.com/vegaprotocol/product/commit/e01a8759eeb38fecd8c55e0c77d96e3ca92cd75c

C0deMunk33:
> What would cause it to fail this check? That’s what we need to capture. Otherwise this looks good to me.

The only reason it would fail is if the event queue gives bad information to the node, but the node must verify all the info before accepting the event. I've added these verifications to the list.

edd:
@barnabee / @ashleyvega as previous reviewers, the ball is back in your court!

ashleyvega (approved):
LGTM

specs/0000-event-queue.md
barnabee: I guess I mean the address of a contract and name of the event may be added in a Vega transaction, i.e. 

```
ProposeMarket { 
    settlementEvent = EthereumEvent {
       contract: 0xblahblahblah
       event: "MyCustomEvent"
       path: "data.value"
    } 
}
```

Where that could even be referring to the Vega bridge contract and events it produces, or any other contract/event on the Ethereum chain. I was wondering what the event type/data model might look like in this case, and if the current design works to be expanded that way.

specs/0000-event-queue.md
barnabee: Again thinking about the oracle scenario really. Where the data payload from the tx/event is used for a calculation in Vega, how do you see that info getting into the core?

I have assumed this is via consensus in the same way it's agreed that the event occurred, but am wondering how that data would make it physically into the transaction, had asssumed that might be this event queue?

barnabee (commented):
@C0deMunk33 I've left a couple of questions on your responses

edd:
@C0deMunk33 this is currently waiting on comments from you. I'll rebase it so it's mergeable again

tamlyn10:
@C0deMunk33 can you give a status update on this spec writing task?

specs/0000-event-queue.md
C0deMunk33: > @C0deMunk33 can you give a status update on this spec writing task?

I need to add the Oracle and market action Triggers thing, but for now the oracle spec takes precedent. @tamlyn10  This is effectivly on hold while some of the deeper/tertiary discussions happen.

jeremyletang:
Updated the code samples to show an example of how would the ERC20 events would work.
Rest of the spec looks good to me.

jeremyletang (approved):
@C0deMunk33 looks good, feel free to merge if you are done with this.

specs/0000-event-queue.md
C0deMunk33: > Where the data payload from the tx/event is used for a calculation in Vega, how do you see that info getting into the core?

This is what @jeremyletang and I have worked out:

The event is picked up by the Event Queue and the tx_hash, block number, and transaction index are sent to the Vega API.

The Validator node takes those values to validate the exact events that core cares about on their local Ethereum node. 
The final values that end up on-chain come from the validators Ethereum node, not from the Event queue. Once a validator has accepted the event, gossip will cause all the other validators to validate that same event locally.  This allows Core to not have to waste cycles scanning for events as opposed to directly calling for the data and one Event Queue could potentially service more than one validator (maybe).

This general pattern holds for BTC and TBD chains.

tamlyn10:
@C0deMunk33 can we merge this?

PR #274, opened 2020-04-16
Clarified multiple amends and optional fields

peterbarrow:
Updating the spec to be more specific about multiple field updates and fields being optional.


edd (approved):
Nice and clear 👍

PR #275, opened 2020-04-29
Add price monitoring spec

### LINKED ISSUE #220: Price monitoring
In GitLab by @tamlyn10 on Feb 25, 2020, 14:36

witgaw:
Closes #220.


specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+- [ ] `vega` switches trading mode from default based on the price monitoring signal.
+- [ ] Mechanism for switching back to default trading mode is implemented.
+- [ ] Order book is processed atomically so that the orders which directly moved the price beyond allowed band get processed again under the fallback trading mode (and no associated trades are generated in default trading mode).
+
+# Summary
+
+There will be times when prices in our markets move by a large amount over a short period of time. While there's nothing wrong with that per se (it may simply be a result of new information being published and market reacting to that), we need to make sure that we minimise the impact of that on the stability of our markets as measured by their volatility and number of traders getting closed-out by the network. One way of achieving this is to put markets into auction mode during those periods as it has been shown that this can result in a more orderly incorporation of new information into the market price.
```
barnabee: To be clear: if we knew that every trade price represented the whole market's real view of the price, we wouldn't use price monitoring. The goal here is to minimise the chance that we are *erroneously* marking positions to market with a larger price move than is necessary.

The dynamics of market price movements are such that price don't always represent the participant's true average view of the price but are instead artefacts of the market microstructure: sometimes low liquidity and/or a large quantity of order volume can cause the price to diverge from the true market price. It is impossible to tell at any point in time if this has happened or not. 

As a result, we assume that relatively small moves are "real" and that larger moves _might_ not be. Price monitoring exists to determine the real price in the latter case.

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+
+Another reason why this is important is the reliance on risk models for calculation of margin requirements for each trade and the fact that those margins are implied by the model's view of the range of prices that are possible in the near future with some specified probability level. To give an example, say the actual market move over last hour is larger than what our model was impling at the time, then the margins that were charged on positions that are still open may be insufficient to cover those moves for traders that were negatively impacted by them (if the margins haven't been updated over that time). To prevent that we need to be able to detect those moves as they occur and be able to change the trading mode on the fly.
```
barnabee: I'm a little unsure on this.  I think really we are only trying not to end up in a situation where our models close people up on a price "blip" / flash crash. I guess if the model is wrong the auction period may be a wake up call to market makers to vote in updated risk params but not sure about that. Would like @tamlyn10 @davidsiska-vega feedback on this.

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+
+To achieve the above we need to be able to check if processing the latest set of orders in market's default trading mode (e.g. continuous trading) would result in the "sensible" price level as implied by the risk model being breached. If that's not the case we need to be able to roll-back processing of those orders, switch to the fallback trading mode (e.g. auction mode) and process those, and any new, orders in that way.
```
barnabee: >we need to be able to check if processing the latest set of orders

We need to check if processing **the next transaction** — we treat each tx (submit/amend/cancel order or the trigger that causes an auction uncrossing) separately here and evalaute price monitoring before commiting to process each.

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+
+![#f03c15](https://via.placeholder.com/15/f03c15/000000?text=+) `TODO (confirm): Once the market is in the fallback trading mode the mid-price will still be update as trading occurs (e.g. via auctions). Throughout this process the risk model will continue to provide projections and once the mid price falls back into the range projected for a given point in time the market will switch trading mode back to market's default.`
```
barnabee: We should change the lingo: rather than "fallback trading mode" we should say the market enters a "temporary auction period"

We do not wait for the price to return to the projected range to return to normal trading. Once the price monitoring triggers have all been tripped and auctions run their course, if the new price is still crazy high or low then that becomes the real new price of the market. Prices are not constrained to a range by the risk model, we just take  more care to double check the price with our participants in an auction the more violently the price looks like it's moving.

Market data in an auction:
- **mid price:** undefined (bid and offer can be crossed, nonsense value)
- **best bid / best offer (ask):** defined as normal (highest bid, lowest ask) and *can be crossed*
- **last traded price:** whatever the LTP was before the auction was entered
- **indicative uncrossing price:** whatever price would be achievd if the auction uncrossed immediately (can be undefined if no crossed orders)

NB: auction spec is here, I reviewed today and added a bunch of comments: https://github.com/vegaprotocol/product/pull/256

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+
+### Note
+
+Price monitoring likely won't be the only possible trigger for changing the trading mode (liquidity monitoring - spec pending - or governance action could be the other ones). Thus the framework put in place as part of this spec should be flexible enough to easily accommodate other triggers and possibly different fallback modes (defined in the market configuration) per trigger.
```
barnabee: The framework should be able to trigger a *temporary period* with any valid trading mode

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+
+# Guide-level explanation
+
+- We need to emit a "significant price change" event if price move over the horizon τ turned out to be more than what the risk model implied at a probability level α.
+  - Take current mid-price S_t,
```
barnabee: For price monitoring we need to evaluate the arrival price of the next transaction — that is, the value that will be the last traded price if we process the next transaction — not the mid price.

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+  - look-up value S_{t-τ} (prices aren't continuous so will need max(S_s : s  ≤ t-τ), call it  S_{t-τ}^*,
+  - get the bounds associated with S_{t-τ}^* at a probability level α:
+    - if S_t falls within those bounds then order book processing is carried out in the default trading mode,
+    - otherwise the orders (if any) which resulted in the price change to S_t get reverted (see point below) and trading mode gets changed to the fallback one.
```
barnabee: orders -> transaction (there's only one evaluated at a time and it doesn't have to be an order)

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+- We need to have "atomicity" in order processing:
+  - When we process orders at t we need to check what the mark price would've been if we processed all of them (S_t).
```
barnabee: again: one transaction
last traded price not mark price (mark price may in future use external sources etc. which we don't care about)

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+  - If it results in "significant price change" event then we want the order book to maintain the state from before we started processing any orders.
+  - Trading mode is then change to the fallback mode and orders are processed.
```
barnabee: Note: if the transaction is not valid in the new period it will be rejected

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+  - Market carries on in the suspended mode until trigger to switch back to deafult is received.
```
barnabee: A market in an auction isn't suspended, that is a separate thing that we may build for some protection situations (it's in the WP) but I think for now we are not building and just going with auctions everywhere.

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+- In general we might have a list of pairs of α, τ.
+  - In that case we take the most narrow interval implied by the above and compare the S_t against it.
+
+# Reference-level explanation
+
+## View from the [vega](https://github.com/vegaprotocol/vega) side
+
+- at the end of the block:
```
barnabee: We definitely don't want to do *anything* based on beginning/end of block

- we don't know how long a block will be and it isn't guaranteed to be uniform or even within some bounds in all cases
- we want Vega to work in situations where the concept of a block doesn't exist

In any case as above we process each tx separately for market monitoring

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+  - pricing engine sends the risk model<sup>[1](#footnote1)</sup> the **LAST** ([![#f03c15](https://via.placeholder.com/15/f03c15/000000?text=+) `TODO (confirm): is this ok or do we need all/max)` mark price and a flag indicating if it was obtained via auction mode or not [![#f03c15](https://via.placeholder.com/15/f03c15/000000?text=+) `TODO (confirm): Do we care how the price was obtained?`]
```
barnabee: We are interested in the last traded price that would be achieved by the tx not mark price

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+  - risk model sends back the price bounds applicable for the next block
```
barnabee: Price bounds for price monitoring should be based on a sliding window in time, i.e. a maximum move over, say 10 minutes (or whatever is configured for the trigger)

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+- during the block, if processing an order would've resulted in a trade at a price that breaches those bounds the market enters an auction mode (and that order is processed along with the other ones in that mode) [![#f03c15](https://via.placeholder.com/15/f03c15/000000?text=+) `TODO (confirm): what's the finest resolution that we care about - do we want to monitor all price within the block (WG: I think so as getting different outcomes depending how we slice the blocks doesn't seem desirable) or just the final price at the end of the block` ].
```
barnabee: Resolution = per tick, we immediately change mode mid-block if the tx triggers monitoring and through that tx and subsequent ones at the market in its new auction period

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+  - the market snaps out of the auction mode once the market clearing price is back within those bounds (either inside the block or in any of the ones that follow) ![#f03c15](https://via.placeholder.com/15/f03c15/000000?text=+) `TODO (confirm)`
```
barnabee: The auction initiated by trigger _n_ ends after a set time (specified in the trigger) unless there is a trigger _n+1_ that is also activated by the price, in which case the auction is extended by the time specified in that trigger.

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+
+## View from [quant](https://github.com/vegaprotocol/quant) library side<sup>[1](#myfootnote1)</sup>
+
+- we get a mark price (unless we want more than one) from trading-core and a timestamp (accurate to say a few seconds).
+- we can use that to build a time series (of required length) and can generate either (depending on which compontent we think should do the actual policing):
```
barnabee: We should calculate and make available to the market an upper and lower price monitoring trigger level (bound). These will be used by other functionality (i.e. market making) and also need to be exposed in the market data API. 

The time series should be maintained with each tick and I'd suggest the bounds are recalculated every time the Vega time (i.e. the timestamp on the block being processed, in the Tendermint case, but that is abstracted away to just "Vega time") changes a the one second resolution (this will mean recalculating the levels approximately once per block but not always, at the current Testnet speed).

barnabee: cc @davidsiska-vega @tamlyn10

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+  - a signal that tolerance has been breached
+  - a tolerance bound to be checked at a future timestamp
+  - a tolerance bound to apply within the next block
+
+### Notes
+
+- we need a probability density function (p.d.f.) from the risk model and the inverse p.d.f.
+- should generally handle a vector of pairs of horizons and probabilities
```
barnabee: Triplets: `{ horizon, probability, auction [extension] duration }`

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,75 @@
...
+- Implement a cut-off on precision so we don’t return spuriously accurate results
```
barnabee: Precision for the price monitoring trigger levels returned should be the number of decimal places configured in the market framework for the market.

barnabee (dismissed):
Thanks @witgaw, this is good, maths looks sound but a few changes on the mechanics of how it works and wording.

specs/0000-price-monitoring.md
tamlyn10: Agree and suggest @candida-d may wish to use what Barney so articulately expressed to go into our FAQ's / protocol explainers.

specs/0000-price-monitoring.md
tamlyn10: Mmm.. this is a good question.  So, Witold has outlined two reasons..

1. Risk models & parameters are "correct" but there is an erroneous price move (not really representing market), so we slow the market down because we don't want to mtm settle (which can be thought of as a rebalancing of margins) because the "market observable" isn't well known by the network / market.

2.  There has been a step change in volatility (for example), so the price is "real" but the network wants to give time to the market makers or calibration engine update risk params so that the network is collecting enough margin to accommodate this change.

I would say that 2. is referring to something different to what we had in mind, and possibly is something we could consider when thinking about the calibration process.. i.e. if there is a "calibration detection" thingy that detects a need to update parameters and give traders more time to either update their collateral balances or get out of their positions if the cost to hold is too high for them.

tamlyn10: I'd add to the above, that perhaps using the risk model to set the "extreme price move level" has muddied the water on the mechanism's intent. The reason it is used, is that of course we need a view of a price move that is large for a market (and the risk model has that view). But in fact, this could be set as a completely separate market parameter by the market makers. If they set it absurdly high, then the market would consider all price moves as "real" and there would be lots more close outs and then most likely liquidity monitoring would kick into effect. I don't think we would ever let market makers control the liquidity monitoring triggers, but in theory,  the price monitoring triggers wouldn't place the network at risk. Am I thinking about this right -  cc @davidsiska-vega ?

barnabee: Yeah I think 2 would be covered if at all by something we might call "risk monitoring" or "calibration monitoring" — thinking aloud, perhaps the quant model would calculate an implied vol over a sliding window period (or periods) and we'd trigger an auction or something if they diverged **and** the calibration was considered stale (i.e. older than the start of the sliding window). We might be able to get away with increasing the margin reqs if the implied vol diverged in the risky direction from the calibration rather than triggering an auction though.

I agree 2 is not an explicit goal of price monitoring at this time, and any benefit it had in that direction is emergent (and welcome.)

tamlyn10: And one final thing to add on mechanism intent - By way of background,  this came out of thinking about traders bullying the markets and "driving up" the price (for example with one huge buy order) then taking their MTM wins that result (they've bought lots on the way up and the final price against which those buys are MTM'd is high so they get a positive MTM) and then taking that released collateral and deploying it to another market then possibly reneging on the first market's inevitably large requirement from them to pay MTM as the market settles back down to the level it had started at before the bully bought a big order.

specs/0000-price-monitoring.md
tamlyn10: I would go one step further and say that we are actually interested in what is the new _mark price_ (=S_t) and the method for calculating that set by market parameter - see [this spec](https://github.com/vegaprotocol/product/blob/master/specs/0009-mark-price.md)

specs/0000-price-monitoring.md
barnabee: @tamlyn10 yes in theory there is no need to use the risk model, the LSE set a manual move amount the same for all stocks in the same market to trigger PM but then it was quite reasonable to say that a main market listed stock shouldn't move more than X% in a period without triggering PM as they are all in a similar class. Wouldn't work on Vega, and introducing more "magic numbers" in parameters for stakehodlers to learn about and need to evalaute in proposals just adds operational risk to the platform so using the risk model to define excessive per market is a good shortcut. 

Re: mechanism intent, I always expected to include PM in the protocol because of seeing markets at the LSE and others (even serious spot markets have something like this). Certainly I haven't been thinking specifically about the case you mentioned as the driver, although I know we've discussed it and it's good to know you think it will mitigate that, I saw a card on market bullying yesterday and thought it was still an open issue/question how to prevent.

tamlyn10: Liquid spot markets that have price monitoring are a nice to have IMO. 
My example shows a sure way of grabbing money with little to no risk, if there is not price monitoring (or some other protection mechanisms in place).
But, this isn't the first time we've battled this difference in perspective :-)

tamlyn10: If we didn't release the MTM gains, that would also be a protector against my example..

tamlyn10: I suggest we move this... to a research ticket? Some good thinking I'd not like to lose..

> Yeah I think 2 would be covered if at all by something we might call "risk monitoring" or "calibration monitoring" — thinking aloud, perhaps the quant model would calculate an implied vol over a sliding window period (or periods) and we'd trigger an auction or something if they diverged **and** the calibration was considered stale (i.e. older than the start of the sliding window). We might be able to get away with increasing the margin reqs if the implied vol diverged in the risky direction from the calibration rather than triggering an auction though.
> 
> I agree 2 is not an explicit goal of price monitoring at this

specs/0000-price-monitoring.md
tamlyn10: Barney, if [part of] the reason for this mechanism is to protect against erroneous MTM swings I would consider any method of setting the mark price to be valid..

specs/0000-price-monitoring.md
tamlyn10: Does this mean if my order to buy x contracts tips the market into an auction, then if that order is:

* it's a LIMIT order with a persistent TIF, it will be added to the auction book, otherwise it will be rejected.
* it's a MARKET order, it will be rejected.
* it's a PEGGED order, it will be rejected.

cc @candida-d it will be worth making this explicit in the guides.

specs/0000-price-monitoring.md
tamlyn10: "price protection mode"?

specs/0000-price-monitoring.md
tamlyn10: as per my above comment, i'm not sure i agree.

specs/0000-price-monitoring.md
tamlyn10: oh this is interesting,  so to clarify, you are saying the _price protection auction call period_ is specified in the trigger (i.e. not by a network/market/risk parameter). I like this because it means we can make longer call periods for more extreme moves.

specs/0000-price-monitoring.md
tamlyn10: The upper and lower bound price levels that trigger an auction should IMO be updated every time the mark price is updated.

tamlyn10: I say this, because in 1 tick of vega time,  the market could move quite a bit and stale price bounds could be annoying.. also, i would expect that this is cheap and easy to do by the risk model since the bounds themselves only rely on the pdf and a view of the current mark price?

specs/0000-price-monitoring.md
barnabee: I'd say it is not, not directly at least. The design goal I have in mind for price monitoring is to ensure that Vega only trades at *reasonably fair prices* (see below), not directly to think about MTM swings or other specific downstream issues stemming from bad prices (if you can prevent bad prices, you can eradicate a whole class of issues, again, see below). 

Further, mark price here specifically could problematic because (a) it doesn't necessarily change after an auction (in future when there are multiple calculation methods), so you don't know if the auction worked; and (b) it might be useful to use future definitions of mark price, (e.g. if it can come from or be bounded by an oracle) on the other side, in triggering price monitoring, i.e. trigger PM if a trade diverges from the mark price by some distance. Note that different mark price calculation may also in future help prevent erroneous MTM swings (by bounding the mark price against divergence from another source, for instance) rather than being the part that we are validating in price monitoring, which in my view is always about the price at which we actually print trades and whether or not to execute an upcoming trade at the specific price it would see. (You can also get a big MTM swing if you use mark price for price monitoring and print a trade far from it which you then value at the mark price, this wouldn't necessarily trip price monitoring if we were using mark price for PM because the mark price may now have moved at all but the trader's slippage casues the problem.)

I realise we've not discussed this much (and @witgaw and others, not at all), so I am going to try to explain why I proposed price monitoring and how I think about designing aspects of the system like this as I've never been explicit about it…

The protocol design philosophy here is not to prevent or mitigate against specific risks/manipulations/problems such as erroneous MTM swings, market bullying, etc. Rather, it is **to create a system that only executes trades at *reasonably fair prices* (for a definition of reasonably fair that we choose and can guarantee).** Any risks, exploit/manipulation vectors, etc. that are not exploitable given our definition of *reasonably fair prices* do not need to be explicitly addressed in other protections, provided that we can show, model, and document that this is the case.

In this case a price is _reasonably fair_ under our current definition if it is either: (a) a sufficiently likely price for the market given the current risk model and calibration; or (b) verified by an auction.

In general, I believe we should always prefer broad principles like this around fairness, execution, etc. over problem specific mitigations as they can be designed and calibrated to catch and mitigate whole classes of problems and are more likely to lead to a simpler and more "antifragile" system.

Please ask questions, etc. on this :)

specs/0000-price-monitoring.md
barnabee: @tamlyn10 yep, that is what I would expect, though a pegged as persistent but not permitted in auctions may be allowed to enter and be immediately parked (that is what we did previously)

specs/0000-price-monitoring.md
barnabee: 'price protection auction period' (which temporarily overrides the trading mode and makes it 'auction') would work for me, I think mode is an overloaded term with trading mode and could cause confusion.

Note: as I see it a liqudiity protection auction period has precedence over a price protection auction period, meaning that if both would be triggered at the same time, it would be the reported period type and if price monitoring was then triggered on exiting (subject to the below) the auction would continue and switch to a price protection auction. If the market is already in price protection but not liquidity protection we wouldn't switch and would let the price protection period(s) run it's course, only switchign to a liquidity protection auction if LP would be triggered after exisint the final price protection auction, if that makes sense.

I'd also suggest time spent in a liquidity protection auction period should be counted towards the time required for any immediately following price protection period so as not to double up (auctions aggregate both liquidity and pricing signals regardless of what triggerd them).

specs/0000-price-monitoring.md
barnabee: @tamlyn10 I am not sure if it is easy in all future risk models. We could say that it is done with the same frequency that the market recalculates margin which I think is configurable (if not, it is expected to be eventually) and is initially done with every price move. Would that work?

specs/0000-price-monitoring.md
barnabee: Yes absolutely, so it might be (massively oversimplified in terms of the movement definition):

\>=1% move in 10 min window -> 5 min auction
\>=2% move in 30 min window -> 15 min auction (i.e. if after 5 min this trigger condiiton is satisfied by the price we'd uncross at, extend auction by 10 mins)
\>=5% move in 2 hour window -> 1 hour auction (if after 15 mins, this is satisfied by the price we'd uncross at, extend auction by another 45 mins)

specs/0000-price-monitoring.md
barnabee: See my other comment here https://github.com/vegaprotocol/product/pull/275/files#r421985031

I do not think we should use mark price for triggering price monitoring. I am quite sure about this, if you disagree we will need a call on it.

specs/0000-price-monitoring.md
tamlyn10: OK, so good point on using definitions of the mark price that help protect against erroneous MTM swings.

I think the problem we've always had on this topic is that you've had your design principle in mind, informed by what you saw at LSE, and in fact, I'm not particularly so concerned about the LSE type motivations.  If we didn't have price monitoring (and in fact I don't think my markets did - or at least, the prices could more than double in a day without the exchange intervening), the way that a market protects itself from erroneous prices is by market forces - i.e. if prices diverge that becomes an opportunity to buy low or sell high for people who hold the belief that it's erroneous. And in normal markets these obligations are enforced by law, so the market participants aren't more at risk because of this kind of trade. I can see why the LSE does it, and your mentality / desire for orderly markets, but as a purist trader it also makes me slightly uncomfortable. I'd rather the erroneous trades were "punished" with market forces than a nanny type mechanism. (To be clear, the penalty is that the person who wants to buy well above where the market really is, gets left with an out of the money contract on their hands...  which they have to pay up for, come settlement - whether by periodic MTM or at expiry).

For me this mechanism is **vital** for while we have **pseudonymous trading** that could leave the market with un-maintained open positions that close-out mechanisms can't fix (because the prices entered at are too far away from where anyone would take the position on at that price). That's what made *me* want price monitoring. And yes, we've always argued about the intent, but ended up both wanting the same thing.

I'd like to note that if we had a credit or identity system or we had some other way of covering the value change of a bad trade, I would probably argue to not do much price monitoring (other than some extreme fat finger protection).

specs/0000-price-monitoring.md
tamlyn10: oh right, yes, I see.. that makes sense.

specs/0000-price-monitoring.md
barnabee: You are correct it is important to create this definition of reasonably fair prices and do something to ensure them because of pseudonymity, that is a good part of my principle. Though an at least equally important basis IMO is around what makes a good or valuable trading platform, and just like the HFT arms race is not beneficial to the market as a whole, so it is not beneficial to have obviously sub-optimal price determination and to "punish" traders like that.

FYI my design principles for price monitoring Vega are not the LSE's design principles for price monitoring. I have just taken a lot of their mechanics. Their reasons are mostly to do with participants getting upset about flash crashes and sometimes regulators too, or I don't think they'd have it.

The problem here is that we need to agree the design principles and philosophy for what we are building and design Vega based on those not on any of our personal reasons or biggest risks/issues. As the originator/architect for the protocol I am going to say that this philosophy/approach has to ultimately be agreed by me. I am happy to discuss and debate these, and change them if convinced, but the basis for the design does need to be consistent and think it is important to be clear on that.

If building price monitoring on this basis doesn't prevent the potential gaming or whatever issues that you see from being a problem then we need to separately evaluate whether changes to the system (price monitoring or otherwise) are needed for those specifically, but my approach is very explicitly not to make the design of price monitoring be or be driven directly by those issues, or indeed by any specific ones.

Happy to have a call later if you want to discuss further.

specs/0000-price-monitoring.md
barnabee: Edited above for clarity.

specs/0000-price-monitoring.md
tamlyn10: yeah i like this..

specs/0000-price-monitoring.md
davidsiska-vega: I think that Barney correctly wrote up the guiding principle behind price monitoring in the first reply above: a large move can be "true view" or it may be an "artefact". The auction is a mechanism to let the market decide and it prevents an "artefact" move create a mess of closeouts. 

We should use the risk-model view of what large is (again what Barney said): yes on the level of price monitoring we can have a different view of what large move is. But why introduce more spurious parameters into the story? Remember it also has to tie in with liquidity measure and funny things could happen if these two become disconnected.

specs/0000-price-monitoring.md
davidsiska-vega: We also probably want to write that if we exited an auction mode then the move monitoring starts at the end of the auction period and then increases until it again reaches the normal period (so that the new price the auction determined is "acknowledged"). To be more precise:
- tau is the monitoring period so we are always checking that  V_{now} - V_{now-\tau} is within acceptable range. 
- if V_{now} escapes this range we go into auction and get V_{auction end} 
- then we should monitor V_{now} - V_{max(auction end, now - \tau}}

In fact as we always start with an opening auction we should always monitor that V_{now} - V_{max(prev auction end, now - \tau}} is within acceptable range.

specs/0000-price-monitoring.md
davidsiska-vega: @barnabee @tamlyn10 In some models(*) the \log(S_T / S_0) = \log(S_T) - \log(S_0) follows a distribution which does not depend on S_0 (aka mark price). So the risk model can calculate everything once, then just do the required logarithmic transformations and only recalculate if risk parameters change. 

(*) All the ones we have implemented now.

In fancier models this is no longer true but we can probably update based on this dependence less frequently. With a bit of caching it should be manageable in general.

specs/0000-price-monitoring.md
davidsiska-vega: I would say the precision should not be configured in the market framework but it should be something the risk model spits out when it's done its calculation (i.e. I, the risk model, managed to do this calculation to this precision).

specs/0000-price-monitoring.md
tamlyn10: I'd rather have the conversation here for the benefit of others and of provenance.

So in terms of design principles, I would assume that foundational to all of yours (as originator/architect) is a desire to build something useful for the real world trading of derivatives. The market using market forces to correct price mismatches (even if by virtue of the market microstructure) is a form of arbitrage and shouldn't be likened to HFT. Much of financial theory relies on market forces to correct prices. Slowing the system down and putting it into auction mode so that can happen more sanely is probably a good thing to do, but the **actual real serious reasons for doing this** - i.e. the negative consequences of an erroneous price (e.g. a  flash crash) are actually not well spelt out either in this spec or in the whitepaper.

I believe I've done a good job above explaining the actual underlying problem with open positions being at higher credit risk in a pseudonymous market that has any time based risk (i.e. not spot instant physical settlement, i.e. derivatives).

To elaborate from a "useful to traders" point of view. 

When you enter an exchange based futures position (particularly if it's to hedge a real world risk), there's a couple of risks you take on.

1. Firmness - the potential for your filled position to degrade in volume because of default from counterparts being fed through to you.

2. Capital - that you can't sustain the mark to market moves to hold your position open.

These are the 2 things I've most thought about when considering a pseudonymous environment with automated liquidations. I actually don't care as a trader if there is a flash crash if the above things remain intact. I do very much care what the protocol is doing in terms of these things.

So, I believe that above, I have outlined solid financial consequences if a) price monitoring doesn't exist + b) collateral release does exist.

I'm fine with you defining all the principles you'd like, but if the Vega protocol doesn't protect against the attack I outlined in my previous message, both of the above things become a risk/threat for traders of this market. Therefore, I think it's worth mentioning in the whitepaper (as we did in Section 6.6) and this spec - because if engineers in a few years time decide to remove price monitoring (because, hey, a market creator may not really want such things and would rather risk market suspension of degradation of the above 2. things I listed above) then I'd like at least part of my legacy to be that I warned you all of simple, easy attack that you might want to think of another mechanism to protect for.

If you think a separate spec that documents this is useful, that's fine with me. I don't think it should be swept aside and lost in this PR messaging.

specs/0000-price-monitoring.md
barnabee: Well the precision at which price levels exist is already configured in the market framework. I am not sure what the monitoring engine is going to do with the information about the model's view of the precision of its outputs? It seems to me all we can do is take the output from the model and take the round/ceil/floor of it as the price level to trigger the auction?

specs/0000-price-monitoring.md
barnabee: I have no idea what you mean by PR messaging.

This is all very impoertant stuff, your points are critical to Vega's design and success. They need to (a) be captured in separate document or repository that rigorously catalogues the types of attacks/issues/risks that the system may face and how the design mitigates them. This should be a reviewed and audited document when we make design changes or protocol additions; and (b) might result in udpates or additions to the whitepaper.

Shoving this stuff in a spec, even short term, is the wrong answer for two reasons: (1) it makes it harder to build clear, good quality specs (and we are already far off what we will need for audits or being able to open source them); and (2) the (very important!) info in these points you're making won't get the proper consideration abd we risk missing stuff if it's just in specs. There's no guarantee that price monitoring alone mitigates these or other risks in all scenarios (unless you believe we have shown this to be true) and we are not likely to cross reference and check off every risk referred to in any spec file for each protocol addiiton or design change. Having a separate document gives us something to review as part of "signing off" a release for Mainnet when its there.

I'm happy to use a GitHub issue or Strategy card to capture those points, or for us to create a starting point for that document/repo, but:

1. We still require a call if you want to discuss changing to mark price instead of the actual trading price
1. I'm not comfortable to say that this features is designed specifically to mitigate those risks. I am not convinced that it does that on its own, and certainly that has not guided the design process for the feature (it may yet guide the _calibration_ of the feature and how we set the parameters that control it, i.e .the moves and period lengths, but that should not be in the spec either).

specs/0000-price-monitoring.md
barnabee: @davidsiska-vega thanks for the info. I am happy for whatever we want to go here in that case and we can deal with it at the time if we hit performance problems later with what we specify now.

specs/0000-price-monitoring.md
barnabee: Hmm, I has assumed other printed prices before the auction but within the window period should be included rather than resetting post-auction. Now I'm not sure - whats your reasoning that we should scrap the prior data points after an auction? Would you do this after all auctions or only price monitoring ones?

specs/0000-price-monitoring.md
tamlyn10: "PR messaging" = "Pull Request messaging" (which is where i think this conversation is happening?).

So yes, I do know that my points are critical to Vega's design and success, given we are trying to make something that's useful to traders. In fact, all our design criteria should be centred in that (which I believe we / you have always tried to do). What I've attempted to express because I believe it's an important perspective is why price monitoring actually matters to traders. Again, I'll restate that I don't think we explain it well in the whitepaper to reflect the consideration of the two things traders who are holding open positions really care about (firmness and capital - that I outlined above). They're more than just risks, they underpin the usefulness of the system, particularly for real world derivatives trading; quite literally, boards will not approve using Vega for hedging real business risk if those two things are considered high or unmanageable risks. So the extent to which price monitoring helps these things is my primary concern. As I said, a flash crash that doesn't impact those things matters far less to me.

So, in summary, I think that your POV is the spec should say
1. price monitoring is to avoid "erroneous prices"

My POV is the spec (and whitepaper) should say
2. price monitoring is to avoid "firmness and capital" risks that arise from "erroneous prices"

I don't much mind if we go with your option, but this conversation has made me relook at our whitepaper and consider how "trader centric" it is. I would say that some of the high level design principles / "desirable properties of a decentralised trading system" should probably have more explicit mention of trader's criteria (for example we don't even mention capital efficiency). 

So, while I brought up a simple (bullying) attack to demonstrate how easily and quickly that the absence of price monitoring would have on a trader's "firmness and capital" for their position, the attack itself is not the whole point I was trying to make. The attack's impact on firmness and capital for traders (as well as the overall sense of fairness of the system) does matter. And I agree, you're right that we need to document them and consider what aspects of our system mitigate those. I would suggest this sit in specs, whitepaper or some level of system design documentation because a tweak of the system here or there should definitely be considered against these.

As to the mark price vs trading price debate, IF we have:
1. the above attacks well documented in a design doc  (and in that I will explain how the simple attack, for now, relies on price monitoring & markprice = tradeprice),
2. a process for only changing the protocol subject to consideration of such docs

(both of which you suggest, I'm just reiterating..)

... then I'm ok with keeping it to be trade price :-)

barnabee: Ah yes, they’re called PRs now not MRs 🤯 🤦‍♂️ 

Sounds like we have an approach then :-)

edd:
@witgaw Will you have time this week to incorporate the changes discussed?

witgaw:
@edd Yeah, I'll go over these this afternoon.

specs/0000-price-monitoring.md
witgaw: Used @barnabee's description and otherwise modified the summary. Please check the new version and add further comments or resolve the conversation.

specs/0000-price-monitoring.md
witgaw: As above, used @barnabee's description and otherwise modified the summary. Also removed the references to mark price. Please check the new version and add further comments or resolve the conversation.

specs/0000-price-monitoring.md
witgaw: Modified as suggested, resolving

specs/0000-price-monitoring.md
witgaw: Modified as suggested, resolving the conversation.

specs/0000-price-monitoring.md
witgaw: Included the concept of arrival price of the next transaction, resolving.

specs/0000-price-monitoring.md
witgaw: This should be addressed now. Please check the new version and add further comments or resolve the conversation.

Please note I haven't included @tamlyn10's comments and possible attack vectors anywhere, not sure where these are meant to go.

specs/0000-price-monitoring.md
witgaw: This should be addressed now. Please check the new version and add further comments or resolve the conversation.

specs/0000-price-monitoring.md
witgaw: Modified as suggested, resolving the conversation.

specs/0000-price-monitoring.md
witgaw: Changed to "next transaction's arrival price" as agreed in a different conversation.

specs/0000-price-monitoring.md
witgaw: Included in the spec, resolving.

specs/0000-price-monitoring.md
witgaw: Included in the spec, resolving.

specs/0000-price-monitoring.md
witgaw: I meant the supported probability level - to avoid situations where the model is queried for a 99.999999% interval which it likely won't be able to provide sensibly. 

Rephrased now to make it clearer. Please check the new version and add further comments or resolve the conversation.

specs/0000-price-monitoring.md
witgaw: Included in the spec, resolving.

specs/0000-price-monitoring.md
witgaw: Included in the spec, resolving.

specs/0000-price-monitoring.md
witgaw: Included in the spec now, resolving.

specs/0000-price-monitoring.md
witgaw: Included in the spec now, resolving.

specs/0000-price-monitoring.md
witgaw: Just to confirm: I'd expect the new 10 minute horizon to be computed every second (or however often we set the bounds to be recalculated), so that a forward-looking time series of bounds is constructed and each of them is considered when their time comes. As opposed to calculating bounds for a 10 minute horizon, checking it for the next 10mins and when the 10mins pass calculating a new one.

Is that in line with your thinking?

barnabee: I think so... if T=14:50:00 and the period is ten mins then the window is all prices from 14:40:00 onwards, if next tx comes in at T'=14:50:03 then window is all prices from 14:40:03, etc. etc. — is that what you're talking about?

specs/0000-price-monitoring.md
barnabee: Ah that makes total sense and I agree, thanks

specs/0000-price-monitoring.md
witgaw: Yes, thanks for confirming.

specs/0000-price-monitoring.md
witgaw: @barnabee I think that @davidsiska-vega's reasoning is that this way we acknowledge that there was a step change/"jump" in the price and we reset the history so as to not mix the 2 regimes (prior to & post jump). Would be good if David could confirm that though.

barnabee: I'm not sure why they are 2 regimes. Also how does this work when we have multiple levels with different length auctions i.e. 1% in10 min = 5min auction, 2% in 30 min = 15 min auction, etc. - for the 2% check you are looking at both what the end price of the 5 min auction would be and the bounds of the prices within 30 min including before the auction, not sure how that works with the reset?

If the auction price still ends up "wrong" for whatever reason we would be erasing all the prior price history. I've always seen it simply as: an auction is a period in which no prices are printed rather than as a reset and determination of a new "one true price". 

@davidsiska-vega would be good to know what your thinking re: the benefits of a reset vs including all traded prices is?

Other random thought: any advantage to compare VWAP of price between now-T and now rather than just the price at now-T? I can't see an obvious advamtage but also not an obvious reason the other way is better either..

barnabee: Re: aknowledging a step change- it may not be a step change. Could be a sustained downward run that trips the trigger too. I'd agree it's more reasonable if we somehow _know_ it's a step change but detecting that seems pretty tough.

davidsiska-vega: Imagine that price is at 100 and a move of 10 is outside 99.9% probability within 1 minute. 

Until t=30 the price stays at 100. At t=30 sec we received an order that would move price to 89 - we triggered an auction which resulted in a price of 88 at t=45 sec. Then there was an order that would trade at 89 at t=46. That’s *still* outside the range. Do we want to trigger an auction? 

My thinking was that no we don’t. When the auction concludes we set the reference price to 88 and start the clock again. 

We could of course not do that and have a few more auctions before the price level of 100 is far enough.

witgaw: Added the resetting assumption to the spec, resolving conversation.

davidsiska-vega:
`- [ ] Risk model returns probability density function (p.d.f.) and the inverse p.d.f.` 

To me it would seem more natural if the risk model implemented only a function that takes as input: (current price - mid or mark, confidence level alpha, time period tau) and returned the limits S_min and S_max such that P(S_\tau < S_max) >= alpha and P(S_\tau > S_min) >= alpha. Example input (100, 1 hour, 0.99) returns S_max = 111.2343 and S_min = 93.456. 

This way it's clear that `core` doesn't need to do any calculations and no post-processing.

witgaw:
> `- [ ] Risk model returns probability density function (p.d.f.) and the inverse p.d.f.`
> 
> To me it would seem more natural if the risk model implemented only a function that takes as input: (current price - mid or mark, confidence level alpha, time period tau) and returned the limits S_min and S_max such that P(S_\tau < S_max) >= alpha and P(S_\tau > S_min) >= alpha. Example input (100, 1 hour, 0.99) returns S_max = 111.2343 and S_min = 93.456.
> 
> This way it's clear that `core` doesn't need to do any calculations and no post-processing.

@davidsiska-vega Ok, good suggestion. However, why "current price - mid or mark" ? Something like (current price, timestamp, confidence level alpha, time period tau) would allow the risk model to hold the history of prices and do the required look-up.

davidsiska-vega:
> > `- [ ] Risk model returns probability density function (p.d.f.) and the inverse p.d.f.`
> > To me it would seem more natural if the risk model implemented only a function that takes as input: (current price - mid or mark, confidence level alpha, time period tau) and returned the limits S_min and S_max such that P(S_\tau < S_max) >= alpha and P(S_\tau > S_min) >= alpha. Example input (100, 1 hour, 0.99) returns S_max = 111.2343 and S_min = 93.456.
> > This way it's clear that `core` doesn't need to do any calculations and no post-processing.
> 
> @davidsiska-vega Ok, good suggestion. However, why "current price - mid or mark" ? Something like (current price, timestamp, confidence level alpha, time period tau) would allow the risk model to hold the history of prices and do the required look-up.

I am sorry I wasn't careful writing this. Indeed the input would be `current price`. What I tried to express that the `current price` could be either `mid price` or `mark price` the spec can decide what makes more sense but from the point of view of the risk model it doesn't matter.  

Also I don't think that in the current risk models setup we need `timestamp` we only need the time period tau. But if you want to be forward looking in terms of possible models we may one day have then we could include `timestamp` as well (but should agree how this time is expressed). By the way I am assuming that the time period tau would be as a year fraction i.e. to input 1 hour you'd input `1/(24*365.25)`.

witgaw (commented):
Approving these as they've already been adressed.

witgaw (commented):
Accepting changes as these have already been addressed.

ashleyvega (commented):
AOK by me.

specs/0000-price-monitoring.md
```diff
@@ -0,0 +1,87 @@
...
+  - Price protection auction is then triggered for a period T.
+- In general we might have a list of triplets: α, τ, T specifying each trigger.
+- Once the price protection auction period finishes, the remaining triggers should be examined and if hit the auction period should be extended accordingly.
+- Then the market continues in the regular fashion.
+
+# Reference-level explanation
+
+## View from the [vega](https://github.com/vegaprotocol/vega) side
+
+- for each transaction:
```
davidsiska-vega: I wasn't mentally present for all the discussions so perhaps this is the best way to do it, in which case ignore my comment. But it's not how I would do it. 

In my mind the trading core should (periodically as it sees fit) obtain from the risk model the maximum move up and the maximum move down for a given time period (and it can keep more than one of the [period, max move up, max move down] triplets). The risk model supplies these based of confidence level `alpha` and on the current mark price. 

Vega core then checks *before each transaction is committed to* whether it would have moved the price out of these bounds: ie it determines new mark price, looks at old mark price (where what old means is determined by the length of the interval) and then:
- if new mark price - old mark price > max move up *then* don't commit transaction and trigger auction
- if old mark price - new mark price > max move down *then* don't commit transaction and trigger auction
- else commit transaction and proceed.

The reason for this setup is that the risk model can be slow (and this is outside what core can control). This way core always have "levels" that it can work and use basic arithmetic to determine whether auction needs to be triggered or not. 

Another reason is that on the quant calculation side we'd rather not have to work with real world objects like `vega time` I'd much rather we stuck to abstractions like year fractions.

witgaw: That would help with allowing some asynchronicity, the only thing is that then core would have to maintain it's own price history (max tau) back, but maybe it's not too big of a problem. I suppose quant library could then look-up the prices elsewhere (not as part of the two way data exchange described in the spec), update the max move up/down when ready and core pulls it in whenever it wants to. Could someone from @vegaprotocol/core comment on the feasibility of this please?

witgaw: @davidsiska-vega one thing, which "old mark price" should core be looking up then? The problem mentioned at during the meeting is that there are usually many prices with same timestamp (within one block), should it be the average over a given timestamp? Clearly same question would still apply to the spec as is, it would just have to be answered at a point of ticketing the work on quant library side.

davidsiska-vega: I see. So maybe the correct setup would to be to have a price monitoring "library" / "module" sitting between core and quant library that does this work. The core will feed it all the prices / timestamps as described in the spec and it will `send back signal informing if the price protection auction should be triggered (and if so how long the auction period should be)`.

It will achieve this by getting communicating with the risk model as described above `(periodically as it sees fit) obtain from the risk model the maximum move up and the maximum move down for a given time period (and it can keep more than one of the [period, max move up, max move down] triplets). The risk model supplies these based of confidence level alpha and on the current mark price.`

It will average the mark prices with the same timestamp for purpose of using as "old price" in the comparison.

witgaw: Cool, thanks @davidsiska-vega. @vegaprotocol/core please provide your input on this entire conversation when you get a chance as we're venturing into implementation territory a lot here.

barnabee: @davidsiska-vega your last comment sounds a lot like what's written already? 

It seems like we need to do some renaming though:

-  the thing responding with the "triggered Y/N" should be the *price monitoring engine* 
- it would be the matching engine (probably... cc @vegaprotocol/core) that sends the arrival price to the price monitoring engine
- the risk model exposes a function(s) to provide what's needed by the *price monitoring engine* to determine if any of the configured limits are breached

Is that it?

edd:
Please could @peterbarrow, @EVODelavega, @jeremyletang look at this before tomorrow evening? This has been waiting for review for a while. 

@pand-dev I've left you off the list as this will probably overlap with the fees implementation work but if you have the time your review is also welcome.

edd:
Also: one of @barnabee or @tamlyn10

specs/0000-price-monitoring.md
jeremyletang: I may not get everything there, but what I think of first is:
you mention first that the core should get the maximum move up and down from the risk models as it see fits. Altho I think this needs somehow to be made explicit I imagine, as you mention the calculation can be done asynchronously? Would it be possible for some nodes running on less faster machine than the rest of the network to have the calculation given to it asynchronously much later than the other nodes, which could make other nodes to think an auction is required already, but some would think not until the asynchronous calcuation is done?

jeremyletang:
That's sounds all pretty clear to me.
Altho I left a comment about the asynchronous calculation done by the risk models as it may need to be clearer on how asynchronous it is / and make sure  all node trigger calculation the same way / wait the same way to get the results.

specs/0000-price-monitoring.md
barnabee: Price monitoring calculations should not ever happen asynchronously. 

In future some risk models will need overall risk calculation to be done asynchronously (i.e. result would presumably need to go through consensus as a transaction) but that should be done as a single asynchronous "risk calculation" the results of which can be used by the price monitoring level calculation if needed, but the PM calc itself is always synchronous using the latest available risk data.

specs/0000-price-monitoring.md
barnabee: @davidsiska-vega fair point, we need to check whether a extension to the auction is needed (i.e. price has triggered the next 'level' = bigger move/longer auction) but once we deciede not and exit the auction I agree the time window needs to be reset.

barnabee:
> That's sounds all pretty clear to me.
> Altho I left a comment about the asynchronous calculation done by the risk models as it may need to be clearer on how asynchronous it is / and make sure all node trigger calculation the same way / wait the same way to get the results.

@jeremyletang I replied to your comment. There should be no asynchronicity in the price monitoring calculation. Some risk models will in future do an async calculation workload, results of which would need to go through consensus (or at least a hash of the results) but ALL of the code that _uses_ the risk model outputs (i.e. for calculating margins, price monitoring, market making stuff, etc.), that calculation that uses the risk model should occur syncronously (i.e. using cached results from the async calc). (Aside: It will very likely be necessary when we get to this to put a market with an asynchronous risk calculation into a temporary auction if it goes too long without a new set of risk calculation outputs coming through consensus, as after some time it will no longer be safe to operate the market without recalculated risk numbers.

barnabee (approved):
Approved once outstanding comments are resolved

peterbarrow (approved):

EVODelavega (approved):

specs/0000-price-monitoring.md
ashleyvega: So, no async price monitoring calculations.

@witgaw Might you be able to add that to the spec and resolve this conversation?

witgaw: @ashleyvega Sure, I'll resolve the remaining conversations and get this ready to be merged this afternoon/evening.

ashleyvega: @witgaw Thanks. Please merge when you're ready (no need for further deliberation :smile:).

witgaw: Incorporated these into the spec now. Resolving the conversation and merging the PR. Thank you all for your contributions.

PR #277, opened 2020-05-04
First fees draft.

davidsiska-vega:


specs/0029-fees.md
```diff
@@ -0,0 +1,42 @@
+
+# WIP Fees on Vega
+
+## Collecting fees
+
+In general the trading fee can be (we could have more general non-affine functions but I don't see why):
+
+fee = "fee offset" + "fee factor" x notional.
```
tamlyn10: Change _notional_ to be a more generic term, configured as a market parameter.

For futures it will be notional = volume * price

specs/0029-fees.md
```diff
@@ -0,0 +1,42 @@
...
+
+Fees are collected in the settlement currency of the market. The "fee offset" is non-negative fixed amount (network parameter?) while the "fee factor" is set as part of the market making staking mechanism (market makers submit their preferred fee factor and their stake-weighted average determines the fee factor). N.B. "fee factor" is per market.
```
tamlyn10: remove word _staking_

specs/0029-fees.md
```diff
@@ -0,0 +1,42 @@
...
+
+In general the trading fee can be (we could have more general non-affine functions but I don't see why):
+
+fee = "fee offset" + "fee factor" x notional.
+
+Fees are collected in the settlement currency of the market. The "fee offset" is non-negative fixed amount (network parameter?) while the "fee factor" is set as part of the market making staking mechanism (market makers submit their preferred fee factor and their stake-weighted average determines the fee factor). N.B. "fee factor" is per market. 
+
+### During Continous Trading
+
+The "aggressor" pays the fee to the "liquidity provider" which, here, are the participants that placed passive orders on the book (limit orders, special MM limit order, ...). Of course, since a market order may be hitting more that one order this has to be split correctly per volume taken from each limit order.
```
tamlyn10: price takers of each trade pay the fee
price makers of each trade receive a portion of the fee

specs/0029-fees.md
```diff
@@ -0,0 +1,42 @@
...
+
+fee = "fee offset" + "fee factor" x notional.
+
+Fees are collected in the settlement currency of the market. The "fee offset" is non-negative fixed amount (network parameter?) while the "fee factor" is set as part of the market making staking mechanism (market makers submit their preferred fee factor and their stake-weighted average determines the fee factor). N.B. "fee factor" is per market. 
+
+### During Continous Trading
+
+The "aggressor" pays the fee to the "liquidity provider" which, here, are the participants that placed passive orders on the book (limit orders, special MM limit order, ...). Of course, since a market order may be hitting more that one order this has to be split correctly per volume taken from each limit order. 
+
+The whole fee should be collected from the "aggressor".
```
tamlyn10: price taker

specs/0029-fees.md
```diff
@@ -0,0 +1,42 @@
...
+
+Fees are collected in the settlement currency of the market. The "fee offset" is non-negative fixed amount (network parameter?) while the "fee factor" is set as part of the market making staking mechanism (market makers submit their preferred fee factor and their stake-weighted average determines the fee factor). N.B. "fee factor" is per market. 
+
+### During Continous Trading
+
+The "aggressor" pays the fee to the "liquidity provider" which, here, are the participants that placed passive orders on the book (limit orders, special MM limit order, ...). Of course, since a market order may be hitting more that one order this has to be split correctly per volume taken from each limit order. 
+
+The whole fee should be collected from the "aggressor". 
+
+Part of the fee (network parameter?) is immediately credited to the "liquidity provider".The remainder of the fee should be transferred to the "market fee pool" (another account similar to but separate from market insurance pool).
```
tamlyn10: price maker

specs/0029-fees.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Auction
+
+During auctions there is no "aggressor" and no "liquidity provider". Each side in a matched trade should contrinute 1/2 of the fee amount. The fee amount should be transferred to the "market fee pool". 
+
+## Distributing fees
+
+The market fee pool has now collected some amount in fees.
+
+Fees should be distributed periodically (could be any period but I would advocate immediately).
```
tamlyn10: Do we want a configurable hold period before fees are distributed to market makers?

specs/0029-fees.md
```diff
@@ -0,0 +1,42 @@
...
+
+As far as I can see, during continuous trading, this should happen after the closeout trades are executed but before new batch of trades are matchd.
+
+During auction mode, I would think that fees can be distributed immediately after the end of auction period (which, if I understand, will result in a bunch of trades).
+
+The "trade liquidity providers" already got their cut earlier, so the only parties with claim to what's in the market fee pool are the node operators and market makers for the market.
+
+A fraction (network parameter?) should be distributed to the node operators (proportionally to the number of Vega tokens they hold and tokens others have delegated to them). This amount should go to a per-staking-node account that however the node operators cannot access immediately (because we may wish to slash their rewards if they are mis-behaving, are not performant enough, have too much stake etc. separate spec to come). For now it is enough that this is a separate account where each staking node operator can see how much has been earned so far.  
+
+The remainder is split between market makers based on their time-weighted market-making stake (another spec).
```
tamlyn10: Add in account. types:

1. Market making fee pool account - 1 per market - gets paid to market makers general account. Do we need this if we are immediately distributing?
2. Infrastructure fee pool account - 1 per Vega - eventually goes to node operators (still to be specced)

specs/0029-fees.md
tamlyn10: fee offset needs to be part of the market - a market parameter?

specs/0029-fees.md
tamlyn10: Fee is an amount in the settlement currency of the market. "fee offset" is an amount in the settlement currency of the market.

tamlyn10: Fee should be collected when trade is finalised.

tamlyn10: Fees should be defined in the market framework... or on the product??

specs/0029-fees.md
tamlyn10: Parameters:
1. proportion that goes to price makers (network parameter)
2. proportion that goes to market makers (network parameter)
3. proportion that goes to nodes (network parameter)

Acceptance criteria:
1. Fees should not be able to be negative

specs/0029-fees.md
barnabee: Entirely remove "stake weighted average" - we shouldn't create two places that describe how to calculate the liquidity component of the fee as then they have to be kept in sync), and I suggest we put the actual calculaiton in the market making spec.

I would say something like:

```
fee = (fee_factor[infrastructure] + fee_factor[maker] + fee_factor[liqudiity]) * trade_valuation
```

Where the fee factors are calculated by:

- infrastructure: staking/governance system/engine (network wide)
- maker: market framework / market making (may be per market or network-wide?)
- liqudiity: market making system (per market)

The method of calculation of these values is out of scope for this spec. If the fees MVP is built before the contributing part of the system that would calculate the fee factor, a static (but changeable by governance as normal) network parameter can be substituted for each missing calculation. Realistic values might be: ff[liquidity] = 0.1%, ff[infrastructure] = 0.05%, ff[maker] = 0.025%.

(NB: I'm using *trade_valuation* to refer to the amount from which we calculate fee, i.e. for futures, the trade's notional value.)

specs/0029-fees.md
barnabee: Either term is fine. I'd vote for aggressive/passive though as it more accurately captures the sentiment in non continuous trading/limit order book scenarios IMO.

specs/0029-fees.md
barnabee: See above.

specs/0029-fees.md
barnabee: See above.

specs/0029-fees.md
barnabee: Yes. My proposal is that:

1. amount to go to each market maker is calculated at time of trade
2. it is distributed to them after the configurable hold period (network param) has elapsed

From an implementation perspective, I'd suggest we created a market making rewards/fees/whatever account for each commited market maker (either per market they're committed to or per asset in which they have commitments) and **transfer the fees there immediately**, then _to their general account after the hold period_, so that (a) we know were the money is in the accounting system and it's not sloshing around in some vague market-wide pool; and (b) the existing APIs allow them to query the rewards they've earned but that haven't yet been paid out.

specs/0029-fees.md
barnabee: Acceptance criteria should also cover that the [temporary] network parameters are used to calculate the fees, and that they can be changed via standard network parameter governance proposals and that those changes take effect at exactly the right time for fee calculations. This will give some confidence that (a) we've built the governance stuff well; and (b) the fee stuff we built will work properly when we replace network params with dynamic calcs.

barnabee: > Add in account. types:
> 
>     1. Market making fee pool account - 1 per market - gets paid to market makers general account. Do we need this if we are immediately distributing?
> 
>     2. Infrastructure fee pool account - 1 per Vega - eventually goes to node operators (still to be specced)

We don't need account type one but I suggest we do need a new account type per market maker for pending fees, either for each market they're committed to or aggregated for each asset that's a fee currency for a market they're commited too, as per https://github.com/vegaprotocol/product/pull/277/files#r420447920

specs/0029-fees.md
barnabee: See below — https://github.com/vegaprotocol/product/pull/277#discussion_r420444798 — I'd suggest renaming fee offset and don't understand how it'd be an absolute amount, I think we need to provide things to multiply the trade value (notional) by. We also need to cover maker/taker as well as liqudiity and infrastructure.

barnabee: > Fees should be defined in the market framework... or on the product??

Not sure what you mean? What would be defined in the market framework/product exactly?

**Infra fee** comes from staking/governance 
**Liquidity fee** comes from market making
**Maker reward** comes from _???_ either a network param, the market framework or market making.

specs/0029-fees.md
```diff
@@ -5,38 +5,40 @@
 
 In general the trading fee can be (we could have more general non-affine functions but I don't see why):
 
-fee = "fee offset" + "fee factor" x notional.
+fee = "fee offset" + "fee factor" x CalculationAmount
 
-Fees are collected in the settlement currency of the market. The "fee offset" is non-negative fixed amount (network parameter?) while the "fee factor" is set as part of the market making staking mechanism (market makers submit their preferred fee factor and their stake-weighted average determines the fee factor). N.B. "fee factor" is per market. 
+Fees are collected in the settlement currency of the market. The "fee offset" is non-negative fixed amount (network parameter?) also set the settlement currency of the market. The "fee factor" is set as as a separate mechanism, and may update regularly.
```
barnabee: I don't understand how fee offset is a fixed amount. Should be another percentage IMO. I'm not sure how the comments work and mine seem to be hidden when viewing the file or GitHub is being screwy but covered this here: https://github.com/vegaprotocol/product/pull/277#discussion_r420444798

specs/0029-fees.md
```diff
@@ -5,38 +5,40 @@
...
 
-### During Continous Trading
+Initially, for futures, the CalculationAmount = notional volume of the trade = size_of_trade * price_of_trade
 
-The "aggressor" pays the fee to the "liquidity provider" which, here, are the participants that placed passive orders on the book (limit orders, special MM limit order, ...). Of course, since a market order may be hitting more that one order this has to be split correctly per volume taken from each limit order.
```
barnabee: Fees should be calculated by and associated with trades not by orders, i.e. for each trade either there is an aggressor (in continuous trading, the newly arriving trade) in which case they pay the full fee or there is no aggressor/it is both (i.e. normal auction) in which case they share the fee equally.

specs/0029-fees.md
```diff
@@ -5,38 +5,40 @@
...
+Fees are collected in the settlement currency of the market. The "fee offset" is non-negative fixed amount (network parameter?) also set the settlement currency of the market. The "fee factor" is set as as a separate mechanism, and may update regularly.
 
-### During Continous Trading
+Initially, for futures, the CalculationAmount = notional volume of the trade = size_of_trade * price_of_trade
 
-The "aggressor" pays the fee to the "liquidity provider" which, here, are the participants that placed passive orders on the book (limit orders, special MM limit order, ...). Of course, since a market order may be hitting more that one order this has to be split correctly per volume taken from each limit order. 
+### During Continuous Trading
 
-The whole fee should be collected from the "aggressor". 
+The "price taker" of each trade pays the fee. An order may cross with more than one order and each trade incurs a fee. NOTE: we need to figure out how the "fee offset" is incurred for multiple trades.
```
barnabee: I don't understand this note or the fee offset, not why we are worrying about orders crossing with more than one thing as it makes no difference.

specs/0029-fees.md
davidsiska-vega: It's a fixed amount to be paid for executing a trade. It can be zero in many markets (certainly for futures it would make sense). If we conclude it's unnecessary then let's get rid of it and we keep just the "factor" part.

barnabee: Ah ok I see. Yeah I hadn’t imagined a fixed amount but I quite like the idea of building it into the market framework as a per market parameter that can be adjusted via governance (which would eventually move from a token weighted vote to market maker weighted) — as you say this might be zero for futures if the other components are contributing something/enough but I’ve definitely seen markets where the fee is the same per trade rather based on trade size. (Another way to achieve this same idea would be for the “trade valuation” aka notional used for fee calf’s to be configurable to a constant in the market framework, I have no opinion either way but this fee offset approach is simpler I suspect.)

That does mean we need to include the rate from
Infrastructure and the maker reward when calculating the fee as well, I think.

specs/0029-fees.md
tamlyn10: Do you understand this now, given David's comment?

specs/0029-fees.md
barnabee: Alternative with fee offset

```
fee = fee_offset + ((fee_factor[infrastructure] + fee_factor[maker] + fee_factor[liqudiity]) * trade_valuation)
```

specs/0029-fees.md
```diff
@@ -0,0 +1,69 @@
...
+
+trade_value_for_fee_purposes:
+* refers to the amount from which we calculate fee, (i.e. for futures, the trade's notional value.)
+* trade_value_for_fee_purposes is defined on the Product and is a function that may take into account other product parameters 
+
+Initially, for futures, the trade_value_for_fee_purposes = notional volume of the trade = size_of_trade * price_of_trade
+
+### Collecting and Distributing Fees
+
+Fees are incurred immediately at the point of trade from the margin balance of the liable trader, and if insufficient balance then the trade cannot occur.
```
ghost: When specifically do we calculate fees? Right before or after a trade was generated?

If insufficient balance, what specifically should happen to the trade and the order fill that generated this trade?
At the moment, Trades in the vega core do not have a state/status. Should failed trades be simply removed if fees could not be collected?

Should the minimum fee be calculated and locked on order submission (and released if order is cancelled or expired)? That could prevent failing trades

ghost:
Will this fee framework be sufficient for deposits/withdrawals and cross account transfers?
Or is this something to be considered later on?

specs/0029-fees.md
```diff
@@ -0,0 +1,69 @@
...
+
+The transfer of fees must be completed and represented in the trader's balances before any risk / margin calculations that use those balances.
+
+infrastructure_fee: transferred to infrastructure fee pool for that asset
+maker_fee: transferred to the relevant party
+liquidity_fee: transferred to market-maker fee pool for that market
+
+Note, further distribution of the infrastructure_fee and liquidity_fee are covered in future specs (market making and validator specs).
```
ghost: For the first iteration, MVP.
Where do we tranfer the fees? Should we allocate an account per market to hold the fees for the moment?

specs/0029-fees.md
```diff
@@ -0,0 +1,69 @@
...
+
+Note maker_fee = 0 if there is no maker, taker relationship between the trading parties.
+
+The fee factors are calculated by:
+
+_Factors_
+infrastructure: staking/governance system/engine (network wide)
+maker: market framework / market making (may be per market or network-wide?)
+liquidity: market making system (per market)
+The method of calculation of these values is out of scope for this spec. If the fees MVP is built before the contributing part of the system that would calculate the fee factor, a static (but changeable by governance as normal) network parameter can be substituted for each missing calculation. Realistic values might be: fee_factor[liquidity] = 0.1%, fee_factor[infrastructure] = 0.05%, fee_factor[maker] = 0.025%.
```
ghost: For MVP and later on (after the system that calculates the fee factor is speced or built).
Will it be possible to change those values via existing governance framework or will we need additional proposal types dedicated to fees?
The proposal types that might be sufficient at the moment are "update network parameters" and "update market".

specs/0029-fees.md
davidsiska-vega: For the first iteration MVP there will need to be a per asset "infrastructure fee pool" and the there will need to be per market "market-maker fee pool". The fee spec only talks about how funds are accumulated in these two types of account. Other specs (market making and staking) will talk about how and when assets from these pools are withdrawn and credited to other parties. 

I think this part makes it clear, but if we need more details let me know.
> infrastructure_fee: transferred to infrastructure fee pool for that asset
> maker_fee: transferred to the relevant party
> liquidity_fee: transferred to market-maker fee pool for that market

specs/0029-fees.md
davidsiska-vega: This part needs an iteration. At the moment I am thinking about a mechanism that would calculate the fee amount needed and add it on top of the margin requirement at the time it goes on the book. That way the margin account always has enough for the fee to be collected. 

If it's a market order it all makes sense. 
If it's a limit order - once it ends up sitting on the book as a passive order the extra amount for fee will be released using the already implemented mechanism for margin. 

This would circumvent the problem of having to cancel a trade once it's been matched. @barnabee @tamlyn10 comment please

specs/0029-fees.md
```diff
@@ -0,0 +1,69 @@
...
+During normal auctions there is no "price maker" both parties are "takers". Each side in a matched trade should contribute 1/2 of the infrastructure_fee + liquidity_fee. Note that this does not include a maker fee. 
+
+
+### Frequent Batch Auctions
+
+Order that entered the book in the current batch are considered aggressive orders. This means that in some cases both sides of a trade will be aggressors in which case the fee calculation for normal auctions applies. Otherwise, the fee calculation for continuous trading applies.
+
+### Position Resolution 
+
+During position resolution all of the parties being liquidated share the total fee for the network order, pro-rated by the size of position. As for fees in other cases, the fee is taken out of the margin account for the liable traders (the insurance pool is not used to top up fees that cannot be paid).
```
ghost: Are fees calculated the same way in this case: `total_fee = infrastructure_fee + maker_fee + liquidity_fee` ?

There is probably no `maker_fee` and the meaning of `liquidity_fee` is not clear to me in this case.
When we close out traders (referred in the core as "NETWORK_CLOSE_OUT_BAD"), Vega network is the aggressor.

What actions are we supposed to take if the margin account for the liquidated traded is not able to cover the fees?

davidsiska-vega: Yes, the fees are calculated the same way. 

In particular the `price maker` (who will receive the `maker_fee`) is the party whose order is being hit by the network when executing the closeout. 

AFAIK Position resolution: 
1) nets off orders against each other so the `maker_fee` is irrelevant (but can still be used it's just a transfer of equivalent amount that goes in both directions)
2) sources the remaining required liquidity from the book - in this case whoever is providing the liquidity should get the `maker_fee` and the delinquent party should be paying.

The `liquidity_fee` is to cover the costs of our market makers who deposit bond to provide liquidity on the market. 

If the margin account is insufficient to cover the fee then the fee (or part of it) is *not* going to get paid. I would suggest that in this case we first pay out the `maker_fee` (or as much as possible), then then `infrastructure_fee` (or as much as possible) and finally the `liquidity_fee`. 

The reason that the insurance pool should *not* be used is that the market makers / infrastructure providers are responsible for specifying a safe market. So they should make sure the risk parameters are such that the margin can cover the losses most of the time. If they keep missing out on fees because the margin is insufficient then they're motivated to set more conservative risk parameters.

specs/0029-fees.md
ghost: We will need to allocate:
 - infrastructure fee account when we create new asset (triggered via a governance proposal);
 - market level account when we create new market (triggered via a governance proposal).

Something we need to be explicit on the implementation tickets.

Perhaps outside of this spec, but will need to be able to validate the amounts on these accounts. I suppose simply tracking fees (via a new field "fee") on each trade plus the transfers should be sufficient.
Is this something we want to expose to users in any way?
Something that would answer the questions of "how much fees did this trade cost me?" and "how much was the total fee for this order?". It is normally an important information. Since we have a non-flat, variable fees framework, this may ease some concerns?
Should we give users fee estimates on order submission?

barnabee:
> Will this fee framework be sufficient for deposits/withdrawals and cross account transfers?
> 
> Or is this something to be considered later on?

There are no fees for deposits/withdrawals on Vega (you’ll pay gas on Ethereum though, as a user). For transfers if/when we implement them we’d need to decide on any fee to be charged and how it’ll be calculated.

specs/0029-fees.md
barnabee: Great questions:

- We should definitely add fees to trades, perhaps as "buyer_fee" and "seller_fee" and expose in APIs
- Users should be able to udnerstand the breakdown of the fee to the three components (perhaps by querying for fee payment transfers by trade ID, if we put enough metadata in the transfer API to easily see the type and associated trade — other solutions welcome to), though this API might be outside the core.
- Fee estimates would be a very good idea and could be done fairly easily if we had an API that exposed the current fee for a trade of size 1.

barnabee: Other than for the last point I don't think we need any new API endpoints, though @tamlyn10 may have other thoughts...

specs/0029-fees.md
barnabee: Existing framework should be used for MVP. For the actual calculations down the line, these will be controlled with a combination of the new staking and market making commitment transactions (when those are built) or through the existing governance proposal types via market / network parameters that are used by the calcs.

specs/0029-fees.md
ghost: > perhaps by querying for fee payment transfers by trade ID, if we put enough metadata in the transfer API to easily see the type and associated trade — other solutions welcome to

Our transfer API is very limited at the moment, consisting exclusively of single gRPC function `TransferResponsesSubscribe`, allowing to subscribe to all transfers performed by the node. No GraphQL or REST.
We might need to expand that API (regardless where the API implementation resides).

barnabee: Yeah, I am ok that a stream (I guess event bus driven?) will be the only API comign out of the core but traders will definitely need to be able to query transfers on a read node. Probably not top priority but it's going to be more and more important as Console develops out and as people use the APIs to really _do_ stuff. Particularly if, as may be the case, it makes the most sense to use these APIs as a way to query fees rather than make specific endpoints for fees (and every other transfer) type.

specs/0029-fees.md
```diff
@@ -25,17 +25,27 @@ _Factors_
...
 
 trade_value_for_fee_purposes:
 * refers to the amount from which we calculate fee, (i.e. for futures, the trade's notional value.)
 * trade_value_for_fee_purposes is defined on the Product and is a function that may take into account other product parameters 
 
 Initially, for futures, the trade_value_for_fee_purposes = notional volume of the trade = size_of_trade * price_of_trade
 
+### Securing Fees ### 
+
+The total_fee required for each trade can be calculated when the order is submitted but before it's placed on the book.
```
barnabee: For each _order_, assuming it is fully filled.

specs/0029-fees.md
```diff
@@ -25,17 +25,27 @@ _Factors_
...
 
 Initially, for futures, the trade_value_for_fee_purposes = notional volume of the trade = size_of_trade * price_of_trade
 
+### Securing Fees ### 
+
+The total_fee required for each trade can be calculated when the order is submitted but before it's placed on the book. 
+At this stage the order may end up being aggresive (market order, limit order or pegged order that's moving) so the party submitting the trade will be liable to pay the fee. 
+Thus the total_fee amount should be trasferred from the general account to the margin account *before* the trade is placed on the order book.  
+If the general account has insufficient balance then the order will be rejected should not make it onto the book (pegged order should be cancelled). 
+For the purposes of securing fees we should treat amends as new orders (so collect the fee on order amend as above).
```
barnabee: Surely we would only want to collect a fee delta, if applicable?

specs/0029-fees.md
davidsiska-vega: Yes. I'll make the edit.

specs/0029-fees.md
davidsiska-vega: Amends: 
- if they change price then we should secure the full fee (on the assumption the whole will trade) because they may become the aggressor due to the price change. 
- if they change volume then it means the order was sitting on the book (price making) and changing the volume doesn't change that so no fee should be needed.  

Am I thinking about this correctly? It's nearly 10pm... see the new commit when it comes through.

specs/0029-fees.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Securing Fees ### 
+
+The total_fee required for each trade can be calculated when the order is submitted but before it's placed on the book. 
+At this stage the order may end up being aggresive (market order, limit order or pegged order that's moving) so the party submitting the trade will be liable to pay the fee. 
+Thus the total_fee amount should be trasferred from the general account to the margin account *before* the trade is placed on the order book.  
+If the general account has insufficient balance then the order will be rejected should not make it onto the book (pegged order should be cancelled). 
+For the purposes of securing fees we should treat amends as new orders (so collect the fee on order amend as above).
+
+If the order does not result in a trade, the order becomes a limit order on the book, and the fee amount will be released from the margin account automatically if the value of the margin account is above the release level (see 0010-margin-orchestration.md, Releasing collateral section).
```
tamlyn10: @davidsiska-vega @barnabee what about orders that move from being passive orders in continuous trading to the auction automatically? They will then need to have a fee amount incurred again?

specs/0029-fees.md
tamlyn10: Regarding this.. it seems like these would be good acceptance criteria. I will add. 

> Great questions:
> 
> * We should definitely add fees to trades, perhaps as "buyer_fee" and "seller_fee" and expose in APIs
> * Users should be able to udnerstand the breakdown of the fee to the three components (perhaps by querying for fee payment transfers by trade ID, if we put enough metadata in the transfer API to easily see the type and associated trade — other solutions welcome to), though this API might be outside the core.
> * Fee estimates would be a very good idea and could be done fairly easily if we had an API that exposed the current fee for a trade of size 1.

tamlyn10: Fee estimates would be good, but I'm not sure about how that gets done? Simply making it be for a trade size of 1 doesn't get around us needing to calculate where the resulting trade would be given current order book?

For limit orders they are trivial since there is a limit price and volume.
For sell market orders the first price in the order book that would be hit will give the upper bound on the fee. 
For buy market orders, the fee will increase depending on how much volume is in the order.

specs/0029-fees.md
tamlyn10: @davidsiska-vega I'm thinking through the amends and agree with what you've written.  There's only one tiny edge case which I **_assume_** isn't the case but thought I'd mention here for absolute explicitness:

@peterbarrow if an amend to increase an order's volume is directly preceded by the order trading.. I presume that means the amend will be rejected - as there is no longer an order to amend. Is that correct? I can't find explicit mention of that in the amends spec, but perhaps it's just obvious!

In that case, I agree with David's dot points!

specs/0029-fees.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Position Resolution 
+
+During position resolution all of the parties being liquidated share the total fee for the network order, pro-rated by the size of position. As for fees in other cases, the fee is taken out of the margin account for the liable traders (the insurance pool is not used to top up fees that cannot be paid).
+
+## Acceptance Criteria
+
+[ ] - Fees are collected during continuous trading and auction modes and distributed to the appropriate accounts, as described above.
+[ ] - Fees are debited from the margin account on any market orders that during continuous trading, the price maker gets the appropriate fee credited to their margin account and the remainder is split between the market making pool and staging pool.
+[ ] - Fees are debited from the margin account on any "aggressive / price taking" limit order that executed during continuous trading, the price maker gets the appropriate fee credited to their margin account and the remainder is split between the market making pool and staging pool.
```
tamlyn10: Only the portion of the limit order that will trade - not the full amount.

specs/0029-fees.md
barnabee: Perhaps better would be to expose the current fee _rate_ and let the client (which has all the info you discuss above as well as order book depth data) decide what to display. Even limit orders aren't totally trivial because if they cross and incur fees they still behave (for fees) like your description of market orders, we just have an absolute upper/lower bound on the fees as well as the probabilistic bounds given by the current state of the book. The price monitoring levels also private a boundary for the fees the order can incur on entry at the extremes. I guess that would allow a best-guess fee based on current book sliippage plus a range (perhaps shown if you hover) to be calculated. (Also, ANY order that persists on the book can incur fees at 50% of the rate unless it is good for the current period / normal trading only, if it trades in an auction.)

specs/0029-fees.md
barnabee: Agree, I was imagininig fees could be incurred any time (technically they can in an auction but I think we just have to assume that existing margin will cope? 🤔)

In fact I think we only want to collect fees in one case of amends: you amend the price to make the order more aggressive. It's the only time your amend could possibly trade, and then only if it's amended such that it crosses with the best price on the other side, which we could also trivially detect. This may be worth doing for (a) future sharding stuff; and (b) reusing amend (if we want to) for peg repricing, as they will likely be guranteed to never end up crossed and requiring fees by design. 

@tamlyn10 if the order trades it's not on the book and I think any amend will be rejected. I'm not sure if this is explicit. (As an aside, some platforms inc. the LSE allow you to increase volume on an order [for up to 60 seconds] _after_ it trades in which case it will be added back to the book. I don't think we should do this.)

specs/0029-fees.md
```diff
@@ -0,0 +1,88 @@
...
+[ ] - Fees are debited from the margin account on any "aggressive / price taking" limit order that executed during continuous trading, the price maker gets the appropriate fee credited to their margin account and the remainder is split between the market making pool and staging pool. 
+[ ] - Fees are debited from the margin account on any "agressive / price taking" pegged order that executed during continuous trading, the price maker gets the appropriate fee credited to their margin account and the remainder is split between the market making pool and staging pool. 
+[ ] - During auctions, each side of a trade is debited 1/2 (infrastucture_fee + liquidity_fee) from their margin account. The infrastucture_fee fee is credited to the staking pool, the liquidity_fee is credited to the market making pool.
+[ ] - During continuous trading, if a trade is matched and the agressor / price taker has insufficient balance in their margin account, then the trade doesn't execute.
+[ ] - During auctions, if either of the two sides has insufficient balance in their margin account, the trade doesn't execute.
```
tamlyn10: presumably orders that don't have the fee amount will be optimistically excluded from the auction's order book? cc @barnabee

specs/0029-fees.md
tamlyn10: Thanks @barnabee - I suggest that we amend the amends spec to include specific mention of this, if other venues allow something different.

And agree on everything else you said!

specs/0029-fees.md
davidsiska-vega: We can also try to secure the fee amount for limit orders on the book from the general account, before they go into the auction order book. For orders where  we failed to get the fee amount we exclude can then not include them in the auction.

specs/0029-fees.md
davidsiska-vega: Yes, will edit the spec.

specs/0029-fees.md
tamlyn10: yeah that makes sense to me as per my other comment somewhere (I'm getting very lost with the way Github does reviews)

specs/0029-fees.md
barnabee: I guess we probably want to allocate more margin at the start of an auction for any counterparties where the delta between actual margin allocated and total required for all positions and orders doesn’t cover the fees? I also assume maybe we want the same when transferring, i.e. like with orders normal margin requirements we should probably only make the request if the actual current balance isn’t high enough to cover the required order margin plus fees on top of the open position’s margin?

barnabee: Which is to say we change the current check to add in a fee component, and re-run this for each counterparty (once per party not once per order) on entering an auction?

specs/0029-fees.md
tamlyn10: Yes I agree with that approach.. client is well placed to offer this value add.

specs/0029-fees.md
tamlyn10: I don't know that I've understood your first message at all haha... So, one question what is "total required" at this point? Initial Margin? 

So, to clarify, are you saying that when we go into an auction, we check each party's positions (open + orders that are going to be transferred) and assume the orders are going to trade in the auction, so add to their margin requirement enough to cover that in fees?

Question about this fees incurred in margin approach:  do the margin requirement levels increase by an amount for fees or is the fee requirement a separate "requirement" and we simply collect it into the margin account? I'm assuming the latter but got confused by your description.

davidsiska-vega: @tamlyn10 From the way the spec is written: the required fee amount is *separate* from margin amounts but gets collected into the margin account for each order that goes on the book (or again if we're switching to auction). 

The fee spec doesn't need to talk about releasing it, this will happen using the usual margin account mechanism of when we go above release level. Please edit anything that needs editing in the spec to "unconfuse" it.

tamlyn10: Thanks @davidsiska-vega you're correct, I hadn't absorbed that bit of the spec adequately - apologies. Tho' I don't think I asked about releasing?

So, about the move from limit order on the book to be auction orders.. are we saying that if the current-deployed-margin is > search-level + fees then we just move the orders over?

davidsiska-vega: @tamlyn10 

> about the move from limit order on the book to be auction orders.. are we saying that if the current-deployed-margin is > search-level + fees then we just move the orders over

This would work fine IMHO. 

Though I thought we will debit the general account for the fee amount and transfer to the margin account. And if we're doing that the question is whether on the margin side we want to go to general account and top up to the level of initial margin on the orders (but that's for the auction spec).

barnabee: > So, about the move from limit order on the book to be auction orders.. are we saying that if the current-deployed-margin is > search-level + fees then we just move the orders over?

They are still limit orders on the same order book, so unless their type prohibits them being in an auction period (then they’re expired/cancelled or parked depending on the specifics) so when the order book transitions into an auction period they will be there by default unless we specify something different. 

That is, we should spec that we are going to check each counterparty’s “worst case” margin plus fee requirements for their open position plus active orders that and cancel the orders if they are underfunded.

specs/0029-fees.md
barnabee: GitHub PRs and reviews are a mystery to me 🤷‍♂️ 

By “optimistically exclude”, do you mean cancel? Don’t think we can only exclude from the auction and put them back after because they might trade aggressively on re-entry to the book when normal trading resumes and then they need the fee again.

tamlyn10: Um, yeah I think we're having this conversation in 2 places, now! Let's lock it down tomorrow evening as per David's meeting!

PR #288, opened 2020-05-07
Add incomplete Market Lifecycle spec

edd:
This is a new spec based on a conversation in slack: https://vegaprotocol.slack.com/archives/CEE07A084/p1588844726042400?thread_ts=1588810853.025300&channel=CEE07A084&message_ts=1588844726.042400

> @barnabee: I do not think close outs should be in the auction. If the auction ends at a price where those positions no longer need to be closed there is no point stressing the market’s liquidity by closing them. This is reasonably likely if the trigger for the auction is a series of large/reactionary price moves. 

> I therefore propose:
> • no closeout orders in auctions 
> • no change to close out spec 
> • after auction uncrosses positions are marked to market as normal and closeout trades created if needed at that point


edd:
@barnabee You said in the thread that auctions will not close until there is sufficient liquidity on the book. Please clarify ‘sufficient’, as in what I have written in the spec file in this PR it looks like this could get in to an endless cycle of insufficient liquidity -> close outs -> auction -> continuous trading -> closeouts insufficient liquidity -> auction mode. I think a definition of sufficient will clear that up.

Also we need to spec auctions.

barnabee:
Sufficient will be defined in the liquidity monitoring spec, initially auctions will only be used for opening and frequent batch auction markets. We did a session on auctions, I thought there was a spec!

Both price and liquidity monitoring will be able to put markets into auctions and define when they come out of them, but haven’t been speeded yet and it was assumed at the time of writing the auction spec that auctions would be in well ahead of those things but I guess they’ve not been started yet?

edd:
@barnabee https://github.com/vegaprotocol/product/pull/256 - please review and comment on the auction spec PR.

barnabee:
I will try to have a stab at this over the weekend.

edd:
I will merge this one after the auction spec is merged so that I can properly link the Auction and market open to the relevant section.

barnabee:
@edd I would like to make some updates on this before it is merged, even if very brief, there is a specific design here and it's probably worth the 20 mins to at least get the key points down

barnabee:
Note to self: this needs to cover market opening (initialisation) phase for new markets

tamlyn10:
Barney has posted a WIP link for review.

jeremyletang:
@barnabee @tamlyn10 Should the opening auction duration be setup through the proposal, or should be a network parameter?

tamlyn10:
> @barnabee @tamlyn10 Should the opening auction duration be setup through the proposal, or should be a network parameter?

I believe this should be a network parameter

barnabee:
Yeah, network parameter, maybe more complex on future but simple = good for now.

edd:
I'm merging this and then will probably immediately make changes. It's better than nothing, and mostly good.

PR #289, opened 2020-05-07
WIP: data sourcing (oracle) v1 spec

barnabee:
PR to discuss and complete a data sourcing (oracle) v1 / MVP spec

Prior discussion also here: https://github.com/vegaprotocol/product/issues/177


barnabee:
@C0deMunk33 if you're going to pick this up at some point, it'd probably be worth us having a quick Zoom call, feel free to schedule one as and when you want to get started :)

barnabee:
@C0deMunk33 lets have the Zoom call I suggested above. I think the proposal is unnecessarily narrowly focused on settlement and it'd be good to discuss.

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+### API Data Source
+A data payload that provably came from (and signed by) a "Trusted API".
+Many companies/brands who act as the authority in their industry proudly defend their reputation. 
+This defence extends to their online presence. 
+
+SSL certificates sign every https transaction to and from public APIs. 
+This signature acts as the verifiable signer for this class of data source.
+
+While this is a very centralized type of data source, it is the only way to get the correct and official reports from the most trusted authorities on a given dataset.
+Combined with composite data sources and the contest mechanism, we believe that the API data source is a valuable tool in the Vega arsenal.
```
edd, line 186: I don't quite see how this works in action. Let's say we use IMDB as an API oracle for film gross. There are three ways to get this data:

1. IMDB write the code to publish to Vega as an Oracle
2. Vega nodes reach out to a known endpoint and fetch the data
3. Someone writes a 'connector' that is a trusted source for IMDB data

These three all have flaws, but I'm assuming we're going with 3:
1. A simple HTTPS API call from them to a Vega endpoint wouldn't inherently send a public key. It probably could provide that, but where's the stable endpoint for them to push to hosted? Or are they running a Vega node.
2. This is liable to break as APIs change. Or might be rate limited, or require credentials. I'd be surprised if we were doing this so I won't spend more time here. However, it does mean we can get the certificate details on request, which is also true for....
3. A bridge makes sense and can be operated by anyone. An event is published and has the key that served the data, but we have no way to verify that the data is what the endpoint published. To fix that we'd probably need multiple nodes to fetch the data, match the data and keys. Which could be done as a composite oracle, discarding any weird results and non-matching keys. But I haven't seen that mentioned.

So is it number 3? Should we document this?

edd: cc @barnabee

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+message KeyValuePair {
+  string key = 1;
+  string value = 2;
+}
+
+message OracleEvent {
+  string[] signers;
+  repeated KeyValuePair payload;
+  uint timestamp;
+  oneof oracleEventMetadata {
```
edd, line 132: When will we be able to settle  on a reasonably stable set of Oracle sources for this structure? I can imagine this will go through an early period of being pretty static, then expanding, and then settling down - it's not like new oracle services pop up every day. But for every new On Chain oracle we add we're going to need to do a release. We may also need to have specific versions numbers in the oracle types (e.g. if Chainlink change their data shape we might have to adapt too) and store those in the proto, as we'll need to be able to process both  Chainlink v1 and v2 messages in the meatadata processor forever. Probably..

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+Oracles are simply providers of information. 
+These data sources are the only way for the Vega network to extend its reach into other blockchains or the real world.
+As such, this places a dangerous amount of power in the hands of potentially centralized or compromised parties. 
+This is widely know as The Oracle Problem.
+Despite the ability to both use diverse oracles and also compose them, there will always be the possibility of a rogue oracle or bug in the latest DeFi protocol.    
+To mitigate these risks, Vega has developed a number of safeguards to ensure the greatest flexibility of the network to deal with threats as they occur.
+
+### Changing Oracle Sources
+Oracles may get compromised, shut down, change location/certificates, or otherwise become too risky to use alone before a market has expired. 
+To mitigate these risks, a governance can change oracle parameters. To do this, see [the Governance specification](https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md).
```
edd, line 298: What's an oracle parameter?

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+ 
+Unlike a typical Vega governance vote, this type of vote is open to all participants in the market, provided they have a balance above a threshold of value (configurable).
```
edd, line 300: Is this a network parameter? A market configuration thing?

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+When the oracle emits a value, Vega waits the required time and if no valid dispute occurs the events that rely on this oracle (like settlement) are executed.
+If however, the threshold for dispute is met, a governance process of accepting proposals occurs to decide the final value.
+
+Unlike the rest of Vega, to submit a complaint to contest a data source result a user must be involved in the market being contested.
+Both market makers and market participants may contest the results of an oracle if they have more than a threshold of the market's value. 
+This limitation is to cut down on spamming and sybil attacks.
+
+To contest the process, a halt market vote will be submitted to governance. 
+Provided the user has enough stake in the market, the vote will be put out to governance.
+Once a halt market vote has been successful, the market is locked down and all assets frozen until a `release market` command is voted through governance.
```
edd, line 315: We're probably going to need a Halt Market spec

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+These votes will be and weighted by value in the market and then weighted against market maker votes and network governance voters.  
+
+### Contesting Data Source Reports
+Vega provides a mechanism to dispute reported oracle prices/results.
+
+When the oracle emits a value, Vega waits the required time and if no valid dispute occurs the events that rely on this oracle (like settlement) are executed.
+If however, the threshold for dispute is met, a governance process of accepting proposals occurs to decide the final value.
+
+Unlike the rest of Vega, to submit a complaint to contest a data source result a user must be involved in the market being contested.
+Both market makers and market participants may contest the results of an oracle if they have more than a threshold of the market's value.
```
edd, line 310: What does a complaint look like in data terms? Is it a complaint against the value only? If it's an API source can we complain if the public key associated with the event looks wrong? 

Is there any specific incentive for complaining and being voted correct? Or do we just assume that only people with something at stake will do it?

Is there any punishment for complaining about results?

Are oracles removed from the system ever? What if every result they publish is complained and those complaints are voted for?

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+If however, the threshold for dispute is met, a governance process of accepting proposals occurs to decide the final value.
+
+Unlike the rest of Vega, to submit a complaint to contest a data source result a user must be involved in the market being contested.
+Both market makers and market participants may contest the results of an oracle if they have more than a threshold of the market's value. 
+This limitation is to cut down on spamming and sybil attacks.
+
+To contest the process, a halt market vote will be submitted to governance. 
+Provided the user has enough stake in the market, the vote will be put out to governance.
+Once a halt market vote has been successful, the market is locked down and all assets frozen until a `release market` command is voted through governance.
+While halted, the oracle can be changed by another governance vote. Other market parameters can be updated during this time as well.
```
edd, line 316: What other market parameters?

specs/0000-data-sourcing.md
```diff
@@ -0,0 +1,335 @@
...
+
+The contest process will start automatically if, upon expiry, the oracle is unavailable.
```
edd, line 318: What happens if a complaint passes and we are left without a data source for a market as no one else provides that data? Is the money locked forever?  Is any trading rolled back or is just the final settlement at expiry changes? Is there any downside for the oracle?

edd (commented):
Complaints feedback

tamlyn10:
@C0deMunk33 this is waiting on you to review comments

specs/0000-data-sourcing.md
C0deMunk33: I don't know to be honest. I was like, hey guys, shouldn't market participants be able to vote? The answer came back from everyone "yes"... 

Got any insight into how we would pull off letting market participants vote on things like changing oracles @jeremyletang ? I know it doesn't exist yet, but any sort of input would probably be helpful.

specs/0000-data-sourcing.md
C0deMunk33: " it's not like new oracle services pop up every day"
welcome to DeFi, the De stands for degenerate.

I don't know when, but if you want to task me with locking down a small set, I can do that. You are correct that every support for a new oracle will come with a bespoke connector. 

Oracle versioning noted. To keep with the pattern, everything will be assumed v1 until we make a v2 in which case it would be named v2: chainlinkV2 since for practical purposes there's no connection between v1 and v2 of immutable protocols.

specs/0000-data-sourcing.md
C0deMunk33: 1. not worried about a public key, more looking at 2: the cert details. 

3 would be fine, I can make a smart contract that is managed and configured by the market maker to provide details on how to parse the responses. 

As for composite, that is covered further down. Basically we can sacrifice a bit of security by having a) a contest period/method and b) composit/changable sources.

@barnabee should totally comment here.

specs/0000-data-sourcing.md
C0deMunk33: Oracle parameters for a market... probably should reword that.

specs/0000-data-sourcing.md
C0deMunk33: All great questions. I'll give this some more thoughts, in the meantime can you braindump a couple sentences please @barnabee?

specs/0000-data-sourcing.md
C0deMunk33: https://github.com/vegaprotocol/product/issues/346

specs/0000-data-sourcing.md
C0deMunk33: Dunno, TODO: dig into market definition spec.

specs/0000-data-sourcing.md
C0deMunk33: Seems like we would need to incentivize providing the data. I think that a locked market will motivate the users to find a way to resolve the market.

This is a really good question and I'd like comments from @barnabee  and/or @tamlyn10  here.

C0deMunk33 (commented):
@tamlyn10 I've made my comments. I've got a couple tasks here, but there's a few comments I tagged @barnabee with to get his feedback and one where I tagged you.

Thanks!

specs/0000-data-sourcing.md
barnabee: > You are correct that every support for a new oracle will come with a bespoke connector.

I don't think this works. We should work out some standard formats, i.e. consuming a source that emits JSON plus a path to the value and use those. We definitely can't do new connectors every time. Even if not every oracle, say on Ethereum, emits a format we like an Eth smart contract to consume the event and emit JSON (or whatever) ought to be easy enough to deploy.

specs/0000-data-sourcing.md
barnabee: Honestly I don't see why not do 2? 

It's easier to do a market change governance request to change the API URL than to deploy a whole new connector, and people ought to consume versioned APIs, and perhaps multiple too via composite oracles for redundancy.

specs/0000-data-sourcing.md
barnabee: "Data source references" where used will be provided as part of the market framework, can be changed through market change governance.

specs/0000-data-sourcing.md
barnabee: I think we should define it as a market governance vote initially with same voters as all others. We will address different types of weighting for various governnace votes later and this spec is the wrong place for that discussion. It'll be a whole _thing_, but maybe not before launch 🤷

specs/0000-data-sourcing.md
barnabee: I might well suggest we leave disputes out of the MVP of data sources. In general the data emitted from a source which is "contestable" will be like a governance vote except:

* default results is pass/yes even if no votes
* voting period is from receipt of the data for a period set as a market param (there may be a network param specifying a minimum)
* we still need to define the mechanism for getting at an answer
* we still need to decide what is the incentive to complain when it's wrong and not to otherwise
* we need to work on the security of how to prevent it being too cheap to complain to rig your profit (hence why I'd leave out of the MVP, this needs some thought, it may be the overall token holders need to vote above a certain threshold for the dispute to carry once agreed in principle by a vote of market participants, thus disincentivising them from even trying to manipulate the result, as it'll get shot down by the full network governance step)

specs/0000-data-sourcing.md
barnabee: Closing and suspending markets are out of scope for this spec and should be covered elsewhere.

specs/0000-data-sourcing.md
barnabee: Closing and suspending markets are out of scope for this spec and should be covered elsewhere.

specs/0000-data-sourcing.md
barnabee: Basically the dispute process would have to come up with and pass a vote for a settlement result. It may be a "neautral" result that someone nullifies P&L and returns money pro-rata or something in extreme cases where no better outcome is possible.

barnabee: (Oracle downside beyond community/reputational off chain repercussions will not be possible until duplex oracles which we don't even have a rough design for and won't be added for some time, perhaps not ever.)

barnabee (commented):
Comments. I can't pick request changes for some reaosn 🤷‍♂️

specs/0000-data-sourcing.md
C0deMunk33: Oh I see, you want to move the bespoke connector thing to a smart contract rather than in core... that's doable.

tamlyn10:
Danny to create a meeting for going through this in the next week.

barnabee:
Closing, this is now here: https://github.com/vegaprotocol/product/pull/450

PR #300, opened 2020-06-10
Market depth creation and distribution

### LINKED ISSUE #299: Add spec for market depth generation and distribution
We currently generate market depth data by listening to the orders placed on the orderBuffer and using that information to build up a market depth order book. This fails when we have amends because there is not enough information in the orders to be able to update the book correctly. 

We have discussed a better way to handle this using the new event bus system. This spec will highlight the features of the new system

peterbarrow:
New spec to define how we will handle market depth data collection and distribution

### DIFF
--- a/0000-market-depth-calculation.md
+++ b/0000-market-depth-calculation.md
@@ -11,17 +11,17 @@ Specification PR: https://gitlab.com/vega-protocol/product/pull/300
 
 
 # Summary
-The market depth builder receives a stream of events from the core from which it builds up an market depth structure for each given market. This structure is used to provide a market depth stream and full market depth dump to any clients which has requested/subscribed to it. 
+The market depth builder receives a stream of events from the core from which it builds up a market depth structure for each given market. This structure is used to provide a market depth stream and full market depth dump to any clients which have requested/subscribed to it. 
 
 # Guide-level explanation
-When the core processes orders, cancels, amends or auction states, it generates one or more events which are sent out via the event-bus. 
+When the core processes an external action such as an order, cancel, amend or changing auction state, it generates one or more events which are sent out via the event-bus. 
 
 The market depth module subscribes to all the event types in the market-event and order-event streams. From the events received from these event streams, we build up a market depth structure for each market which will be a representation of the orderbook stored in the core. 
 
 Clients connect to a vega node and subscribe to a MarketDepth stream via gRPC or GraphQL for a specific market. This stream will contain all the updates occuring to the market depth structure and will contain a sequence number with each update. The client then makes a request to get a snapshot dump of the market depth state. This dump will contain the full market depth structure at the current time along with a sequence number for the current state. The client will then apply all updates that have a sequence number higher than the original dump to the market depth structure to keep it up to date.
 
 
-# Reference-level explanation (This is the main portion of the specification. Break it up as required.)
+# Reference-level explanation
 
 The market depth builder needs to receive enough information from the core to be able to build the market depth structure to have exactly the same price and volume details as the order book stored in the matching-engine. Therefore any change to the order book in the matching engine must generate one or more events that can be used to update the market depth order book in the same way.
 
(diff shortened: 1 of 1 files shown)


0000-market-depth-calculation.md
```diff
@@ -0,0 +1,129 @@
...
+# Acceptance Criteria
+- The market depth builder must be able to handle all available order types
+- The construction of the market depth structure must not impact the core performance.
+- Entering and leaving auctions must be handled correctly
+- All subscribed clients must receive all the data neccessary to build their own view of the market depth
+- There should be no batching in sending out market depth updates
+
+
+# Summary
+The market depth builder receives a stream of events from the core from which it builds up an market depth structure for each given market. This structure is used to provide a market depth stream and full market depth dump to any clients which has requested/subscribed to it.
```
edd: Typo: client -> clients

0000-market-depth-calculation.md
```diff
@@ -0,0 +1,129 @@
...
+- Entering and leaving auctions must be handled correctly
+- All subscribed clients must receive all the data neccessary to build their own view of the market depth
+- There should be no batching in sending out market depth updates
+
+
+# Summary
+The market depth builder receives a stream of events from the core from which it builds up an market depth structure for each given market. This structure is used to provide a market depth stream and full market depth dump to any clients which has requested/subscribed to it. 
+
+# Guide-level explanation
+When the core processes orders, cancels, amends or auction states, it generates one or more events which are sent out via the event-bus.
```
edd: This *feels* like a complete list. Is it? If not, indicate it with ‘e.g.’ or something, or otherwise... it’s probably fine.

0000-market-depth-calculation.md
```diff
@@ -0,0 +1,129 @@
...
+    type PriceLevel struct {
+        Price       int64
+        Volume      uint64
+        NumOfOrders uint64
+		Side        bool
+    }
+
+An update message is:
+
+    type UpdateMarketDepth struct {
```
edd, line 88: Missing sequence number [blocker if true]

0000-market-depth-calculation.md
```diff
@@ -0,0 +1,129 @@
...
+        Send updates to all subscribers for price levels that changed
+	End
+
+
+The client side will perform the following steps to build and keep an up to date market depth structure
+
+	Subscribe to market depth updates
+	Request current market depth structure
+	Forever
+		Receive market depth update
```
edd: Indentation might need to be fixed. Also might not. Up to you

0000-market-depth-calculation.md
```diff
@@ -0,0 +1,129 @@
...
+
+* Create a new order in an empty order book, verify MD
+* Cancel an existing order, verify MD
+* Amend an existing order for both price and quantity, verify MD
+* Cancel an order and replace it with the same order values, verify the MD sees an update
+* Enter into auction
+* Leave auction
+* Do nothing for many minutes, make sure subscribers do not timeout/fail
+* Send a large spike of order updates, make sure the system does not stall
+
```
edd, line 140: Add: Sequence number increments for each emitted book update
Add: Updates that are not received/processed by the client are not buffered on their behalf

0000-market-depth-calculation.md
```diff
@@ -0,0 +1,129 @@
...
+# Guide-level explanation
+When the core processes orders, cancels, amends or auction states, it generates one or more events which are sent out via the event-bus. 
+
+The market depth module subscribes to all the event types in the market-event and order-event streams. From the events received from these event streams, we build up a market depth structure for each market which will be a representation of the orderbook stored in the core. 
+
+Clients connect to a vega node and subscribe to a MarketDepth stream via gRPC or GraphQL for a specific market. This stream will contain all the updates occuring to the market depth structure and will contain a sequence number with each update. The client then makes a request to get a snapshot dump of the market depth state. This dump will contain the full market depth structure at the current time along with a sequence number for the current state. The client will then apply all updates that have a sequence number higher than the original dump to the market depth structure to keep it up to date.
```
edd, line 21: How is the sequence number generated? When is it updated?

I’m guessing ‘It increases by one for each event’, but it’s worth clarifying.

edd (commented):
Some niggles and one change requested - I think the sequence number is missing in the code sample. Otherwise this is a good summary and I understand how this is intended to work 👍 

We might end up moving this to another folder as it sits on the boundary of ‘Core functionality’ and ‘implementation detail’, I think. It’s both, and that’s fine.

EVODelavega (approved):

0000-market-depth-calculation.md
```diff
@@ -0,0 +1,132 @@
+Feature name: feature-name
```
mattrussell36: feature name missing

edd:
@peterbarrow what do you think is left to do before we can merge this spec?

peterbarrow:
One issue we need to resolve is the building up of the cumulative volume data. While it is easy to do that in the Vega node side. It does not make much sense if we are going to send out only changed price levels to the client. For a 10 level side of the book if we add a new order at level 5, we need to recalculate the cumulative volume at level 6-10. We would then have to send those levels out to the client. If we decide to let the client handle the cumulative volumes we can just send out the update to level 5 and they can perform the update themselves. It's a trade off between how much data we want to emit from the node against the amount of processing the client will perform.

peterbarrow:
Taking a look at the BitMEX streaming price data, they do not have the cumulative volume supplied so I think it makes sense to expect the client to build up this information if they need it. The client is the only party to know how many rows of the order book they need so it would be able to calculate just the information it needs and no more.

peterbarrow:
@mattrussell36 @campbellssource @edd Can I get some feedback from you about the cumulative volume please.
Would you be happy to handle it all on your side?

mattrussell36 (approved):

edd:
@peterbarrow We'll build it client side

PR #301, opened 2020-06-11
Cancel orders spec

### LINKED ISSUE #134: Write spec for cancelling orders
In GitLab by @tamlyn10 on Sep 16, 2019, 15:25

As per Slack message: https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1568643724025700?thread_ts=1568642627.016800&cid=CAHA5EX0F

For OrderCancellation, it should probably be one of order ID, market ID, party ID or potentially always a party ID plus optionally either order ID or market ID, because for cancellation we should support:
1) cancel specific order by ID
2) cancel all orders I have on a market
3) cancel all my orders on the network
I’m not sure if this is what’s implemented atm

peterbarrow:
Creation of the spec to define how the different types of cancels are performed in the core.


jeremyletang:
Looks pretty good to me as it is.
Just thinking of one thing, the margins are not re-calculated when an order is cancel, maybe it's worth mentionning in the specs?
Also they are not re-calculated as of now, where we can cancel only a single order at the time, would this be the same thing with the 2 others way of cancelling orders?

0000-cancel-orders.md
```diff
@@ -0,0 +1,83 @@
+Feature name: feature-name
+Start date: YYYY-MM-DD
+Specification PR: https://gitlab.com/vega-protocol/product/pull/301
+
+# Acceptance Criteria
+
+- An order cancelled by orderID+marketID+partyID will be removed from the order book and an order update message will be emitted
+- All orders for a given partyID will be removed from a single market if a cancel all party orders per market message is sent
+- All orders for a given party across all marekts will be removed from the vega system when a cancel all orders message is sent
```
edd: Typo: marekts

0000-cancel-orders.md
```diff
@@ -0,0 +1,83 @@
...
+- Orders which are not currently on the orderbook but are being held `offline` due to being in auction should also be affected by cancels.
+
```
edd: Can I mix and match these? Can I submit a partyID and an orderID and have an order cancelled whatever market it is on? (Not covered below either.) (I guess because it’s not in a list even further down, it’s not supported. This should be explicit.

0000-cancel-orders.md
```diff
@@ -0,0 +1,83 @@
+Feature name: feature-name
+Start date: YYYY-MM-DD
+Specification PR: https://gitlab.com/vega-protocol/product/pull/301
+
+# Acceptance Criteria
+
+- An order cancelled by orderID+marketID+partyID will be removed from the order book and an order update message will be emitted
+- All orders for a given partyID will be removed from a single market if a cancel all party orders per market message is sent
+- All orders for a given party across all marekts will be removed from the vega system when a cancel all orders message is sent
+- Orders which are not currently on the orderbook but are being held `offline` due to being in auction should also be affected by cancels.
```
edd: + - A cancellation for a party that does not match the party on the order will be rejected
+ - A cancellation without a party ID will be rejected

edd (commented):
Found a pretty trivial edge case. Well, really, just something that isn’t explicitly forbidden. Also added some acceptance criteria, but they might be better off as tests

0000-cancel-orders.md
peterbarrow: It's not part of the current list but if you need it we can add it?

0000-cancel-orders.md
peterbarrow: Nice, added.

0000-cancel-orders.md
```diff
@@ -0,0 +1,86 @@
...
+- Cancel by partyID and marketID - This removes all the orders for a given party in the given market.
+- Cancel by partyID - This removes every order for that given party in the vega system. 
+
+Parked orders are affected as part of direct cancels or cancels that sweep over a market/system.
+
+
+# Reference-level explanation
+
+## Cancel by orderID, partyID and marketID
+The orderbook is looked up using the marketID and then we issue a cancel on that orderbook. Validation takes place to make sure the partyID supplied matches the partyID stored with the order. At most a single order will be cancelled using this method. As the order price is not supplied in the cancel and the order book stores all the orders via price level, the market has a separate map linking all orderIDs to their position in the order book. This allows cancellations to be performed efficiently.
```
barnabee: When we say "party ID supplied" we should be explicit that it is the party ID in each case is owned by the signer of the transaction

0000-cancel-orders.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Cancel by partyID and marketID
+The orderbook is looked up using the marketID. We then scan through all the orders on that orderbook looking for orders which match the partyID. Depending on the number of orders in the orderbook, this might be an inefficient operation. Each matching order is cancelled.
```
barnabee: I would expect that a lookup table of party -> orders is needed based on what I've seen elsewhere.

barnabee (changes requested):
Ok to merge when comments addressed. No need to re-review on my part.

tamlyn10:
> Just thinking of one thing, the margins are not re-calculated when an order is cancel, maybe it's worth mentionning in the specs?
> Also they are not re-calculated as of now, where we can cancel only a single order at the time, would this be the same thing with the 2 others way of cancelling orders?

I think this is a really important point from @jeremyletang . If we batch cancel orders, and orders are all we have, then the margin requirement should immediately drop.

0000-cancel-orders.md
peterbarrow: I have added a statement to make this explicit.

0000-cancel-orders.md
peterbarrow: I have updated the spec to mention the per market partyID->order lookup table.

edd (approved):

PR #312, opened 2020-06-30
Added wash trade order status table

### LINKED ISSUE #307: Updating order status specs for wash trading
There is nothing in specs to cover the status codes for wash trading.
We need to add the possible cases for this

peterbarrow:
At any time during the processing of an order, we might come across a wash trade. This will result in the aggressive trade being rejected. I have updated the order status spec to show the new order state when this situation occurs.


campbellssource (approved):

specs/0024-order-status.md
```diff
@@ -57,3 +57,11 @@ Note: The last row in the table below is added for clarity rather than being a l
 |      GTT      | Partial |    No   |         No        |        Yes        |      Stopped     |
 |      GTT      |   Yes   |    No   |         No        |         No        |      Filled      |
 |      GTT      |   Yes   |   Yes   |         No        |         No        |      Filled      |
+
+## Wash trading
+If an order would be filled or partially filled with an existing order from the same traderID, the order is rejected. Any existing fills that happen before the wash trade is identified will be kept.
```
campbellssource, line 63: Could you detail what the “rejected reason” would be. If you haven’t already checked fix. e.g “wash trade”.

campbellssource (dismissed):

specs/0024-order-status.md
peterbarrow: I've added the reason to the table, is that what you mean?

peterbarrow: Or do you mean this: https://www.onixs.biz/fix-dictionary/5.0/tagNum_103.html

campbellssource: At the moment in console when an order is rejected you hover over the Rejected text in the order table and you get a tool tip that says "Insufficient collateral" or something. I'm not sure where these reasons are documented (they are not here at the moment).
Screenshot 2020-07-09 at 12 01 43

campbellssource: Interesting that There isn't a rejected reason for wash trades on Fix. Perhaps it is hidden behind one of the other reasons.

edd:
@campbellssource is this good to merge now?

campbellssource:
> @campbellssource is this good to merge now?

I've resolved my comment as there isn't currently any talk of rejection reasons in specs. I was expecting there to be a list like this https://github.com/vegaprotocol/vega/blob/develop/gateway/graphql/schema.graphql#L1107 in specs

PR #322, opened 2020-07-14
Add liquidity monitoring spec

### LINKED ISSUE #309: Add liquidity monitoring spec
Write a sepc for liquidity monitoring

witgaw:
WIP

Some of the things to consider:
- What is the trigger for entering the auction mode?
- What is the trigger for leaving the auction mode?
- Are we counting the parked orders not market as good for auction towards the liquidity (e.g. MMs pegged orders)?

### DIFF
--- a/specs/0000-liquidity-monitoring.md
+++ b/specs/0000-liquidity-monitoring.md
@@ -1,0 +1,21 @@
+Feature name: liquidity-monitoring
+Start date: 2020-07-14
+Specification PR: 
+
+# Acceptance Criteria
+
+- [ ]
+- [ ]
+- [ ]
+
+# Summary
+
+WIP
+
+# Guide-level explanation
+
+
+# Reference-level explanation
+
+# Test cases
+
--- a/specs/0000-liquidity-monitoring.md
+++ b/specs/0000-liquidity-monitoring.md
@@ -1,6 +1,6 @@
 Feature name: liquidity-monitoring
 Start date: 2020-07-14
-Specification PR: 
+Specification PR: https://github.com/vegaprotocol/product/pull/322
 
 # Acceptance Criteria
 
@@ -14,6 +14,28 @@ WIP
 
 # Guide-level explanation
 
+## Measuring the liquidity
+
+### Liquidity required
+
+Lower bound on liquidity requirement measured as a constant multiple of Open Interest [Do we want to apply some smoothing e.g.: average or max open interest over a rolling window? If so do we want to apply some operation to the measure of liquidity supplied?]
+
+### Liquidity supplied
+
+Committed & supplied Siskas [What are these, do we want to count just the MM-posted liquidity or overall?]
+
+## Trigger for entering an auction
+
+When liquidity supplied < liquidity required
+
+## Trigger for exiting the auction
+
+When liquidity supplied > c * liquidity required,
+where c is a constant greater than 1 (part of the market configuration), to reduce the chance of another auction getting triggered soon after.
+
+## What happens during the auction?
+
+[Is there any trading?]
 
 # Reference-level explanation
 
--- a/specs/0000-liquidity-monitoring.md
+++ b/specs/0000-liquidity-monitoring.md
@@ -1,43 +1,52 @@
-Feature name: liquidity-monitoring
-Start date: 2020-07-14
+Feature name: liquidity-monitoring\
+Start date: 2020-07-14\
 Specification PR: https://github.com/vegaprotocol/product/pull/322
 
-# Acceptance Criteria
+## Summary
 
-- [ ]
-- [ ]
-- [ ]
+Liquidity in the market is not only a desirable feature from a trader's point of view, but also an important consideration from the risk-management standpoint. Position of a distressed trader can only be liquidated if there's enough volume on the order book to offload it, otherwise a potentially insolvent party remains part of the market.
 
-# Summary
+Similarly to [price monitoring](0032-price-monitoring.md), we need to be able to detect when the market liquidity drops below the desired level, launch a liquidity auction and terminate it when the market liquidity level is back at a sufficiently high level.
 
-WIP
+## Liquidity auction network parameters
 
-# Guide-level explanation
+**c<sub>1</sub>** - constant multiple of open interest estimate triggering the commencement of liquidity auction.
 
-## Measuring the liquidity
+**c<sub>2</sub>** - constant multiple of open interest estimate triggering the termination of liquidity auction (such that c<sub>2</sub> > c<sub>1</sub>).
 
-### Liquidity required
+**p** - period of time used for the open interest estimate.
 
-Lower bound on liquidity requirement measured as a constant multiple of Open Interest [Do we want to apply some smoothing e.g.: average or max open interest over a rolling window? If so do we want to apply some operation to the measure of liquidity supplied?]
+## Glossary
 
-### Liquidity supplied
+**Open interest** - volume of all long positions.
 
-Committed & supplied Siskas [What are these, do we want to count just the MM-posted liquidity or overall?]
+**Open interest estimate** - maximum open interest in the market captured between the current time t and t-[p](#Liquidity-auction-network-parameters)
+
+## Required liquidity
+
+Minimum liquidity requirement measured as a constant multiple c<sub>1</sub> (market parameter) of max Open Interest over period p
+
+## Supplied liquidity
+
+Count (probability of trading weighted) liquidity committed via market making orders. Please see the [liquidity measurement spec](0034-prob-weighted-liquidity-measure.ipynb) for details.
 
 ## Trigger for entering an auction
 
-When liquidity supplied < liquidity required
+When liquidity supplied < c<sub>1</sub> * [open interest estimate](#Glossary).
+
+Similarly to [price monitoring](0032-price-monitoring.md), the auction should be triggered pre-emptively. That is, the transaction that would have triggered the liquidity auction should be re-processed once auction mode is on. Please note that this transaction may be cancelled if it's type is not valid for auction, however even in that case the market should still go into the auction mode to send a signal that it requires more liquidity.
 
 ## Trigger for exiting the auction
 
-When liquidity supplied > c * liquidity required,
-where c is a constant greater than 1 (part of the market configuration), to reduce the chance of another auction getting triggered soon after.
+When liquidity supplied ≥ c_2 * [open interest estimate](#Glossary), \
+where c_2 > c_1, to reduce the chance of another auction getting triggered soon after.
+
+During the liquidity monitoring auction new or existing market makers can commit more liquidity through the special market making order type and enable this by posting enough margin - see market making mechanics spec (WIP) for details. These need to be monitored to see if auction mode can be exit.
 
 ## What happens during the auction?
 
-[Is there any trading?]
+The auction proceeds as usual. Please see the auction spec for details.
 
-# Reference-level explanation
+## Test cases
 
-# Test cases
-
+* Market with no market makers enters liquidity auction immediately.
--- a/specs/0000-liquidity-monitoring.md
+++ b/specs/0000-liquidity-monitoring.md
@@ -10,17 +10,17 @@ Similarly to [price monitoring](0032-price-monitoring.md), we need to be able to
 
 ## Liquidity auction network parameters
 
-**c<sub>1</sub>** - constant multiple of open interest estimate triggering the commencement of liquidity auction.
+**c<sub>1</sub>** - constant multiple of [liquidity demand estimate](#Glossary) triggering the commencement of liquidity auction.
 
-**c<sub>2</sub>** - constant multiple of open interest estimate triggering the termination of liquidity auction (such that c<sub>2</sub> > c<sub>1</sub>).
+**c<sub>2</sub>** - constant multiple of [liquidity demand estimate](#Glossary) triggering the termination of liquidity auction (such that c<sub>2</sub> > c<sub>1</sub>).
 
-**p** - period of time used for the open interest estimate.
+**p** - period of time used for the [liquidity demand estimate](#Glossary).
 
 ## Glossary
 
 **Open interest** - volume of all long positions.
 
-**Open interest estimate** - maximum open interest in the market captured between the current time t and t-[p](#Liquidity-auction-network-parameters)
+**Liquidity demand estimate** - use maximum open interest in the market captured between the current time t and t-[p](#Liquidity-auction-network-parameters) for now. **Note** that we may use a different way to estimate liquidity demand in the future, hence it's important that it's properly abstracted away without a need to refactor it's downstream usage should we change the way we calculate it.
 
 ## Required liquidity
 
@@ -32,13 +32,13 @@ Count (probability of trading weighted) liquidity committed via market making or
 
 ## Trigger for entering an auction
 
-When liquidity supplied < c<sub>1</sub> * [open interest estimate](#Glossary).
+When liquidity supplied < c<sub>1</sub> * [liquidity demand estimate](#Glossary).
 
 Similarly to [price monitoring](0032-price-monitoring.md), the auction should be triggered pre-emptively. That is, the transaction that would have triggered the liquidity auction should be re-processed once auction mode is on. Please note that this transaction may be cancelled if it's type is not valid for auction, however even in that case the market should still go into the auction mode to send a signal that it requires more liquidity.
 
 ## Trigger for exiting the auction
 
-When liquidity supplied ≥ c_2 * [open interest estimate](#Glossary), \
+When liquidity supplied ≥ c_2 * [liquidity demand estimate](#Glossary), \
 where c_2 > c_1, to reduce the chance of another auction getting triggered soon after.
 
 During the liquidity monitoring auction new or existing market makers can commit more liquidity through the special market making order type and enable this by posting enough margin - see market making mechanics spec (WIP) for details. These need to be monitored to see if auction mode can be exit.


barnabee:
> WIP
> 
> Some of the things to consider:
> * What is the trigger for entering the auction mode?
> * What is the trigger for leaving the auction mode?
> * Are we counting the parked orders not market as good for auction towards the liquidity (e.g. MMs pegged orders)?

1) I think in Edinburgh we came up with a way of deciding how many siskas comitted + supplied we need at a minimum (the "liquidity maintenance" level, if you will). As soon as the amount [available | supplied] (see below) drops below this, the market should enter an indefinitely long liquidity seeking auction. The calculation for v1 may have been as simple as a constant _k_ [a network parameter] times the market's open interest (??? cc @tamlyn10 @davidsiska-vega). I think we may also have discussed that this should be an average or max open interest over a rolling window to avoid crazy volatility in this level?

2) I am reasonably sure that however we calculate the trigger level, the level to exit the liquidity auction we discussed was to set it at some buffer amount above the trigger for entry, that should also be configurable via a network parameter.

3) We should definitely count all MM supplied liquidity. In fact I have a vague recollection that we may have decided *only* to count liquidity committed and supplied by MMs, but I am not sure if that's where we ended up or just something we discussed.

witgaw:
> > WIP
> > Some of the things to consider:
> > 
> > * What is the trigger for entering the auction mode?
> > * What is the trigger for leaving the auction mode?
> > * Are we counting the parked orders not market as good for auction towards the liquidity (e.g. MMs pegged orders)?
> 
> 1. I think in Edinburgh we came up with a way of deciding how many siskas comitted + supplied we need at a minimum (the "liquidity maintenance" level, if you will). As soon as the amount [available | supplied] (see below) drops below this, the market should enter an indefinitely long liquidity seeking auction. The calculation for v1 may have been as simple as a constant _k_ [a network parameter] times the market's open interest (??? cc @tamlyn10 @davidsiska-vega). I think we may also have discussed that this should be an average or max open interest over a rolling window to avoid crazy volatility in this level?
> 2. I am reasonably sure that however we calculate the trigger level, the level to exit the liquidity auction we discussed was to set it at some buffer amount above the trigger for entry, that should also be configurable via a network parameter.
> 3. We should definitely count all MM supplied liquidity. In fact I have a vague recollection that we may have decided _only_ to count liquidity committed and supplied by MMs, but I am not sure if that's where we ended up or just something we discussed.

Thanks for your comments @barnabee. Regarding 3, what would such auction involve then? If market makers fulfil their liquidity primarily via the pegged orders, then would there be any trading in the auction (e.g. non-MM participants trading among themselves, or MM trying to fulfil their commitment via orders valid for auction) or is it just a period of temporary market suspension while MMs add enough pegged orders that will become active as soon as the auction is ended?

davidsiska-vega:
So is this an auction solely for market makers to commit more stake? And are they bidding on fees?

barnabee:
> So is this an auction solely for market makers to commit more stake? And are they bidding on fees?

@davidsiska-vega the auction itself features no market making mechanisms, it's a normal trading auction period but yes it exists in order to protect the market until more liquidity is committed (and stake/bonds to go with it). Market makers will generally need to commit more stake and bid on fees using _exactly the same method they would use during normal trading_ in order for the auction to end. (The spec for the method of committing stake and bidding fees is out of scope for this spec. though, of course.) 

Depending on the precise calculation method used for the liquidity requirement it may _also_ be possible for the auction to end if the amount of liquidity required drops below the amount supplied. For example, if the trades made uncrossing the auction would reduce the open interest far enough that the existing liquidity commitments would suffice.

davidsiska-vega:
@barnabee So it's an auction 
- it's triggered if commited liquidity is low, 
- during the auction we are waiting for MMs to commit and others can submit prices but no trades are executed, 
- once MMs have committed enough liquidity then the auction period ends, 
- if there are any crossed prices we generate trades and then switch to continuous trading. If there are no crossed prices then no trades are generated but the auction still ends and we go to continuous trading.

barnabee:
> @barnabee So it's an auction 
> 
> - it's triggered if commited liquidity is low, 
> 
> - during the auction we are waiting for MMs to commit and others can submit prices but no trades are executed, 
> 
> - once MMs have committed enough liquidity then the auction period ends, 
> 
> - if there are any crossed prices we generate trades and then switch to continuous trading. If there are no crossed prices then no trades are generated but the auction still ends and we go to continuous trading.  

That sounds exactly right. @tamlyn10?

witgaw:
Ok, so it's a single period auction (where the period is non-deterministic as it depends on liquidity getting supplied)? Because if that's a case then that: **"For example, if the trades made uncrossing the auction would reduce the open interest far enough"** would never happen (unless I'm misunderstanding it). Or is it an auction of duration driven by some parameter that gets repeated as long as the exit trigger isn't hit?

barnabee:
> Ok, so it's a single period auction (where the period is non-deterministic as it depends on liquidity getting supplied)? Because if that's a case then that: **"For example, if the trades made uncrossing the auction would reduce the open interest far enough"** would never happen (unless I'm misunderstanding it). Or is it an auction of duration driven by some parameter that gets repeated as long as the exit trigger isn't hit?

I'm not sure what you mean @witgaw. If we wanted to apply the check you quoted we'd need to calculate the open interest we'd have _if_ we ended the auction and uncrossed after each change to the order book and recalculate the liquidity commitment requirement, to decide whether the requirement was now low enough to be met by the existing committed liquidity. We could also do that recalculaiton periodically rather than every tx if we wanted. 

If we did this, we may also consdier applying rules that only allowed orders that reduce open interest to be placed during the auction. We could even apply rules like this during normal trading for a market that was close to hitting a liquidity monitoring auction (i.e. within a threshold set by a network parameter) to reduce the chance of it getting their faster.

My suggestion for public Testnet is that we perhaps **do none of these** and simply require that more liquidity be committed and the threshold met that way, in order to minimise the dev work in core. It'd also be good to model how this behaves before doing something more advanced anyway — there is a strong incentive to supply liquidity when these auctions hit I think, because someone making up the difference could basically set the fee level, at least for a while, so it would pretty much always result in a big fee spike but also perhaps a quick return to trading because of it.

witgaw:
@davidsiska-vega (cc @barnabee) I added the details discussed during today's meeting to the spec (and polished it a bit).

I've also added the details that we have discussed to the auction spec (https://github.com/vegaprotocol/product/pull/256) and the market maker mechanics spec (https://github.com/vegaprotocol/product/pull/263).

Things that I think we might still want to include/clarify:
- Should the open interest estimate freeze during the auction, or should we continue to update it (it will be constant during the auction)?

Could you please have a look at it when you have a moment, make any changes you wish, unWIP it and ask someone from engineering to have a look at it and merge if they are happy?

barnabee:
> * Should the open interest estimate freeze during the auction, or should we continue to update it (it will be constant during the auction)?

I think this question is answered by the answer in the discussion above at https://github.com/vegaprotocol/product/pull/322#issuecomment-658833550

I've suggested that:

1. we may want to re-evaluate open interest as it would be post-uncrossing, allowing people to close out volume in an auction and cause the market to reopen; but
2. we should do the simplest version for now and leave that for a future, post-public Testnet iteration and revisit later once we understand how it behaves in practice.

@davidsiska-vega @witgaw are you OK with that?

witgaw:
@barnabee I'm ok with your suggestions. What I meant is that if we estimate the liquidity demand as the max open interest over period p then in some cases that estimate will be falling during the auction due to the flow of time itself (as the more extreme observation will leave the window and more and more of it will be occupied by the the constant OI from the auction period itself). 
This is perhaps undesirable and we should use whatever the max open interest over a period p upto the beginning of the auction as the estimate of the liquidity demand throughout the auction.

barnabee:
> @barnabee I'm ok with your suggestions. What I meant is that if we estimate the liquidity demand as the max open interest over period p then in some cases that estimate will be falling during the auction due to the flow of time itself (as the more extreme observation will leave the window and more and more of it will be occupied by the the constant OI from the auction period itself).
> This is perhaps undesirable and we should use whatever the max open interest over a period p upto the beginning of the auction as the estimate of the liquidity demand throughout the auction.

Hmm, yes, good point. We should at least stop moving the window forward during an auction, I agree.

witgaw:
Edit to the above: Perhaps I'm talking nonsense above. As long as we're fine with the auction period in some cases ending as soon as c_2 * estimate of liquidity demand drops below the committed liquidity (even in cases where committed liquidity wasn't increased during the auction) then we're fine. Only for that to happen we'd need some sort of a lagged window [t-p-l, t-l), which would be equivalent to saying we're ok with liquidity demand exceeding the committed liquidity for no longer than l.

If l=0 then using max doesn't really change anything (we might as well just use instantaneous open interest) as we trigger the auction as soon as the violating (too high) open interest appears. 

It's also late so I may be making some errors in reasoning here...

barnabee:
Hmm yeah that is an option, that basically means that after the auction had been open for duration _p_, _c2_ would be based only on the current open interest rather than any more sophisticated estimate.

I don't think this would need any lagged window (or I'm not understanding what you mean). The open interest is still there during the auction so once we'd been in an auction for _p_ we'd just be taking the max of one constant amount of OI, but that ought to still work.

Pros of this over stopping the window:
- On average markets would reopen sooner
- If the market has an unusual OI spike this wouldn't taint the requirement forever

Cons
- If the market was at lower than usual OI it could go back to an auction quickly (this seems like an edge case because you'd usually expect to hit liquidity monitoring due to increasing OI not lower OI, this would mean a market maker went bust probably)
- It's potentially easier to game somehow… force another auction soon? I dunno, as you say it's late. I think we could live with that too.

So I am happy with either but leaning towards the very simple period that, if the auction is long enough, only includes the OI that went into the auction. 

cc @tamlyn10 @davidsiska-vega

witgaw:
What I meant with the lagged window is that without it it's irrelevant if we use the max or just the current open interest as in this iteration (the initial spec) there won't be any way for open interest to decrease before the exit trigger is activated (once the auction has commenced). 

So if there's some "threshold open interest" level such that committed liquidity = c_1 * "threshold OI", we enter an auction as soon as we go above the "threshold OI" and stay there "forever", that is until the auction exit trigger is activated by enough liquidity getting committed and OI can change again. 
Whereas with the lagged window [t-p-L,t-L) the OI "has" L amount of time to drop to or below the "threshold OI" to make the auction period last at most p (unless more liquidity is committed before).

witgaw:
2nd thing I forgot to add to the list:

> Things that I think we might still want to include/clarify:
> 
> Should the open interest estimate freeze during the auction, or should we continue to update it (it will be constant during the auction)?

2. During a meeting with @edd and @ashleyvega (among others) we briefly discussed network vs market parameters. Based on that conversation I'd be a bit more comfortable if we made c_1, c_2 and p market parameters (rather than network as is currently speccd) with default values supplied to the users in case they can't be bothered to think about them. Basically I'd rather considerations like initial users' lack of experience wouldn't drive our design.

barnabee:
I'm still not quite sure I understand. The reason for using max over a period is to make the system often require _more_ commitment that would be needed to support the instantaneous current OI because that will be volatile and could allow market makers to get out of their commitments on a small drop in OI in a market that actual needs them there. 

It seems OK to eventually have the max over _p_ become the current OI, though the upshot is we'll probably make _p_ longer. Even several days or a week may not be unreasonable. (If we keep this rough model I could see us changing to look back for ever but decay the OI from each point in time with age.)

If we don't get enough new liquidity committed to cover _c2 * current OI_, i.e. the perpetual state once we enter an auction and the **lowest** threshold we'll require, then the only way the market could reopen is in future if we allowed _current OI_ to be reduced (per my suggestion aboe, by evaluating what would be closed in the auction ended, which we will not implement for the first version). 

The auction therefore needs to last forever if noone steps up with more liquidity commitment. Note: I think this is unlikely. If you're a market maker and your capital is tied up and you are not making any MM profit or fee revenue, but cannot exit the commitment, you are likely to put more money in to unstick things. Particularly given that you get to bid up the fees as high as you like and receive more revenue. The only time noone would do that or come in afesh is in a market that's not trading — _this is a risk though_, because if there's OI but little active trading that might not be fun for an MM 🤷

barnabee:
> 1. During a meeting with @edd and @ashleyvega (among others) we briefly discussed network vs market parameters. Based on that conversation I'd be a bit more comfortable if we made c_1, c_2 and p market parameters (rather than network as is currently speccd) with default values supplied to the users in case they can't be bothered to think about them. Basically I'd rather considerations like initial users' lack of experience wouldn't drive our design.

Not sure about othes but I didn't recommend network params for users' benefit. I recommended them because I don't think the liquidity system should work differently from one market to the next. If the measure (max OI over a period) is the wrong one to multiply the network parameter constants by because it doesn't account for enough of the difference between markets then we should aim to design a better measure, not fiddle with the constants IMO. It's pretty likely this is the case, and I expect us to want to iterate it how we set the threshold, but requiring market creators to choose and maintain magic numbers and do it well on a per market basis is something that should be a last resort IMO.

witgaw:
> > 1. During a meeting with @edd and @ashleyvega (among others) we briefly discussed network vs market parameters. Based on that conversation I'd be a bit more comfortable if we made c_1, c_2 and p market parameters (rather than network as is currently speccd) with default values supplied to the users in case they can't be bothered to think about them. Basically I'd rather considerations like initial users' lack of experience wouldn't drive our design.
> 
> Not sure about othes but I didn't recommend network params for users' benefit. I recommended them because I don't think the liquidity system should work differently from one market to the next. If the measure (max OI over a period) is the wrong one to multiply the network parameter constants by because it doesn't account for enough of the difference between markets then we should aim to design a better measure, not fiddle with the constants IMO. It's pretty likely this is the case, and I expect us to want to iterate it how we set the threshold, but requiring market creators to choose and maintain magic numbers and do it well on a per market basis is something that should be a last resort IMO.

Ah perfect, I misunderstood the motivation to make it a network parameter then. Let's ignore my comment and thanks for clarifying. I wish those could be threads and we could close this one so others don't have to spend time reading it (like we did back in GitLab:))

witgaw:
> I'm still not quite sure I understand. The reason for using max over a period is to make the system often require _more_ commitment that would be needed to support the instantaneous current OI because that will be volatile and could allow market makers to get out of their commitments on a small drop in OI in a market that actual needs them there.
> 
> It seems OK to eventually have the max over _p_ become the current OI, though the upshot is we'll probably make _p_ longer. Even several days or a week may not be unreasonable. (If we keep this rough model I could see us changing to look back for ever but decay the OI from each point in time with age.)
> 
> If we don't get enough new liquidity committed to cover _c2 * current OI_, i.e. the perpetual state once we enter an auction and the **lowest** threshold we'll require, then the only way the market could reopen is in future if we allowed _current OI_ to be reduced (per my suggestion aboe, by evaluating what would be closed in the auction ended, which we will not implement for the first version).
> 
> The auction therefore needs to last forever if noone steps up with more liquidity commitment. Note: I think this is unlikely. If you're a market maker and your capital is tied up and you are not making any MM profit or fee revenue, but cannot exit the commitment, you are likely to put more money in to unstick things. Particularly given that you get to bid up the fees as high as you like and receive more revenue. The only time noone would do that or come in afesh is in a market that's not trading — _this is a risk though_, because if there's OI but little active trading that might not be fun for an MM 🤷

Ah yes, obviously - forgot we also use the estimate of liquidity demand to decide if M&Ms can leave the market. I fixated on auction triggers a bit too much, in that case it 100% makes sense to use the max - again thanks for clarifying...

(In that case the lagged window would be counterproductive as they could just observe it themselves and escape while they can, so let's forget about it altogether).

witgaw:
@edd we're happy with the spec now. Could you please nominate someone from engineering to read through it, ask for any details or clarification if need be and merge once everything's been resolved?

witgaw:
@peterbarrow Sorry for assigning you without any comment, but that was during a catch-up with @edd and I forgot to revisit it afterwords. Would you be happy to review the spec when you have a moment?

peterbarrow:
This spec looks good to me, but I have a few non blocking questions/comments:
* What if the window period is changed to a larger period, we will not have the sliding window of historic open values to refer to.
* Keeping a sliding window of open volumes won't be fun as we will need to record the value after every position change and will also have to prune it after every time update. Working out the new max value after the previous max leaves the window could require a huge sweep through all the history which depending on the size of the window might be a big task.
* We are trying to decouple the core from the blockchain but the only timer we have to update the core is from the blockchain, so the trigger to see if we are outside of the liquidity threshold is driven by the blocktime. How accurate do we need the sliding window to be?

davidsiska-vega:
Hi @peterbarrow, 

1) that's not such a big deal, you can just use the length of data you have until the history expands sufficiently. So let's say you're keeping one day and suddenly a vote changes it to seven days. Then, initially, it will still only be one day, tomorrow it will be two days and only in 6 days time you'll get to the full seven days.
1) as you say: you can keep the running maximum with a time-stamp of when the maximum was achieved. To update the running maximum you only need to check the current open interest. When the time-stamp indicates that your running maximum is now outside the window over which you're looking you need to look at all the "history" to choose the new maximum. But we do *not* need the history to contain *every* position change. We can decide to keep roughly 100 (or 1000) evenly spaced samples through the window. You could have a bell that chimes every say one minute (if the window is 24 hours this would correspond to 60 times per hour times 24 hours = 1440 samples in the window) and you'd only update the running maximum once the bell chimed. If we wanted to avoid unrepresentative values then during the "no-update" period we could keep a running average open volume that could be updated on every position change - but that's a simple calculation requiring keeping track of just two numbers: the running average so far and the number of samples used to calculate the prev value.
1) which brings us to the needed accuracy: it doesn't need to be very accurate (we're not talking milliseconds). It only needs to be accurate *relative to the window size* and we can reject windows that are too short. 

@barnabee @witgaw comments? 

Cheers,
David

barnabee:
1. 💯 

2. I'd probably do something like this, to avoid ever scanning the whole list. There might be a better algorithm though 🤷 

    a. Maintain a sorted list data structure (I guess a B Tree or something) of `(time, open interest)` tuples sorted by open interest.
    b. Each time open interest changes, you can remove **the entirety** of the list with a lower or equal open interest value than the current value because none of those can possibly ever be the max over any future sliding window.
    c. As David says, we don't need 100% precision in this so you could for instance maintain a single max value in a variable and commit that to the list as described in (b) only periodically (say once a second or once a minute, this could be configurable in the code or a network param OR better still be a set _fraction of the window length_), after which you reset the max value variable to the current value.
    d. Delete any values from the top of the list that are older than the sliding window length. (You could delete from the whole list in this way by maintaining a time ordered set of pointers, depends on memory vs CPU I think, but I suspect if CPU is the bottleneck you could get away without doing this.)
    e. Your value to use is then always the top of the list

3. Yeah, covered above I think in (c), we can effectively advance the window in chunks of a set length as described and lose accuracyin the time domain to make stuff faster. As @davidsiska-vega says, we need the accuracy relative to the window size only and could therefore calculate the "chunk length" based on a set fraction of the window size (say 100 or 1000 chunks per window ought to do it).

tamlyn10 (approved):

PR #323, opened 2020-07-21
WIP: Mechanism setting fees and rewarding market makers

### LINKED ISSUE #331: Setting fee prices and splitting MM rewards

davidsiska-vega:
Hi @barnabee, @tamlyn10, 

If you have a minute to read this and leave some comments... (or even better we can have a quick call) but this is round one of how we're proposing to set fees and reward market makers.

Cheers,
David


specs/0000-setting-fees-and-rewarding-mms.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Timing market fee changes
+
+Initially (before market opened) the maximum open interest is by definition zero (it's no possible to have a position on a market that's not opened yet). Hence by default the initial fee is the one supplied by the market maker who commited stake *first*. 
+Once the market opens (opening auction starts) a clock starts ticking. We have a period over which we measure the maximum open interest to estimate liquidity demand. This is a network parameter as per [liquidity monitoring](????-liquidity-monitoring.md) spec. Every time this period elapses the market fee is re-evaluated. This is written with the assumption that this parameter is something between 24 hours and 7 days. If it's significantly longer then this doesn't really work so well. On the other hand I don't want to introduce more parameters for now... and we don't want to update fees continuously as this will create another source of unpredictability for users.
```
tamlyn10: I would suggest that we separate the "calculation sliding window length" with how quickly we actually update the fees. Also, in previous conversations we've discussed having a new fee set but only come into effect in the future (that's also a future iteration of margins IIRC).

For example, we may find that a 24 hour maximum open interest is a good indicator, but we only want to reset fees weekly. This would mean every week the network would do a "fee reset".. but that the rolling window it uses is just the 24 hours prior.

tamlyn10 (commented):
just left a comment..

specs/0000-setting-fees-and-rewarding-mms.md
davidsiska-vega: I was trying to keep things simple with as few parameters as possible. I agree with both the question is do we want it now?

specs/0000-setting-fees-and-rewarding-mms.md
```diff
@@ -0,0 +1,110 @@
...
+[MM 1 liquidity, MM 1 fee]
+[MM 2 liquidity, MM 2 fee]
+...
+[MM N liquidity, MM N fee]
+```
+where `N` is the number of market makers who have committed to supply liquidity to this market. Note that `MM 1 fee <= MM 2 fee <= ... <= MM N fee` because we demand this list of pairs to be sorted. 
+
+We now find smallest integer `k` such that `c_2 x [liquidity demand estimate] < sum from i=1 to k of [MM liquidity i]`. In other words we want in this ordered list to find the market makers that supply the liquidity that's required. If no such `k` exists we set `k=N`.
+
+Finally, we set the fee for this market to be the fee `MM k fee`.
```
davidsiska-vega: @barnabee I am  still not comfortable with the last MM on the fee offer stack setting the fee, I think that having a weighted average of those providing the “required” liquidity would be better.

This is because I think in a market that’s under-supplied with liquidity someone could come in with a stake and “kill” the market by setting a ridiculously high fee.

As our mechanism for estimating liquidity demand is rather crude (open interest times a constant) this could potentially be gamed... take a big(ish) position, imply high OI the commit with a market making order demanding high fee killing the market. Of course if all works well someone else may come in and supply the liquidity at a lower fee but... it's harder to maliciously shift stake-weighed (or equity-like-share-weighted) average.

specs/0000-setting-fees-and-rewarding-mms.md
tamlyn10: I propose that we include this distinction as I don't think that it makes sense to simplify the number of parameters if it means that down the track a full rework of the code would be needed to get the behaviour we actually want? It also allows us much more flexibility when using the Scenario Runner to ascertain what "works" for a market. I'd be happy to leave off the "implement in the future" bit, but I do think the different time window parameters could be useful over the next few months.

specs/0000-setting-fees-and-rewarding-mms.md
tamlyn10: Yeah, I'm becoming convinced that the marginal fee mechanism works well when it represents the marginal cost of supplying the final “thing” i.e. when people are competing to provide that last thing and if you don't go low enough you'll be excluded from the sale. But, in this situation we let anyone provide liquidity so I could be convinced by David's argument and the potential for gaming it.

specs/0000-setting-fees-and-rewarding-mms.md
barnabee: I expected the fee would be calculated in real time based on the sliding window and the current liquidity commitments. Are you sure we'd only want to recalculate it periodically?

specs/0000-setting-fees-and-rewarding-mms.md
barnabee: I'm still pretty uncomfortable with the averaging model. If the market is in auction and I can't push the fee up enough to want to supply liquidity to reopen it, why would I bother?

Even if it reopens with a higher fee, someone else can immediately bid down the fee anyway...

However we could mitigate even that by having:

- a minimum liquidity auction duration; and
- a set time from sufficient liqudiity being committed to market reopening, so other liquidity providers would be "on notice" that it'll be reopening with a high fee and can get in ahead and reduce the fee before it reopens

A further option would be to set the level to exit a commitment even higher than the auction end level to increase the likelihood that I'm stuck with my commitment at a _lower_ fee than I bid.

davidsiska-vega: This isn't about a liquidity auction. This is normal market operation where market makers commit stake.

specs/0000-setting-fees-and-rewarding-mms.md
davidsiska-vega: We can have it real-time with a sliding window *but* it adds another thing that changes in real time and my impression was that users were not comfortable with this? 

Philosophically I am happier with the fee changing in real time because why hinder the flow of information through the market.

specs/0000-setting-fees-and-rewarding-mms.md
davidsiska-vega: But let's leave this one as it is. I don't have a good model to justify my worry. Let's just assume that if someone supplies liquidity at stupidly high fee there will be someone motivated to come in at a lower fee. If it's a viable market then this will likely happen and all is good. If it's not a viable market maybe killing it with a high fee is not a disaster. 

The solution to the problem may be a better estimate of liquidity demand rather than changing the mechanism here. We can experiment changing the "liquidity demand" estimate to eg a combination of open interest plus traded volume over a window. This way if you set a fee that's too high then eventually the measure "liquidity demand" will drop and your high fee will get ignored in the offer stack.

specs/0000-setting-fees-and-rewarding-mms.md
tamlyn10: To be honest, I had thought it was real time too, but then wondered if I'd missed something in the conversation last week (and I recalled that we'd previously discussed a window between a new fee being set and being effective). For simplicity, then, let's do it real time and then add in smoothing adjustments down the track. We can be transparent with the community that this is deliberate and we can ask how they'd like it to work.

barnabee: Cool, looks like real-time it is then!

specs/0000-setting-fees-and-rewarding-mms.md
```diff
@@ -0,0 +1,118 @@
...
+1. If the `liquidity demand estimate = 11.5` then `c_2 x [liquidity demand estimate] = 115` which means that the needed liquidity is given by MM 1 and MM 2, thus `k=2` and so the market fee is  `MM 2 fee = 0.75%`. 
+1. If the `liquidity demand estimate = 123` then `c_2 x [liquidity demand estimate] = 1230` which means that even putting all the liquidity supplied above does not meet the estimated market liquidity demand and thus we set `k=N` and so the market fee is `MM N fee = MM 3 fee = 3.75%`. 
+
+## Timing market fee changes
+
+Initially (before market opened) the maximum open interest is by definition zero (it's no possible to have a position on a market that's not opened yet). Hence by default the initial fee is the one supplied by the market maker who commited stake *first*. 
+Once the market opens (opening auction starts) a clock starts ticking. We have a period over which we measure the maximum open interest to estimate liquidity demand. This is a network parameter `t_liquidity_window`. The fee is continuosly re-evalueated using the mechanism above. 
+
+## Calculating market value proxy
```
davidsiska-vega: @barnabee Review from here... thx!

davidsiska-vega: Below we compare the stake with traded notional over a window which I've not really thought through yet 100% (which is why I liked the future fee income, but as we discussed this analogy breaks down elsewhere).

specs/0000-setting-fees-and-rewarding-mms.md
```diff
@@ -0,0 +1,118 @@
...
+This will be used for determining what "equity like share" does commiting market making stage at a given time lead to. 
+It's calculated, with `t` denoting time now, as follows:
+```
+total_stake = sum of all mm stakes
+active_time_window = [max(t-t_liquidity_window,0), t]
+
+traded_value_over_window = total trade value for fee purposes of all trades executed on a given market the active_time_window
+
+market_value_proxy = max(total_stake, traded_value_over_window)
+```
```
tamlyn10, line 71: so this means market won't be able to "shrink"...?

edd:
@edd to rebase

PR #324, opened 2020-07-24
Add a parameters section to the Market spec

### LINKED ISSUE #293: Network parameter spec
https://github.com/vegaprotocol/product/issues/292#issuecomment-636798841

edd:
- Add empty table
- Add description of section


edd:
Alright @barnabee & @tamlyn10, I've made a table in here. It feels like it's not right, but what I did was:

- Search the specs for parameters and add them
- Look at the parameters defined in code: https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md#types-of-proposals

As you can tell from the Section, column, they sort of feel like they might actually belong better spread across specs. Also there are probably some params that I haven't added from this doc yet. Unsure.

barnabee:
Ah, I see, yes @edd maybe I was unclear in the thread on Slack (and this explains your comment) but I think **network parameters need their own spec** and it's just _market parameters_ that should be in the market framework spec (and risk and product parameters, which are included within the market definition in the market framework and use the market change governance action should be in the relevant specs for the risk model and product).

edd:
> Ah, I see, yes @edd maybe I was unclear in the thread on Slack (and this explains your comment) but I think **network parameters need their own spec** and it's just _market parameters_ that should be in the market framework spec (and risk and product parameters which are included within the market definition in the market framework and use the market change governance action should be in the relevant specs for the risk model and product).

Haha - that makes sense. Easy to change!

tamlyn10:
@edd I'm going to finish this PR now.. it will be a first draft that people can add to..

edd:
> @edd I'm going to finish this PR now.. it will be a first draft that people can add to..

Thank you @tamlyn10

edd:
https://github.com/vegaprotocol/product/blob/master/specs/0028-governance.md#when-a-proposal-is-enacted - pre-enactment period is a network parameter too

PR #328, opened 2020-08-04
Add auction details

peterbarrow:
Updating the order combinations to include the new `Good For` values for auctions and continuous trading

### DIFF
--- a/specs/0014-order-types.md
+++ b/specs/0014-order-types.md
@@ -45,26 +45,33 @@ Notes on scope of current version of this spec:
 	1. **Immediate Or Cancel (IOC):** an order that trades as much of its volume as possible with passive orders already on the order book (assuming it is crossed with them) and then stops execution. It is never placed on the book even if it is not completely filled immediately, instead it is stopped/cancelled.
 	1. **Fill Or Kill (FOK):** an order that either trades all of its volume immediately on entry or is stopped/cancelled immediately without trading anything. That is, unless the order can be completely filled immediately, it does not trade at all. It is never placed on the book, even if it does not trade.
 
+### Good For / validity:
+
+*Good for must be explicitly provided, there in no default.*
+
+1. **Good For Auction (GFA):** This order will only be accepted by the system if it arrives during an auction period. Otherwise it will be rejected.
+1. **Good For Continuous (GFC):** This order will only be accepted by the system if it arrived during continuous trading. Otherwise it will be rejected.
+
 
 ### Valid order combinations
 
 ##### Continuous trading
 
-| Pricing method | GTT | GTC | IOC | FOK |
-| -------------- |:---:|:---:|:---:|:----|
-| Limit          | Y   | Y   | Y   | Y   |
-| Pegged         | Y   | Y   | Y   | Y   | 
-| Market         | N   | N   | Y   | Y   |
+| Pricing method | GTT | GTC | IOC | FOK | GFA | GFC |
+| -------------- |:---:|:---:|:---:|:---:|:---:|:---:|
+| Limit          | Y   | Y   | Y   | Y   | N   | Y   |
+| Pegged         | Y   | Y   | Y   | Y   | N   | Y   |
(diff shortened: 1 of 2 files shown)


specs/0014-order-types.md
```diff
@@ -45,26 +45,33 @@ Notes on scope of current version of this spec:
...
-| Pricing method | GTT | GTC | IOC | FOK |
-| -------------- |:---:|:---:|:---:|:----|
-| Limit          | Y   | Y   | N   | N   |
-| Pegged         | N*  | N*  | N*  | N*  | 
-| Market         | N   | N   | Y   | Y   |
+| Pricing method | GTT | GTC | IOC | FOK | GFA | GFC |
+| -------------- |:---:|:---:|:---:|:----|:---:|:---:|
+| Limit          | Y   | Y   | N   | N   | Y   | N   |
+| Pegged         | N*  | N*  | N*  | N*  | Y   | N   |
+| Market         | N   | N   | Y   | Y   | Y   | N   |
```
campbellssource: I read this as Market orders are possible for Auctions?
This was not my understanding @peterbarrow . I'm not sure about Pegged orders

peterbarrow: Good catch, in fact the original table was wrong as it was allowing IOC and FOK market orders in auction.
I have updated to fix market orders.
I'm confused about pegged orders, I'll go and have a read and get back to you on that bit.

specs/0014-order-types.md
```diff
@@ -45,26 +45,33 @@ Notes on scope of current version of this spec:
 	1. **Immediate Or Cancel (IOC):** an order that trades as much of its volume as possible with passive orders already on the order book (assuming it is crossed with them) and then stops execution. It is never placed on the book even if it is not completely filled immediately, instead it is stopped/cancelled.
 	1. **Fill Or Kill (FOK):** an order that either trades all of its volume immediately on entry or is stopped/cancelled immediately without trading anything. That is, unless the order can be completely filled immediately, it does not trade at all. It is never placed on the book, even if it does not trade.
 
+### Good For / validity:
+
+*Good for must be explicitly provided, there in no default.*
```
barnabee: How do I specify an order is GTT/GTC/other TIF? Just by leaving out this field altogether?

specs/0014-order-types.md
peterbarrow: Updated pegged order options after discussion with Barney:

Pete Barrow:
Pegged Orders and Auctions: It says in the spec that pegged orders are parked when moving from continuous trading into an auction. Can new pegged orders be accepted during auctions and if so are they put straight into the parked area?

barney:
Good question, I think letting them be placed and immediately parked sounds right

Pete Barrow:
Related question, in that case would we send the pegged order in with GFA or GFC? Feels like GFC as that's 
when the order can match.

barney:
Pegged order TIF could be GFC, GTT, GTC, etc. but not GFA as pegs are always parked in an auction so it wouldn’t make sense.

Pete Barrow:
So a new pegged+GFA order would be rejected, great I'll update the specs.

specs/0014-order-types.md
peterbarrow: We have 2 fields, one for GTC,GTT, IOC and FOK the other for GFA and GFC.

So an order for continuous trading with a limited time would have TIF=GTT and GoodFor=GTC

barnabee: What about a normal GTT (doesn't care, unless it trdes it stays on the book if we enter an auction, stays on the book after the auction ends)? This is the most common and important case.

I think if we decide we want to support the additional GFA/GFC+GTT combo (which isn't something I think we're asking for, even if it could have its uses, though I am happy enough for us to have it too if it's not a big burden), it seems like we'd be better off making additional TIF enums than adding a field?

peterbarrow: I have removed the GoodFor field and added GFA and GFN to the TIF in the specs.

campbellssource:
I think this is good to merge. And wanted to link to it when writing a ticket.

barnabee (approved):

PR #335, opened 2020-08-06
WIP: 318 spec for mm order type

### LINKED ISSUE #318: Spec for market maker orders
Use this as a guide... 

https://docs.google.com/document/d/1V3YTlSOTU9Do_dmYhqtOqLO6lee_UeRHS_ikHoD7v4E/edit#

davidsiska-vega:
@tamlyn10 This is where I got to with the MM order type. I think it's ready, at a high level. It may need an iteration with someone from engineering.

### DIFF
--- a/specs/0000-market-making-order-type.md
+++ b/specs/0000-market-making-order-type.md
@@ -7,11 +7,52 @@ Market makers can commit to provide liquidity to a market by submitting this spe
 The liquidity is measured in "siskas" as set out in the [Probability Weighted Liquidity Measure](0034-prob-weighted-liquidity-measure.ipynb) specification. 
 
 As part of the market making order the market maker submits the following:
-1) their desired fee for the market and
-1) something that implies what liquidity commitment (in siskas they are making). This can take two forms: either they submit volume-at-distance from market mid price from which liquidity is calculated (details follow below) or they submit the amount of liquidity to commit and how they want to distribute it (details follow below).
-1) Morover, optionally, they can submit mid price (this will change where their orders are placed on the order book or nothing else). Otherwise, market mid-price is used.
+1) their desired fee for the market, 
+1) the amount of liquidity to commit and how they want to distribute it (details follow below) and
+1) optionally, they can submit mid price (this will change where their orders are placed on the order book or nothing else). Otherwise, market mid-price is used.
 
+The liquidity commitment they make (in siskas) is converted using, a per-asset network parameter `siskas_to_bond_in_asset_X` as follows: 
+``` mm_bond = liquidity_commitment x siskas_to_bond_in_asset_X.```
 
+The volume implied by these MM order types is *not* placed on the book during auctions. 
 
+## Specifying liquidity in terms of fraction-of-commitment-at-a-distance
 
+The market maker submits:
+1) market ID 
+1) desired fee
+1) liquidity commitment
+1) optional mid price
+1) sell-side: list of `[distance from mid, fraction-of-sell-side-commitment]` pairs. 
+1) buy-side: list of `[distance from mid, fraction-of-buy-side-commitment]` pairs. 
 
+We check that the sum across the list of the `fraction-of-***-side-commitment` is `1`, with `***=[sell,buy]`. If not reject the order.
+We check that they have 
+
+``` mm_bond = liquidity_commitment x siskas_to_bond_in_asset_X```
+
+in their general account for the relevant asset; if not the order is rejected. If yes, then the amount is transferred into their per-asset market making bond amount.  
+
+## Converting fraction-of-commitment-at-a-distance into volume
+
+Assumption: Either the MM order specifies mid price or there is a mid price for the book. If not a price monitoring auction should be triggered (!!! Barney, Tamlyn, Witold, true? false? !!!!). 
+
+For each entry on the buy and sell side list we now do the following:
+
+Given `mid_price`, `distance_from_mid` and  `fraction-of-commitment` we obtain the `probability_of_trading` at that `distance_from_mid` at that `mid_price` from the risk model, see [Quant risk model spec](0018-quant-risk-models.ipynb). 
+
+The volume implied by the oder at that distance from mid price is then 
+
+``` volume = liquidity_commitment x fraction-of-commitment / probability_of_trading```. 
+
+This volume should be added to the order book subject to margin account check (and perhaps transfer from general to margin account). If there is still insufficient margin then the required amount should be transferred from the market maker bond account. Additionally there may be a penalty applied as per the [slashing spec](????-????.md). 
+
+## Amending the MM order:
+
+Market makers are always allowed to ammend:
+- 2. and 4.,
+- 5. and 6. - as long as the relevant fractions add up to `1`, 
+
+Market makers are allowed to increase the liquidity commitment 3. subject to being able to top up the bond commitment from the general account (check the new commitment amount, and transfer the difference from the general account to the bond account). 
+
+Market makers are allowed to decrease the liquidity commitment 3. subject to there being sufficient liquidity committed to the market so that we stay above `liquidity supplied >= c_2 x liquidity demand estimate` see [liquidity monitoring spec](0035-liquidity-monitoring.md). This means we calculate the supplied liquidity with the amended order plus all other committed market makers, if the above inequality is still satisfied the order is accepted, otherwise rejected.
--- a/specs/0000-market-making-order-type.md
+++ b/specs/0000-market-making-order-type.md
@@ -2,7 +2,7 @@
 
 ## Summary 
 
-When market makers commit to providing liquidity they are required to submit a set of valid buy orders and sell orders [market making mechanics](????-mm-mechanics.md). This commitment will ensure that they are eligible for portion of the market fees as set out in [Setting Fees and Rewarding MMs](????-setting-fees-and-rewarding-mms.md).
+When market makers commit to providing liquidity they are required to submit a set of valid buy shapes and sell shapes [market making mechanics](????-mm-mechanics.md). This commitment will ensure that they are eligible for portion of the market fees as set out in [Setting Fees and Rewarding MMs](????-setting-fees-and-rewarding-mms.md).
 
 
 ## Market making order features
@@ -23,7 +23,7 @@ The network will translate these shapes into order book volume by creating an or
 
 Each entry must specify:
 
-1. **Liquidity proportion:** the relative proportion of the commitment to be allocated at a price level. Note, the network will normalise the liquidity proportions of the refined order set (see below).
+1. **Liquidity proportion:** the relative proportion of the commitment to be allocated at a price level. Note, the network will normalise the liquidity proportions of the refined order set (see below). This must be strictly positive number.
 
 2. A **price peg:** , as per normal [pegged orders](), a price level specified by a reference point (e.g mid, best bid, best offer) and an amount of units away. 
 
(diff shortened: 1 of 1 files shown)


specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,58 @@
...
+The liquidity is measured in "siskas" as set out in the [Probability Weighted Liquidity Measure](0034-prob-weighted-liquidity-measure.ipynb) specification. 
+
+As part of the market making order the market maker submits the following:
+1) their desired fee for the market, 
+1) the amount of liquidity to commit and how they want to distribute it (details follow below) and
+1) optionally, they can submit mid price (this will change where their orders are placed on the order book or nothing else). Otherwise, market mid-price is used.
+
+The liquidity commitment they make (in siskas) is converted using, a per-asset network parameter `siskas_to_bond_in_asset_X` as follows: 
+``` mm_bond = liquidity_commitment x siskas_to_bond_in_asset_X.```
+
```
tamlyn10, line 16: @davidsiska-vega what is the rationale for this being a **per-asset** network parameter?  Worth noting that if it is, we need to make sure that when a new asset is added on Vega that this network parameter is set. cc @barnabee

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,58 @@
...
+## Specifying liquidity in terms of fraction-of-commitment-at-a-distance
+
+The market maker submits:
+1) market ID 
+1) desired fee
+1) liquidity commitment
+1) optional mid price
+1) sell-side: list of `[distance from mid, fraction-of-sell-side-commitment]` pairs. 
+1) buy-side: list of `[distance from mid, fraction-of-buy-side-commitment]` pairs. 
+
```
tamlyn10, line 21: I think that [the way we last discussed this](https://docs.google.com/document/d/1V3YTlSOTU9Do_dmYhqtOqLO6lee_UeRHS_ikHoD7v4E/edit#heading=h.eypadvah1g9f) using [order books sheet](https://docs.google.com/spreadsheets/d/10kQxWBtJ5czoke4hDHEBqAV_dzo3OS1r-7VK_fVbtCc/edit#gid=364089557) from MM data sheet mucking spreadsheet we worked out that the market maker only specifies the distance from peg OR the fraction of liquidity commitment - i.e. not both. I will push a new commit with my proposed amendments.

tamlyn10: image

Screenshot of example order...

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,58 @@
...
+## Summary 
+
+Market makers can commit to provide liquidity to a market by submitting this special order type. This commitment will ensure that they are eligible for portion of the market fees as set out in [Setting Fees and Rewarding MMs](????-setting-fees-and-rewarding-mms.md). 
+
+The liquidity is measured in "siskas" as set out in the [Probability Weighted Liquidity Measure](0034-prob-weighted-liquidity-measure.ipynb) specification. 
+
+As part of the market making order the market maker submits the following:
+1) their desired fee for the market, 
+1) the amount of liquidity to commit and how they want to distribute it (details follow below) and
```
tamlyn10: If this is part of every MM order, fee change transaction (i.e. those actions all live together), we should have an option for them to leave it empty if they don't want to adjust their stake (and the same goes for fees). I'm not sure how this all plays out if they want to amend their orders and their stake together but the orders are malformed for whatever reason - do we still amend the stake? 

I know we've discussed MM transactions for orders vs stake vs fees but don't think we landed on whether to make 1tx handle all of them. @barnabee do you recall or have a preference?

specs/0000-market-making-order-type.md
tamlyn10: One thing I'd note about above screenshot is that we were thinking that we were going to have 1 x transaction for buy orders and one for sell orders. I don't recall why - @barnabee may?

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,58 @@
...
+We check that they have 
+
+``` mm_bond = liquidity_commitment x siskas_to_bond_in_asset_X```
+
+in their general account for the relevant asset; if not the order is rejected. If yes, then the amount is transferred into their per-asset market making bond amount.  
+
+## Converting fraction-of-commitment-at-a-distance into volume
+
+Assumption: Either the MM order specifies mid price or there is a mid price for the book. If not a price monitoring auction should be triggered (!!! Barney, Tamlyn, Witold, true? false? !!!!). 
+
```
tamlyn10, line 36: We should permit pegging to best bid / offer on either side. 

## Logic for missing peg price
This is an interesting one, because if a market has 0 open interest, then we don't want to be putting it into a liquidity auction if there is not BBOs. I would vote that in this case, we make all market making orders be the "least competitive but compliant" version and allow them to amend them from there (because they can specify using proportions at price levels).




Then, if there is no best bid or offer then liquidity monitoring should be triggered IMO... (this may require a

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,58 @@
...
+``` volume = liquidity_commitment x fraction-of-commitment / probability_of_trading```. 
+
+This volume should be added to the order book subject to margin account check (and perhaps transfer from general to margin account). If there is still insufficient margin then the required amount should be transferred from the market maker bond account. Additionally there may be a penalty applied as per the [slashing spec](????-????.md). 
+
+## Amending the MM order:
+
+Market makers are always allowed to ammend:
+- 2. and 4.,
+- 5. and 6. - as long as the relevant fractions add up to `1`, 
+
```
tamlyn10, line 147: As per above screenshot I think we were going to just normalise whatever numbers they provided.

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,58 @@
...
+Market makers are allowed to increase the liquidity commitment 3. subject to being able to top up the bond commitment from the general account (check the new commitment amount, and transfer the difference from the general account to the bond account). 
+
+Market makers are allowed to decrease the liquidity commitment 3. subject to there being sufficient liquidity committed to the market so that we stay above `liquidity supplied >= c_2 x liquidity demand estimate` see [liquidity monitoring spec](0035-liquidity-monitoring.md). This means we calculate the supplied liquidity with the amended order plus all other committed market makers, if the above inequality is still satisfied the order is accepted, otherwise rejected.
```
tamlyn10: Just realised we also need to add in auto-refreshing of market maker orders.

specs/0000-market-making-order-type.md
tamlyn10: OK just discovered this covered in this doc which I missed meeting around.. https://docs.google.com/document/d/15_gXcUorYmCgLnBquQxZ5Vx_amtemMCSOFqSoj73U_g/edit which covers off some of this.

specs/0000-market-making-order-type.md
tamlyn10: Add in price monitoring threshold as price limit on either side?

specs/0000-market-making-order-type.md
barnabee: Agree I think it should be a single network parameter. The asset units cancel out anyway I think as siskas are an asset-dependent unit?

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Summary 
+
+When market makers commit to providing liquidity they are required to submit a set of valid buy orders and sell orders [market making mechanics](????-mm-mechanics.md). This commitment will ensure that they are eligible for portion of the market fees as set out in [Setting Fees and Rewarding MMs](????-setting-fees-and-rewarding-mms.md).
+
+
+## Market making order features
+
+Market maker orders are a special order type with the following features:
+- Is a batch order: allows simultaneously specifying multiple orders in one message/transaction
```
barnabee, line 11: Perhaps we should leave the batch part to the MM mechanics spec and just specify what a valid single order row looks like so we don't get duplicative here?

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,76 @@
...
+- Initially all are pegged orders but other price types may be available in future
+- Are always priced limit orders that sit on the book
+- Are “post only” and do not trade on entry
+- The order is always refreshed after trading (once the tx is processed so not refreshed before closeouts, etc.) based on the above requirements so that the full commitment is always supplied (spec coming)
```
barnabee: NB: "refresh" occurs at the same tame as peg price updates

specs/0000-market-making-order-type.md
barnabee: @tamlyn10 it's either distance OR quantity plus a liquidity fraction IIRC, the other of price/quantity is calculated from the two provided values

specs/0000-market-making-order-type.md
davidsiska-vega: That's right the asset units cancel out do it can be a network param.

specs/0000-market-making-order-type.md
davidsiska-vega: To me that only worked with some implicit assumption and I forgot what it was. It seems deeply unnatural to me. 

Specifying what fraction of your liquidity at what distance remains the most natural to me which is why I wrote it up that way.

specs/0000-market-making-order-type.md
tamlyn10: OK, since this has now moved over to the mm-mechanics spec, I will amend there.

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,76 @@
...
+***PLUS EITHER:***
+
+**Price peg:** the price level specified by a reference point (e.g mid, best bid, best offer) and an amount of units away (initially we offer this specified in ticks. In future this could be specified as a % or number of standard deviations as per risk model distribution).
+
+***OR***
+
+**Size:** the size of the order at the given price
+
+```
+# Example 1:
```
barnabee, line 31: I can redo this how I see it

specs/0000-market-making-order-type.md
barnabee: I would like to stick to what we discussed and agreed, or reopen the discussion

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,76 @@
...
+
+Given the price peg information (`price-peg-reference`, `number-of-units-from-reference`) and  `liquidity-normalised-proportion` we obtain the `probability_of_trading` at the resulting order price, from the risk model, see [Quant risk model spec](0018-quant-risk-models.ipynb). 
+
+The volume implied by the order at that distance from mid price is then 
+
+``` volume = liquidity_obligation x liquidity-normalised-proportion / probability_of_trading```. 
+
+## Adding MM orders to the book:
+
+The set of buy orders are applied to the resulting order price and volume.
```
barnabee: Should be explicit. The calculation above is performed along with the peg re-pricing, whenever pegs are repriced as per pegged order spec

specs/0000-market-making-order-type.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Amending the MM order:
+
+Market makers are always allowed to amend their orders by submitting a market maker network transaction with a set of revised orders (see [market making mechanics spec](./0000-mm-mechanics.md))
```
barnabee: Note: futiure an individual order's params may be amendable by using it's order ID and an Amend tx

barnabee (changes requested):

peterbarrow (approved):
This all makes sense to me but I would like to see all the questions already listed resolved before this is merged as they will impact the way the spec is used.

PR #336, opened 2020-08-10
new mm mechanics

### LINKED ISSUE #334: Market making mechanics
Similar to the margin orchestration, this spec explains how it all fits together.

tamlyn10:


specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,191 @@
...
+
+The point of market making on Vega is to incentivise people to place orders on the market that maintain liquidity on the book. This is done via a financial commitment and reward + penalty mechanics, and through the use of a special batch order type that automatically updates price/size as needed to meet the commitment and automatically refreshes its volume after trading to ensure continuous liquidity provision.
+
+## Market making network transaction
+
+Any Vega participant can apply to market make on a market by submitting a transaction to the network which includes the following
+
+1. Market ID
+1. COMMITMENT AMOUNT: liquidity commitment amount (in settlement asset of the market) 
+1. FEES: nominated fee amount (TODO: UNITS?). Default equals zero.
```
witgaw: Maybe "FEES: nominated liquidity fee factor. Default equals zero." to keep it in line with the fees spec?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,191 @@
...
+1. ORDERS: a set of _buy orders_ and _sell orders_ to meet the market making obligation.
+
+Accepted if all of the following are true:
+- [ ] The participant has sufficient collateral in their general account to meet the size of their nominated commitment amount, as specified in the transaction.
+- [ ] The market is active
+- [ ] The nominated fee amount is not less than zero.
+- [ ] There are a set of valid buy and sell orders (see MM orders spec)       
+
+Invalid if:
```
witgaw: I'd write "Invalid if any of the following is true:" to avoid confusion.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,191 @@
...
+### Network automatically amends commitment amount
+The network will amend the commitment amount if they market maker's obligations have not been met (see section below)
+
+## FEES
+
+### Nominating and amending fee amounts
+
+The network transaction is used by market makers to nominate a fee amount which is used by the network to calculate the [liqudity_fee](./0029-fees.md) of the market.
+- [ ] A nominated fee amount must be greater than or equal to zero. The units of the fee amount are a percentage. A number greater than 1 is permitted.
+- [ ] If nominated fee amount is malformed or less than zero, the network should assign a default amount of zero.
```
witgaw: Wouldn't it be safer to just reject the transaction in that case?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,191 @@
...
+
+Market makers may amend their nominated fee amount by submitting a market maker transaction to the network with a new fee amount. If the fee amount is valid, this new amount is used. Otherwise, the nominated fee amount is set to zero, as per above criteria.
+
+### How fee amounts are used
```
witgaw, line 117: Shouldn't this section just refer https://github.com/vegaprotocol/product/pull/323?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+Invalid if:
+- [ ] Tx settlement asset does not equal settlement asset of market
+- [ ] Liquidity commitment amount is less than or equal to zero
+- [ ] Nominated fee amount is less than zero
+- [ ] Acceptance criteria from ORDERS spec is not met.
+
+## COMMITMENT AMOUNT
+
+### Processing the commitment
+When a commitment is made the liquidity commitment amount is transferred to the market making bond account (new account type, 1 per market) for the market.
```
witgaw: Is this one account for all market maker bonds? If so, wouldn't keeping them separate make it easier to deal with slashes, commitment changes etc.?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+***Case:*** `margin-shortfall > market-maker-commitment-amount - market-maker-bond-penalty` 
+
+The network will consider the market maker to be distressed and immediately place it into position resolution.
+
+***Case:*** `margin-shortfall >= market-maker-commitment-amount - market-maker-bond-penalty` 
+
+The network will:
+1. Confiscate an amount equal to `market-maker-bond-penalty` from the market's _market maker bond account_ and add it to the insurance pool.
+1. Transfer an amount equal to `margin-shortfall` from the market's _market maker bond account_ into the market maker's margin account.
+1. Adjust the market maker's commitment amount to match the amount netted from the penalty: `actual-commitment-amount = previous-commitment-amount - market-maker-bond-penalty`
```
witgaw: Shouldn't it be: `actual-commitment-amount = previous-commitment-amount - market-maker-bond-penalty - margin-shortfall`?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+Note: in future versions, the degree to which a market maker has met their obligation may be measured over a rolling time period (specified by a network parameter, initially set to 24 hours) and calculated as an average of their (attained outcome - obligation). The penalty would then be proportional to the amount of failure and the amount of time in breach.
+
+In the following cases, we assume that the `market-maker-bond-penalty` is not available to cover the shortfalls of the market maker.
+
+***Case:*** `margin-shortfall > market-maker-commitment-amount - market-maker-bond-penalty` 
+
+The network will consider the market maker to be distressed and immediately place it into position resolution.
+
+***Case:*** `margin-shortfall >= market-maker-commitment-amount - market-maker-bond-penalty`
```
witgaw: Shouldn't it be "<=" ?

witgaw (commented):
Looks ok to me, added some comments.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+Any Vega participant can apply to market make on a market by submitting a transaction to the network which includes the following
+
+1. Market ID
+1. COMMITMENT AMOUNT: liquidity commitment amount (in settlement asset of the market) 
+1. FEES: nominated fee amount (TODO: UNITS?). Default equals zero.
+1. ORDERS: a set of _buy orders_ and _sell orders_ to meet the market making obligation.
+
+Accepted if all of the following are true:
+- [ ] The participant has sufficient collateral in their general account to meet the size of their nominated commitment amount, as specified in the transaction.
```
barnabee, line 23: Add and the margin required for their liquidity provision orders?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+- [ ] The market is active
```
barnabee: Or proposed, or pending (need to make clear you can commit during proposal period and after a proposed market is approved but before it's created)

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+- [ ] The nominated fee amount is not less than zero.
```
barnabee: Should we also have a global "MAX_LIQUIDITY_FEE" network parameter, which the bid must be <=?

I think maybe necessary to prevent attack where someone comes into a low liquidity market with a fee of 100% (or extremely high). Would assume noone would trade until someone else comes in but if I could force a liquidation then the MMs take all the value in fees.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+- [ ] There are a set of valid buy and sell orders (see MM orders spec)
```
barnabee: "valid buy and sell liquidity provision orders ..." might be better?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+The point of market making on Vega is to incentivise people to place orders on the market that maintain liquidity on the book. This is done via a financial commitment and reward + penalty mechanics, and through the use of a special batch order type that automatically updates price/size as needed to meet the commitment and automatically refreshes its volume after trading to ensure continuous liquidity provision.
+
+## Market making network transaction
+
+Any Vega participant can apply to market make on a market by submitting a transaction to the network which includes the following
+
+1. Market ID
+1. COMMITMENT AMOUNT: liquidity commitment amount (in settlement asset of the market) 
+1. FEES: nominated fee amount (TODO: UNITS?). Default equals zero.
+1. ORDERS: a set of _buy orders_ and _sell orders_ to meet the market making obligation.
```
barnabee: We should think about naming. Do we call this orders or something like "bid liquidity" and "offer liquidity"?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+## COMMITMENT AMOUNT
+
+### Processing the commitment
+When a commitment is made the liquidity commitment amount is transferred to the market making bond account (new account type, 1 per market) for the market.
+
+- Market maker bond account:
+	- [ ] Each active market has one bond account per settlement asset for that market.
+    - [ ] When a market maker is approved, the size of their staked bond is immediately transferred from their general account to the market's bond account.
+    - [ ] Only the network may withdraw collateral from this account.
```
barnabee: What does this mean. If I reduce my committment and it's accepted I can withdraw? If we consider that the network then I'd say increasing my commitment or creating a new one is also done by the network so only the network can transfer to or from the account. I think this is mostly unhelpful though as everything in the system can only be done by the network.

Suggest something like this instead: "a market maker can only transfer funds to or from this account by submitting a valid transaction to create, increase, or decrease their commitment to the market, which must be validated and pass all checks (e.g. including those around minimum liquidity commitment required, when trying to reduce commitment)"

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+    - [ ] Collateral withdrawn from this account may only  be transferred to either:
```
barnabee, line 61: Add: the market maker's margin account, in the event that they fall below the maintenance level and have zero balance in their general account.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+      - [ ] The insurance pool of the market (in event of slashing)
+      - [ ] The market maker's general account (in event of market maker reducing their commitment)
+
+### Market maker proposes to amend commitment amount
+The commitment transaction is also used to amend any aspect of their market making obligations.
```
barnabee: We should clarify in this section that to end their commitment a MM must submit an accepted Commit Liquidity transaction with a commitment size of zero. We should also clarify that in this case it cannot contain any liquidity bids/offers and if accepted all their liquidity provision orders will be cancelled off the book and any bond returned, margin returned only if they now have no position/orders at all or if the margin now falls below the recalculated release level.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
+# Market maker mechanics
+
+The point of market making on Vega is to incentivise people to place orders on the market that maintain liquidity on the book. This is done via a financial commitment and reward + penalty mechanics, and through the use of a special batch order type that automatically updates price/size as needed to meet the commitment and automatically refreshes its volume after trading to ensure continuous liquidity provision.
+
+## Market making network transaction
```
barnabee: We should call the the `Commit Liquidity` transaction IMO

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+	- [ ] Each active market has one bond account per settlement asset for that market.
+    - [ ] When a market maker is approved, the size of their staked bond is immediately transferred from their general account to the market's bond account.
+    - [ ] Only the network may withdraw collateral from this account.
+    - [ ] Collateral withdrawn from this account may only  be transferred to either:
+      - [ ] The insurance pool of the market (in event of slashing)
+      - [ ] The market maker's general account (in event of market maker reducing their commitment)
+
+### Market maker proposes to amend commitment amount
+The commitment transaction is also used to amend any aspect of their market making obligations.
+A participant may apply to amend their commitment amount by submitting a transaction for the market with a revised commitment amount.
```
barnabee, line 68: Use name i.e. "by submitting a `Commit Liquidity` transaction"

If they are amending an existing commitment, the participant may omit / leave blank the orders section to use their existing orders (these will be scaled to the new commitment automatically) or provide a new valid set of bid and offer liquidity orders.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+      - [ ] The market maker's general account (in event of market maker reducing their commitment)
+
+### Market maker proposes to amend commitment amount
+The commitment transaction is also used to amend any aspect of their market making obligations.
+A participant may apply to amend their commitment amount by submitting a transaction for the market with a revised commitment amount. 
+
+`proposed-commitment-variation = new-proposed-commitment-amount - old-commitment-amount`
+
+***Case:*** `proposed-commitment-variation >= 0`
+A market maker can always increase their commitment amount as long as they have sufficient collateral in the settlement asset of the market to meet the new commitment amount.
```
barnabee: and any margin increase for their revised orders

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+If they do not have sufficient collateral:
+    - [ ] the previous market making commitment is retained
+    - [ ] any other details contained within the new transaction is ignored by the network, i.e.
```
barnabee: "the entire transaction is rejected, and any other details..."

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+        - [ ] the new fee bid is ignored
+        - [ ] the new market maker orders are ignored
+
+
+***Case:*** `proposed-commitment-variation < 0`
+The network needs to calculate whether the market maker may lower their commitment amount and if so, by how much.
+
+`maximum-reduction-amount = max(1/siskas_to_bond_in_asset_X x (total-market-making-liquidity - (c_2 x liquidity-demand-estimate)), 0)`
```
barnabee: I think we should calculate the c1 and c2 levels and expose these on APIs (say, the market data API, or a new market making data API). Will be pretty fundamental to know this.

I also think we should maybe calculate and mostly talk about these in terms of _base asset commitment (bond) requirement_ rather than siskas — only a slight reaarrangement I know, but I think will aid education and understanding. We can calculate and show siskas but at the moment siskas aren't a universal unit, they're a different unit per asset and until we find a way to talk about that (i.e. normalise to siskas/USD or something?) I think they're perhaps going to be confusing.

Not sure this changes this section BTW, I just thought about it here and didn't want to lose. Will delete if it's covered elswhere.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+where:
+
+`total-market-making-liquidity` is the sum of all market making obligations [TODO: LINK TO BELOW] that the market has, measured in siskas and using the previous-commitment-amount for the proposing market maker.
+
+`liquidity-demand-estimate` is a measure of the market's current liquidity requirements, as per the calculation in the [liquidity monitoring spec](0035-liquidity-monitoring.md).
+
+`actual-reduction-amount = min (abs(proposed-reduction-amount), maximum-reduction-amount)`
```
barnabee: I had assumed we'd reject a reduction that was too large outright and wait for a valid new commitment to be entered. It seems cleaner and easier to reason about to me that if a `Commit Liquidity` transaction isn't rejected then my liquidity commitment is equal to the commitment I requested.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+`new-actual-commitment-amount =  old-commitment-amount - actual-reduction-amount ` 
+
+i.e. market makers are allowed to decrease the liquidity commitment subject to there being sufficient liquidity committed to the market so that it stays above the market's minimum liquidity threshold. The above formulae result in the fact that if `maximum-reduction-amount = 0`, then `actual-reduction-amount = 0` and therefore the market maker is unable to reduce their commitment amount.
+
+When `actual-reduction-amount > 0`:
+
+- [ ] the difference between their previous commitment and new commitment is transferred back to their general account, i.e.
```
barnabee: Unless the current balance of their bond is below par i.e. they have been penalised + bond has been used as margin. So the difference between their current bond balance and new commitment is transferred rather than diff to their previous commitment.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+`transferred-to-general-account-amount =  old-commitment-amount - new-actual-commitment-amount `
+
+- [ ] the revised fee amount and set of orders are processed.
+
+When `actual-reduction-amount = 0`:
```
barnabee: Add the entire `Commit Liquidity` transaction is rejected.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+    - [ ] the previous market making commitment is retained
+    - [ ] any other details contained within the new transaction is ignored by the network, i.e.
+        - [ ] the new fee bid is ignored
+        - [ ] the new market maker orders are ignored
+
+### Network automatically amends commitment amount
+The network will amend the commitment amount if they market maker's obligations have not been met (see section below)
```
barnabee: Is this right? I think the amount committed ought to remain the same and the _actual amount supplied / bond balance_ would be what's amended, and if they realise a general account increase (deposite, settlement, reduce other commitment, etc.) the actual bond would be topped back up to the commitment amount.

Unless we discussed this and agreed something else that I forgot?

specs/0000-mm-mechanics.md
barnabee: Yes. Always reject if we can't do what the TX asks for IMO.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+## FEES
+
+### Nominating and amending fee amounts
+
+The network transaction is used by market makers to nominate a fee amount which is used by the network to calculate the [liqudity_fee](./0029-fees.md) of the market.
+- [ ] A nominated fee amount must be greater than or equal to zero. The units of the fee amount are a percentage. A number greater than 1 is permitted.
```
barnabee: Per above comment too, should we have a MAX set by network param?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+- [ ] If nominated fee amount is malformed or less than zero, the network should assign a default amount of zero.
+
+Market makers may amend their nominated fee amount by submitting a market maker transaction to the network with a new fee amount. If the fee amount is valid, this new amount is used. Otherwise, the nominated fee amount is set to zero, as per above criteria.
```
barnabee: See above

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+The [liqudity_fee](./0029-fees.md) of a market on Vega takes as an input, a [fee factor[liquidity]](./0029-fees.md) which is calculated by the network, taking as an input the data submitted by the market makers in their market making network transactions (see [this spec]() for more information on the specific calculation).
+
+
+### Distributing fees between market makers
+When calculating fees for a trade, the size of a market maker’s commitment along with when they committed and the market size are inputs that will be used to calculate how the liquidity fee is distributed between market makers. See [this spec]() for the calculation of the split.
+
+
+## ORDERS
+
+In a market  maker proposal transaction the participant must submit a valid set of orders, comprised of:
```
barnabee, line 127: "liquidity provision orders"

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+1. A set of valid buy orders
```
barnabee: "A set of buy orders which must be valid pegged orders of type liquidity provision order"

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+1. A set of valid sell orders
```
barnabee: "A set of sell orders which must be valid pegged orders of type liquidity provision order"

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+Market maker orders are a special order type described in the [market maker order spec](). Validity is also defined in that spec.
```
barnabee: Per discussion on Slack - probably rename market maker orders to luqidity provision orders or whatever we decide.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+
+A market maker can amend their orders by providing a new set of orders in the market maker network transaction. If the amended orders are invalid, the previous set of orders will be retained.
```
barnabee: `Commit Liquidity` transaction

If amending orders only, the commtiment amount, fee bid can be omitted.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+**During continuous trading:**
+A market maker complies with their market making obligation by:
+1. Submitting valid _market maker orders_ in the market maker network transaction.
+1. Holding sufficient collateral to meet the usual margin obligations associated with these orders. 
+
+Since market maker orders automatically refresh, a market maker is only non-compliant when they have insufficient capital to meet the margin requirements of these orders.
+
+**During auction:**
+- Market maker obligation during auction (including market commencement auction):
+	- [ ] Market makers are not required to place orders during an auction period.
```
barnabee: Not sure this is worded quite right. A market maker with a valid commitment should be required to have a valid set of liqudity provision orders at all times as part of their commitment. These will just be parked in an auction so they will not be on the book, and a market maker is not required to place any _additional_ orders during the auction or participate in the auction.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+    - [ ] Market maker orders that are placed during an auction call period are parked and reinstated when the limit order book is reinstated.
+	- [ ] At conclusion of auction period call period, market makers obligations are reinstated.
+
+### Non-compliance
+
+If at any point in time, the market maker has insufficient capital to meet their margin requirements arising from their market making orders and open positions, the network will utilise their market making commitment, held in the _market maker bond account_ to cover their commitment, and penalise them at a proportional rate.
```
barnabee: Does this happen before or after the usual cancellation of [other] orders, position resolution of open position. Assume before, but would we want to cancel non-liquidity-providing orders first?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+Let `market-maker-bond-penalty = bond-fine-parameter * margin-shortfall` be the amount of commitment that has been slashed, where `bond-fine-parameter` is a network parameter and the `margin-shortfall` refers to the absolute value of the amount of margin that the market maker was unable to cover through their margin and general account.
```
barnabee: Probably prefer the term penalty to fine? i.e. "liquidity penalty factor"?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+Note: in future versions, the degree to which a market maker has met their obligation may be measured over a rolling time period (specified by a network parameter, initially set to 24 hours) and calculated as an average of their (attained outcome - obligation). The penalty would then be proportional to the amount of failure and the amount of time in breach.
```
barnabee: I'm not sure how this would work. Are you saying that perevious breaches/breach history would be used when a new or further breach occurred to decide how much penalty to apply?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+In the following cases, we assume that the `market-maker-bond-penalty` is not available to cover the shortfalls of the market maker.
+
+***Case:*** `margin-shortfall > market-maker-commitment-amount - market-maker-bond-penalty` 
+
+The network will consider the market maker to be distressed and immediately place it into position resolution.
```
barnabee: I think we would be better off saying after the penalty is applied we transfer the shortfall (or as much of it as available) from the bond account to the margin (so it can be used in MTM or taken to insurance pool, otherwise it just sits in bond account) and then things proceed as normal i.e. if it is still in shortfall then position resolution will occur. 

Should avoid special cases like if X then immediately go to position resolution where we can reuse exisitng flow and logic, for simplicity/consistency.

specs/0000-mm-mechanics.md
barnabee: Per my above comment, I think this is the only case

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+
+***Case:*** `margin-shortfall >= market-maker-commitment-amount - market-maker-bond-penalty` 
+
+The network will:
+1. Confiscate an amount equal to `market-maker-bond-penalty` from the market's _market maker bond account_ and add it to the insurance pool.
```
barnabee: Transfer penalty amount (or everything is less) to the insurance pool. I'd avoid terms like confiscate here in the spec. It's just a transfer.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+1. Transfer an amount equal to `margin-shortfall` from the market's _market maker bond account_ into the market maker's margin account.
```
barnabee: Continue as normal. With margin management - we should amend collateral search to search first general account then their bond account for the market.

specs/0000-mm-mechanics.md
barnabee: As aboce (and I think what @witgaw is getting at) — I don't think we amend their commitment amount but record the actual level of commitment they are making.

specs/0000-mm-mechanics.md
tamlyn10: I think the show should go on.. their fee nomination is only one of many. If it is set to zero, traders just get cheap prices for a bit..

specs/0000-mm-mechanics.md
tamlyn10: yeah, I hadn't spotted that David had done that already, a later commit fixed it..

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,184 @@
...
+1. Adjust the market maker's commitment amount to match the amount netted from the penalty: `actual-commitment-amount = previous-commitment-amount - market-maker-bond-penalty`
+
+
+## Network parameters 
+`bond-fine-parameter` - used to calculate the penalty to market makers when they fail to meet their obligations.
+`market-size-measurement-period` - used in fee splitting
+
+
+## What data do we keep relating to market making?
+1. List of all market makers and their commitment sizes and their “equity-like share” for each market (https://github.com/vegaprotocol/product/pull/323/files)
```
barnabee: Add their actual amount provided (current bond balance)

barnabee (changes requested):

specs/0000-mm-mechanics.md
tamlyn10: I'm actually OK to use account management up to the engineers.. originally I'd specced separate accounts, then Barney's doc said one. I don't mind.

specs/0000-mm-mechanics.md
tamlyn10: Actually last we discussed this the idea was to keep the commitment the same.. to give them a chance to top back up their margin.. this does remind me, however, that we don't talk about this account periodically doing a "search" to top itself back up. mmm.

specs/0000-mm-mechanics.md
tamlyn10: why?

specs/0000-mm-mechanics.md
tamlyn10: yeah, I wasn't clear on how the ordering of this would work. we could accept the commitment and then go through the normal flow of using their bond account for margin if needed.

specs/0000-mm-mechanics.md
tamlyn10: When we have a market lifecycle spec this will be easier to define! (on your list I believe ;-)

tamlyn10: (but I'll add your wording meanwhile)

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,227 @@
...
+
+The point of market making on Vega is to incentivise people to place orders on the market that maintain liquidity on the book. This is done via a financial commitment and reward + penalty mechanics, and through the use of a special batch order type that automatically updates price/size as needed to meet the commitment and automatically refreshes its volume after trading to ensure continuous liquidity provision.
+
+Important note on wording:
+* market making / liqudity COMMITMENTs are the amount of stake a market maker places as a bond on the market to earn rewards.
+* the COMMITMENT is converted to a liquidity OBLIGATION, measured in siskas.
+
+This is the actual outcome...
+* the amount of bond that a market maker actually has, compared to what they COMMITTED is referred to as ACTUAL STAKE.
+* the measure of liquidity that a market maker actually supplies is the ACTUAL STAKE referred to as the SUPPLIED liquidity (measured in siskas).
```
edd: **question (blocking)** I think there's a copy and paste error here - remove ACTUAL STAKE?

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,227 @@
...
+
+## Commit liquidity network transaction
+
+Any Vega participant can apply to market make on a market by submitting a transaction to the network which includes the following
+
+1. Market ID
+1. COMMITMENT AMOUNT: liquidity commitment amount
```
edd: **thought (non-blocking)** It's obvious with some thought, but I think it's worth clarifying that this is in the settlement asset for the market.

specs/0000-mm-mechanics.md
```diff
@@ -0,0 +1,227 @@
...
+* the amount of bond that a market maker actually has, compared to what they COMMITTED is referred to as ACTUAL STAKE.
+* the measure of liquidity that a market maker actually supplies is the ACTUAL STAKE referred to as the SUPPLIED liquidity (measured in siskas).
+
+## Commit liquidity network transaction
+
+Any Vega participant can apply to market make on a market by submitting a transaction to the network which includes the following
+
+1. Market ID
+1. COMMITMENT AMOUNT: liquidity commitment amount 
+1. FEES: nominated [liquidity fee factor](./0029-fees.md).
```
edd: **thought (non-blocking)** This is cleared up later, but as someone who understands fees this surprised me. I assumed they'd earn the fee that is configured for the market. It might be worth making this a link to the fees section below to help get people to the answers they might have on this field.

specs/0000-mm-mechanics.md
tamlyn10: fixed..

specs/0000-mm-mechanics.md
tamlyn10: haha.. i just deleted that as that doesn't need to be specified in the transaction. I'll make a note to make that explicit.

specs/0000-mm-mechanics.md
tamlyn10: I don't know how to link to a line.. but, I also think that it's more important it links to the fee spec?

specs/0000-mm-mechanics.md
tamlyn10: done

specs/0000-mm-mechanics.md
tamlyn10: Unbelievably, I don't recall if we got resolution on this in the end?

specs/0000-mm-mechanics.md
tamlyn10: I don't understand this sentence:  _I also think we should maybe calculate and mostly talk about these in terms of base asset commitment (bond) requirement rather than siskas — only a slight reaarrangement I know, but I think will aid education and understanding._

specs/0000-mm-mechanics.md
tamlyn10: amended

specs/0000-mm-mechanics.md
tamlyn10: added

PR #337, opened 2020-08-12
Update 0029-fees.md

witgaw:
@tamlyn10 (cc @candida-d) pretty sure "notional volume of the trade" should read "notional value of the trade", could you please check if you agree and merge?

### DIFF
--- a/specs/0029-fees.md
+++ b/specs/0029-fees.md
@@ -34,7 +34,7 @@ trade_value_for_fee_purposes:
 * refers to the amount from which we calculate fee, (e.g. for futures, the trade's notional value = size_of_trade * price_of_trade)
 * trade_value_for_fee_purposes is defined on the Product and is a function that may take into account other product parameters 
 
-Initially, for futures, the trade_value_for_fee_purposes = notional value of the trade = size_of_trade * price_of_trade. For other product types, we may want to use something other than the notional volume. This is determined by the Product.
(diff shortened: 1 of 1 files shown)


specs/0029-fees.md
```diff
@@ -34,7 +34,7 @@ trade_value_for_fee_purposes:
 * refers to the amount from which we calculate fee, (e.g. for futures, the trade's notional value = size_of_trade * price_of_trade)
 * trade_value_for_fee_purposes is defined on the Product and is a function that may take into account other product parameters 
 
-Initially, for futures, the trade_value_for_fee_purposes = notional volume of the trade = size_of_trade * price_of_trade. For other product types, we may want to use something other than the notional volume. This is determined by the Product.
+Initially, for futures, the trade_value_for_fee_purposes = notional value of the trade = size_of_trade * price_of_trade. For other product types, we may want to use something other than the notional value. This is determined by the Product.
```
tamlyn10, line 38: I think this is right but we don't define the "size_of_trade", which is actually the notional volume per contract * number of contracts in the trade... and we have to make sure we get our inverses correct if the settlement asset is not teh underlying. It's a pain! I'm trying to get away from using contract terminology but haven't figured it all out, yet. So, I think it's fine to make this change but I will need to work this out really soon.

tamlyn10 (approved):

PR #340, opened 2020-08-20
Update 0028-governance.md

edd:
Attempt to clarify how the participation rate is determined. The spec is intentionally vague so that the calculation has changed, but was also vague enough that the current implementation  is valid but incorrect. First we'll clear up the spec. Then we'll fix the implementation

- Add acceptance criteria

Slack ref: https://vegaprotocol.slack.com/archives/CD4AKTFQT/p1597751818310200


edd:
@barnabee Review this please

PR #341, opened 2020-08-21
Update auctions spec: Add Market Depth requirement

edd:
> Auction Market Data question: I can't find anything in the auction spec so I'll ask here and add the details in. Do we provide normal MarketDepth data during an auction along with extra indicative price+volume? I thought in a normal auction only the best bid/ask + IP/IV was given but as our orders are all visible to any node runner should we publish the full depth?

Specifically require that Market Depth API works in auction mode

- Add acceptance criteria
- Update API section to clarify requirements for existing APIs

Slack ref: https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1598023283003000


barnabee:
Yes, market depth should be available regardless of trading mode. 

Some exchanges have certain auctions that are “level 1 only” (ie no order book / market depth data) but most auctions share market depth, and we aren’t planning on implementing those that don’t because (a) no need; and (b) the we can’t keep order data private technically at the moment.

PR #365, opened 2020-09-29
Add max deposit size

edd:
Add a new acceptance criteria for max deposit size, required for 🤠 net

NOTE: This is a *draft* because the requirement for 🤠 is not yet well understood. It will probably be a *lifetime* max. Or total monetary value. We are unsure. But it'll be *something*, to be decided by @barnabee.

Spec call to be arranged by @tamlyn10 for @C0deMunk33 / @barnabee etc

### DIFF
--- a/specs/0031-ethereum-bridge-spec.md
+++ b/specs/0031-ethereum-bridge-spec.md
@@ -120,3 +120,9 @@ Blacklisting is simply removing an asset from the whitelist
 *  A valid multisig bundle can be passed to the setDepositMinimum function to successfully set a deposit minimum for a given asset
 *  an invalid multisig bundle is rejected by the setDepositMinimum function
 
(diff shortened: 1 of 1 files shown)


C0deMunk33:
so in order to preempt this, Jeremy suggested I add a "max deposit" function to the bridge. This has been added and follows the same pattern as "min deposit", but only applies to individual deposits, not lifetime. This version will go through audit.

Once audit is complete, I will add in the lifetime limits into that mechanism so each asset can be set individually.

tamlyn10:
Note, this is going to be handled in a spec of "training wheels" etc. for WWN. https://github.com/vegaprotocol/research-internal/issues/191

tamlyn10:
New spec here: https://github.com/vegaprotocol/product/pull/467

PR #375, opened 2020-10-06
WIP: Update Market Lifecycle 

tamlyn10:
@davidsiska-vega @barnabee @witgaw  would you guys be able to review this WIP PR to see if you agree with my interpretation of market lifecycle stuff?


tamlyn10:
I'll keep adding to it, but table and the figure are the main things I want to get agreement on initially.

specs/0026-auctions.md
```diff
@@ -14,16 +14,24 @@ They are mostly useful in less liquid markets, or in specific scenarios where a
...
-A market cannot be in both mode at the same time and will trade ever in a auction or continuous trading. There will be the normal trading mode, configured in the market framework, and a period mode which temporarily overrides it. For example a market may be configured to be a Frequent Batch Auction market, but be in an Auction Period triggered by liquidity monitoring.
+## Auction config
 
-In the first implementation, auction periods will be used at opening only as price monitoring will not exist.
+All auctions have a `min_auction_length`, which defines the minimum `call period` for an auction.
 
-There will be a network parameter that sets the minimum allowable duation of the auction period at market creation. New market proposals must specify a period of at least this duration.
+Any auction that would be less than `min_auction_length` seconds should not be started (e.g. if the market is nearing the end of its open period / active trading).
```
davidsiska-vega: Is the `min_auction_length` a network parameter?

specs/0026-auctions.md
```diff
@@ -14,16 +14,24 @@ They are mostly useful in less liquid markets, or in specific scenarios where a
...
+
+## Opening auctions (at creation of the market)
+
+A newly created market will start in auction mode, then once the auction comes to an end the market will switch to the default trading mode, and will stay like until there's a need for it to go in auction mode again (e.g: based on the price changes).
```
davidsiska-vega: "stay like until" feels like there's a word missing.

specs/0026-auctions.md
```diff
@@ -80,7 +88,20 @@ Additional Time in Force order options need to be added: only good for normal tr
...
+
+where `calculated_end_time` is the call period of the auction.
+
+If the market is going to expire the auction must end and uncross at or before expiry.
+
+Any auction that would be less than `min_auction_length` seconds should not be started.
+
+### Exiting during opening auction
+
+The auction should not exit unless at least one trade an ending criteria for the opening auction rather than restart it, we already extend if liquidity criteria aren’t met and I think these things will play better together if we simply have the opening auction last until we are ready to start trading, which given the above should include having had a trade.'
```
davidsiska-vega: "The auction should not exit unless at least one trade an ending criteria" perhaps missing "matching" or a similar word?

davidsiska-vega: I am not sure how "extend if liquidity criteria aren’t met" would work. 

In my imagination it was: end opening auction. 
If there are no trades resulting, restart the auction for the same period. 
If the opening auction results in open interest and stake is zero start a liquidity auction.

specs/0026-auctions.md
```diff
@@ -80,7 +88,20 @@ Additional Time in Force order options need to be added: only good for normal tr
...
+
+Please note auction should only ever be exited if doing so wouldn't trigger entry triggers of any other auction types defined for the market (e.g. price monitoring auction, liquidity monitoring auction). Otherwise the auction gets extended by the period prescribed by the appropriate trigger(s).
```
davidsiska-vega: Why? Let's say I have a price monitoring auction, it ends, uncrosses and results in a price that would trigger price monitoring. Then I start this new price monitoring auction.

specs/0043-market-lifecycle.md
```diff
@@ -1,24 +1,55 @@
...
+| Rejected          |   Yes          |  No        | Governance vote fails/loses                               | N/A                                                    
+| Pending           |   Yes          |  Yes       | Governance vote passes/wins                               | Governance vote (to close) OR enactment date reached
+| Active            |   Yes          |  Yes       | Enactment date reached and usual auction exit checks pass | Governance vote (to close) OR maturity of market      
+| Suspended         |   Yes          |  Yes       | Price monitoring or liquidity monitoring trigger          | Monitoring definition
+| Closed            |   Yes          |  No        | Governance vote (to close)                                | Governance vote (to reopen)                           
+| Trading Terminated|   No           |  No        | Market parameter setting closing date OR defined on the product | Settlement event commences                       
+| Settled           |   No           |  No        | Settlement event concludes                                | N/A                                             
 
+<br>
```
davidsiska-vega: Markdown is not HTML so instead of <br> you can just use newline.

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +60,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
 
   - Passed (yes votes win & thresholds met) → Pending
-  - Failed (no votes win or thresholds not met) → Pending
+  - Failed (no votes win or thresholds not met) → Rejected
 
 **Behaviour:**
 
 - Participants can vote for or against the market
-- Liquidity providers can make, change, or exit commitments
+- Liquidity providers can make, change, or exit commitments (proposer can't commit below proposer minimum)
```
davidsiska-vega: Where /  how is the proposer minimum defined?

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +60,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
 
 
-### Problematic states and their resolutions
+### Trading Terminated
 
-#### Insufficient liquidity to close out a trader
-In the situation that there is insufficient liquidity to close out a trader, the market is put in to a [liquidity sourcing auction](./0026-auctions.md). This will be triggered by [liquidity monitoring](./0035-liquidity-monitoring.md).
+A market terminates trading at some point prior to, or at, the settlement of the product. Markets in this state accept no trading but retain the positions and margin balances that were in place after processing the expiry trigger (which may itself generate MTM cashflows, though for futures it doesn't). A market moves from this termination state to settled when enough information exists and the triggers are reached to settle the market. This could happen instantly upon trading terminating, though usually there will be a delay, for instance to wait for receipt and acceptance of data from a data source (oracle). An example of an instant transition would be where the trigger for terminating trading and the settlement are the publishing of a specific price from another market on the Vega network itself (same shard)
```
davidsiska-vega: From risk point of view: the gap between "trading terminated" and "settlement time" should always be lower than \tau. Otherwise it's very likely that margin balances won't cover settlement.

specs/0026-auctions.md
tamlyn10: Good question.. @barnabee ?

specs/0026-auctions.md
tamlyn10: Gracious, I had pasted some Slack or Git chat and forgot to edit properly.. let me fix and have  you review!

specs/0026-auctions.md
tamlyn10: yeah.. agree..  please see revised edits

specs/0043-market-lifecycle.md
tamlyn10: it's not yet.. line 16 talks of it as a future feature.. it will be a network param in the future I guess..

tamlyn10: @davidsiska-vega  fact, I'm not sure why this is a future feature.. should we not require a minimum of 1 ccy-siska of liquidity? cc @barnabee

specs/0043-market-lifecycle.md
tamlyn10: well, actually no.. it's the gap between "trading terminated"  and "product expiry" that's important..  settlement can happen days after - the calc of what is owed is about expiry. However, expiry is actually defined on the oracle.. hence me excluding it from this lifecycle description.

I think that maybe we have to restrict that "trading terminated" market parameter to make sure that `oracle-data-time` - `trading-termination-time` >= tau

specs/0043-market-lifecycle.md
davidsiska-vega: The special market making order type rounds up when it comes to volume (or it should anyway). So any non-zero stake implies non-zero liquidity commitment which implies non-zero volume on the book. So strictly speaking, I don't see a need for minimum even in the future. This is from the point of view how the liquidity mechanics work. 

There may be a "spam protection" angle or an "DOS attack vector" angle. Say that there are 10^5 liquidity providers in a market each with a miniscule stake - does the core grind to a halt cycling through these and distributing fees?

From this point of view I would say that we should have network-wide-per-asset parameters that say what a minimum commitment is in each asset which would be say roughly USD 1000. 

And I agree - it doesn't have to be future feature, let's discuss and spec what we agree now.

specs/0043-market-lifecycle.md
davidsiska-vega: @tamlyn10 I see and I agree with you. Please add this restriction and mark as resolved.

davidsiska-vega (approved):

specs/0043-market-lifecycle.md
tamlyn10: Yeah, my point is that at the moment you can create markets without submitting a liquidity transaction (because that feature hasn't commenced implementation yet). I think we need to make sure that as part of implementing liquidity, we have this specced properly (and so far it's not, anywhere).  For example, we've not decided whether they need to submit the 2 tx's separately but within x time of eachother, or reference the 1 in the other, or allow them to embedd a liquidity submission tx in with the governance proposal (that would have my vote).

specs/0026-auctions.md
barnabee: Yes, its purpose is to prevent auction calls that are too short given the network latency/granularity, so should be some multiple of the worst case expected block time at some confidence level, which I think is best maintained by governance voting. We should update and be explicit here that it's a network param, possibly explain why it exists too.

specs/0026-auctions.md
```diff
@@ -14,16 +14,24 @@ They are mostly useful in less liquid markets, or in specific scenarios where a
...
 
-In the first implementation, auction periods will be used at opening only as price monitoring will not exist.
+All auctions have a `min_auction_length`, which defines the minimum `call period` for an auction.
 
-There will be a network parameter that sets the minimum allowable duation of the auction period at market creation. New market proposals must specify a period of at least this duration.
+Any auction that would be less than `min_auction_length` seconds should not be started (e.g. if the market is nearing the end of its open period / active trading).
+
+## Opening auctions (at creation of the market)
+
+A newly created market will start in auction mode, then once the auction comes to an end, the market will switch to the default trading mode, and will stay like that until there's a need for it to go in auction mode again (e.g: based on the price changes).
```
barnabee: Think this should now be something like:

>  A market that has passed the governance vote and is in Pending state will be in an auction period. The auction will never uncross while the market is in a Pending state, and only does so when it moves to another state (i.e. becomes Active, see Market Lifecycle for criteria for transiiton out of Pending state).

specs/0026-auctions.md
```diff
@@ -14,16 +14,24 @@ They are mostly useful in less liquid markets, or in specific scenarios where a
...
+A market cannot be in both modes at the same time and will trade either in an opening auction mode or in the defaul trading mode (e.g. continuous trading). This default trading mode is configured in the market framework, and a period mode can temporarily override it. For example a market may be configured to be a Frequent Batch Auction market, but be in an Auction Period triggered by liquidity monitoring.
+
+The enactment period of the governance proposal refers to the time between the proposal being accepted and active trading commencing, therefore inclusive of the opening auction period (see [market lifecycle spec](./0043-market-lifecycle.md)).  A governance network parameter will set the minimum allowable enactment period for new market proposals.
```
barnabee: Maybe ref. Pending state here too

specs/0026-auctions.md
```diff
@@ -80,7 +88,20 @@ Additional Time in Force order options need to be added: only good for normal tr
 
 ## Exiting the auction mode
 
-Please note auction should only ever be exitted if doing so wouldn't trigger entry triggers of any other auction types defined for the market (e.g. price monitoring auction, liquidity monitoring auction). Otherwise the auction gets extended by the period prescribed by the appropriate trigger(s).
+```auction_end_time = min(calculated_end_time, market_expiry)```
+
+where `calculated_end_time` is the call period of the auction.
```
barnabee: Worth noting this may be ∞ (e.g. for liquidity monitoring)

specs/0026-auctions.md
```diff
@@ -80,7 +88,20 @@ Additional Time in Force order options need to be added: only good for normal tr
...
+
+If the market is going to expire the auction must end and uncross at or before expiry.
```
barnabee: Current term here seems to be "Trading Terminated" rather then expired, so I think this should be updated, but see comment below in Market Lifecycle spec.

specs/0026-auctions.md
```diff
@@ -80,7 +88,20 @@ Additional Time in Force order options need to be added: only good for normal tr
...
+
+Any auction that would be less than `min_auction_length` seconds should not be started.
```
barnabee: Should cover here and/or in Market Lifecycle (maybe just Market Lifecycle, thinking about it) what happens if this occurs in a pending/new market (i.e. because we're that close to the end date of the market).

specs/0026-auctions.md
```diff
@@ -80,7 +88,20 @@ Additional Time in Force order options need to be added: only good for normal tr
...
+
+### Exiting during opening auction
+
+The auction should not exit unless:
+
+- there has been at least one trade on the market
```
barnabee: Should we be adding somewhere that during an auction we should not accept an order from a party that is crossed with another order from the same party, to "prevent"[1] wash trading. 

[1] well, barely, as I can use sybils, but it at least puts a cost on it and I end up with positions and margins and have to deal with other people/prices in the market afterwards, plus if I close out straight away we're back in an auction if OI = 0, as I understand it from recent discussions.

specs/0043-market-lifecycle.md
```diff
@@ -1,24 +1,47 @@
 # Market Lifecycle
 
+All markets have a "trading mode" (plus its coniguration) as part of the [market framework](0001-market-framework.md). When a market is Active (i.e. it is open for trading), it will be in a trading period. Normally, the trading period will be defined by the trading mode (additionally, this is one period for the life of the market once it opens, but in future, trading modes may specify a schedule of periods). When created, a market will generally start in an opening auction period. Markets can also enter exeptional periods of either defined or indefinite length as the result of triggers such as price or liquidity monitoring or a governance vote (this spec does not specify any triggers that should switch periods, only that it must be possible).
```
barnabee: probably should be:

> for the life of the market _unless changed by governance_

specs/0043-market-lifecycle.md
```diff
@@ -1,24 +1,47 @@
...
 
 A market can progress through a number of states through its life. The overall market status flow is shown in the diagram below.
 
-![](0029-market-lifecycle-flow-diagram.svg)
+
+DAVID/BARNEY - note, I've suggested we don't use "matured" or "expired" but rather "trading terminated" which is the terminology of CME to describe when that happens at end of trading of a market. Then, after the expiry via the oracle data, the last state could be "settled", though I'd prefer to be more explicit that it's the final settlement (for when we one day have interim settlements).
 
 
-### Active markets
+| State              | New LP         | Trading    | Entry                                                     | Exit
```
barnabee: I think we should add a trading mode column per the 3 bullets above under the "Market status" heading, and it should be either "N/A", "Opening auction", "Exceptional auction" (or similar), or "Normal trading".

specs/0043-market-lifecycle.md
```diff
@@ -1,24 +1,47 @@
...
-![](0029-market-lifecycle-flow-diagram.svg)
+
+DAVID/BARNEY - note, I've suggested we don't use "matured" or "expired" but rather "trading terminated" which is the terminology of CME to describe when that happens at end of trading of a market. Then, after the expiry via the oracle data, the last state could be "settled", though I'd prefer to be more explicit that it's the final settlement (for when we one day have interim settlements).
 
 
-### Active markets
+| State              | New LP         | Trading    | Entry                                                     | Exit                                                  
+|--------------------|----------------|------------|-----------------------------------------------------------|-------------------------------------------------------
+| Proposed          |   Yes          |  No        | Governance vote valid                                     | Governance proposal period ends                       
+| Rejected          |   Yes          |  No        | Governance vote fails/loses                               | N/A
```
barnabee: New LP = No

specs/0043-market-lifecycle.md
```diff
@@ -1,24 +1,47 @@
...
+|--------------------|----------------|------------|-----------------------------------------------------------|-------------------------------------------------------
+| Proposed          |   Yes          |  No        | Governance vote valid                                     | Governance proposal period ends                       
+| Rejected          |   Yes          |  No        | Governance vote fails/loses                               | N/A                                                    
+| Pending           |   Yes          |  Yes       | Governance vote passes/wins                               | Governance vote (to close) OR enactment date reached
+| Active            |   Yes          |  Yes       | Enactment date reached and usual auction exit checks pass | Governance vote (to close) OR maturity of market      
+| Suspended         |   Yes          |  Yes       | Price monitoring or liquidity monitoring trigger          | Monitoring definition
+| Closed            |   Yes          |  No        | Governance vote (to close)                                | Governance vote (to reopen)
```
barnabee: New LP = No

barnabee: I also don't think reopen should be allowed. IMO it should be a new proposal, it's simpler and it ensure it get's properly reviewed like any other market opening.

specs/0043-market-lifecycle.md
```diff
@@ -1,24 +1,47 @@
...
+| Rejected          |   Yes          |  No        | Governance vote fails/loses                               | N/A                                                    
+| Pending           |   Yes          |  Yes       | Governance vote passes/wins                               | Governance vote (to close) OR enactment date reached
+| Active            |   Yes          |  Yes       | Enactment date reached and usual auction exit checks pass | Governance vote (to close) OR maturity of market      
+| Suspended         |   Yes          |  Yes       | Price monitoring or liquidity monitoring trigger          | Monitoring definition
+| Closed            |   Yes          |  No        | Governance vote (to close)                                | Governance vote (to reopen)                           
+| Trading Terminated|   No           |  No        | Market parameter setting closing date OR defined on the product | Settlement event commences                       
+| Settled           |   No           |  No        | Settlement event concludes                                | N/A
```
barnabee: I'm wondering whether we need to differentiate between **settled** and **closed** — I suspect not, and if that's the case let's drop either one of them or pick something new?

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+- No exit. End state.
+
+**Behaviour:**
+
+- Nothing can happen to the market with this status - it is an inert state.
+
 ### Pending
 
-Once creation of a market is approved via a governance proposal, or (in future) when a market is scheduled to be created as part of a series of auto-generated markets, it enters a pending state prior to being created. A Pending market becomes an Active market when both the enactment date is reached and it has met the minimum liquidity commitment requirement. A market will be cancelled before creation if the maximum time allowed (network param) to collect liquidity is excdeed, if the market reaches its expiry date (if applicable) whilst in a pending state, or via a governance vote.
+If a "new market" governance proposal is successful the market is opened into a "pending" state, where the price determination method is an auction period with a callPeriod that ends at the enactment date, specified in the governance proposal.
```
barnabee: Do we want to make it clear that this is a state for _any market that is due to be created_ and that currently this means by governance proposal, but in future there may be automated market creations? e.g. a series of markets, creation from an oracle/data source, etc.

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
 **Exit:**
 
-- Enactment date is reached and liquidity stake committed >= minimum required launch liquidity → Active
-- Expiry date is reached and liquidity stake committed < minimum required launch liquidity → Cancelled
-- Maximum time in Pending (a network parameter) is reached and liquidity stake committed < minimum required launch liquidity → Cancelled
-- Market change governance vote approves closure/cancellation of market → Cancelled
+- Auction period ends when either of the following occur:
+
+  - Enactment date is reached and the usual [ending of auction checks pass](./0026-auctions.md) → Active
+  - Market change governance vote approves closure/cancellation of market → Cancelled
```
barnabee: Cancelled is not described or in the table

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
 
 **Behaviour:**
 
 - Liquidity providers can make, change, or exit commitments
-- No trading is possible, no orders can be placed (except the liquidity provider order/shape that forms part of their commitment)
-- No market data (price, etc.) is emitted, no positions exist on the market, and no risk management occurs
```
barnabee, line 65: I had mentioned this for clarity in my draft, maybe worth adding here?

> Only market data related to auctions (indicative uncrossing price/volume, etc.) are emitted, no positions exist on the market, and no risk management occurs

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+- Auction orders are accepted as per [any regular auction period](./0026-auctions.md).
+- Margins on orders as per auction based instructions in [margin calculator spec](./0019-margin-calculator.md).
+
 
+### Active
 
+Once the enactment date is reached the market becomes Active. This status indicates it is trading via it's normally configured trading mode according to the market framework (continuous trading, frequent batch auction, RFQ, block only, etc.). The specification for the trading mode should describe which orders are accepted and how trading proceeds. The market can become Expired via a product trigger (for futures, if the trading termination date is reached) and can be temporarily suspended automatically by various monitoring systems ([price monitoring](./0032-price-monitoring.md), [liquidity monitoring](./0035-liquidity-monitoring.md)). The market can also be closed via a governance vote (market parameter update) to change the status to closed.
```
barnabee: it's → its

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+
+**Entry:**
+
+- Enactment date reached
+- Conditions specified in [price monitoring](./0032-price-monitoring.md) and [liquidity monitoring](./0035-liquidity-monitoring.md) are met for the market to exit the suspended status back to Active.
```
barnabee: Above you used this to specify exiting an auction, which I quite like, would it also work here?

> the usual [ending of auction checks pass](./0026-auctions.md)

specs/0043-market-lifecycle.md
barnabee: expired → something about trading terminated

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+
+**Exit:**
 
+- Price, liquidity or other monitoring system triggers suspension → Suspended
+- Trading termination is triggered (defined by a market parameter) → Trading Terminated
```
barnabee: > by a product trigger (for futures, if the trading termination date, set by a market parameter, is reached)

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+
+**Entry:**
+
+- Price, liquidity or other monitoring system triggers suspension → Suspended
+
+**Exit:**
 
-#### Insufficient liquidity to close out a trader
-In the situation that there is insufficient liquidity to close out a trader, the market is put in to a [liquidity sourcing auction](./0026-auctions.md). This will be triggered by [liquidity monitoring](./0035-liquidity-monitoring.md).
+- Conditions specified in [price monitoring](./0032-price-monitoring.md) and [liquidity monitoring](./0035-liquidity-monitoring.md) are met for the market to exit the suspended status back to Active.
```
barnabee: Same comment as above re: using this instead, it means only one place to change if we add more checks:

> the usual [ending of auction checks pass](./0026-auctions.md)

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+
+**Behaviour:**
+
+- Governance vote to close a market passes → Closed
+- Orders may be cancelled, no new orders accepted.
+
+
+### Trading Terminated
+
+A market terminates trading at some point prior to, or at, the settlement of the product. Markets in this state accept no trading but retain the positions and margin balances that were in place after processing the expiry trigger (which may itself generate MTM cashflows, though for futures it doesn't). A market moves from this termination state to settled when enough information exists and the triggers are reached to settle the market. This could happen instantly upon trading terminating, though usually there will be a delay, for instance to wait for receipt and acceptance of data from a data source (oracle). An example of an instant transition would be where the trigger for terminating trading and the settlement are the publishing of a specific price from another market on the Vega network itself (same shard)
```
barnabee: > A market that has a finite lifetime and thus a defined final settlement terminates trading at some point prior to, or at, the final settlement driven by the product.

specs/0043-market-lifecycle.md
barnabee: We may not always have a way to know that `oracle-data-time - trading-termination-time >= tau` as it may just not be published data, even if it is, you're trusting the proposer at proposal time.

So if we add a check for this we'll basically be saying "enter a time and we'll validate that it's reasonable, but the governance voters have to validate that the oracle feed supplies a data point for the time the proposer said" whereas without the validation it will be "governance voters have to check they believe oracle feed specified will supply a data point that meets the tau criteria"… not hugely different.

There may still be value in the slightly more explicit version, but just wanted to be clear that we cannot get certainty on this until after the fact, and are trsuting the governance process either way.

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+
+**Entry:**
+
+- Triggered by market parameter OR defined on the product.
```
barnabee: It's just triggered by the product. The futures product uses the market parameter to effect the trigger, and this can be noted, though, i.e. something like this wording again:

> by a product trigger (for futures, if the trading termination date, set by a market parameter, is reached)

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+
+**Exit:**
+
+- Settlement dependencies met (i.e. oracle data received) → Settled
+
+**Behaviour:**
+
+- No trading occurs, no orders are accepted
+- Mark to market settlement is performed if required after termination is triggered then never again
+- No market data are emitted
```
barnabee: Is this right? (I know I wrote it...) 

Wondering if it should be something like: 

> A single set of market data may be emitted for the final settlement data (e.g. settlement mark price), after which no market data are emitted.

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+- No risk management of price/liquidity monitoring occurs
```
barnabee: of → or

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+
+
+### Settled
+
+Once the required data to calculate the settlement cashflows is available for an Expired market, these cashflows are calcualted and applied to all traders with an open position (settlement). The positions are then closed and all orders cleared. All money held in margin accoutns after final settlement is returned to traders' general accounts. The market can be deleted entirely at this point, from a core perspective. Any insurance pool funds are distributed as per the insurance pool spec.
```
barnabee: calcualted → calculated

barnabee: Probably should link section of insurance pool spec on funds redistribution

specs/0043-market-lifecycle.md
```diff
@@ -29,47 +52,150 @@ Markets created via [governance proposals](./0028-governance.md#1-create-market)
...
+- No exit. End state.
+
+**Behaviour:**
 
-At the auction end, all positions are marked to market, and closeout trades will be created as usual.
+- No trading occurs, no orders are accepted
+- All final settlement cashflows are calculated and applied (settled)
+- Margins are trannsferred back to general accounts
+- Insurance pool funds are redistributed
+- Market is over and removed.
```
barnabee: Probably should link section of insurance pool spec on funds redistribution

barnabee (changes requested):
Think this is looking really good. Few small things but nothing major.

barnabee:
I can update the diagram in OmniGraffle if you need me to @tamlyn10 - pls just let me know what to change when ready, if it's not obvious.

specs/0026-auctions.md
tamlyn10: ok, thanks, will include, and resolve this.

specs/0043-market-lifecycle.md
tamlyn10: @barnabee have you had any thoughts on this?

specs/0043-market-lifecycle.md
tamlyn10: @barnabee to clarify, you mean that the `oracle-data-time` represents the product expiry (i.e. not the time the data is expected to be delivered).. I'm sure you do, just writing for future selves.

So yes, agree that this is all encapsulated in the governance vote, anyway. But I do think that there is value in making this explicit, if just for the teaching it provides of what margins actually are.

specs/0043-market-lifecycle.md
tamlyn10: OK, so you think that this should always be a parameter that is set in reference to the oracle data date? I have a slightly odd feeling about that, not sure why... perhaps because in future, there may be products that settle against more that one underlying.

tamlyn10:
@barnabee I'll take you up on offer to edit diagram - let's do as final thing once we have agreement on above points.

specs/0043-market-lifecycle.md
tamlyn10: Mmm... I think it's useful to know if a market has been settled by expiry rather than closed and then arbitrarily (or at least unspecced) settlement / unwind of positions procedure.

specs/0043-market-lifecycle.md
tamlyn10: changed to closed

specs/0043-market-lifecycle.md
tamlyn10: no positions exist on the market, and no risk management occurs doesn't seem correct to me. If it's a price monitoring auction positions do exist and orders are charged margin.. as per the 0019 margin spec.

specs/0043-market-lifecycle.md
tamlyn10: i encorporated it..

specs/0043-market-lifecycle.md
tamlyn10: there isn't just one place tho'.. is there? or should there be.. mmmm

specs/0043-market-lifecycle.md
barnabee: I think I'd be happy with something like a per-asset) minimum proposer liquidity stake_ as a network parameter, to provide a barrier to entry that means we only accept sufficiently "serious" proposals to create markets and supplied liquidity.

Re @davidsiska-vega's DOS angle… good point, I think we should have a per-asset absolute _minimum liquidity stake_ for all providers too, which would be needed to prevents 1000s of 0.000000000000000001 ETH stakers screwing up the system.

tamlyn10: So @barnabee this would form part of the validation of the proposal itself?

specs/0043-market-lifecycle.md
tamlyn10: I think I made this edit anyway.. so will resolve this.

specs/0043-market-lifecycle.md
barnabee: Isn't this describing the Pending state? (Genuine question - the Github view is full of comments and diffs that makes it hard to see the whole doc.)

If so then it can't be in price monitoring.

specs/0043-market-lifecycle.md
barnabee: There used to be: https://github.com/vegaprotocol/product/pull/375/files#diff-9993b28fdf1ab342cc13c1be8a941439355efb8054883586f1db86da001c94acL83

I don't really see why we wouldn't want there to be? It makes sense to me that there would be a standard set of conditions to enter and exit "Suspended" mode, and that these would be enumerated and refernced in a single place (either in auction or market lifecycle spec), so we only have one place to update if we add new monitoring capabilities.

specs/0043-market-lifecycle.md
barnabee: No, I think this is **never** triggered directly by a parameter. It is triggered by code in the product implementation, it just so happens that for futures the code references a product parameter (which all sit in the market framework). I don't think this would relate to the oracle, it's a parameter that the product needs to decide whether to trigger the state change.

specs/0043-market-lifecycle.md
tamlyn10: OMG I find Github soooo difficult to do this review. I often don't know where in the file the snippet sits! If it is pending, I agree with you and will change if so!

specs/0026-auctions.md
barnabee: I don't think this is correct. the [price monitoring spec.](https://github.com/vegaprotocol/product/blob/master/specs/0032-price-monitoring.md) states explicitly that:

- price monitoring is extended if further levels are breached (i.e. you'd never uncross price monitoring then go back into another price monitoring auction).
- "price monitoring is meant to stop large market movements that are not "real" from occuring, rather than just detect them after the fact. To that end, it is necessary to pre-process every transaction and check if it triggers the price monitoring action" — that is, we would not process an uncrossing and generate trades if they would trigger price monitoring.

So I think it's correct that auctions shouldn't be exited if price monitoring would be triggered, but should be extended with the call duration defined as per the price monitoring spec.

specs/0026-auctions.md
edd: Also: Please clarify how an opening auction acts if it has not had a single trade. I know the high level @the auction happens again' - presumably for the same period?

campbellssource: And what if it never ends? Is there a max amount of time or attempts before it becomes a failed market? What happens to peoples liquidity if/when it fails?

tamlyn10: Seriously.. Github???? Where do you hide these comments.. 

@edd yes, we need to clarify that - it's not explicit. I'm going to suggest that after the enactmentDate, any single trade could be the trigger to make it exit (subject to other exit conditions such as liquidity).

@campbellssource good question. LP's may add and remove their stake over this period (unclear if the proposor can) and yes, with no trade, the market may just remain in a pending state forever. However, I'd expect it to be closed by governance vote (this is unspecced).

barnabee: We _could_ have a _max opening auction extension_ network param so that market creation/opening is "cancelled" if it doesn't get created within reasonable time of the planned creation. This could be infinite/extremely large to mimic the current functionality but having it would allow community experimentation and engagement through governance.

specs/0043-market-lifecycle.md
barnabee: Yep... I actually can't tell either. GH is infuriating!

specs/0043-market-lifecycle.md
barnabee: Minimum proposer liquidity stake = proposal validation
Minimum liquidity stake = new/amended commitment validation

barnabee (commented):
Final comments as far as I can tell

specs/0026-auctions.md
```diff
@@ -14,16 +10,25 @@ They are mostly useful in less liquid markets, or in specific scenarios where a
...
+
+1. General auctions
+1. Opening auctions: purpose is to calibrate a market / help with price discovery when a new market is started.
+1. Frequent batch auctions: a trading mode that may be set as the default / normal mode that has trading occur only through repeated auctions (as opposed to continuous trading)
+
+## Auction config
+
+All auctions have a `min_auction_length`, which defines the minimum `call period` for an auction.
+
+Any auction that would be less than `min_auction_length` seconds (network parameter) should not be started (e.g. if the market is nearing the end of its open period / active trading). This is to prevent auction calls that are too short given the network latency/granularity, so should be some multiple of the worst case expected block time at some confidence level, which is best maintained by governance voting (hence being a network parameter).
```
witgaw: What if the auction length is less than `min_auction_length`, but there's still plenty of time until end of open period? Shouldn't we just run the auction until it's at least ``min_auction_length` long? 

Otherwise one can never be sure if price monitoring triggers will work as that network parameter can be changed later on effectively rendering them inactive.

tamlyn10:
Adding a link to the google sheet from yesterday's meeting where we should all contribute any changes we notice that will need implementation work. This will help with prioritisation. https://docs.google.com/spreadsheets/d/1XVFcvuiEvYcU2JPKxIsYZp1InRABSP0koWSM_RR8I-Y/edit?usp=sharing

specs/0026-auctions.md
barnabee: Hmm yes I think for price monitoring it should be the price monitoring spec that specificies the end time as `max(now + configured_period_length, now + min_auction_length)` and then this auction logic does not need to contain a specific case for price monitoring and can remain generic.

tamlyn10: @witgaw could you make any changes that arise from this or anything else you come across in the PR?

specs/0043-market-lifecycle.md
tamlyn10: I'm just reviewing this and I think that I'm not going to make any market creation / market framework spec changes for this. I'll copy the thinking into an analysis ticket to not lose it.

barnabee: Tamlyn's ticket is here FYI https://github.com/vegaprotocol/product/issues/392

tamlyn10:
I will put in a meeting for Monday 2nd November. Core to read through the spec prior. Tamlyn to make further updates as in spreadsheet.

specs/0026-auctions.md
witgaw: Sure, will do.

specs/0026-auctions.md
```diff
@@ -14,16 +10,25 @@ They are mostly useful in less liquid markets, or in specific scenarios where a
...
-As mentioned earlier, this specification introduces new trading modes. A first one which purpose is to calibrate a market / help with price discovery when a new market is started. A second one meant to be trading only through auction called `Frequent batch auction`.
+As mentioned earlier, this specification introduces new trading modes. 
+
+1. General auctions
+1. Opening auctions: purpose is to calibrate a market / help with price discovery when a new market is started.
+1. Frequent batch auctions: a trading mode that may be set as the default / normal mode that has trading occur only through repeated auctions (as opposed to continuous trading)
+
+## Auction config
+
+All auctions have a `min_auction_length`, which defines the minimum `call period` for an auction.
```
jeremyletang: Where / how is this defined?
At the proposal stage as part of the proposal? Should we reject a proposal with a min_auction_length shorter than the network parameter?

If the min_auction_length network parameter change after the market is started what's supposed to happen and the min_auction_lenght of the market become shorter than the network parameter?

Also if not mistaken price monitoring auction duration is calculated dynamically, so what should happen if the duration calculated is shorted than the network parameter?

specs/0026-auctions.md
```diff
@@ -80,7 +86,27 @@ Additional Time in Force order options need to be added: only good for normal tr
 
 ## Exiting the auction mode
 
-Please note auction should only ever be exitted if doing so wouldn't trigger entry triggers of any other auction types defined for the market (e.g. price monitoring auction, liquidity monitoring auction). Otherwise the auction gets extended by the period prescribed by the appropriate trigger(s).
+Auction periods may be ended which an uncrossing and the creation of any resulting trades due to:
```
jeremyletang: which -> with?

specs/0026-auctions.md
```diff
@@ -80,7 +86,27 @@ Additional Time in Force order options need to be added: only good for normal tr
...
+
+- the auction call period end time being reached (if such a time is set); or 
+- other functionality (related to the type of auction period) that triggers the end of auction.
+
+Auction periods do not end if the resulting state woudld immediately cause another auction to occur. Instead an _extension period_ is added to the auction of the relevant type. For example, if a liquidity monitoring auction would be triggered at the end of an opening auction, then the auction continues and the market is in a "liquidity monitoring extension [auction]".
+
+
+### Ending at scheduled end time
+
+If the auction period specifies and end time and the market is about to transition to the "Trading Terminated" state before the auction end time is reached, then the auction must end immediately before the transition occurs.
```
jeremyletang: I'm not sure to understand this whole sentence?
I can see only 1 reference to Trading Terminated in this file, does that mean, that if the market is about to reach expiry before an auction end time is reached, we need to get back to the normal trading mode before we expire?

specs/0028-governance.md
```diff
@@ -151,10 +151,18 @@ We allow users to submit proposals covering 3 types of governance action:
 
 ## 1. Create market
 
-A proposal to create a market contains a compelte market specification as per the Market Framework (see spec) that describes the market to be created. 
+This action differs from from other governance actions in that the market is created and some transactions (namely around liquidity provision) may be accepted for the market before the proposal has succesfully passed. The lifecycle of a market and its triggers are covered in the [market lifecycle](./0043-market-lifecycle.md) spec.
+
+A proposal to create a market contains 
+1. a complete market specification as per the Market Framework (see spec) that describes the market to be created. 
+1. a liquidity provision commitment via LP commitment data structure, specifying stake amount, fee bid, plus buy and sell shapes [see lp-mechanics](0044-lp-mechanics.md).
```
jeremyletang: So this is literally required as part of the proposal right?

if the liquidity provision commitment submitted is not enough / the party does not have enough funds the proposal should be rejected?

specs/0028-governance.md
```diff
@@ -151,10 +151,18 @@ We allow users to submit proposals covering 3 types of governance action:
...
+
+A proposal to create a market contains 
+1. a complete market specification as per the Market Framework (see spec) that describes the market to be created. 
+1. a liquidity provision commitment via LP commitment data structure, specifying stake amount, fee bid, plus buy and sell shapes [see lp-mechanics](0044-lp-mechanics.md). 
+The stake commitment must exceed the `minimum_proposal_stake_amount` which is a per-asset parameter.
+1. an enactment time that is at least the *minimum auction duraton* after the vote closing time (see [auction spec](./0026-auctions.md))
 
 All **new market proposals** initially have their validation configured by the network parameters `Governance.CreateMarket.All.*`. These may be split from `All` to subtypes in future, for instance when other market types like RFQ are created.
 
+Note that additional requirements on validity of the governance proposal may in future include a requirement that the proposer also commits to [supply liquidity to the market](./0044-lp-mechanics.md). More details in the [market lifecycle spec](./0043-market-lifecycle.md).
```
jeremyletang: Well how does that work with my previous comment, this seems to imply that the things at line 158 are not required as of now?

specs/0026-auctions.md
jeremyletang: I see it's in the list of market states..

jeremyletang:
Would it be possible to commit this file: specs/0029-market-lifecycle-flow-diagram.graffle in a format that can be shown in the github UI?

specs/0026-auctions.md
barnabee: * min auction length is a network param and not provided on the proposal
* a proposal should be rejected if it would require an auction shorter min_auction_length
* for price/liquidity monitoring, etc. the auction must last for at least the min_auction_length and therefore we can avoid checking other conditions until that length is reached
* if the parameter is changed it needs to be re-applied to any current auctions, this means that shortening it could trigger an auction ending

I've added these in the latest commit to the spec

specs/0026-auctions.md
barnabee: fixed

specs/0026-auctions.md
barnabee: updated the spec, the auction uncrosses immediately in this case and does not go back to normal trading

specs/0028-governance.md
barnabee: yes, updated spec

specs/0028-governance.md
barnabee: removed, this note was from a previous version of the spec when we hadn't specced liqudiity yet

PR #393, opened 2020-10-28
Updated parameter ranges and default values.

davidsiska-vega:
Hi @jeremyletang , @tamlyn10 , 

This should clear up @jeremyletang 's query about valid ranges and default values. @tamlyn10 please check whether it makes sense as you wrote this spec. Shouldn't take more than a couple of minutes. 

Cheers,
David

### DIFF
--- a/specs/0044-lp-mechanics.md
+++ b/specs/0044-lp-mechanics.md
@@ -191,10 +191,11 @@ Important: a trader's general account should be periodically searched to top bac
 
 
 ## Network parameters 
-- `bond-penalty-parameter` - used to calculate the penalty to liquidity providers when they fail to meet their obligations.
+- `bond-penalty-parameter` - used to calculate the penalty to liquidity providers when they fail to meet their obligations. 
+Valid values: any decimal number `>= 0` with a default value of `1`.  
 - ~~`market-size-measurement-period` - used in fee splitting~~
-- `maximum-liquidity-fee-factor-level` - used in validating fee amounts
-- `stake_to_ccy_siskas` - used to translate a commitment to an obligation (in siskas)
+- `maximum-liquidity-fee-factor-level` - used in validating fee amounts. Any decimal number `>0` and `<=1`. Default value `1`.
+- `stake_to_ccy_siskas` - used to translate a commitment to an obligation (in siskas). Any decimal number `>0` with default value `1`.
 
 ## What data do we keep relating to liquidity provision?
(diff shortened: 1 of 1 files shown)


specs/0044-lp-mechanics.md
```diff
@@ -191,10 +191,11 @@ Important: a trader's general account should be periodically searched to top bac
 
 
 ## Network parameters 
-- `bond-penalty-parameter` - used to calculate the penalty to liquidity providers when they fail to meet their obligations.
+- `bond-penalty-parameter` - used to calculate the penalty to liquidity providers when they fail to meet their obligations. 
+Valid values: any decimal number `>= 0` with a default value of `1`.
```
tamlyn10: yeah, this is probably OK but I'd expect us to be setting this more at around the 10% level typically.

specs/0044-lp-mechanics.md
```diff
@@ -191,10 +191,11 @@ Important: a trader's general account should be periodically searched to top bac
 
 
 ## Network parameters 
-- `bond-penalty-parameter` - used to calculate the penalty to liquidity providers when they fail to meet their obligations.
+- `bond-penalty-parameter` - used to calculate the penalty to liquidity providers when they fail to meet their obligations. 
+Valid values: any decimal number `>= 0` with a default value of `1`.  
 - ~~`market-size-measurement-period` - used in fee splitting~~
-- `maximum-liquidity-fee-factor-level` - used in validating fee amounts
-- `stake_to_ccy_siskas` - used to translate a commitment to an obligation (in siskas)
+- `maximum-liquidity-fee-factor-level` - used in validating fee amounts. Any decimal number `>0` and `<=1`. Default value `1`.
```
tamlyn10: We probably need to add the use of this into the description above. I actually don't know what it's for...

tamlyn10 (changes requested):

jeremyletang (approved):
Sounds good to me.

specs/0044-lp-mechanics.md
davidsiska-vega: I'll set the default to 10%. This means that if your margin shortfall is 200 then the penalty will be 20.

specs/0044-lp-mechanics.md
davidsiska-vega: It's for validating the lp-provider batch order, namely the fee amount. I'll clarify this.

PR #395, opened 2020-10-28
Rename parameters

witgaw:
Rename parameters to be more in line with convention used in code.


witgaw (commented):
Will merge now, discussed the change with David on Slack already.

PR #409, opened 2020-11-04
Adding amends details for pegged orders

### LINKED ISSUE #408: Add amends support for pegged orders
The spec for pegged orders does not say if they support amends. Here are the features they support:
 * Amend size
 * Amend reference
 * Amend offset
 * Amend GTT->GTC and GTC->GTT
 * Amend expiry time

peterbarrow:
Added details about what can be amended for pegged orders

Closes #408


witgaw (approved):
Looks good to me.

PR #410, opened 2020-11-09
Clarify that we count all orders towards liquidity

witgaw:
To me, the spec reads like only the orders submitted via `liquidity provider orders` get counted towards the provided liquidity, but based on this conversation it's all of the orders on the book submitted by a party that's a liquidity provider:
https://github.com/vegaprotocol/vega/issues/2399#issuecomment-723174708

Made a few minor edits to clarify that.

### DIFF
--- a/specs/0044-lp-mechanics.md
+++ b/specs/0044-lp-mechanics.md
@@ -153,15 +153,16 @@ Note here "ccy" stands for "currency" and that the liquidity measure units are a
 
 **During continuous trading:**
 A liquidity provider complies with their liquidity provision obligation by:
+1. Submitting valid orders of any type to the book.
 1. Submitting valid _liquidity provider orders_ in the liquidity provider network transaction.
-1. Holding sufficient collateral to meet the usual margin obligations associated with these orders. 
+1. Holding sufficient collateral to meet the usual margin obligations associated with all the submitted orders.
 
 Since liquidity provider orders automatically refresh, a liquidity provider is only non-compliant when they have insufficient capital to meet the margin requirements of these orders.
 
 **During auction:**
 - liquidity provider obligation during auction (including market commencement auction):
-    - [ ] liquidity provider pegged orders that are placed during an auction call period are parked and reinstated when the limit order book is reinstated.
-	- [ ] At conclusion of auction period call period, liquidity provider's pegged orders are reinstated.
+    - [ ] liquidity provider [GFN](./0014-order-types.md) orders (both orders submitted as part of _liquidity provider orders_ and any other orders submitted in regular way) that are placed during an auction call period are parked and reinstated when the limit order book is reinstated.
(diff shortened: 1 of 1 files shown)


specs/0044-lp-mechanics.md
```diff
@@ -153,15 +153,14 @@ Note here "ccy" stands for "currency" and that the liquidity measure units are a
...
 
 **During auction:**
-- liquidity provider obligation during auction (including market commencement auction):
-    - [ ] liquidity provider pegged orders that are placed during an auction call period are parked and reinstated when the limit order book is reinstated.
-	- [ ] At conclusion of auction period call period, liquidity provider's pegged orders are reinstated.
+- liquidity provider can still submit/amend/cancel any orders during the auction. However, depending on type of the auction some of them might be rejected (e.g. a GFN limit order will be rejected in all auction modes, a request to decrease _liquidity provision obligation_ will be denied during the liquidity monitoring auction as if there was enough liquidity in the market the auction wouldn't have happened). The supplied liquidity will be calculated assuming the market is in it's default trading mode - so even though the pegged orders will be parked for the duration of the auction the liquidity that they would've supplied under normal trading mode gets counted towards the supplied liquidity.
```
witgaw, line 163: @tamlyn10 do you agree with the change? I talked with @davidsiska-vega about this and we've decided to remove references to order mechanics (how they get parked and unparked and all that as this should already be covered in auction spec and we don't want to duplicate it).

I substituted it with a description of how supplied liquidity can be affected by the LPs during auction and how we calculate it. Please feel free to commit any edits you may want to make directly to this.

barnabee: During an auction can we not simply assume that a committed LP's supplied liquidity is always equal to the amount calculated from their stake? It doesn't seem like we would need to calculate the amounts supplied by orders at all? 

What am I missing?

tamlyn10: Yeah, I agree with that @barnabee but correct me if I'm wrong, @witgaw , I think the drafting is trying to say that a liquidity provider won't be penalised even though they're not providing liquidity in the auction.

witgaw: Ok, can I just change it to:
> **During auction:**
> Liquidity provider obligation is not evaluated.

?

edd: @barnabee this is waiting for feedback from you

barnabee: I'm ok with that if @tamlyn10 and @davidsiska-vega have no objections. I'd like their sanity check though, in case I missed something.

tamlyn10:
@peterbarrow could you confirm whether this spec should be merged?

tamlyn10:
(and then merge)

peterbarrow (approved):

PR #417, opened 2020-11-12
Clarified pegged orders are not part of best price and volume calc

peterbarrow:
Updated pegged order and market data spec to clarify that pegged orders should not be used to calculate the best prices and volume values.


barnabee:
Question on whether we should publish the "full" best prices as well as the one that excludes pegs in the market data messages. I feel like the answer is probably yes… cc @tamlyn10

tamlyn10:
I'm confused - do you mean the Best bid / Best offer at a point in time? Why would pegged orders not be used for this? They essentially are, at a point in time, a limit order, right? In any case, Barney's suggestion works for me.

barnabee:
@tamlyn10 There is a bunch of volume not included in best price calculations on many platforms, it is just that there's usually no way for traders to know those orders existed at all until they trade (and even then they often don't know what type they were or total size). On Vega people can find out about every order if they read the incoming transactions so we don't have that option hence if we calculate the best price you'd have seen on a normal exchange, and that we need to use for peg pricing, it doesn't include all the orders you know about (although we could treat them as hidden/not shown in our APIs, someone would eventually build better APIsperhaps privately, so it's better IMO that we give the data to all in a nice format up front to level the playing field).

The firm/real/non-derived best price is possibly useful to traders because it's the best "explicit" price (i.e. one that a trader actively chose) and it is the one that's used for pegs, so traders probably will want to know it, however it also seems like the best price including everything may indeed be the best one to show normally (which is different to other platforms).

So I think I've convinced myself:

- best bid/offer: calculated using _every_ order and disseminated in the market data streams/APIs we have now
- firm/static best bid/offer: excludes derived prices like pegs, disseminated via a new field, used for peg offsets

@peterbarrow does that seem reasonable and in line with your thinking?

peterbarrow:
I agree with what @barnabee has said above but I have one more question, do we supply the volumes with both sets of prices?

barnabee:
@peterbarrow I think we need volume at the best prices that include all orders at least, but if we calculate both then I'd sat lets supply both.

tamlyn10:
@peterbarrow I'm happy for this to be merged - but do you mind checking over the conflict and resolving, then merging? Thank you!

PR #421, opened 2020-11-23
WIP trusted party oracles

tamlyn10:


specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+## Proposing Changes to Oracle
+
+A future feature requirement - any participant may propose a change to a market's oracle data source via on-chain governance. This includes:
+1. Proposing a change in oracle type - e.g. a non TPO / NDS
+1. Proposing a change to the oracle values, such as a change in the sourcePubkeys or filters.
+
+
+## Contingencies
+
+### What happens when Oracle Data doesn't arrive?
```
jeremyletang: Is this a market parameter? Or defined per market?

jeremyletang: Should we also default to this if no oracle is provided for a market? Or is it mandatory to provide one?

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Futures
+A [Cash Settled Direct Futures](./xxxx) product requires specification of a settlementPriceSource for settlement of the product (at expiry). TODO: Move this to the WIP built in futures spec!!
+
+```
+Future {
+    ... 
+    settlementPriceSource:  { 
+        signedMessage: {
+	    sourcePubkeys: ["A VEGA_PUBKEY_HERE"],
```
jeremyletang: would we want some maybe slightly more complex datastructure to specify the pubkeys?
I'm thinking of something like metada, just string this would allow maybe to add information about the pubkey, just so people may know/help identify what's the entity behind the pubkey providing the oracle data?

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+- A valid TPO that has an existing commitment to provide data is referred to in this section as an Active TPO
+
+#### Inactive Trusted Party Oracle
+
+- A valid TPO that has no existing commitments to provide data is referred to in this section as an Inactive TPO
+    - Note: that most accounts on the network that have ever traded thus count as 'Inactive TPOs' 
+
+### Invalid Trusted Party Oracle
+A invalid Trusted Party Oracle (TPO) is:
```
jeremyletang, line 127: Would we want something as well like: The pubkey needs to have some funds deposited on vega, like governacne token or something?

jeremyletang: OK this is answered line 114.

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
+# Trusted Party Oracles
+Trusted Party Oracles are the name for the Minimum Viable Data Source. Rather than implement the entire [Data Sourcing spec](https://github.com/vegaprotocol/product/pull/289), we are going to add a Vega transaction that represents a data result, which can be used when [settling a market at expiry](./0004-settlement-at-instrument-expiry.md).
```
barnabee: This is not _instead of_ it is the first part and a key building block of. Point is this shouldn't be considered throwaway or something that is a non-strategic part of the protocol and I think this sentence may give that impression.

barnabee: > add a Vega transaction that represents a data result, which can be used when [settling a market at expiry](./0004-settlement-at-instrument-expiry.md)

I would reword to say this spec adds:

- a transaction that allows arbitrary signed data to be submitted to the Vega blockchain (creating a stream of data that can be identified by a data source definition)
- a way to define a data source that reads specific data from such a stream (and discards other data)
- the ability to define the settlement oracle for a market to use such data source definition (NB: there can and will be other places the EXACT SAME features described above will be used, but for now, it is just settlement data)

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+_Trusted Party Oracles_ refer to a type of oracle that may be specified on a market. This data source consists of nominating one or more Vega participants (pubKeys) who are permitted to submit and attest to the relevant data required by the market. In initial markets on Vega (and beyond) the nominated Vega participants submit the data at market expiry.
```
barnabee: They can submit data at any time. Not all of it will be used, and a single stream of submissions could be used by many markets.

For example, the Coinbase oracle API provides a stream of signed messages for many different crypto prices, so for one source we could be receiving many messages that aren't used by any given market, for instance if someone set a bot to submit the latest Coinbase oracle signed messages to the Vega chain once per hour. (Hence the "filters" section in the definition to determine which messages to look at.)

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+Data is supplied through a transaction being posted directly on the Vega network that is signed by one (or more) of the valid signers for that specific data source.
```
barnabee: One or more transactions

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+Note: With this type of oracle there’s no incentive in the Vega data source system, you’re trusting the keyholder(s) at settlement.
+
+*NOTE: This is the only version available in testnet, and initially will require only one of the specified keys to sign and submit the data transaction**
+## Specifying TPO on Products
+When a market is proposed, the settlement data source is specified as part of specifying the instrument. This data source may be a TPO. Note, a TPO may submit data that is used by multiple markets. Each market may use a different field of the data.
```
barnabee: Per above, different field or a differerent message, selected from the stream of messages by the filters

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Futures
+A [Cash Settled Direct Futures](./xxxx) product requires specification of a settlementPriceSource for settlement of the product (at expiry). TODO: Move this to the WIP built in futures spec!!
+
+```
+Future {
+    ... 
+    settlementPriceSource:  { 
+        signedMessage: {
```
barnabee, line 36: We should move this "signedMessage:…" part out of the Futures section and have a section which is "defining a TPO data source", then this section should make clear that in "settlementPriceSource" for Futures we should accept **any valid data source definition** (the obvious next one will be to allow the same filters/field spec to be run on Vega's own [Market Data](https://github.com/vegaprotocol/product/blob/master/specs/0021-market-data-spec.md) stream.

The Futures bit really should be in the spec for the Futures product, actually.

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+	    sourcePubkeys: ["VEGA_PUBKEY_HERE"],
+	    field: "price",
+	    dataType: "decimal",
+	    filters: [ 
+	        { "field": "feed_id", "equals": "BTCUSD/EOD" },
+	        { "field": "mark_time", "equals": "31/12/20" }
```
barnabee, line 42: We'll probably need to add types to the filters, and to support string, number, and at least one (possibly several) date/time formats. We should check if we need to support number-as-string here too (or decide if we would quietly accept a number-as-string in a numeric field...

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+}
+```
+
+
+## Blockchain Transaction for Oracle Data Submission
+
+This data is submitted at expiry of the market.
+
+```
+SubmitOracleData {
```
barnabee, line 55: We should probably support this key/value IMO but I also want to make sure that we support the [Open Oracle](https://github.com/compound-finance/open-oracle) message format used by the messages in [Coinbase oracle](https://docs.pro.coinbase.com/#oracle) and others - I am not 100% sure of the encoding format, it may be JSON in which case we might need to flatten json keys/array indices to paths like `key.1.subkey` or it might be that it's encoding specifically to their `kind, timestamp, key, value` format in which case those become the keys.

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+```
+    feed_id: ZZZZ
+    mark_time: 31/12/2020 23:59:59
+    price: 100.2
+    volume: 100,000
+````
+
+
+### Validation:
+1. If the _mark_time_ of the transaction does not equal the _mark time_ specified on the Product (e.g. `settlementPriceSource.filters.field.mark_time` ) then this data is not valid for use in the settlement action.
```
barnabee, line 87: If any of filters do not match (or are of the wrong type) the message is ignored and not passed as a valid data point to whatever defined the source.

barnabee: Also if the specified field is of the wrong type, the data is not passed as a valid data point, but this should create a warning event i.e. "data event passed filter but format of XX field doesn't match data soruce definition" as the format may have changed and it could be necessary for a [market, in the case of settlement] governance action to change the data source.

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Proposing Changes to Oracle
+
+A future feature requirement - any participant may propose a change to a market's oracle data source via on-chain governance. This includes:
```
barnabee, line 91: This is not really part of this spec. The inclusion of a valid data source definition in the market proposal/market framework should be defined by the spec for the Futures product and normal rules apply re: governance actions.

So to be clear: this would be expected to be addressed as part of market change governance actions.

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+1. Proposing a change in oracle type - e.g. a non TPO / NDS
+1. Proposing a change to the oracle values, such as a change in the sourcePubkeys or filters.
```
barnabee, line 93: The entire data source definition should be replaceable in a change. Includes the above and any other valid changes.

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+## Contingencies
+
+### What happens if Oracle Data doesn't arrive?
+For the minimum requirement, a (network parameter) timeout should be implemented, whereby if the oracle price data doesn't arrive, the market is settled by another means. See [settlement at expiry spec](./0004-settlement-at-instrument-expiry.md).
```
barnabee, line 106: Not sure about this... I think the minimum would be to simply do nothing until something arrives and allow a market param change to specify a new oracle. In future voting to "close" the market instead may also be possible.

barnabee (changes requested):

specs/trusted-party-oracles.md
tamlyn10: Makes sense

specs/trusted-party-oracles.md
```diff
@@ -0,0 +1,114 @@
...
+
+
+## Network Parameters
+1. Timeout, after which the oracle data source is considered undelivered and the Product will settle itself according to contingencies.
+## Acceptance
+
+### Trusted Party Oracle
+
+- A trusted party oracle is a vega pubic/private keypair
+    - Identified by its public key
```
barnabee, line 116: Needs to cover more of the spec

barnabee:
~Closing, rebased branch into https://github.com/vegaprotocol/product/pull/289~

barnabee:
Data source/oracles now here: https://github.com/vegaprotocol/product/pull/450

PR #422, opened 2020-11-23
Shift peg prices when outside the valid price range

### LINKED ISSUE #412: Include price constraints in liquidity obligations
As per Slack discussion here.. https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1605106283478800?thread_ts=1605103710.470400&cid=CAHA5EX0F

```
if (order.isPegged && order.isFromLiquidityCommitment) {
    order.price = best_price_of(order.price, worst_non_pm_triggering_price)
}
```

witgaw:
Add the bit saying that if the price for a given peg is outside the valid price range as implied by the price monitoring module then it should get shifted "in" so that it's at the last valid tick for that side of the order book.

Closes #412

### DIFF
--- a/specs/0038-liquidity-provision-order-type.md
+++ b/specs/0038-liquidity-provision-order-type.md
@@ -83,6 +83,7 @@ Given the price peg information (`peg-reference`, `number-of-units-from-referenc
 ``` volume = ceiling(liquidity_obligation x liquidity-normalised-proportion / probability_of_trading / price)```. 
 
 where `liquidity_obligation` is calculated as defined in the [market making mechanics spec](./0044-lp-mechanics.md) and `price` is the price level at which the `volume` will be placed.
+Note: if the resulting price for any of the entries in the buy / sell shape is outside the valid price range as provided by the price monitoring module it should get shifted to the valid price that's furthest away from the mid for the given order-book side.
 
 ```
 Example: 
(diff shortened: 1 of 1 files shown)


tamlyn10:
@barnabee just to check this is as per agreed and then merge..

specs/0038-liquidity-provision-order-type.md
```diff
@@ -83,6 +83,7 @@ Given the price peg information (`peg-reference`, `number-of-units-from-referenc
 ``` volume = ceiling(liquidity_obligation x liquidity-normalised-proportion / probability_of_trading / price)```. 
 
 where `liquidity_obligation` is calculated as defined in the [market making mechanics spec](./0044-lp-mechanics.md) and `price` is the price level at which the `volume` will be placed.
+Note: if the resulting price for any of the entries in the buy / sell shape is outside the valid price range as provided by the price monitoring module it should get shifted to the valid price that's furthest away from the mid for the given order-book side.
```
barnabee: I think we should be explicit how to calculate the price here or make sure we are referring a specific part of the price monitoring spec. and linking there.

barnabee (changes requested):
One update to be explicit about the valid price otherwise looks good

specs/0038-liquidity-provision-order-type.md
tamlyn10: @witgaw do you mind addressing this comment then merging?

witgaw: Clarified now I believe:

> Note: if the resulting price for any of the entries in the buy / sell shape is outside the valid price range as provided by the price monitoring module (the min/max price that would not trigger the price monitoring auction per triggers configured in the market, see [price monitoring](./0032-price-monitoring.md#view-from-quant-library-side) spec for details) it should get shifted to the valid price that's furthest away from the mid for the given order-book side.

Will merge tomorrow morning if I don't hear any objections before.

witgaw: Now merging for real!

witgaw:
Merging as I wasn't challenged on this for a week.

PR #425, opened 2020-11-23
Define a party, mainly by linking to other specs

### LINKED ISSUE #424: What is a party
Make a spec

edd:
Closes #424

### DIFF
--- a/specs/0000-party.md
+++ b/specs/0000-party.md
@@ -1,0 +1,28 @@
+# Party 
+
+A party is...
+
+
+# Summary
+
+A party may have:
+- 0 or more [accounts](./0013-accounts.md)
+    - These accounts may contain a balance of a single [asset](./0040-asset-framework). This is known as [collateral](./0005-collateral.md)
+
+A party can:
+- [Submit an order](./0025-order-submission.md)
+- [Amend an order](./0026-amends.md)
+- [Cancel an order](./0033-cancel-orders.md)
+- [Submit a governance proposal](./0028-governance.md)
+- [Vote on a governance proposal](./0028-governance.md)
+- [Commit to provide liquidity on a market](./0038-liquidity-provision-order-type)
+
+# Guide-level explanation
+
+
+# Acceptance Criteria
+
+
+## Criteria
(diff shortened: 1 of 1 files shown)


edd:
@jeremyletang There's a section in here for 'Special parties'. Currently I'm only aware of 'network'

1. Does 'network' have a keypair? I think **No**
    1. If so, who has the private key
1. Is a transaction submitted to the chain on behalf of 'network'? I think **No**
    1. If it is, how is it signed?
1. Are there other special things we add in as 'parties'?
1. I have a vague memory of a '*' and a '!' appearing in some fields. Do they relate to special parties?

davidsiska-vega:
@jeremyletang  I think this is "good enough" (TM) and "better than nothing" (TM).

edd:
@jeremyletang are there any other 'special parties' like `network`? I think no.

jeremyletang:
> @jeremyletang There's a section in here for 'Special parties'. Currently I'm only aware of 'network'
> 
>     1. Does 'network' have a keypair? I think **No**
>        
>        1. If so, who has the private key
> 
>     2. Is a transaction submitted to the chain on behalf of 'network'? I think **No**
>        
>        1. If it is, how is it signed?
> 
>     3. Are there other special things we add in as 'parties'?
> 
>     4. I have a vague memory of a '*' and a '!' appearing in some fields. Do they relate to special parties?

1. No
2. No
3. No
4. No (relate to accounts, it's only internals or at least should make sense only internally)

jeremyletang (approved):
Sounds good to me.
One thing that I'm not sure needs to be done there, but what should be the representation of a party via the API.

As of now we have a type:
```
Party {
   string ID;
}
```

Shoudl we simplify this an replace this type containing a single field (where actually ID is the public key of the party), by a simple string everywhere?

This would slightly simplify API, and even some places internally in the core.

Also I think that was briefly discussed somewhere already @edd @barnabee

edd:
> This would slightly simplify API, and even some places internally in the core.
> Also I think that was briefly discussed somewhere already @edd @barnabee

That sounds good to me. Any objection @barnabee?

I don't think that changes anything in the spec, so now I have updated it to reflect the special party clarifications, let's merge.

Issue #426, opened 2020-11-24
Clarify in LP specs that existing orders can still be cancelled/amended


peterbarrow:
The spec says the user cannot amend or cancel orders that are created as part of the liquidity providing shape. It does not mention what can be done to existing orders that are used as part of the commitment.

After a chat on slack we are going to allow those orders to be amended/cancelled like ordinary orders that are not part of the liquidity commitment.

PR #427, opened 2020-11-24
Allow modifying of existing orders (#426)

peterbarrow:
Any orders that were already placed that are not part of the LP shape can be cancelled and amended like a normal order.

### DIFF
--- a/specs/0038-liquidity-provision-order-type.md
+++ b/specs/0038-liquidity-provision-order-type.md
@@ -141,7 +141,9 @@ Liquidity providers are always allowed to amend their shape generated orders by 
 
 No cancellation of orders other than by lowering commitment as per [[Liquidity Provisioning mechanics spec](./0044-lp-mechanics.md).
 
-Existing orders that are used as part of the liquidity commitment can be cancelled and amended just like normal orders on the book which are not part of liquidity provisioning.
+Any attempt to directly cancel or amend an order generated from the LP shape will be rejected.
+
+Existing persistent orders (limit or pegged) that are used as part of the liquidity commitment can be cancelled and amended just like normal orders on the book which are not part of liquidity provisioning.
 
 
 ## Network Parameters:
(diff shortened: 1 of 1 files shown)


specs/0038-liquidity-provision-order-type.md
```diff
@@ -141,6 +141,8 @@ Liquidity providers are always allowed to amend their shape generated orders by
 
 No cancellation of orders other than by lowering commitment as per [[Liquidity Provisioning mechanics spec](./0044-lp-mechanics.md).
 
+Existing orders that are used as part of the liquidity commitment can be cancelled and amended just like normal orders on the book which are not part of liquidity provisioning.
```
tamlyn10: I think we should be clear this is about limit orders..

tamlyn10 (changes requested):
one minor comment - if fixed then good to merge

specs/0038-liquidity-provision-order-type.md
peterbarrow: I understood it to mean all persistent orders which includes limit and pegged orders?

barnabee: @peterbarrow yeah, I think that's correct, the only orders that can't be amended are those that are derived from the liquidity commitment, as far as I know (i.e. this should include any order that would normally be amendable)

tamlyn10: ah i thought it referred to this question.... https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1606217324042000

davidsiska-vega:
@peterbarrow I've not seen this one when dealing with the #426 so I ended up creating a different pull request. Decide which one you want to keep and delete the other one (the other one being https://github.com/vegaprotocol/product/pull/459 ). 

D

peterbarrow:
Has been superseded, closing.

Issue #431, opened 2020-11-30
Confirm in specs how to calculate the uncrossing price


peterbarrow:
The uncrossing price for an auction is calculated as:

(minimum price of the uncrossing range + maximum price of the uncrossing range)/2

tamlyn10:
@peterbarrow could you confirm if this is still open / requiring attention?

peterbarrow:
The spec was updated and merged, closing ticket.

PR #432, opened 2020-11-30
Added uncrossing price formula (#431)

peterbarrow:
Added the formula for working out the auction uncrossing price


tamlyn10 (approved):
Thanks for making this more clear @peterbarrow

PR #435, opened 2020-12-07
Modify the price monitoring spec

### LINKED ISSUE #434: Price monitoring: let's imply bounds from the "monitoring price"
@witgaw following a discussion with @tamlyn10 and @barnabee we are thinking that always implying the bounds from the monitoring price is probably the more intuitive thing to do. We should also do volume weighted averages as you mentioned. 

This shouldn't hold up the release so you can update the spec and then create a core ticket and update the implementation... 

The other thing is, as you mentioned, to use a multiplicative scaling factor for deriving the bounds instead of an additive delta.

witgaw:
Modify the price monitoring spec as per #434: modify the engine to use the reference price (past price per trigger) rather than latest price when syncing with the risk model and to store the moves as multiplicative (rather than additive) factors internally. Also added reference price caching not to look it up more than once. 

Closes #434

### DIFF
--- a/specs/0032-price-monitoring.md
+++ b/specs/0032-price-monitoring.md
@@ -75,9 +75,13 @@ Price monitoring engine will interface between the matching engine and the risk 
 Specifically:
 
 - Price monitoring engine averages all the prices received from the matching engine that have the same timestamp.
-- It periodically (in a predefined, deterministic way) sends the accrued price history to the risk model and obtains the set of max up/down moves per period τ and the associated level α (it can keep more than one of the [τ, max move up, max move down] triplets).
+- It periodically (in a predefined, deterministic way) sends the:
+  - the probability level α,
+  - period τ,
+  - the associated reference price
+to the risk model and obtains the max up/down scaling factors per each of the specified triggers. The reference price is the latest price such that it's at least τ old or the earliest available price should price history be shorter than τ.
 - It holds the history of average prices looking back to the maximum τ configured in the market.
-- Everytime a new price is received from the matching engine the price monitoring engine checks all the [period, max move up, max move down] triplets relevant for the timestamp, looks-up the associated past (averaged) price and sends the signal back to the matching engine informing if the received price would breach the min/max move prescirbed by the risk model.
+- Everytime a new price is received from the matching engine the price monitoring engine checks all the [τ, up factor, down factor] triplets relevant for the timestamp, looks-up the associated past (averaged) price and sends the signal back to the matching engine informing if the received price would breach the min/max move prescirbed by the risk model.
 - The bounds corresponding to the current time instant and the arrival price of the next transaction will be used to indicate if the price protection auction should commence, and if so, what should its' period be (see below).
 - To give an example, with 3 triggers the price protection auction can be calculated as follows:
   - \>=1% move in 10 min window -> 5 min auction,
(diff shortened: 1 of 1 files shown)


specs/0032-price-monitoring.md
```diff
@@ -75,9 +75,13 @@ Price monitoring engine will interface between the matching engine and the risk
 Specifically:
```
davidsiska-vega, line 75: On l. 61 we might want to say:  "Specifies how often (expressed in seconds) the price monitoring scaling factors should be updated by the risk model." instead of "Specifies how often (expressed in seconds) the price monitoring bounds should be updated by the risk model."

davidsiska-vega: On l. 73 we might want to say: "It will communicate with the risk model with a predefined frequency to inform the risk model of the latest price history and obtain a new set of scaling factors used to calculate min/max prices from the reference price" instead of "It will communicate with the risk model with a predefined frequency to inform the risk model of the latest price history and obtain a new set of min/max price move bounds."

specs/0032-price-monitoring.md
```diff
@@ -75,9 +75,13 @@ Price monitoring engine will interface between the matching engine and the risk
...
 
 - Price monitoring engine averages all the prices received from the matching engine that have the same timestamp.
```
davidsiska-vega: Say what kind of average (volume weighted perhaps).

specs/0032-price-monitoring.md
```diff
@@ -88,9 +92,9 @@ Specifically:
 
 ## View from [quant](https://github.com/vegaprotocol/quant) library side
 
-- The risk model calculates the bounds per current price, horizon τ and confidence level α beyond which a price monitoring auction should be triggered.
+- The risk model calculates the bounds per reference price, horizon τ and confidence level α beyond which a price monitoring auction should be triggered.
```
davidsiska-vega, line 95: Do we want to say scaling factors here as well?

specs/0032-price-monitoring.md
witgaw: Thanks, added now.

specs/0032-price-monitoring.md
witgaw: Thanks, modified now.

specs/0032-price-monitoring.md
witgaw: I think it's ok to leave as is and just calculate the factors internally (within the PM engine) as: `upFactor`:= `maxPrice`/`refPrice`, `downFactor` := `minPrice`/`refPrice`.

davidsiska-vega: But I think we can get the factors directly from the risk model by asking for  `min / max` for price of `1`. At least that's what it looks like to me from https://drive.google.com/file/d/11AGvFkzpoSqJJ2vczdRMvd4blGkTFsdX/view?usp=sharing .

witgaw: Yes, the two methods are equivalent. I can change it to yours if you want, but it will only make a performance difference (less memory usage) if we implement caching on the risk model's side (well that and those two divisions that would be saved).

witgaw: Though I think just going `upFactor`:= `maxPrice`/`refPrice`, `downFactor` := `minPrice`/`refPrice`, and back to `maxPrice`=`upFactor`*`refPrice` etc is more readable to someone without much knowledge about what the model is doing.

davidsiska-vega: If you think your approach is clearer than let's go with that.

specs/0032-price-monitoring.md
```diff
@@ -70,14 +70,18 @@ Likewise, pre-processing transactions will be needed as part of the [fees spec](
...
-- It holds the history of average prices looking back to the maximum τ configured in the market.
-- Everytime a new price is received from the matching engine the price monitoring engine checks all the [period, max move up, max move down] triplets relevant for the timestamp, looks-up the associated past (averaged) price and sends the signal back to the matching engine informing if the received price would breach the min/max move prescirbed by the risk model.
+- Price monitoring engine averages (weighted by volume) all the prices received from the matching engine that have the same timestamp.
+- It periodically (in a predefined, deterministic way) sends the:
+  - the probability level α,
+  - period τ,
+  - the associated reference price
+to the risk model and obtains the max up/down scaling factors per each of the specified triggers. Please note the scaling factors can be either additive or multiplicative depending on the risk model used. The reference price is the latest price such that it's at least τ old or the earliest available price should price history be shorter than τ.
```
davidsiska-vega: "Additive" scaling factor sounds confusing, doesn't it?

davidsiska-vega: Additive offsets and multiplicative factors perhaps?

witgaw: Thanks, changed now.

davidsiska-vega (approved):

PR #445, opened 2021-01-06
Add currency units to target stake

### LINKED ISSUE #444: target stake
As per this Slack discussion, target stake is unitless but that doesn't work when using it. David knows what to do!

davidsiska-vega:
Added currency units to target stake and clarified (but not changed) liquidity monitoring spec. Closes #444 .

### DIFF
--- a/specs/0035-liquidity-monitoring.md
+++ b/specs/0035-liquidity-monitoring.md
@@ -8,37 +8,30 @@ Liquidity in the market is not only a desirable feature from a trader's point of
 
 Similarly to [price monitoring](0032-price-monitoring.md), we need to be able to detect when the market liquidity drops below the desired level, launch a liquidity auction and terminate it when the market liquidity level is back at a sufficiently high level.
 
+Note that there is one-to-one correspondence between the total stake committed by liquidity providers (LPs), see [LP mechanics](specs/0044-lp-mechanics.md) spec, and the total supplied liquidity. Thus even though below we will be comparing `target_stake` with `supplied_stake` what this really monitors is amount of liquidity supplied. 
+Note that [target stake](0041-target-stake.md) is defined in a separate spec.
+
 ## Liquidity auction network parameters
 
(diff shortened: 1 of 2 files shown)


specs/0035-liquidity-monitoring.md
```diff
@@ -8,37 +8,30 @@ Liquidity in the market is not only a desirable feature from a trader's point of
 
 Similarly to [price monitoring](0032-price-monitoring.md), we need to be able to detect when the market liquidity drops below the desired level, launch a liquidity auction and terminate it when the market liquidity level is back at a sufficiently high level.
 
+Note that there is one-to-one correspondence between the total stake committed by liquidity providers (LPs), see [LP mechanics](specs/0044-lp-mechanics.md) spec, and the total supplied liquidity. Thus even though below we will be comparing `target_stake` with `supplied_stake` what this really monitors is amount of liquidity supplied.
```
witgaw, line 11: Is this really true? The liquidity (at least the probability weighted order book volume measure of it) may fluctuate as the mark price and order-book composition changes even holding the total stake constant.

davidsiska-vega: From 0044:

The stake they commit implies liquidity obligations. This is derived using a single network parameter stake_to_ccy_siskas as follows: lp_liquidity_obligation_in_ccy_siskas = stake_to_ccy_siskas x stake. Note here "ccy" stands for "currency" and that the liquidity measure units are actually a currency e.g. ETH or USD. This is because it's basically volume x probability of trading x price of the volume and price of the volume is in the said currency.

So the lp_liquidity_obligation is in one-to-one correspondence to stake. How much volume it means is posted at each level changes with the price fluctuations.

witgaw: Ok, fair enough. Thanks for clarifying.

witgaw (approved):
Looks good to me.

PR #449, opened 2021-01-10
Release overviews

tamlyn10:
Trialling a new way capture where we specify the scope of releases that is trackable, links to project boards and easy for non-developers to reference. This will replace the former way of doing this in the handbook, Notion and strategy cards.


releases/3-oregon-trail.md
```diff
@@ -0,0 +1,37 @@
...
+
+
+## Core
+
+### Features
+    - Settlement at expiry
+
+    - Spam protection
+
+    - Full deposit and withdrawal via bridge
```
barnabee: I don't think anything is missing from this now, core-wise

PR #450, opened 2021-01-12
Data sourcing

barnabee:
This is the REAL, FINAL PR that will be data sourcing.

- `0045-data-sourcing.md` — data sourcing framework
- `0046-data-source-signed-message.md` — signed message data source
-  `0047-data-source-filter.md` — filters
- `0048-data-source-internal.md` — basic internal data sources


specs/0045-data-sourcing.md
```diff
@@ -0,0 +1,173 @@
...
+a) Data sources must be able to emit the following data types:
+1. Integer
+1. Floating point
+1. DecimalFromInteger (i.e. how Ethereum stores decimal places, need to specify number of decimal places with this type)
+1. Integer/Floating/DecimalFromInteger from string (we need to support the cases where numbers come from APIs in string format / are too large for basic number types to represent)
+1. Strings
+1. Date/Time
+1. Boolean
+1. Empty (ignore all values, source is used as a trigger only)
+1. Key value pairs (needed so we can have price(s) and timestamp, etc. for filtering)
```
jeremyletang: I'm not sure to see the point in having 4 different sources of number? 
I think that as we have all the control over how this is designed, and in any case people implementing oracle for us will have to implements what our needs are, I would think we should try something simpler, and support a single number type, probably only strings, and if there's decimals, e.g for price, they should be expressed with all the decimal the asset support, so for 1 ethereum it would "1,000000000000000000". all decimal explicitly set.

Basically that should be the only format for number that we support:
Integer/Floating/DecimalFromInteger from string (we need to support the cases where numbers come from APIs in string format / are too large for basic number types to represent)

specs/0045-data-sourcing.md
```diff
@@ -0,0 +1,173 @@
...
+
+Note that for number types the system should convert appropriately when these are used in a situation that requires Vega's internal price/quote type using the configured decimal places, etc. for the market.
+
+Note that we should support all the number types as we want to enable the community to easily submit data from various sources and not worry about conversion, and different market types will have different data types for settlement, and different underlying source assumptions.
```
jeremyletang: Well, while I undertsand the point, I still think we should not support X way of doing the same thing. It just brings confusion, and a more complex implementation for us.

specs/0045-data-sourcing.md
```diff
@@ -0,0 +1,173 @@
...
+
+b) Future types (when we create new products and add calculation features we will need these):
+1. List of items of any type (i.e. list of floats)
+
+
+## 5. Tracking active data sources
+
+Vega will need to keep track of all "active" defined data sources that are referenced either by markets that are still being managed by the core (i.e. excluding Closed/Settled/Cancelled/other "end state" markets) or by other data source definitions.
+
+Data sources that are no longer active as defined above can be discarded.
```
jeremyletang: Should we do anything about data source provider that would keep sending data once the e.g market using it have ended?

Should we reject the transaction even before it reach the blockchain, any action to be taken about this?

specs/0045-data-sourcing.md
```diff
@@ -0,0 +1,173 @@
...
+
+
+## 6. Types of data source
+
+The following data sources have been defined:
+1. Internal basic data sources (Vega time, direct value) [TODO: link]
```
jeremyletang: link?

specs/0045-data-sourcing.md
```diff
@@ -0,0 +1,173 @@
...
+1. [signed message](./0046-data-source-signed-message.md)
+1. Time triggered (at a certain date/time)
+1. Filters (exclude certain events basd on conditions and boolean logic against the fields on the data such as equals, simple comparisons, is/is not in a predefined list, an MVP of this functionality is needed to allow TPOs to be practical)
```
jeremyletang: Would be good to have comparison explicitly defined, as they will require implementation on our side, and to know what types specificaly allow what comparisons.

specs/0046-data-source-signed-message.md
```diff
@@ -0,0 +1,100 @@
...
+Note: With this type of oracle there’s no incentive in the Vega data source system, you’re trusting the keyholder(s) and any modifiers or verification applied through the [data source framework](./0045-data-sourcing.md) at settlement.
+
+*NOTE: This is the only external oracle available initially in Vega, and initially requires only one of the specified keys to sign and submit the data transaction. This means that initially it will only be possible to construct external oracles on Vega in which one or more third party entities/systems must be trusted. This will change with modifiers that allow combinations of data sources, verification of data stream via governance votes, and data sources that birdge to events included on other blockchains.**
+
+
+## Defining the data source
+
+A data source must define:
+
+- Public keys (and key algorithm to be used?) that can sign and submit values for this oracle
```
jeremyletang: Should we restrict this to vega public keys initially?
For every algorithm that we want to support an implementation needs to be implemented in vega basically.

jeremyletang (commented):
Do we get a new governance proposal to create data sources? Or does these data sources get create as part of something else e.g the declaration of the future in the market proposal?

specs/0045-data-sourcing.md
barnabee: > in any case people implementing oracle for us will have to implements what our needs are

This is simply not true. We are not expecting "people to implement oracles for us" so much as people to implement oracles and for us to want to use them. We want to work easily with as many protocols and solutions as possible including the OpenOracle standard (used by Coinbase oracle), the Ethereum blockchain, other blockchains that we will bridge to, protocols with their own networks like Band Protocol, ChainLink, etc.

If those protocols and networks come up with a signed bundle of data, we want to be able to process that signed bundle directly, not require a step to convert it and re-sign (introducing more risk of corrupted oracles) and not to require that the protocol writes some special case code for us, as far as possible.

If it is difficult to cover all the ways numbers could be expressed or there are edge cases/issues that require more thought then we could spend the time to do some research and identify which types we need to support most urgently and move the others to future work, but as numbers seem relaticely simple we had assumed we might just implement all the obvious cases.

specs/0045-data-sourcing.md
barnabee: Yes, we should reject any data source tx that is not explicitly required, so this would include a tx:

- For a pubkey never used in a data source
- For a pubkey only used in data sources referenced by markets (or other things) that are no longer being managed by the core (i.e. once a marked is in Closed or Settled or Cancelled state according to the market framework) or before the enactment date of the market proposal
- If a data source combines a primary source (like a signed message) with a filter (for instance saying we are only interested in messages where `ticker = GBPUSD` and `timestamp = 20211231T23:59:00`) then the _complete data source definition_ defines the source and can be used to accept/reject transactions, so even if this data source is active, transactions from the same pubkey would be rejected if the ticker and timestamp were not correct, then the whole data source would be discarded once it is no longer referenced by a market

Ideally all of this would happen before the tx is scheduled in a block. This would allow someone to, for instance, create a service that regularly (say, hourly or daily) posts the signed bundles from the Coinbase Oracle to the Vega network, and Vega would automatically accept data that is going to be used by a market and reject the rest.

specs/0045-data-sourcing.md
barnabee: ⏳ this will be added in the next set of commits as is still WIP

specs/0045-data-sourcing.md
barnabee: Agree and we will do this in the next commits. We'll try to keep the number of required comparisons as small as possible to begin with.

specs/0046-data-source-signed-message.md
barnabee: We at a minimum need to support data signed by Vega public keys and [Coinbase Oracle](https://docs.pro.coinbase.com/#oracle) (and therefore the OpenOracle standard) as we would like users to be able to post these signed messages *directly to Vega* so that the signing key you are trusting is Coinbase, etc. not the person submittng this message to Vega.

I suspect this means supporting two signing algorithms initially.

specs/0045-data-sourcing.md
barnabee: See [comment above](https://github.com/vegaprotocol/product/pull/450/files#r560158026), if it is a big deal to build domething more generic and anticipate operating in an open environment then lets do the work to understand what formats we really want to support (we can reach out to community and some investors for this, but it will take a little while to do).

This is one of the key points where we integrate with other systems and proejcts (similar to the bridge) and our ability to work with existing solutions _**easily**_ will have a huge impact on our success. Partnerships and joint activities are one of the best ways to encourage adoption and we very regularly have people asking about whether X can be used as an oracle on Vega, so we really don't want to be forcing development work to integrate if simply supporting conversion from a different numeric type would avoid it.

However, I am happy to accept that if some of the types aren't needed at all we don't have to build them, so if you'd like us to do that analysis please let us know.

specs/0045-data-sourcing.md
jeremyletang: OK, I was going from the assumption we would have adaptors like peices of software building the data out of other oracle sources.

Based on this I understand the needs of being able to support as much as possible format, but it sounds like a pretty unrealistic / complex task to take on.

To me it's not just about numbers format or anything, it's about the level of flexibility we want to offer / support. and the more we do, the more error prone we would be, the more complexity we bring, and more time we spend building "parser" for data of any kinds instead of building other complex pieces of the software.

specs/0046-data-source-signed-message.md
jeremyletang: I see, that was coming from my assumption we/peope would build adaptors to the data from differents sources somehow.

specs/0045-data-sourcing.md
barnabee: Yeah, I agree that we need to find a middle ground but that middle ground is probably **not** "there's one way and only one way and it's Vega's way that is different to the rest of the ecosystem"

Let's work out how we can define this better thougn.

specs/0045-data-sourcing.md
barnabee: We now have a filter spec in this PR, we only need very simple comparisons initially: 
https://github.com/vegaprotocol/product/pull/450/files#diff-1bda4a253db6b9f9617c103665e0e7cf24d99fdee1d2f7e46a75479803992f23

specs/0046-data-source-signed-message.md
barnabee: That will be necessary if the sources are very weird, but OpenOracle is a standard that we could implement and support a number of sources. If this looks really tough, we could split the work into two parts:

1. Support Vega keys and protobuf list of key/value pairs
2. Support OpenOracle / Coinbase Oracle (I think uses Ethereum ABI format or something, so that code _might_ be reusable when we get to Ethereum bridge data sources) and the crypto aglorithm needed to sign these

We could deliver an MVP with just 1, but it will certainly impact our ability to run joint initiatives and do things in the community with solid integrations.

tamlyn10:
Follow up meeting scheduled for next week to resolve this analysis ticket: https://github.com/vegaprotocol/product/issues/462

@barnabee @tamlyn10 to complete the remaining bits of this PR.

qustavo:
@barnabee @tamlyn10 I wonder if this spec should include a strategy or mechanism to deal with errors (networking, formatting/encoding, signing), or that's exclusively up to the implementation?.
By dealing with errors I mean errors coming from the oracles

tamlyn10:
@barnabee to push changes today for final review from @jeremyletang (and others).

tamlyn10:
@barnabee to address @gchaincl comment directly in spec.

specs/0045-data-sourcing.md
barnabee: I have tightened the specs and tried to make the limited initial scope clearer

specs/0045-data-sourcing.md
barnabee: Link added

specs/0045-data-sourcing.md
barnabee: I have tried to make the limited initial scope clearer. We may need _several_ ways to do things but it is not unlimited and they should be more tightly scoped and clearly defined now.

barnabee:
> @barnabee @tamlyn10 I wonder if this spec should include a strategy or mechanism to deal with errors (networking, formatting/encoding, signing), or that's exclusively up to the implementation?.
> By dealing with errors I mean errors coming from the oracles

@gchaincl good question, I guess my thinking is that network level stuff would be defined at a lower level than these specs, as we are using the same transaction capabilities as other Vega blockchain transactions, so that probably doesn't live here [1]. Also, we ought to reject malformed messages. There is some limited reqiurement, mentioned in the spec, that we emit some log/event if data is accepted and ought to be passed to a product (for settlement for instance), but the data is of the wrong type. But other than that I can't think what might be needed here - please shout if there is something I'm missing though.

[1] In future (but definitely not now) if we were to implement something like an HTTPS data source scheme (i.e. like the bridge a quorum of validators must query an endpoint around some time and agree they see the same value) then there may be some error handling needed. But that is for later :)

barnabee:
@jeremyletang @gchaincl @C0deMunk33 @tamlyn10 

💥  This is (hopefully) the final version of the data sourcing specs! 

⏳  I will merge at EOD tomorrow (Thurs 4th Feb 2021) unless there are new and unresolved review comments or someone akss me to hold off.

qustavo:
@barnabee let's say coinbase api goes down, do we need to send a special Vega Tx? or we just don't send data and wait until the API is back. A more complex error might be that for whatever reason coinbase updated their api breaking our relay. The end result for w/e reason will be no Tx broadcasting, if we reach that point, I wonder what's the expected behavior

jeremyletang:
@ValentinTrinque you may want to take a look at this as well.

barnabee:
> @barnabee let's say coinbase api goes down, do we need to send a special Vega Tx? or we just don't send data and wait until the API is back. A more complex error might be that for whatever reason coinbase updated their api breaking our relay. The end result for w/e reason will be no Tx broadcasting, if we reach that point, I wonder what's the expected behavior

## Let's look at some scenarios...

### If the tx doesn't arrive: CB API is down
We have no way to if this is the case vs other non-arrival. If it is temporarily down it will arrive later, and people could probably use a CB API status page to tell that is the issue. If it's down as in gone forever, we can't tell it's not the latter and would need governance to be invoked to change the data source to one that will work (incidentally, we have the idea of "value" as a data source - [see here](https://github.com/vegaprotocol/product/pull/450/files#diff-c87deea3309be0e7ec0040a8f07ad0ccb0ce83e420e9eead041498a1213e0228R13-R24), to allow governance to directly vote in the data they want rather than pick another source, if that's the optimal outcome)

### Transaction doesn't arrive part 2: noone submits the signed data from the CB API to Vega
In this case, people generally have an economic incentive (all their margin etc. is tied up in the market until settlement) to go get the data from the API and post it. If it's there it'll find its way in.

### API still publishes data but signing key changes
This looks like the first case above, because the original data source never publishes.

### API still publishes data but data format changes
This looks like the first case above. If the data get's filtered out, because of the change we may not be able to tell, but it may trigger a warning message, i.e. for a [type mismatch](https://github.com/vegaprotocol/product/pull/450/files#diff-1f4a0b602f14ba89e19ad82b674312a2d0ba461d329f348056e086036450c564R75-R79) either on the network/event bus or on the node the TX was submitted to (i.e. it comes back as an error in submission and perhaps appears in logs). In this case the same applies as if it simply didn't arrive, but someone might have been alerted to the issue by the error.


## Final thought...
We _could_ add an `expected_by` optional field on any data source which would simply trigger an event bus event if the data source remained unfulfilled after that date. But I am unsure whether that's something that ought to live on chain. I guess it could be pretty useful to flag in Console / to general users of the protocol that something is up 🤔 

cc @tamlyn10 @campbellssource @jeremyletang what do you think of this?

qustavo:
Thanks @barnabee for the examples.
> would need governance to be invoked to change the data source to one that will work

This could be used for other reasons, not only because we lost contact with the Oracle, i.e we found a better one. Does the possibility of changing a Data source at _runtime_ present a security/trust issue?

barnabee:
> Thanks @barnabee for the examples.
> 
> > would need governance to be invoked to change the data source to one that will work
> 
> This could be used for other reasons, not only because we lost contact with the Oracle, i.e we found a better one. Does the possibility of changing a Data source at _runtime_ present a security/trust issue?

It presents a _risk_ but not an issue. We require community governance for many things (including approving the market and data source in the first place) so it is a class of risk we have throughout the protocol, and even extends to trusting the quorum of validators, etc. 

The governance framework is designed for each proposal type to have sub-categories of proposal precisely so that potentially dangerous changes can have much higher thresholds. So we would expect it to be a pretty high barrier to switch the data source or _any_ product parameters for a market as that can **completely change the market** and that affects the value of everyone's positions. It's defiintely an attack vector, so we need such changes to be pretty hard at a minimum.

jeremyletang (approved):

PR #460, opened 2021-01-20
Minimum lp stake per asset

### LINKED ISSUE #364: Minimum liquidity provider stake commitment before market goes to opening auction
What is the minimum stake a ~~market maker~~ liquidity provider must commit to a market before it's allowed to go into an opening auction? 
1. Zero? Then immediately after opening auction a liquidity auction will be triggered and maybe that's good enough?
2. Enough for `max_oi = 1` ie  `stake >=  v_for_target_stake x rf` (see `0041-target-stake.md`)? 
3. We add *another network parameter* something like "mimimum_viable_oi"  and say that `stake >=  mimimum_viable_oi x v_for_target_stake x rf`? 
4. Have per asset minimum stake (asset parameter) to launch a market. 

I like 1. for its simplicity as to me it seems we don't have to worry and just go with we've got in the protocol already. 

2. is OK but not so different from 1. since finally as soon as some meaningful trading happens we're likely to end up in a liquidity auction. 

3. is pointless in my opinion because markets can be proposed with one contract unit not having much value and so it could be gamed (in my mind) and it adds another network parameter (which should be avoided if possible). 

4. This is more thinking about avoiding a market proposal spam - if you need to commit some ~ 1 000 USD to launch a market you're going to be a bit serious about it. 

@barnabee @tamlyn10 @witgaw Please let's hear your thoughts and add conclusion to the market lifecycle spec.

davidsiska-vega:
Closes #364. 

@jeremyletang @C0deMunk33, please review how this pull request talks about specifying minimum LP commitment per asset on the asset framework / proposal side of things.

### DIFF
--- a/specs/0040-asset-framework.md
+++ b/specs/0040-asset-framework.md
@@ -17,6 +17,8 @@ In the case of an ERC20 token for example, only the contract address of the toke
 
 The asset specific part is represented by the source field, which can be one of the different source of assets supported by vega.
 
+Note that the `minLPstake` field sets the minimum liquidity commitment required for any market using the assest as settlement asset, see [LP mechais spec](0044-lp-mechanics.md).
+
 ```protobuf
 syntax = "proto3";
 
@@ -30,6 +32,7 @@ message Asset {
   string symbol = 3;
   string totalSupply = 4;
   uint64 decimals = 5;
+  decimal minLPstake = 100.01; 
 
   oneof source {
     BuiltinAsset builtinAsset = 101;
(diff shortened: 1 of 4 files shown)


tamlyn10:
@jeremyletang to review / edit

specs/0040-asset-framework.md
```diff
@@ -30,6 +32,7 @@ message Asset {
   string symbol = 3;
   string totalSupply = 4;
   uint64 decimals = 5;
+  decimal minLPstake = 100.01;
```
barnabee: This probably needs to be string same as totalsupply? cc @jeremyletang

jeremyletang: hum didnt catch this one...
let me edit.

jeremyletang: done.

PR #464, opened 2021-01-25
Add Pegged order to protocol-glossary

ValentinTrinque:
Add `Pegged order` definition to protocol glossary.


edd:
**issue (blocking)**: My first thought was 'I think this belongs in `trading-and-protocol-glossary.md` as it's not specific to Vega. But I'm not actually sure that differentiation is useful.

Could you add this and Vega Time in to `trading-and-protocol-glossary.md` instead, delete `vega-glossary` and then we can merge this pr. Otherwise: good description.

ValentinTrinque:
Yes, I can add `Vega Time` to  `trading-and-protocol-glossary.md`.

edd:
@jeremyletang just pointed out that you did make the change in the right file. My mistake :laughing:. Sorry. So if you move in Vega Time as well, we'll delete that file once and for all and I'll merge. Thanks.

edd (approved):

edd:
Now to go and fix #463 ...

PR #465, opened 2021-01-26
Add ABI/ABI encoding to DLT glossary

edd:


tamlyn10:
ABI = Abstract Binary Interface

PR #467, opened 2021-01-27
WIP Training wheels (aka safety limits) for Vega

davidsiska-vega:
This  will eventually be a spec for what limits we want to be able to place on Vega and how these will interact with token holders voting on things. @C0deMunk33 .

### DIFF
--- a/non-protocol-specs/0002-training-wheels.md
+++ b/non-protocol-specs/0002-training-wheels.md
@@ -1,27 +1,23 @@
 # Limits to make first mainnet deployments of Vega safer
 
-Philosophy:
-- limits on deposits: how much do we allow people to risk 
-- limits on withdrawals: limits the danger that due to a bug in the system someone can drain all the asssets 
+### Philosophy:
+- limits on deposits: reduce how much do we allow people to risk initially 
+- limits on withdrawals: reduce the probability that due to a bug in the system someone can drain all the asssets 
 
-Questions: 
-- should we create limits in `core` that can be circumvented with multiple accounts? Yes! It sends a message. 
+### Questions: 
+- should we create limits that can be circumvented with Sybil accounts? Yes! It sends a message. 
 - how do we compare across assets? Maximum deposit is a multiple of minimum stake per asset (already in asset in framework) or constant per asset - for the life of the vega chain, changeable via governance
 
-What do we want to limit:
-- maximum deposit per asset from any given Ethereum address and to any Vega address
-- withdrawals amount and timing? e.g. withdrawals only allowed 1 week after last settlement?
-- withdrawals over certain amount during a time period subject to Vega token holder vote? If a transaction isn't approved(*) within a time limit it will be automatically addressed to a new Ethereum address controlled by Vega token multisig - allowing DAO-like approach to distributing this.
+### Deposit limits
+- maximum deposit per asset from any given Ethereum address and to any Vega address. The limits per asset should be configurable via governance. Limits should be part of whitelisting a new asset. 
 
-(*) by default approved i.e. unless we have more than 50% votes against the withdrawal proceeds 
+### Withdrawal limits
+- withdrawals over certain amount during a defined time period will be delayed (e.g. 72 hours). During that delay period the Vega validators can blacklist the bundle to prevent a suspicious withdrawal by signing a multi-sig transaction (this could be done manually by agreement between validators or via governance action on Vega blockchain).
 
-- do we need a lever to stop all withdrawls are automatically moved to DAO? 
+- there is BIG RED BUTTON available to the validators that stops the bridge from processing deposits / withdrawals (and a smaller green button which can resume this by re-signing the original or a new bridge contract).  
 
-- there is BIG RED BUTTON on the ethereum side which can move all locked assets to the DAO. From there they can be manually distributed subject to vote. The purpose is to deal with catastrophic failure of Vega chain or loss of validators or, or, or anything that could have gone wrong.
+- orderly withdrawal of funds at the end of life of Vega network is in (link to TODO limited network life spec). This will cover insurance pool money too.
 
-- work out how orderly withdrawal of funds happens at the end of life of Vega network (another spec?)
-
-- insurance pool money after all settlement and Vega net shutdown will go into DAO.
 
 
 
--- a/non-protocol-specs/0002-training-wheels.md
+++ b/non-protocol-specs/0002-training-wheels.md
@@ -1,24 +1,77 @@
-# Limits to make first mainnet deployments of Vega safer
+# Decentralised limits and controls
 
-### Philosophy:
-- limits on deposits: reduce how much do we allow people to risk initially 
-- limits on withdrawals: reduce the probability that due to a bug in the system someone can drain all the asssets 
-- users can circumvent limits via sybil attack. This is known and accepted.
-- should we create limits that can be circumvented with Sybil accounts? Yes! It sends a message. 
-- how do we compare across assets? Maximum deposit is a multiple of minimum stake per asset (already in asset in framework) or constant per asset - for the life of the vega chain, changeable via governance
+This spec describes a set of limits and controls that must be supported by deployments of the Vega system to mitigate the risk of loss or misappropriation of funds early in the life of the system while it is relatively less well tested and more likely to contain major bugs or mechanism design issues. The aim is to achieve this by implementing features which reduce both the expected magnitude and the probability of financial loss, from the perspective of participants interacting with the Vega protocol.
+
+These limits are expected to be used in early deployments of Vega and are desiogned to be raised and/or removed over time via the governance protocol as the security and the robustness of the implementation is demonstrated.
+
+
+## Principles
+
+These features:
+- Must not introduce points of centralisation. They are therefore designed to be operated by decentralised governance protocols in keeping with the core Vega protocol.
+- Should be as simple as possible and should operate correctly independent of whether any aspect of the core protocol is functioning as expected (i.e. they should allow recovery from issues in which someone is able to exploit bugs in products, margins, data sourcing, settlement, collateral, etc.). 
+- Are designed to be resilient to failure of the Vega chain and operable assuming only the continued existence of an honest quorum of validators (with their private keys) and the multisig control and bridge pool contract holding control over assets on the Ethereum chain.
+- Are risk-based: they do not need to prevent certain things happening or provide absolute guarantees so much as to reduce the probabilities and magnitudes.
+- Signal the system's readiness and risk level: should be implemented and communicated and raised over time in such a way that the risks inherent in the system are clear, and that going beyond the limits (e.g. using sybils) requires sufficient understanding and effort that it would be done with clear knowledge of the risks.
+- Are intended for early alpha versions of the Vega system and are expected to be implemented almost entirely in the Ethereum bridge smart contracts, no thought has been given to implementations for other chains.
+
+
+## Required limits and controls
+
+We have identified three types of limit/control that will together achieve these aims:
+
+- Deposit limits reduce the exposure of individual participants as well as the total funds at risk
+- Withdrawal controls reduce the probability that funds acquired in error on Vega can be taken outside the control of the system before the error can be fixed
+- A system wide deposite/withdrawal stop provides the ability to buy extra time to investigate or fix any issues 
+
 
 ### Deposit limits
-- maximum deposit per asset from any given Ethereum address and to any Vega address. The limits per asset should be configurable via governance. Limits should be part of whitelisting a new asset. 
+
+These limits restrict the risk that can be easily taken by each participant. They can be overcome by creating multiple keypairs on Vega and the host chain, but per the principles above and given the impact of gas costs, this is not a problem.  
+
+- There will be a `maximum lifetime deposit` configured for each asset. This amount will be stored in the host chain's bridge contract system and set during the whitelisting process for adding a new asset to Vega.
+- It should be possible to amend the `maximum lifetime deposit` via a Vega governance transaction (update asset). This feature may be omitted in an MVP while the network is reset regularly.
+- Any attempt to deposit funds where `total funds deposited by sender address > maximum lifetime deposit` must be rejected.
+- Any attempt to deposit funds where `total funds deposited to receiver address > maximum lifetime deposit` must be rejected.
+
+
 
 ### Withdrawal limits
-- withdrawals over certain amount during a defined time period will be delayed (e.g. 72 hours). During that delay period the Vega validators can blacklist the bundle to prevent a suspicious withdrawal by signing a multi-sig transaction (this could be done manually by agreement between validators or via governance action on Vega blockchain).
 
-- there is BIG RED BUTTON available to the validators that stops the bridge from processing deposits / withdrawals (and a smaller green button which can resume this by re-signing the original or a new bridge contract).  
+These limits reduce the risk that someone who is able to exploit implementation bugs or protocol design issues is able to acquire and withdraw funds that are not intended for them. 
 
-- orderly withdrawal of funds at the end of life of Vega network is in (link to TODO limited network life spec). This will cover insurance pool money too.
+- A single `withdrawal delay period` to apply for all assets will be stored in the bridge contract system. This will default to 120 hours (5 days) and may be changed via multisig control.
+- There will be a `withdrawal delay threshold` configured for each asset. This amount will be stored in the host chain's bridge contract system and set during the whitelisting process for adding a new asset to Vega.
+- It should be possible to amend the `withdrawal delay threshold` via a Vega governance transaction (update asset). This feature may be omitted in an MVP while the network is reset regularly.
+- Any withdrawal bundle created where `withdrawal amount > withdrawal delay threshold` will be rejected by the bridge if `time since bundle creation <= withdrawal delay period` (the bundle must therefore contain a timestamp of its creation, which must be validated by nodes before they sign the bundle)
+- An API is required to list all pending withdrawals (i.e. those that have not been executed on the bridge and have not expired) on the Vega chain across all public keys. This is required to allow the community to identify transactiosn that require withdrawals to stopped pending investigation
+- [TODO: maybe delete, see comment on PR] A quorum of validators may sign a transaction bundle which, when submitted to the bridge contract system, will halt a withdrawal with a given ID. Once a valid transaction of this type is received, any bundle with the specified ID must be rejected.
 
 
+### Global bridge stop
+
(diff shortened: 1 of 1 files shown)


C0deMunk33:
In general, the less on-chain (eth-side) the better. 

Notes:
"should we create limits in core that can be circumvented with multiple accounts? Yes! It sends a message."
- agreed, we can't prevent sybil, but we can let people know that if you want to sybil, you have to make a conscious decision to risk more of your money.

"how do we compare across assets? Maximum deposit is a multiple of minimum stake per asset (already in asset in framework) or constant per asset - for the life of the vega chain, changeable via governance"
- this should not be changable by governance for WWN, constant per asset makes most sense. I can code this into the contract and we can hard-set this before launch.

"maximum deposit per asset from any given Ethereum address and to any Vega address"
- I will add this to the post-audit smart contract itself.

"withdrawals amount and timing? e.g. withdrawals only allowed 1 week after last settlement?"
- we have withdrawal expiry built in already, just set that in governance

"withdrawals over certain amount during a time period subject to Vega token holder vote?"
- I will have a hard time coding this, but if CORE delays issuing of the signature bundle, this can be easily accomplished. 

"do we need a lever to stop all withdrawls are automatically moved to DAO?"
- if we want this, I can wire in a BIG RED BUTTON to halt withdrawals that requires multisigControl

"here is BIG RED BUTTON on the ethereum side which can move all locked assets to the DAO"
- no? this doesn't exist yet and all it would be able to do is kill despoits/withdrawals. It can't move anything. (the asset pool exists because of this problem). if we do this, we need some sort of emergency "disable multisig control" that is triggered by... someone... 

"insurance pool money after all settlement and Vega net shutdown will go into DAO."
- this will require CORE to write a withdrawal order to it's target.

barnabee:
> * this should not be changable by governance for WWN, constant per asset makes most sense. I can code this into the contract and we can hard-set this before launch.

WWN will be "launched" multiple times, and assets will be added. I think it makes sense to make this part of the allow-listing of a new asset, even if not _changeable_ by governance initially.

barnabee:
> * no? this doesn't exist yet and all it would be able to do is kill despoits/withdrawals. It can't move anything. (the asset < pool exists because of this problem). if we do this, we need some sort of emergency "disable multisig control" that is triggered by... someone...

this could all be triggered by the validaotrs so I think it could use multisig control

davidsiska-vega:
@tamlyn10 - ready for review... (and @jeremyletang  if you're interested).

tamlyn10 (approved):

peterbarrow (approved):

non-protocol-specs/0002-training-wheels.md
```diff
@@ -0,0 +1,27 @@
+# Limits to make first mainnet deployments of Vega safer
+
+### Philosophy:
+- limits on deposits: reduce how much do we allow people to risk initially 
+- limits on withdrawals: reduce the probability that due to a bug in the system someone can drain all the asssets 
+
+### Questions: 
+- should we create limits that can be circumvented with Sybil accounts? Yes! It sends a message. 
+- how do we compare across assets? Maximum deposit is a multiple of minimum stake per asset (already in asset in framework) or constant per asset - for the life of the vega chain, changeable via governance
```
edd: **question (blocking)**: Decision required. Is it? What multiple?

non-protocol-specs/0002-training-wheels.md
```diff
@@ -0,0 +1,27 @@
+# Limits to make first mainnet deployments of Vega safer
+
+### Philosophy:
+- limits on deposits: reduce how much do we allow people to risk initially 
+- limits on withdrawals: reduce the probability that due to a bug in the system someone can drain all the asssets 
+
+### Questions:
```
edd:
```suggestion
```

Remove header

non-protocol-specs/0002-training-wheels.md
```diff
@@ -0,0 +1,27 @@
+# Limits to make first mainnet deployments of Vega safer
+
+### Philosophy:
+- limits on deposits: reduce how much do we allow people to risk initially 
+- limits on withdrawals: reduce the probability that due to a bug in the system someone can drain all the asssets 
+
```
edd, line 26:
```suggestion
- users can circumvent limits via sybil attack. This is known and accepted.
```

non-protocol-specs/0002-training-wheels.md
```diff
@@ -0,0 +1,27 @@
...
+### Questions: 
+- should we create limits that can be circumvented with Sybil accounts? Yes! It sends a message.
```
edd: Remove, accept above change

non-protocol-specs/0002-training-wheels.md
```diff
@@ -0,0 +1,27 @@
...
+### Philosophy:
+- limits on deposits: reduce how much do we allow people to risk initially 
+- limits on withdrawals: reduce the probability that due to a bug in the system someone can drain all the asssets 
+
+### Questions: 
+- should we create limits that can be circumvented with Sybil accounts? Yes! It sends a message. 
+- how do we compare across assets? Maximum deposit is a multiple of minimum stake per asset (already in asset in framework) or constant per asset - for the life of the vega chain, changeable via governance
+
+### Deposit limits
+- maximum deposit per asset from any given Ethereum address and to any Vega address. The limits per asset should be configurable via governance. Limits should be part of whitelisting a new asset.
```
edd: **question (blocking)**: Where should this block happen. Governance param suggests in Vega, but if it's in Vega then the deposit has already happened to the smart contract, and thus needs to be returned

non-protocol-specs/0002-training-wheels.md
```diff
@@ -0,0 +1,27 @@
...
+
+### Questions: 
+- should we create limits that can be circumvented with Sybil accounts? Yes! It sends a message. 
+- how do we compare across assets? Maximum deposit is a multiple of minimum stake per asset (already in asset in framework) or constant per asset - for the life of the vega chain, changeable via governance
+
+### Deposit limits
+- maximum deposit per asset from any given Ethereum address and to any Vega address. The limits per asset should be configurable via governance. Limits should be part of whitelisting a new asset. 
+
+### Withdrawal limits
+- withdrawals over certain amount during a defined time period will be delayed (e.g. 72 hours). During that delay period the Vega validators can blacklist the bundle to prevent a suspicious withdrawal by signing a multi-sig transaction (this could be done manually by agreement between validators or via governance action on Vega blockchain).
```
edd: **question (blocking)**: Needs more detail. What does this look like in practice? @C0deMunk33 ?

non-protocol-specs/0002-training-wheels.md
```diff
@@ -0,0 +1,27 @@
...
+
+- there is BIG RED BUTTON available to the validators that stops the bridge from processing deposits / withdrawals (and a smaller green button which can resume this by re-signing the original or a new bridge contract).  
+
+- orderly withdrawal of funds at the end of life of Vega network is in (link to TODO limited network life spec). This will cover insurance pool money too.
```
edd: **nitpick (blocking)**: Please add a link before merging.

edd: **question (blocking)**: Please document a workflow of how this works. The end of a network life to me is when validators stop producing blocks, at which point obviously they can't multisig anything. This needs some design thought.

edd (changes requested):
I don't think this is ready to be merged. It should be detailed enough that we can plan an implementation off it, and it isn't there yet.

barnabee:
Thanks @edd, I'll make some updates and be sure to get your _seal_ of approval before merging

non-protocol-specs/0002-training-wheels.md
Vegaklaus: I can generalise the text I put on Discord about Sybil attacks for that case so we can put something in an FAQ on why we accepted this

non-protocol-specs/0002-training-wheels.md
barnabee: Expected to be in bridge. My understanding is supported assets are added via governance but this also triggers an Ethereum "allowlisting", I expect the parameter would be passed as part of the same governance.

non-protocol-specs/0002-training-wheels.md
barnabee: Will clarify and reword

non-protocol-specs/0002-training-wheels.md
barnabee: @edd I am trying to work out wha'ts missing here, will try splitting this to the wording suggested below but please clarify what you'd like to see here if this doesn't do it, or I'll add this when I make updates:

- each withdrawal will be subject to some (possibly 0) amount of delay between the withdrawal bundle being created and the withdrawal being committed to the Vega (and possible Ethereum) chain(s) and therefore publicly broadcast and "known" to anyone who cares to see it
- the delay will be definable by threshold based on withdrawal amount as a multiple of the deposit limit (so it is cross-asset and thresholds need to be defined once for the whole system), so larger withdrawals can be subject to longer delays
- during that delay period multisigcontrol will be able to prevent a withdrawal succeeding via an Ethereum transaction

non-protocol-specs/0002-training-wheels.md
barnabee: Spec does not exist yet, do you want us to delay merging (once issues are fixed) until it does?

barnabee: The point of that comment is that this documentation is not in the scope of this spec. Does not seem like the other spec should be blocking this one merging to me, as the functionality described herein has no dependency on this, it's just a helpful note.

edd: I would prefer that, but I'll live without

non-protocol-specs/0002-training-wheels.md
edd: That's most of the detail that is required. For point 1, the multisig process has no interaction with ethereum, so point 3 becomes more interesting - what is broadcast on Ethereum that blocks the withdrawal. If it's a withdrawal ID, that can't be known before the bundle is broadcast on the Vega event bus, as otherwise there's no way for the person sending this TX to know the ID. So I guess it would have to be a Vega address to ban from withdrawals? That's my first thought, but it's easily dismissable. 

tldr more thought on point three please. Where does that TX happen and what data does it contain and who, in practice, broadcasts it?

barnabee: > the multisig process has no interaction with ethereum

I'm not sure I understand this... the multisig process creates a tx that will be processed on Ethereum so it can interact with it by including a parameter like `withdrawalCreatedTimestamp` that can be queried by the bridge to determine whether to allow the withdrawal bundle to be processed in a block.

My understanding is it's possible for humans operating the validator nodes to manually sign ad-hoc transactions. This would be done by a quorum of validators signing something like a call to `blockWithdrawal(tx_bundle_id)`. Then you can refuse to accept the bundle on Ethereum before the delay expires, or ever, if its ID was blocked.

Some of these things feel a bit like implementation details rather than specs though?

What we want is: 
- The delay between generation of the withdrawal bundle which others can know about and the funds leaving the bridge pool
- The ability of a manual quorum-of-validator signed transaction submitted on Ethereum before that delay expires to prevent the withdrawal happening

edd: # My summary of what I think you have proposed
That helps somewhat. So a withdrawal request is created by the user, which will be visible and have an ID, but the multisig bundle will not start being created until `timeout` is reached, at which point it will be signed - even if it is going to be blocked.

Then we're going to add a new function to the bridge smart contract** that allows a multisig bundle to block a withdrawal by the vega event ID. This signature bundle gathering process will be started manually by a Vega validator runer, and submitted to Ethereum manually by a Vega validator runner

* This specifically relates to your point here that uses the phrases 'committed' and 'known', which it's worth being clear about because they are not specific:
> each withdrawal will be subject to some (possibly 0) amount of delay between the withdrawal bundle being created and the withdrawal being committed to the Vega (and possible Ethereum) chain(s) and therefore publicly broadcast and "known" to anyone who cares to see it

# Alternative
If the delay is added, then can't we simply make the WithdrawalBlock action a Vega-only TX that doesn't interact with the smart contract? That way we don't need to add a new function to the solidity, or require validators to spend gas.

barnabee: > That helps somewhat. So a withdrawal request is created by the user, which will be visible and have an ID, but the multisig bundle will not start being created until timeout is reached, at which point it will be signed - even if it is going to be blocked.

No I think validators can sign the bundle immediately. It can be checked by the bridge contract when the withdrawer tries to submit it to Ethereum and if it's too early the Ethereum transaction would be rejected (and I think not incur any fees) and they would submit again later. I'm pretty sure it is fine to sign it before you know if it will be blocked and before the delay is up.

We want to ensure blocks work even if someone has managed to bring down the Vega chain, so the delay/block needs to be enforced on Ethereum. Only time a validator (or us!) incurs fees is if there's a need to block something.

barnabee: @C0deMunk33 please correct me if any of what I'm stating/proposing here differs from what you think we discussed

C0deMunk33:
So here's what I distilled from all that, we want:
1. Delayed withdrawal - configurable delay before contract accepts withdrawal order
2. Multisig controlled halt order that can be issued any time before withdrawal/expiry

barnabee:
> So here's what I distilled from all that, we want:
> 
>     1. Delayed withdrawal - configurable delay before contract accepts withdrawal order
> 
>     2. Multisig controlled halt order that can be issued any time before withdrawal/expiry

@edd in reply to [this thread on Slack](https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1614621212079400?thread_ts=1614614672.077900&cid=CAHA5EX0F):

Yes, that's what we want. It is quite close to the original text, are you happy for me to change the wording to this? Are there other details you would like to see included from the discussion above?

barnabee:
@C0deMunk33 can you confirm my [assumptions here](https://github.com/vegaprotocol/specs/pull/467#discussion_r584757389) are ok?

edd:
> It is quite close to the original text, are you happy for me to change the wording to this? Are there other details you would like to see included from the discussion above?

Yeah it’s quite close, so yes I think it’s the clarifications from above that need to get included. 

# Delayed withdrawal
Withdrawals above a configured size (see below) will be forced to wait for a configured delay. This is being introduced to protect us from end users abusing a previously unknown loophole and running off with all the collateral. The logic lives mostly in the bridge smart contract.

The delay period defaults to 72 hours in order to give validators a chance to skim all the pending deposits. If any are suspicious looking, they can be blacklisted by withdrawal ID

## Multisign halt order
1. A validator identifies a withdrawal bundle they would like to blacklist
2. Validator submits a new Vega TX - `multisigWithdrawalhalt`
3. Validators gather multisig bundle
4. Anyone submits the multisigned halt signature bundle to the smart contract
5. If the withdrawal ID is submitted after the configured wait, it will be rejected
 
## User Workflow
1. A user requests a withdrawal from the system
1. As it does now, Vega will sign a multisig bundle to pass to the user, for submission to the smart contract on Ethereum
1. When the user submits the multisig bundle to the bridge contract:
    1. Submission of the multisig bundle to the bridge contract before `delayed_time`, but below `delay_deposit_value` will be accepted and processed as before
    1. Submission of the multisig bundle to the bridge contract before `delayed_time`, but at or above  will be rejected
    1. Submission of the multisig bundle to the bridge contract at or after `delayed_time`:
        1. Unless explicitly blacklisted, will be accepted

## Limitations
- A user could submit multiple withdrawals just below the configured limit. As users pay the gas costs on Ethereum, this would be expensive, and so we’re not protecting against that
- This will be annoying for people who have genuinely traded well and want to exit the system. Documentation will clarify that this limit will go away.

## Configuration

There are no new network parameters for the system, as the delay and value limits are defined in the bridge smart contract. The smart contract will need to allow: 
- delayed_deposit_value - the value (defined once for all assets, representing the tokens)* above which withdrawals will be delayed
- delay_time - the amount of time (defined in seconds) that must have passed before the bridge will process a withdrawal request


* This is unclear. I mean not defined in dollars/gbp. It’s early 🤷 

> > So here's what I distilled from all that, we want:
> > ```
> > 1. Delayed withdrawal - configurable delay before contract accepts withdrawal order
> > 
> > 2. Multisig controlled halt order that can be issued any time before withdrawal/expiry
> > ```
> 
> @edd in reply to [this thread on Slack](https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1614621212079400?thread_ts=1614614672.077900&cid=CAHA5EX0F):
> 
> Yes, that's what we want. It is quite close to the original text, are you happy for me to change the wording to this? Are there other details you would like to see included from the discussion above?

barnabee:
I am significantly rewriting/improving this spec and have come across a question that might result in a change (simplification). I discussed this briefly with @tamlyn10 earlier but would like everyone's input here… 

With the delay-plus-block transaction option, given that:

- If there is an issue that's possibly been exploited it could be exploited in other withdrawals (both in progress and not-yet-made), including those that are beneath the threshold size, so any issue being investigated would likely require stopping other transactions or a significant ovehread of checking them all; and
- As the delay is finite (and would be kept as short as reasonably practical), any serious investigation (and possible fix) would quite likely require employing the 🛑 BIG RED STOP BUTTON 🛑 until the investigation is complete and if necessary, a fix is deployed.

SO, given these points, do we need to build the withdrawal halt feature at all or would the delay plus the ability to temporarily stop all withdrawals and deposits be just as good (or even preferable)?

@C0deMunk33 @davidsiska-vega @edd @tamlyn10

edd:
> SO, given these points, do we need to build the withdrawal halt feature at all or would the delay plus the ability to temporarily stop all withdrawals and deposits be just as good (or even preferable)?
 
I prefer this approach as it's simpler. I assume that the 'Allow Deposits or not' flag is still set with a multisig bundle on the smart contract, rather than being implemented in the nodes?

barnabee:
> I prefer this approach as it's simpler. I assume that the 'Allow Deposits or not' flag is still set with a multisig bundle on the smart contract, rather than being implemented in the nodes?

I believe it would implemented using the fact that we have a simple asset pool contract that has a designated "asset controller" aka bridge contract that accepts bridge deposits and executes withdrawals. Multisig control has the ability to change this controller contract (including setting it to no contract), and so the "stop deposits and withdrawals" flag is in reality _using an action that is already built_ to remove/restore the bridge's ability to interact with the asset pool.

So as I understand it, we'd simplify twice by not building the "halt individual withdrawal" logic and also not needing any new features for stopping withdrawals that don't exist already. We just need to build the delays, for the withdrawal limits, in that case. 

@C0deMunk33 please correct me if this is wrong

edd (approved):
New version is nice and clear, thanks

barnabee:
Merging, further changes can be addressed w/ new issues

PR #483, opened 2021-02-09
Update to liquidity auctions

davidsiska-vega:
Closes #482

### DIFF
--- a/specs/0035-liquidity-monitoring.md
+++ b/specs/0035-liquidity-monitoring.md
@@ -6,11 +6,17 @@ Specification PR: https://github.com/vegaprotocol/product/pull/322
 
 Liquidity in the market is not only a desirable feature from a trader's point of view, but also an important consideration from the risk-management standpoint. Position of a distressed trader can only be liquidated if there's enough volume on the order book to offload it, otherwise a potentially insolvent party remains part of the market.
 
-Similarly to [price monitoring](0032-price-monitoring.md), we need to be able to detect when the market liquidity drops below the safe level, launch a "liquidity seeking" auction (in which, due to the liquidity mechanics](./0044-lp-mechanics.md), there is an incentive through the ability to set fees, to provide the missing liquidity) and terminate it when the market liquidity level is back at a sufficiently high level.
+Similarly to [price monitoring](0032-price-monitoring.md), we need to be able to detect when the market liquidity drops below the safe level, launch a "liquidity seeking" auction (in which, due to the [liquidity mechanics](./0044-lp-mechanics.md), there is an incentive through the ability to set fees, to provide the missing liquidity) and terminate it when the market liquidity level is back at a sufficiently high level.
 
-Note that, except where LP orders are off the book if there are no limit orders to peg the price to, there is one-to-one correspondence between the total stake committed by liquidity providers (LPs), see [LP mechanics](specs/0044-lp-mechanics.md) spec, and the total supplied liquidity. Thus even though below we will be comparing `target_stake` with `supplied_stake` what this really monitors is amount of liquidity supplied. 
+Note that as long as all pegs that LP batch orders can peg to exists on the book there is one-to-one correspondence between the total stake committed by liquidity providers (LPs), see [LP mechanics](specs/0044-lp-mechanics.md) spec, and the total supplied liquidity. 
+Indeed 
+```
+lp_liquidity_obligation_in_ccy_siskas = stake_to_ccy_siskas ⨉ stake.
+```
+Thus it is sufficient to compare `target_stake` with `supplied_stake` while also ensuring that `best_bid` and `best_offer` are present on the book (*).
 Note that [target stake](0041-target-stake.md) is defined in a separate spec.
 
+(*) Having `best_bid` and `best_offer` implies that `mid` also exists. If, in the future, [LP batch orders](0038-liquidity-provision-order-type.md) are updated to allow other pegs then the protol must enforce that they are also on the book.
 
 ## Liquidity auction network parameters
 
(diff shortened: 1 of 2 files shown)


barnabee (approved):
Pushed some tweaks, approved

barnabee (approved):
Approve

specs/0035-liquidity-monitoring.md
```diff
@@ -6,31 +6,47 @@ Specification PR: https://github.com/vegaprotocol/product/pull/322
...
+
+(*) If we are in an auction then this should be measured by building a shadow order book: pretend that the auction uncrosses, keep limit orders left over from the auction uncrossing that do not expire due to being "Good For Auction", and unpark the LPs special orders.  
 
-When `supplied_stake` ≥ `target_stake`.
+Similarly to [price monitoring](0032-price-monitoring.md), the auction should be triggered pre-emptively. That is, the transaction that would have triggered the liquidity auction should be [re-]processed once auction mode is on. Please note that this transaction may be cancelled if it's type is not valid for auction, however even in that case the market should still go into the auction mode to send a signal that it requires more liquidity.
+
+
+## Trigger for exiting the auction
 
+With `target_liquidity` and `present_firm_liquidity` defined as above we exit if
```
witgaw: Isn't `present_firm_liquidity` redundant now?

specs/0035-liquidity-monitoring.md
```diff
@@ -6,31 +6,42 @@ Specification PR: https://github.com/vegaprotocol/product/pull/322
...
 `supplied_stake` is the sum the stake amounts committed by all the LPs in the market (see [LP mechanics](specs/0044-lp-mechanics.md)). 
 
 ## Trigger for entering an auction
 
-When `supplied_stake` < c<sub>1</sub> x `target_stake`, 
-where 0 < c<sub>1</sub> < 1, to reduce the chance of another auction getting triggered soon after e.g. c<sub>1</sub> = 0.7. 
+Here 0 < c<sub>1</sub> < 1, to reduce the chance of another auction getting triggered soon after e.g. c<sub>1</sub> = 0.7. The parameter c<sub>1</sub> is a network parameter.
 
-Similarly to [price monitoring](0032-price-monitoring.md), the auction should be triggered pre-emptively. That is, the transaction that would have triggered the liquidity auction should be re-processed once auction mode is on. Please note that this transaction may be cancelled if it's type is not valid for auction, however even in that case the market should still go into the auction mode to send a signal that it requires more liquidity.
+The auction is triggered when
```
witgaw, line 31: I'd start the paragraph with this and move "Here 0 < c<sub>1</sub> < 1 (...)" below it, reads more easily to me that way.

witgaw (changes requested):
Overall looks good to me, just two comments.

specs/0035-liquidity-monitoring.md
```diff
@@ -6,31 +6,42 @@ Specification PR: https://github.com/vegaprotocol/product/pull/322
...
+```
+supplied_stake < c_1 x target_stake OR there is no best_bid OR there is no best offer.
+```
+Similarly to [price monitoring](0032-price-monitoring.md), the auction should be triggered pre-emptively. That is, the transaction that would have triggered the liquidity auction should be [re-]processed once auction mode is on. Please note that this transaction may be cancelled if it's type is not valid for auction, however even in that case the market should still go into the auction mode to send a signal that it requires more liquidity.
```
witgaw, line 37: This is comment is not related to other changes in this PR, but came up during my work on this and thought we could take care of it here too:

The pre-emptive bit should exclude LP closeouts, right? As in we want to closeout anyway (because the guy just can't afford margin) and switch to auction. Or are we just assuming that a closeout isn't an explicit transaction and hence it's ok?

barnabee: I think that'scorrect, yeah. If the price move is real I think we probably shouldn't reject/delay until the auction the whole transaction that caused the price move, which we'd do if pre-empting. It seem probably right that we fill the order or whatever and then closeout the LP if needed and enter a liquidity auction. It's definitely not the only way we could handle it but it feels right....

Main alternative is we detect the LP would be closed out and enter an auction. But this seems bad because from the PoV of the auction there's no reason to be there and we end up creating a situation where some prices become less "valid" / not hittable due to LPs margin position, which I don't like.

cc @tamlyn10 @davidsiska-vega

davidsiska-vega:
@barnabee @witgaw I have another question: we say 
``` 
supplied_stake is the sum the stake amounts committed by all the LPs in the market (see LP mechanics).
```
but in fact isn't the relevant amount the LPs bond account balance (so the sum of all the LPs bond account balances)? The fact that an LP "committed" to something but then they don't have the margin / bond balance to keep commitment doesn't seem right to me.

barnabee:
> @barnabee @witgaw I have another question: we say
> 
> ```
> supplied_stake is the sum the stake amounts committed by all the LPs in the market (see LP mechanics).
> ```
> 
> but in fact isn't the relevant amount the LPs bond account balance (so the sum of all the LPs bond account balances)? The fact that an LP "committed" to something but then they don't have the margin / bond balance to keep commitment doesn't seem right to me.

No because we put **all** the orders for their commitment on the book even if their balance is lower because of penalties IIRC. They are either supplying their full commitment or nothing cos they got closed out. As in:

> 4. If the liquidity provider's orders or positions were closed out, and they are therefore no longer supplying the liquidity implied by their Commitment: In this case the liqudity provider's Commitment size is set to zero and they are no longer a liquidity provider for fee/reward purposes, and their commitment can no longer be counted towards the supplied liquidity in the market. (From a Core perspective, it is as if the liquidity provider has exited their commitment entirely and they no longer need to be tracked as an LP.)

Also we removed the definition of `supplied stake` from LP mechanics.

specs/0035-liquidity-monitoring.md
davidsiska-vega: I think all close-outs (LPs and other parties) should happen if there is volume on the book and only then should we enter an auction.

davidsiska-vega: If we think the specs need clarifying I suggest a separate issue because it will impact both price monitoring and liquidity monitoring and is separate  to issue #483.

witgaw: OK, thanks for clarifying. I'll create a separate update to the spec to cover that.

witgaw (approved):
Looks great now, thanks for incorporating my suggestions.

PR #484, opened 2021-02-16
Add clarification for entering monitoring auctions

witgaw:
Clarify when orders causing **liquidity auction** should be triggered pre-emptively and when should auction be launched after the change to the order book state has already occurred.

Modify **price monitoring** spec to only go into auction mode if the "offending" transaction is valid for auction. Otherwise the transaction should get rejected, an appropriate event get sent and market should continue in its current trading mode. See https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1613130964258500.


witgaw (commented):
Ready for review.

specs/0035-liquidity-monitoring.md
```diff
@@ -34,7 +34,22 @@ total_stake < c_1 x target_stake OR there is no best_bid OR there is no best off
...
+### Increasing target stake
+
+If an incoming order would match orders on the book resulting in trades increasing `target_stake` so that liquidity auction gets triggered then:
+
+- if the incoming order would stay on the book in auction mode the auction should get triggered pre-emptively (the order doesn't get matched in market's current trading mode, market switches to auction mode and the incoming order gets added to the book once market is in auction mode).
+- if the incoming would not stay on the book in auction mode then the order gets rejected, appropriate event gets sent to signal to the market that insufficient `supplied_stake` has prevented the market from increasing the open interest and the market remains in the current trading mode.
```
barnabee, line 42: I would like to check on this @davidsiska-vega @tamlyn10 @witgaw 

Do we want to reject an order that should trade because it would increase OI so that a liquidity auction is triggered, or should we let the trade happen and trigger the auction? It feels odd not to do a trade when there's liquidity and this would be the only case where we do that (in other cases they could trade when the auction ends).

I think this may be the right call but just want to know we've all looked at this and agreed with it, and I can't remember if the original discussion covered this fully...

barnabee (approved):
Approved but please check the comment on line 42

specs/0035-liquidity-monitoring.md
davidsiska-vega: I think this is fine. 

In the future we can tinker with how we measure the target stake so that, for example, this event would temporarily increase the target stake (so that if another LP wants to go in but only bidding higher fee then there is space for them). 

For now I agree with the way it's written.

specs/0032-price-monitoring.md
```diff
@@ -18,13 +18,13 @@ The dynamics of market price movements are such that prices don't always represe
...
...ice monitoring is meant to stop large market movements that are not "real" from occuring, rather than just detect them after the fact. To that end, it is necessary to pre-process every transaction and check if it triggers the price monitoring action. If pre-processing the transaction doesn't result in the trigger being activated then it should be "committed" by generating the associated events and modifying the order book accordingly (e.g. generate a trade and take the orders that matched off the book). On the other hand if the trigger is activated and the submitted transaction is valid for auction mode, the entire order book **along with that transaction** needs to be processed via price protection auction. If the transaction which activate the trigger is not valid for auction, then it should get rejected and market should continue in the current trading mode. Auction period associated with a given distribution projection horizon and probability level will be specified as part of market setup. Once the auction period finishes the trading should resume in regular fashion (unless other triggers are active, more on that in [reference-level explanation](#reference-level-explanation)).
```
davidsiska-vega, line 21: A meta comment. Using new lines more often (e.g. per sentence) will allow better targeting of comments. 

The actual comment: `If the transaction which activate the trigger is not valid for auction, then it should get rejected and market should continue in the current trading mode. ` Do I understand correctly that e.g. a market order that would push mark price out of bounds will be *rejected* but *no auction* will be triggered?

barnabee: Yes that's correct. Because you are going to discard the order and execute nothing either way, there isn't much point entering an auction. No order would be added to the book that would push the price into the monitoring auction, so it's an unnecessary auction. If someone comes along with a limit order that would cause the same move then an auction makes sense.

We need the rejection reason to be clear on why.

NB: this is the same behaviour that I've seen/used in the past in other venue with similar features

specs/0035-liquidity-monitoring.md
```diff
@@ -34,7 +34,22 @@ total_stake < c_1 x target_stake OR there is no best_bid OR there is no best off
...
+
+### Decreasing supplied stake
+
+If the [liquidity provision transaction would decrease](0044-lp-mechanics.md#liquidity-provider-proposes-to-amend-commitment-amount) `supplied_stake` so that liquidity auction gets triggered then the liquidity provision amendment should be rejected and market should stay in it's current trading mode.
```
davidsiska-vega, line 46: I feel the spec `0044` already covers this case:
```
maximum-reduction-amount = total_stake - target_stake
```
and
`The above formulae result in the fact that if `maximum-reduction-amount = 0`, then `actual-reduction-amount = 0` and therefore the liquidity provider is unable to reduce their commitment amount.`

davidsiska-vega (approved):
If the comments I made have already been taken into account and discussed elsewhere then I am happy with the PR as is.

PR #486, opened 2021-02-19
Add draft V2 token spec

barnabee:


non-protocol-specs/0002-token-v2.md
```diff
@@ -0,0 +1,58 @@
...
+Version 2 of the $VEGA token replaces the existing token and providse two crucial upgrades:
+
+1. It allows for more sophisticated unlocking (vesting) schedules to be implemented and enforced on chain
+1. It places the total supply of the token in the control of  network/token holder governance
+
+
+## 1. Token supply
+
+1. Supply is "fixed" initially at 64,999,723
+1. Supply **can** be increased by the community (e.g. multisig control or token holders voting...?)
```
MM0819: I don't understand why Multisig control would be permitted to alter the supply of tokens, when tokens are intended to govern the constituents of Multisig control. It feels like the only reason to do this would be a token holder vote?

non-protocol-specs/0002-token-v2.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. Deployment/activation of the token contract will automatically issue tokens based on the wallet balances of the V1 token at deployment time
+1. The contract owner must be able to assign vesting tranche for auto-issued tokens
+1. Each address receiving auto-issed tokens may have a different vesting tranche
+1. Auto-issued tokens with no assigned vesting tranche must remain locked
+
+
+## 3. Issuance mechanics
+
+1. The contract owner must be able to assign the ability to issue tokens to new wallets or smart contract addresses ("permitted issuers")
```
MM0819, line 31: I would recommend that the contract owner is a Gnosis Multisig contract once this goes live, as it has a lot of power.

non-protocol-specs/0002-token-v2.md
```diff
@@ -0,0 +1,58 @@
...
+1. The contract onwer must be able to assign limits per permitted issuer address for total issuance by that address
+1. The contract onwer must be able to change the issuance limit for a permitted issuer address
+1. When issued, tokens are always issued to a vesting category/tranche OR must be locked until a tranche is assigned
+1. Tokens must remain locked until they unlock per the vesting (unlock) rules below, as applied for the tranche they are assigned to (the rules can mean they are immediately unlocked)
+1. Tokens cannot be recalled/clawed back or re-issued once issued
+
+
+## 4. Vesting/unlock tranches and rules
+
+1. New tranches can be created at any time by the contract "owner"
```
MM0819, line 43: Can a tranche be applied to tokens which have already had a tranche applied to them, and that tranche has subsequently completed vesting? For example, if I receive 1 token, which I'm permitted to sell in six month's time, can the owner apply six months to this token unilaterally without my permission at six months + 1 minute?

non-protocol-specs/0002-token-v2.md
```diff
@@ -0,0 +1,58 @@
...
+1. Each tranche has a `cliff duration` (i.e. 1 month, 4 months, immediate)
+1. Each tranche has a `vesting duration` (i.e. 6 months, 9 months, immediate/all at once)
+1. The countdown to unlock (i.e. start of the cliff) is triggered by a manual smart contract call with a `trigger start date/time` parameter, which may be in the past or future
```
MM0819, line 46: Clarification point: this is the owner calling a function to effectively say "public sale complete", which is the trigger for lockups per the current contracts with investors?

I presume this does not mean for every single vesting tranche someone needs to call a function to trigger it?

non-protocol-specs/0002-token-v2.md
```diff
@@ -0,0 +1,58 @@
...
+1. During an upgrade, all addresses with balances should retain their original balance 
+1. During an upgrade, all unlocked tokens should remain unlocked
+
+
+## 6. Security and control
+
+1. The contract owner must not be able to change anything about tokens that have been issued and unlocked
+1. The contract owner must not be able to change the supply of the token
+1. The contract owner must not be able to effect a token contract upgrade
+1. The contract owner must not be able to modify the functionality of the token contract, except to renounce ownership entirely, for example by assigning a null address as the new contract owner (which would mean the contract owner no longer held any powers)
```
MM0819: Assigning a null address seems like a bad idea given the power the owner has. I would think this contract always wants to be owned by someone. Therefore, changing owner would involve proposing an address, and the change takes place only when the new owner accepts, which proves they own the private key for the new owner address.

non-protocol-specs/0002-token-v2.md
barnabee: Mutlisig control is controlled by the validators who are picked by token holders, so it's pretty much the same same really

I think @C0deMunk33 has some preference for multisigcontrol over token holder voting…

non-protocol-specs/0002-token-v2.md
barnabee: Probably a good idea (though out of scope for the spec I guess, we may want to note that this should be possible)

non-protocol-specs/0002-token-v2.md
MM0819: So if validators increased supply by 10x, could it have any negative impact on Vega network governance? E.g. voting thresholds to pass governance changes? If thresholds only apply to staked tokens, and if Multisig control can only increase supply, but not issue those tokens to anybody, then I think this sounds okay.

non-protocol-specs/0002-token-v2.md
barnabee: 1. Will clarify that unlocked tokens can never become locked, it is one way and irreversible
2. I am 50/50 on whether locked tokens should be re-trancheable OR whether tranche cliff/vesting durations should be changeable (perhaps only before the trigger is activated). I think ideally to reduce risk we would be able to change them, but is it worth the complexity?

non-protocol-specs/0002-token-v2.md
barnabee: It is possible not all tranches will have the same trigger. I assumed we need to do it per tranche but we could do somethign jazzier to avoid that. There will probably be a total of <10 tranches, definitely <20.

non-protocol-specs/0002-token-v2.md
barnabee: We need to remove all control from our org long term, so probably it's making ownership be in the same hands as the "community" control functions. Tranches etc. are not necessary in future but _if_ the community decides to increase supply they'll need to also be able to issue the new tokens somewhere (likely to the Vega bridge so the protocol can allocate them).

non-protocol-specs/0002-token-v2.md
MM0819: I'd say it's okay to change a tranche before it triggers, and after that the rule specified in (1) applies. I.e. can never change, and once unlocked can never be re-tranched.

non-protocol-specs/0002-token-v2.md
barnabee: New spply has to get issued somewhere... I think it'd need to be / usually be the bridge but that address can change via multisigcontrol IIRC so whoever's increasing supply also probably needs to be able to choose where to issue it, I guess.

barnabee: In general Vega security assumptions assume you can trust multisigcontrol - if multisigcontrol is evil that means the Vega chain is evil anyway so it doesn't really matter to prevent multisigcontrol having absolute power because it already does... (@C0deMunk33 correct me if I'm wrong there...)

barnabee: After a chat I will add a time limit so that the ability to increase supply would not be possible _at all_ before a certain date, something like mid 2022. This would prevent early, less secure/well tested networks getting messed with in some way via this feature.

C0deMunk33: Since the network can presumably always just vote in a new token we may need an adjustable supply. The real question, as you said @barnabee is: where do the newly created tokens go?
couple ideas that spring to mind:
VEGA inflates for all wallets at the same rate. Increasing supply without diluting any one hodlrs value, like a stock split
and/or 
the new supply is 100% awards for things (grants/validators/user incentives) so those voting don't automatically benefit.

Re: validator security having absolute power... yeah, I don't really see another way, it's decentralized in this way or not at all at the moment... it's an open protocol however so I dunno.

non-protocol-specs/0002-token-v2.md
```diff
@@ -0,0 +1,58 @@
...
+
+1. Supply is "fixed" initially at 64,999,723
+1. Supply **can** be increased by the community (e.g. multisig control or token holders voting...?) 
+1. Supply cannot be changed by the contract owner/administrator
+
+
+## 2. Migration from Token V1
+
+1. Deployment/activation of the token contract will automatically issue tokens based on the wallet balances of the V1 token at deployment time
+1. The contract owner must be able to assign vesting tranche for auto-issued tokens
```
C0deMunk33: maybe something like "must be able to assign [user/hodler/investor... to] vesting tranche

non-protocol-specs/0002-token-v2.md
barnabee: I can't see any reason we would want to inflate wallets. The reason to allow inflation would be so that new rewards could be awarded and/or to supply an on-chain treasury, or similar things.

We would want the new supply to end up in the bridge for the Vega chain to do whatever it wants with — on Vega they'd appeasr in a network owned account and be dispersed by the protocol over some period of time. At that point it would just be another piece of protocol design how the newly minted tokens are distributed. May be grants, validaotr/trader/LP incentives, ... or whatever, but that would all happen on Vega. 

This would all be probably >18 months in the future, if it happens at all, but it's important that we could support it if the community voted to do it.

non-protocol-specs/0002-token-v2.md
C0deMunk33: governance can always just abandon this altogether. so to plan for that and putting the contract out of the control of the vega team is a good idea.

I will make a better writeup of this, but I think the token and vesting need to be broken into 2 contracts.

C0deMunk33 (commented):
I need to write this up, but here's what I'm thinking:
image

since it will spend more time NOT vesting than vesting leave the token as totally standard, ship all the initially minted tokens to a lockup that "we" control that meets the spec. then at some point in the future the mint mechanism "unlocks" on the token itself.

MM0819:
> I need to write this up, but here's what I'm thinking:
> image
> 
> since it will spend more time NOT vesting than vesting leave the token as totally standard, ship all the initially minted tokens to a lockup that "we" control that meets the spec. then at some point in the future the mint mechanism "unlocks" on the token itself.

This is the same design pattern I've seen/used in the past for vesting schedules. 

I was chatting with @barnabee over the weekend about the approach for this, and we got into the detail about how staking will work. Given the plan is to allow people to delegate / stake locked tokens, how would it work if they are held by a smart contract rather than an address controlled by the owner?

barnabee:
@C0deMunk33 @lewisb11 I have pushed some updates taking into account the above comments

non-protocol-specs/0002-token-v2.md
barnabee: Added to section 6

non-protocol-specs/0002-token-v2.md
barnabee: Added this

non-protocol-specs/0002-token-v2.md
barnabee: Clarified this below

tamlyn10:
@C0deMunk33 to review ...

C0deMunk33:
Thanks for the poke @tamlyn10 
So far everything looks good, I have no further questions until I hit that spot in the code. I am assuming the spec is correct and coding against it until I hit snags (which is how many of the Barney comments come to be).

This is my top priority at the moment and I am heads down on it.

C0deMunk33:
So after some trial and error I have come to the conclusion that the balances and tranches and vesting are so entangled that the idea of keeping balances separate in an asset pool is just not feasible as the logic that is most likely to have a problem and need upgrading is the very logic that runs the whole thing.

There's no way to upgrade already vesting tranches that doesn't also break the "no clawback" so if we need new features they will only apply to new tranches in a new contract, you just have to forward some of the tokens the v2 vesting contract.

barnabee:
> So after some trial and error I have come to the conclusion that the balances and tranches and vesting are so entangled that the idea of keeping balances separate in an asset pool is just not feasible as the logic that is most likely to have a problem and need upgrading is the very logic that runs the whole thing.

Cool, so vesting contract both holds the tokens and has the logic that releases them?

> There's no way to upgrade already vesting tranches that doesn't also break the "no clawback" so if we need new features they will only apply to new tranches in a new contract, you just have to forward some of the tokens the v2 vesting contract.

That seems fine.

Any view on whether this spec is mergeable?

edd:
Does any of this change anything about [Simple Staking](https://github.com/vegaprotocol/specs/issues/492) ?

barnabee:
@edd I don't think so unless @C0deMunk33 wants to say this part of the spec needs changing https://github.com/vegaprotocol/specs/pull/486/files#diff-51dc38a0d88e67e732cdeaaca4c391724c82571badb7f210b038cece8310ef11R72-R83

C0deMunk33:
https://github.com/vegaprotocol/Vega_Token_V2

First draft. Starting on Tests.

C0deMunk33:
> @edd I don't think so unless @C0deMunk33 wants to say this part of the spec needs changing https://github.com/vegaprotocol/specs/pull/486/files#diff-51dc38a0d88e67e732cdeaaca4c391724c82571badb7f210b038cece8310ef11R72-R83

So there's this:
image

which manages depositing and removing stake. I'm attempting to get @Vegaklaus to agree to 1) no slashing and 2) mid-epoch removing of stake just causes the award from that period to be lost.
if we can do those 2 things, staking is permanently out of control of MultisigControl. See screenshot above.

edd:
Seems to me like this is ready to merge @barnabee ?

edd:
Given @C0deMunk33 has implemented most of this already over in https://github.com/vegaprotocol/Vega_Token_V2, I’m going to merge this. If https://github.com/vegaprotocol/specs/pull/486#issuecomment-790952859 still needs addressing, it’s a new pr.

PR #487, opened 2021-02-23
Add Dynamic Validator Spec

### LINKED ISSUE #485: Tendermint interface (weighting) Validator CRUD
Short version: we need a spec for the blockchain commands required for validators to join and leave the network. @gchaincl's summary is below:

1. **Tendermint Interface (weighting)** 

Needs spec :page_facing_up: 

@gchaincl

* Create tendermint Validator api calls (AddValidator, RemoveValidator, UpdateValidator (changes stake)) (Note: not transactions, this is the linkage between us and tendermint)
* Handle Add New Validator
  * Weigth redistribution
* Handle Remove New Validator
  * Handle no validators left
* Handle Update New Validator
  * Remove + Add ?
* Integrate testing with Tendermint's Validator API

### LINKED ISSUE #575: Dynamic Validator Spec
Add Dynamic Validator Spec to come in a milestone after Oregon Trail

qustavo:
Note: Using `0000` prefix. Will use the right one when PR is approved

Closes #485

### DIFF
--- a/specs/0000-dynamic-validators.md
+++ b/specs/0000-dynamic-validators.md
@@ -1,14 +1,13 @@
-Dynamic Validator Set
-
+Feature name: Dynamic Validator Set
 
 # Acceptance Criteria
 
-• A new validator joins the set with a given power.
-• A validator leaves the set.
-• Each time the validator set changes, the power is redistributed based on the
+* A new validator joins the set with a given power.
+* A validator leaves the set.
+* Each time the validator set changes, the power is redistributed based on the
 validator the validators stake proportion.
-• Validator set should be reflected by the [Tendermint Validators API](http://localhost:26657/validators).
-• Validator set should be stored in the persistent storage so that it can be restored upon a node recovery.
+* Validator set should be reflected by the [Tendermint Validators API](http://localhost:26657/validators).
+* Validator set should be stored in the persistent storage so that it can be restored upon a node recovery.
 
 # Summary
 
@@ -63,7 +62,7 @@ type Set interface {
    Remove(id Pubkey)
 
    // Iterate calls fn for each validator on the Set.
-   Iterate(fn func(id pubkey, power Decimal))
+   Iterate(fn func(id Pubkey, power Decimal))
 }
 
 // Store is a generic storage interface. It's designed to integrate well with
--- a/specs/0000-dynamic-validators.md
+++ b/specs/0000-dynamic-validators.md
@@ -1,0 +1,84 @@
+Dynamic Validator Set
+
+
+# Acceptance Criteria
+
+• A new validator joins the set with a given power.
+• A validator leaves the set.
+• Each time the validator set changes, the power is redistributed based on the
+validator the validators stake proportion.
+• Validator set should be reflected by the [Tendermint Validators API](http://localhost:26657/validators).
+• Validator set should be stored in the persistent storage so that it can be restored upon a node recovery.
+
+# Summary
+
+The current network considers a fix number of validators and signers (see
+[Multisig Control]()?), meaning that the initial validator set (defined in the
+Genesis Block) is immutable. However, since there is an economic incentive for
+being a validator, new participants will join and old participants will leave.
+
+# Guide-Level Explanation
+
+As stated in the [summary](#summary) the Genesis block contains an initial set
+of validators which (may) vary over time. Since these changes are part of the
+consensus, and they change the `AppState`, they are ultimately triggered by a
+`DeliverTx` Transaction. The effect of such transaction might not happen
+immediately, but recorded and executed after a new [`epoch`](TODO: More on this)
+happens.
+
+## Adding a new Validator
+
+The process starts with an event from the Bridge which atests that a the
+contract (TODO: Add Reference) has been executed. Once this reaches Vega
+network the event is included in a block (via `DeliverTx`).
+
+## Removing a Validator
+
+In this case the process starts with a Vega transaction sent by the validator
+who wants to leave the network and unstake its tokens.
+
+# Reference-Level Explanation
+The application may set the validator set during `InitChain`, and update it
+during
+[`EndBlock`](https://docs.tendermint.com/master/spec/abci/abci.html#endblock)
+method. Updates to the Tendermint validator set can be made by returning
+`ValidatorUpdate` objects in the [`ResponseEndBlock`](https://docs.tendermint.com/master/spec/abci/apps.html#validator-updates).
+
+The application object will contain a `validator.Set` object which handles the validator set. The object is constructed by the Application constructor.
+Adding and Removing a validator should invoke `Add` and `Remove` methods in the `processor.Processor` module.
+To update the set `Iterate` will be executed by the `ABCI` application inside the `EndBlock` method to construct a new validator Set.
+
+# Pseudo-code / Examples
+```go
+package validator
+
+// Set 
+type Set interface {
+   // Add adds a new validator to the set and recomputes the remaining
+   // validators power.
+   Add(id Pubkey, stake Decimal)
+
+   // Remove removes a validator from the set and recomputes the remaining
+   // validators power.
+   Remove(id Pubkey)
+
+   // Iterate calls fn for each validator on the Set.
+   Iterate(fn func(id pubkey, power Decimal))
+}
+
+// Store is a generic storage interface. It's designed to integrate well with
+// the current storage interfaces defined by the persistance layer.
+type Store interface {
+   Set(key []byte, value []byte)
+   Get(key []byte) []byte)
+}
+
+// NewSet returns a Set given a store using MaxTotalVotingPower defined by Tendermint.
+func NewSet(s Store) Set {
+   return NewSetWithMaxVotingPower(s, abci.MaxTotalVotingPower)
+}
+
+// NewSetWithMaxVotingPower manages validators distributing power based on max.
+func NewSetWithMaxVotingPower(s Store, max uint64) Set {
+}
+```
--- a/specs/0000-dynamic-validators.md
+++ b/specs/0000-dynamic-validators.md
@@ -4,14 +4,13 @@ Feature name: Dynamic Validator Set
 
 * A new validator joins the set with a given power.
 * A validator leaves the set.
-* Each time the validator set changes, the power is redistributed based on the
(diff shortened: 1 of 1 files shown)


edd:
Reference to card that this spec covers: https://github.com/orgs/vegaprotocol/projects/65#card-54637947

edd:
@Vegaklaus @barnabee I feel like this spec is in a place where it needs you to decide more detail about how this’ll work. I think @gchaincl has covered the how on the implementation detail but I don’t think these feel correct:

https://github.com/vegaprotocol/specs/blame/4c8ba1bb915896700f144b07c087e7796dcfc11e/specs/0000-dynamic-validators.md#L29-L33

This doesn’t feel right to me. While the paper mentions that an ETH based event could be a backup, I think this primarily needs to happen on Vega to meet the outline below:

8C24AB8D-E0EF-473B-A41A-7ADEC677F887

Some *somehow* someone signals their intent to be a validator. While I think that process *could* be outside this spec, I am working on the assumption that there’s probably a new Vega TX that they can send that says ‘I want to be a validator’ and then at epoch end the set is recalculated as covered elsewhere in this spec. 

https://github.com/vegaprotocol/specs/blame/4c8ba1bb915896700f144b07c087e7796dcfc11e/specs/0000-dynamic-validators.md#L36-L37

This doesn’t cover the case where a validator is removed for being malicious, or if they just shut down and disappeared entirely. Someone else will need to be able to run the TX. This is covered here somewhat:

99BB829B-E2EA-4495-A76B-0A1F70D1EFF5

Handy link to research paper:
https://github.com/vegaprotocol/research-internal/blob/master/validator_rewards/ValPol.pdf

TLDR: @Vegaklaus / @barnabee please could you decide how you want these two sections to work, then we’ll work out how to finish this spec from there

edd:
@Vegaklaus @barnabee 
2988D291-FFFF-43B7-A964-750E48FA81B1

This spec also raises some questions about Epochs. I think we can start here, but it might become its own spec.

- @barnabee how long should epochs be to start with?
- @barnabee what unit is an epoch measured in?
- @barnabee Is the epoc length changeable through governance or set at genesis or hard coded?
- @Vegaklaus presumably we hold some sort of buffer of potential changes that is then unwound at epoch start?
- @Vegaklaus / @barnabee if epoch end changes the validator set, which requires an update to the multisig contract, how do we progress with the epoch if that update has to be run manually and has not been? 
  - If it’s run automatically, who pays?
    - And how much gas do they pay?
      - Where does that gas money come from?
        - What if the network is really congested and that transaction is not mined for the duration of the epoch?
  - @Vegaklaus / @barnabee how does this change if we also have a Bitcoin bridge
  - @Vegaklaus / @barnabee  how does this change if we also have a bridge for another Ethereum token standard?

Relevant brainstorming card: https://github.com/orgs/vegaprotocol/projects/65#card-54637996

qustavo:
Thanks @edd for the clarification, I naively assumed that the moment _someone_ staked certain amount, it immediately was signaling its intend to become a validator.
What's not clear to me is that, to proclaim my intent to become a validator _someone_ needs a _sufficient stake_, while that stake might be sufficient at the moment of proclamation, it might not be enough at the moment of epoch completion. Wouldn't it be sufficient to satisfy the stake requirement upon epoch completion?

qustavo:
@edd @Vegaklaus @barnabee about Epoch
My understanding is that we have block producers and blockchain (Ethereum) Witnesses being ran by the same entities: Validators, but these are different roles that could be decoupled. Therefore, while Epoch can be used to synchronize Pegs-In and Pegs-Out, the validator set can be updated immediately  (i.e: _emergency_defund_).

Vegaklaus:
Not sure I fully understand what you mean, and why the decoupling of the two roles in necessary; can you elaborate a bit more ?  In my understanding the validator set could be updated at any time already, but won't to make the life of the clients easier - with the exception of the emergency defund, which creates a little mess as the price for a fast reaction; am I missing something here on block-production ?

qustavo:
After having a call, @Vegaklaus clarified my questions.
Epochs are Tendermint specific.
Even though witnessing Ethereum and produce Tendermint blocks are different roles it makes sense that they are run by the same node.

barnabee:
Are we sure epochs are Tendermint specific @Vegaklaus? I thought they were used to control the speed of changes to the validator set coming from changes in staked/delegated amounts in the Vega protocol, and thus like all the staking/delegation transactions and rewards, would be agnostic to the consensus layer… or have I got that wrong?

Vegaklaus:
@barnabee I think there's a different interpretation of what 'specific' means. They're not Tentermint specific in the sense that we need to open up the protocol or have a hard time implementing them on another blockchain if necessary. However,  it easiest to measure time in Vega/Tendermint blocks (as opposed to real time or Ethereum blocks). The consensus has to play a role to synchronise things (i.e., if the validator weight changes, all validators need to adapt the new setting at the same block to not confuse the protocol), so the easiest implementation is 'One epoch is 1000 vega (tendermint) blocks. This also makes it easier for light clients to follow, they then just need to check every 1000th block to keep updated on who the validators are (assuming all necessary information is put into the first block of a new epoch), rather than having to search around for change information.

barnabee:
Hmm I see, does that mean you can hold up a validator change if block production is slower than expected? And does that matter? I guess we might need to make sure to test the performance extremes of core pretty heavily if relying on block production taking a certain amount of time...

Vegaklaus:
Yes, this makes the changes dependent on the block time; the reasoning was that the exact (real) time doesn't matter that much, and what we keep constant instead is the number of blocks (and thus the revenue generated) per epoch.
Now we could use an external trigger (Ethereum, time), and build a synchronization mechanism on the vega chain (i.e., every validator who sees the trigger puts it on the chain, and the block containing the x-th trigger message , n/3<x<n, is then the start of the new epoch. While that keeps real time constant(ish), it now means that a faster chain has more blocks per epoch (and vice versa), so I'm not sure that is any better

barnabee:
Time wise there's a timestamp in the block, so I assumed we'd use first block with `timestamp > epoch_end_time`, but I'm not sure how that compares to waiting _n_ blocks… whatever you think is best

qustavo:
@barnabee I misused the terms here. Whatever method we use (block height or block time), epochs reference the Tendermint blocks (or any other consensus mechanism that  Vega uses).
Regarding block timestamp or height, my understanding is that we could use absolute or relative block time and/or number of blocks since a given block. To avoid slow block production would using a `what-ever-happens-first` approach makes sense @Vegaklaus ?

edd:
Given that the above conversation appears to have come to a conclusion, @barnabee & @Vegaklaus please could you answer all of https://github.com/vegaprotocol/specs/pull/487#issuecomment-784451497 questions and then we can get them in to the spec

qustavo:
@edd here are my 2cents on you questions, I'd also like to hear from @barnabee and @Vegaklaus 

> how long should epochs be to start with?

once a day approx, although this might be changed through governance.

> what unit is an epoch measured in?

As mentioned above it can be block height or block time, whatever we choose it will always be approximated

> Is the epoc length changeable through governance or set at genesis or hard coded?

Should be initially declared in the genesis and then changeable through governance 

> presumably we hold some sort of buffer of potential changes that is then unwound at epoch start?

Yes

> if epoch end changes the validator set, which requires an update to the multisig contract, how do we progress with the epoch if that update has to be run manually and has not been?

This is a really question, and I don't know. I'd think that the onchain fees should be pays by the validators, money is taken out of a special fund for fees, updated based on the price gas (on ethereum) and the average fee from the last N block (on bitcoin)

Vegaklaus:
> * @barnabee how long should epochs be to start with?
Roughly 24 hours sounds like a good long term value. Maybe with some shorter value right after genesis to allow things to fall into place faster.
> * @barnabee what unit is an epoch measured in?
Block height and block time both works. While I initially had block height in mind, I tend more to block time now, but if someone sees a good reason for block height that works too.
> * @barnabee Is the epoc length changeable through governance or set at genesis or hard coded?
I'd say definitively changeable.
> * @Vegaklaus presumably we hold some sort of buffer of potential changes that is then unwound at epoch start?
All events that can trigger a change should be on the chain somewhere. Should be easier to keep them in a buffer rathern than going through the chain though, so yes (if I understand the question right).
> * @Vegaklaus / @barnabee if epoch end changes the validator set, which requires an update to the multisig contract, how do we progress with the epoch if that update has to be run manually and has not been?

The epoch should be unaffected; in the worst case, we have different thresholds for the multisig and vega internally for a while, which most of the time shouldn't be an issue. 
One risk (which would primarily come from too long epochs) is if we need to change the validator set too drastically, in which case the whole communication with the multisig becomes messy. 
>   
>   * If it’s run automatically, who pays?
This is still a vbit under discussion, my latest status is (please check @C0deMunk33)
 - if a new validator joins, they pay to join.
 - on joining, a validator pays a deposit that is (likely) big enough to pay for them leaving, and then gets the rest back once they left

>     
>     * And how much gas do they pay?
>       
>       * Where does that gas money come from?
>         
>         * What if the network is really congested and that transaction is not mined for the duration of the epoch?
Assuming validator changes are not massive, we can tolerate quite some time of minor inconsistencies between the vega weights and the smart contract ones (they're somewhat different as it is anyhow). There's two special scenarios I see
 - If a validator is voted out 'in anger' (caught cheating, run by the Lizard people, ...), we want them out fast. This could
    be done by then using their entire deposit to pay a much higher gas price than necessary
- If we get a massive change (say, all NATO states outlaw validators, and suddenly a lot of them need to stop operating
  at once) there's a bigger issue anyhow. As for gas, that means though the we have a lot of deposit to draw from to pay the gas price, so we probably can afford to bid rather high.
>   * @Vegaklaus / @barnabee / @C0deMunk33 how does this change if we also have a Bitcoin bridge
>   * @Vegaklaus / @barnabee / @C0deMunk33 how does this change if we also have a bridge for another Ethereum token standard?
I'd say it's getting more expensive, but shouldn't change radically.
> 
> Relevant brainstorming card: https://github.com/orgs/vegaprotocol/projects/65#card-54637996

tamlyn10:
@Vegaklaus and @gchaincl to complete and get it ticketing ready...

davidsiska-vega:
@edd assigning to you

gordsport:
Closing as duplicate #766

PR #497, opened 2021-03-04
Add rule limit to price monitoring spec

edd:
Slack ref: https://vegaprotocol.slack.com/archives/C8WS7M50E/p1614876167205400?thread_ts=1614873827.199500&channel=C8WS7M50E&message_ts=1614876167.205400

This limit does not currently exist, so it will be a dev ticket to add the limit.


davidsiska-vega:
@davidsiska-vega pull your finger out...

PR #498, opened 2021-03-05
Updated market lifecycle diagram

campbellssource:
Updated To:
0029-market-lifecycle-flow-diagram

Previous:
Screenshot 2021-03-05 at 10 48 19

Also changed file number to be the same as spec.


tamlyn10:
@campbellssource thanks for this! Btw you can also get to "Cancelled" state from a governance vote to close the market.

tamlyn10:
Also, the "maturity event" isn't quite the trigger for the "Trading Terminated".. the table says:

> Defined by the product (i.e. from a product parameter, specified in market definition, giving close date/time)

edd:
> @campbellssource thanks for this! Btw you can also get to "Cancelled" state from a governance vote to close the market.

You *should* be able to. But I don't think this is implemented. Which is an oversight:

https://github.com/vegaprotocol/vega/blob/8fb63848073fac5c7bbd31ab2af8eedcb1b636b8/gateway/graphql/schema.graphql#L1967-L1976

https://github.com/vegaprotocol/vega/blob/8fb63848073fac5c7bbd31ab2af8eedcb1b636b8/proto/governance.proto#L88

campbellssource:
Guilty of raising the PR a little early. Will make it a draft and revise

campbellssource:
To much to rebase, just going to create a fresh branch

PR #500, opened 2021-03-09
Mid price is to be used when calculating volume for LP spec order

davidsiska-vega:
This is a spec clarification following the discussion with @tamlyn10 and @witgaw yesterday about LP orders implying silly volumes. See issue https://github.com/vegaprotocol/vega/issues/3113 in core.


witgaw (approved):
LGTM

barnabee (approved):

PR #503, opened 2021-03-11
Add spec for Validator Tendermint Performance Metrics (#491)

### LINKED ISSUE #491: Validators: Simple performance metrics hooks
> 2. **Performance measurement hooks
> We need to be able to measure per validator if a validator is useful or not.

# @Vegaklaus’ outline
 My first thought would be
- A function that devies on who contributed to a block derived from chain data (if the signatures certifying a block are in the order of arrival of the corresponding messages, that could be added)
- A validator keeping statistical data on the order and time of tendermint-internal messages arriving and for now logging that data
- point-to-point communication times between validators for reference

# Scope
> Performance measurement to collect data for future phases

See #475. Scope set by @barnabee [here](https://github.com/orgs/vegaprotocol/projects/65#card-55923190)

ashleyvega:
This PR adds a spec for Validator Tendermint Performance Metrics

Closes #491.

### DIFF
--- a/specs/0000-validator-tendermint-metrics.md
+++ b/specs/0000-validator-tendermint-metrics.md
@@ -8,13 +8,24 @@ Start date: 2021-03-11
 - [ ] For each validator, a list of vote arrival times is available
 - [ ] After a network has been active for some minimum amount of time (say 100 blocks), statistics for vote time (mean, standard deviation, percentiles) are available for each validator
 - [ ] By looking at the above statistics, it is clear which validators (if any) are at the end of high-latency network connections.
+- [ ] If possible, also provide statistics on the point-to-point connection speeds between validators (e.g., measuring the time
+      between sending a message and receiving an ack, if that exists.
+
+Listing all voters is preferable if the API allows that reliably
+The protocol halso as a pre-vote which is interesting (ig the API gives us that data, too)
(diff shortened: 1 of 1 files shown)


ashleyvega:
@Vegaklaus @barnabee Please take a quick look at this draft PR.

(It's my first spec, so be gentle.)

davidsiska-vega:
@barnabee @Vegaklaus could you please leave some comments on this draft pr... cheers!

ashleyvega:
And @gchaincl  - please review too, since you're Mr Tendermint :smile:.

specs/0000-validator-tendermint-metrics.md
```diff
@@ -0,0 +1,92 @@
+Feature name: Validator Tendermint Performance Metrics
+Start date: 2021-03-11
+
+# Acceptance Criteria
```
qustavo, line 4: Is this planned to be implemented as an API endpoint of the same node, or it should be an external process that gather the information and crunch the numbers?

ashleyvega: I hope that tendermint does/can provide enough info via API that an external process can gather this info. If not, we could add/improve tendermint API endpoints.

gordsport:
Closing as duplicate of #762

PR #517, opened 2021-03-31
WIP: On-Chain Treasury - trading rewards

### LINKED ISSUE #502: On-chain fund
Write spec for on-chain fund, ideally with some test case scenarios

https://github.com/vegaprotocol/research-internal/issues/197

### LINKED ISSUE #504: Trading rewards to link to on-chain fund
As per task described here:

https://github.com/vegaprotocol/research-internal/issues/197

tamlyn10:
Closes #504


davidsiska-vega:
@davidsiska-vega after it's ready.

tamlyn10:
Meeting booked for Friday 14th May morning

protocol/0000-on-chain-treasury-trading.md
```diff
@@ -0,0 +1,53 @@
...
+A participant's:
+1. total trading activity on the network
+1. contribution to bootstrapping trading activity in new markets (separate to liquidity provision)
+1. trading activity when markets are in protective auctions
+
+[Are these useful?] A participant's:
+1. use of special orders to create trading (e.g. pegged orders)
+1. contribution to increasing or decreasing open interest
+1. supplying of the orders that get filled for liquidations
+1. good behaviour - never used or "donated to" insurance pool
```
EVODelavega: Using the insurance pool is bad behaviour? What about loss socialisation? It's not the fault of the trader that their MTM win had to come from the insurance pool, but rather the fact that the traders holding an opposite position couldn't cover the MTM loss

jeremyletang: Well I think that meant used the insurance pool from the side of the party in fault, in order to pay the good behaving party darling.

protocol/0000-rewards.md
```diff
@@ -0,0 +1,97 @@
...
+            else:
+                Proportions[acc] = 0
+            return Proportions
+		}
+    return [eligible_accounts, proportions]
+  }
+}
+```
+
+## Rewards
```
ValentinTrinque: Discussed during spec review:

Rename Reward to Reward Scheme here, to dedicate the term Reward to the result of the Reward Function.

protocol/0000-reward-functions.md
```diff
@@ -0,0 +1,72 @@
...
+_Reward Pot:_ Fixed total pot amount (N), all distributed at the reward distribution point
+
+_Reward Calculation:_ between the reward commencement time and the reward distribution time, any party who has made a deleegation at least once will receive N/(total count of eligible participants) tokens. Everyone receives the same amount of tokens.
+
+_Acceptance Criteria:_
+
+- [ ] a party that locks money on the bridge but does not delegate it, will not receive any reward
+- [ ] a party that delegates and then undelegates during the period is considered an eligible participant
+- [ ] all eligible participants receive the same size reward
+- [ ] the reward pot is empty at the conclusion of the reward distribution
```
edd: If we are running both 5.1 and 5.2, they both empty the 'pot' (for clarity: a pot is a single treasury account for each asset)(see above comment about the definition of a pot that may solve this)

protocol/0000-on-chain-treasury.md
```diff
@@ -0,0 +1,65 @@
+# Network Treasury
+
+The Network Treasury is a set of accounts (up to 1 per asset supported by the network via ther asset framework) that are funded by parties, deposits, or by direct transfers (e.g. a portion of fees, or from insurance pools at market closure). 
+The funds in the network treasury are spent either by direct governance action (transfer) or by mechanisms controlled by governance, such as a periodic transfer into a reward pot.
```
edd: We need a concise, reasonable definition of a 'pot'. From reading the initial two lines here, a pot is a single reward account on each asset. However looking at 5.1 and 5.2, it suggests that each reward empties the pot - which would mean that those two cannot run simultaneously, which doesn't seem right?

edd (changes requested):

protocol/0000-on-chain-treasury.md
edd: If it's the same as a reward pool below, the language needs to be standardised

protocol/0000-rewards-overview.md
```diff
@@ -0,0 +1,87 @@
+# Reward framework
+
+Vega core will initially specify a number of Reward Functions. In the future there will be an interace for code / bytecode / WASM to implement arbitrary reward function. 
+
+Rewards are deployed using the Reward Functions. Individual Reward Pools are automatically created when an asset is transferred to the Reward.
```
edd: There is one treasury account per asset. Is this also specifying that each reward will have an account asset?

protocol/0000-reward-functions.md
tamlyn10: My understanding was there would be 1 x pot per reward scheme which is different to the top level treasury account which is 1 per asset.

protocol/0000-rewards-overview.md
tamlyn10: @edd could you take a look at the figure I've included with this PR and see if you agree with what I've mapped out in terms of 1 account per reward? If so, I can add more clarity about that reward account like when it gets created, deleted etc.

barnabee:
Merging post review with @davidsiska-vega @edd

PR #518, opened 2021-03-31
Update life cycle diagram

### LINKED ISSUE #558: Update market lifecycle flow diagram (0029)
## Issue
In order to support the writing of the MVP for settlements and expiry in the context of the Terra integration we need the market [lifecycle flow diagram](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0029-market-lifecycle-flow-diagram.svg) to be updated and referenced in the [market lifecycle spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0043-market-lifecycle.md).

## Possible Tasks?
- [ ] Update the [lifecycle flow diagram](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0029-market-lifecycle-flow-diagram.svg)
- [ ] reference in the [market lifecycle spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0043-market-lifecycle.md).

campbellssource:
0043-market-lifecycle-flow-diagram

New diagram (above) to replace existing one (below)

Untitled

closes: https://github.com/vegaprotocol/specs-internal/issues/558


edd (approved):

tamlyn10:
@campbellssource thanks for this. I think we're nearly there.

**Transitions from Pending State:** 

**Cancelled**
A market becomes Cancelled when a Market Proposal is successful and conditions are not met to transition the Market to the Active state during the Pending period, and one of the following apply:

1. the market reaches a timeout for the length of time in the Pending state; or
2. the Instrument hits expiry; or
3. governance votes (by approving a proposal to close the market) not to create the market after all.

To represent this, should we rather do 3 arrows coming off the pending state:

1. "successful governance proposal to close market?" ---->yes---> "cancelled"
2. "timeout for the length of time in the Pending state?" ---->yes---> "cancelled"
3. "auction end. Are conditions met?" (as you have)

I realise you'd done the first one as a separate action, but I personally found that confusing.

campbellssource:
@tamlyn10 what about this:
Screenshot 2021-04-09 at 13 11 37
I've changed auction to time out.

I'd like to keep the external actions yellow boxes: I think as a group they all work together. If I got rid of the yellow box on "propose to cancel" I probably need to change all the yellow boxes.

does that may you thing any different or should I give it another crack?

tamlyn10:
@campbellssource - let me mull this a bit - any chance could I have access to a copy of the file and have a play around with it?

campbellssource:
The original is in the branch and I've just DM'd it to you

candida-d:
@tamlyn10, @campbellssource - how's this coming along?

edd:
@tamlyn10 still needs to update this

campbellssource:
@tamlyn10 I've had another go at the diagram based on your feedback. What do you think?

0043-market-lifecycle-flow-diagram

tamlyn10:
I love this @campbellssource ! I find it really easy to read. Under Product trigger, could we rather put "Product.TradingTerminated" or something like that? I'd like to avoid the concept of maturity (this is not defined on the product).

campbellssource:
0043-market-lifecycle-flow-diagram

@tamlyn10 I went with "Product trigger" without an example. 
Rationale: already have "Monitoring trigger" which doesn't provide an example.

tamlyn10:
Perfect ❤️  I absolutely love this.

PR #522, opened 2021-03-31
Coded features review: LP reward distribution

witgaw:
Placeholder PR, work not started yet.


witgaw:
@davidsiska-vega I believe it's a good start of the reward split testing with a health chunk of fees and total stake covered. We may wish to extend these later on, but I think it's enough for now until we have the first implementation of the running test.

davidsiska-vega (approved):

PR #524, opened 2021-03-31
Coded features review: total stake

witgaw:
Total stake checks that the total amount of LP commitments of individual parties is taken into account correctly.

Covered by vegaprotocol/specs-internal#528


davidsiska-vega:
Will be covered by vegaprotocol/vega#4811.

PR #526, opened 2021-04-03
Add LP spec clarification

witgaw:
This doesn't add anything new to the spec, just clarifies a few points. I realise it pollutes the spec to some extent and increases dependence between specs (e.g. now if we were to change the spec governing the closeout behaviour we'd have to remember to change the note added to the LP spec), but in my opinion the benefits (clarifying behaviours from the intersection of the specs) outweigh the potential problems, especially when clarifications like these are clearly labelled as "Note:" subsections etc. - happy to be proven wrong though, let me know what you think.


davidsiska-vega (approved):
Sounds reasonable.

barnabee (approved):

PR #533, opened 2021-05-05
Add minimum probability of trading parameter.

### LINKED ISSUE #499: Push LP special order volume inside range that has reasonable prob of trading
Any part of the shape of LP special order that has very small (let's say < 10e-7) probability of trading leads to horrendous margins.

Moreover it may happen that it will lead to numerical instability. 

Add a network parameter : minimum_LP_order_prob_of_trading with default of 10e-7 and push all shapes that are at distance further out inside the range like we do with price monitoring. 

I.e. ask quant model: given mid and min prob of trading, what is the lowest and highest price. Convert this to number of ticks up and number of ticks down. Push all the shape that's outside this range onto the last tick. 

Which spec should this go into? 

Do we also want to say that we'll have different \tau for probability of trading, while we're doing this?

@barnabee FYI

davidsiska-vega:
Closes #499. 

@jeremyletang FYI

### DIFF
--- a/protocol/0038-liquidity-provision-order-type.md
+++ b/protocol/0038-liquidity-provision-order-type.md
@@ -158,7 +158,7 @@ Note that any other orders that the LP has on the book (limit orders, other pegg
 
 ## Network Parameters:
 * mm-time-horizon: market making time horizon to imply probability of trading.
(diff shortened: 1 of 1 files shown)


jeremyletang (approved):

jeremyletang:
work in progress for this is there: https://github.com/vegaprotocol/vega/pull/3457

PR #535, opened 2021-05-07
API Node Spec

### LINKED ISSUE #534: API Node Spec/Design
Add a document to describe how we want to migrate the current API structure into a new executable and what we would like to improve for stage 2.

peterbarrow:
This new spec covers the changes we want to make in taking the API functionality out of the core node and implementing it in a stand alone executable in a new repository. We split the work into 2 stages, one to move the current functionality across and a second to improve and add more features to the API node.

### DIFF
--- a/non-protocol-specs/0003-api-node.md
+++ b/non-protocol-specs/0003-api-node.md
@@ -24,7 +24,10 @@ A DataNode is a stand alone executable that connects to the event-bus stream of 
 
 The DataNode will be developed in two stages, the first stage requires us to move the existing API code into a separate executable so that it can be built, run and tested independently of the (non-)validator nodes. The second stage is taking each data type and optimising how we create, handle and store the information. As well as adding features such as restarting the node and allowing other DataNodes to retrieve a snapshot of state so that they can start up at any point in the lifetime of a Vega network.
 
-The DataNode will consume the events sent through the event-bus stream. Messages on this stream may arrive in in non id order. It is the responsibility of the DataNode to process the messages in the correct order and to verify that all the messages have arrived and none have gone missing.
+The DataNode will consume the events sent through the event-bus stream. Messages on this stream may arrive in non id order. It is the responsibility of the DataNode to process the messages in the correct order and to verify that all the messages have arrived and none have gone missing.
+
+### Blockchain statistics
+Clients are currently able to request blockchain information via a gRPC API. As the DataNode will not have direct access to the blockchain it will need to subscribe to the gRPC stream of a Validator/Non-Validator node and then relay that information into interested clients. While the blockchain stats are only available via gRPC, the DataNode will be able to distribute the data in all three supported formats (gRPC, REST and GraphQL)
 
 ### Stage 1
 A new executable will be developed in it’s own repository that connects to a newly started stream node and requests all event bus messages. It will pass those messages onto the appropriate handlers and they will build up any data structures required to represent the information. The node will allow incoming requests from clients as well as subscriptions for setting up persistent connects that updates are sent down. This node will continue to use the Badger database to store its data as it did prior to this change. The only requests that will be handled by the (non-)validator nodes will be requests for event streams.
--- a/non-protocol-specs/0003-api-node.md
+++ b/non-protocol-specs/0003-api-node.md
@@ -8,7 +8,7 @@ An API node is a read only service that connects to a stream node (non-validator
 
 ## Guide-level explanation
 
-An API node is a stand alone executable that connects to the eventbus stream of a stream node (also called a non-validator node). It constructs any data objects from the incoming data required to provide information to the client such as user positions and market depth. It allows users to make one off requests for data such as information about a market or a particular order. It also allows subscriptions to be started in which the API node sends update messages to the client as events occur in the core. The API node is started at the same time as the stream node so that all events can be captured. The steam node has a small buffer of events that it keeps to allow the API node to query older information should it require it to build up the full local dataset required to be consistent with the core.
+An API node is a stand alone executable that connects to the eventbus stream of a stream node (also called a non-validator node). It constructs any data objects from the incoming data required to provide information to the client such as user positions and market depth. It allows users to make one off requests for data such as information about a market or a particular order. It also allows subscriptions to be started in which the API node sends update messages to the client as events occur in the core. The API node is started at the same time as the stream node so that all events can be captured. This is required for stage 1 so that the API node sees all the event messgaes and can have a complete picture of the validator node state. The steam node has a small buffer of events (several blocks/seconds) that it keeps to allow the API node to connect after the stream node comes up so that the API dataset is consistent with the core.
 
 ## Reference-level explanation
 
@@ -23,7 +23,7 @@ The API node may consist of one or more executables that subscribe to some or al
 
 ## Acceptance Criteria
 ### Stage 1
-* The API node must be a separate executable in it’s own source code repository
+* The API node must be a separate executable in it's own source code repository
 * No API related code must be left in the Core Node
 * The API node must be started within a few seconds of a newly started stream node
 * The API node must be able to handle brief network outages and disconnects
@@ -31,6 +31,7 @@ The API node may consist of one or more executables that subscribe to some or al
 * All information available from the API node must be retrievable via all of the 3 connections types (gRPC, GraphQL and REST)
 * The validator node will only accept requests for event bus subscriptions. All other API requests will be invalid.
 * The event bus stream is only available from the stream node and not the validator or API node
+* All information that is emitted from the stream node is processed by the API node (no data is lost)
 
 ### Stage 2 (for further discussion)
 * The API Node will be split into different services to allow data types to be handled more optimally using better suited tools and technologies.
--- a/non-protocol-specs/0003-api-node.md
+++ b/non-protocol-specs/0003-api-node.md
@@ -1,0 +1,41 @@
+# Feature name: API Node
+## Start date: 2021-05-07
+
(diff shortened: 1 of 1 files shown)


non-protocol-specs/0003-api-node.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Summary
+
+An API node is a read only service that connects to a stream node (non-validator node) and gives users the ability to query the state of the network.The user can request information from the API nodes using GraphQL, gRPC and REST. The user can query the API node for one off snapshots of data or request a subscription of live data.
+
+
+## Guide-level explanation
+
+An API node is a stand alone executable that connects to the eventbus stream of a stream node (also called a non-validator node). It constructs any data objects from the incoming data required to provide information to the client such as user positions and market depth. It allows users to make one off requests for data such as information about a market or a particular order. It also allows subscriptions to be started in which the API node sends update messages to the client as events occur in the core. The API node is started at the same time as the stream node so that all events can be captured. The steam node has a small buffer of events that it keeps to allow the API node to query older information should it require it to build up the full local dataset required to be consistent with the core.
```
cityhunteur: > The API node is started at the same time as the stream node so that all events can be captured

I think we need to clarify this one; I thought this was not a hard requirement; and the API nodes will source the events as they come online.

peterbarrow: Updated, does that make more sense now?

cityhunteur: It does 👍

non-protocol-specs/0003-api-node.md
```diff
@@ -0,0 +1,42 @@
+# Feature name: API Node
+## Start date: 2021-05-07
+
+## Summary
+
+An API node is a read only service that connects to a stream node (non-validator node) and gives users the ability to query the state of the network.The user can request information from the API nodes using GraphQL, gRPC and REST. The user can query the API node for one off snapshots of data or request a subscription of live data.
```
edd:
```suggestion
An API node is a read only service that connects to a stream node (non-validator node) and gives users the ability to query the state of the network. The user can request information from the API nodes using GraphQL, gRPC and REST. The user can query the API node for one off snapshots of data or request a subscription of live data.
```

non-protocol-specs/0003-api-node.md
```diff
@@ -0,0 +1,42 @@
...
+### Stage 1
+* The API node must be a separate executable in it's own source code repository
+* No API related code must be left in the Core Node
+* The API node must be started within a few seconds of a newly started stream node
+* The API node must be able to handle brief network outages and disconnects
+* All API functionality currently in the Core node must be available from the API node
+* All information available from the API node must be retrievable via all of the 3 connections types (gRPC, GraphQL and REST)
+* The validator node will only accept requests for event bus subscriptions. All other API requests will be invalid.
+* The event bus stream is only available from the stream node and not the validator or API node
+* All information that is emitted from the stream node is processed by the API node (no data is lost)
```
edd: **question (blocking)**: This sounds good, but I’d like some more detail on how error handling will work:

- If a stream node stops emitting events forever, what can we do? 
- ~If a stream node is still processing blocks but not emitting events to the DataNode (connectivity, I guess) does it buffer events for the DataNode? If so how long for?~ Answered in stage 2

I assume the answer for stage one is ‘You just stop getting events forever’, which is probably roughly equivalent to now, which is OK. But let’s make it explicit

non-protocol-specs/0003-api-node.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Stage 2 (for further discussion)
```
edd, line 56: **question (non-blocking)**:  Will the DataNode need block-level information that the Event Bus doesn’t currently have?  Off the top of my head the first two clients for this DataNode would be a Block Explorer and the Console needing more historical data. The Block explorer probably wants to get block information (which is in Tendermint, but not coming via the event bus). Is that out of scope?

edd (changes requested):
Some clarification in the acceptance criteria for stage 1, I think. Otherwise it sounds good.

non-protocol-specs/0003-api-node.md
```diff
@@ -0,0 +1,42 @@
...
+The API node may consist of one or more executables that subscribe to some or all of the event stream messages from a stream node. Each executable could handle one set of data (such as orders or market depth) and use the best technology available to store and server up that data. For example the order API node could store all the historic orders in a Postgres database so that the users can query back to the start of the Vega market. Or the market depth API node could store things only in memory as no historic data is required, just the most up to date snapshot of the market. There could be a middleware layer between the stream node and API nodes that caches the event bus data and allows the API nodes to query older events to allow the nodes to be stopped temporarily or for new API nodes to be started to handle increased loads. A simple template would be made available for users to build their own API nodes so that they could handle the information in a bespoke way suited to the rest of their trading system.
+
+
+## Acceptance Criteria
+### Stage 1
+* The API node must be a separate executable in it's own source code repository
+* No API related code must be left in the Core Node
```
edd, line 44: **question (blocking)**: Can we define this more? We’ve split our nodes in to three:

- Validator nodes. **GRPC? (see below),** no REST, no GraphQL.
- Stream nodes: GRPC event bus only.
- DataNodes (the thing this spec is about): GRPC, REST, GraphQL

Can anyone subscribe to the stream node’s event bus? I assume yes, and I assume that it would be up to the node operator to restrict this if they wanted.

Also presumably validators need some way for people to submit TXs? Or are we poking TXs directly in to Tendermint now? If so, which node does this? If it’s not validators that might be confusing. Of course if it is validators it might be confusing too. Either way, clarification required.

edd:
I’ve left a review asking for clarity in a few places. A question that correctly *isn’t* addressed is when should this split out in to a separate repo from the core, given that [vegaprotocol/vega](https://github.com/vegaprotocol/vega) is currently a private repo and this code will be useful to other people. 

My votes is for that being a stage 2 requirement, as per this spec. The reason for that is that the code in the API layers isn’t going to be changed significantly in stage 1, while to adapt in the process of implementing stage 2 will require  more sweeping changes. Let’s use stage 1 to get familiar with the API code and decide what says and what will need to change, and then plan how that migrates over to a repo while planning stage 2 in more depth.

non-protocol-specs/0003-api-node.md
jeremyletang: Ultimately the validator and stream nodes are the same, they just have a difference in configuration which make them part of consensus or not, but they shall exposes same APis.

So now knowing they are the same thing here's the APIs they would make available:
- submit transaction (as it is now pretty much, with the SubmitTx call, no prepare exists anymore).
- some informational things (e.g: healthchecks, statistics)
- The event bus.

jeremyletang:
> I’ve left a review asking for clarity in a few places. A question that correctly _isn’t_ addressed is when should this split out in to a separate repo from the core, given that [vegaprotocol/vega](https://github.com/vegaprotocol/vega) is currently a private repo and this code will be useful to other people.

I think the intent of phase one, is to completely decorelate the core to the APIs, which  the goal is to get it out of the core repo once that is done and is the finally achievement of part 1, Part to is basically possible because the APIs do not exists in the core, and can be built without any constraints from it runing with the core.

Also I'd really approach part 1 with the deliverable being this needs to be fully in another repo, just so we can make clear that we have no more dependencies between core and APIs, and don't get any suprise, or any non expected behaviour later on, when we finally get it out, and realise there was issues in the design or anything, to me it should be worked out of the core from the beginning.

non-protocol-specs/0003-api-node.md
edd: It would be good to clarify these definitions at the top of the file then:

- A *validator node* only exposes the event bus and a submitTx endpoint. All REST, GraphQL and other GRPC endpoints will be removed in stage 1 of this spec.
- A *stream node* is functionally identical to a validator node, but does not participate in the consensus of the underlying blockchain. A stream node can still be used to submit transactions.
- A *data node* is a new node type that subscribes to the event bus of a validator or stream node, and provides the REST, GraphQL and GRPC endpoints that were formerly built in to the core.

jeremyletang: Yes, that sounds like a good addition to it.

edd: Also worth noting that it’s plausible that in future a validator node will get its own GRPC commands that relate only to validator nodes. Which is fine as we are keeping the stream & submitTX endpoints in there directly, so it’s not like anyone is removing it entirely. I’m just being pedantic about the acceptance criteria that ‘No API related code…’ is in here. I guess the AC should be more like:

- The gateway code (GraphQL &  REST) is no longer in the core binary (i.e. validators and stream nodes **cannot** serve REST & GraphQL)
- The current GRPC endpoints should be reduced to a minimal set. All other endpoints should no longer exist in the core binary. The only endpoints should be:
  - Health check
  - Event bus
  - SubmitTransaction

non-protocol-specs/0003-api-node.md
peterbarrow: Good thoughts. We will be able to tell that the streaming node has stopped sending us events as we expect a block time update every second or so. We can monitor that and shutdown with an error if we don't see the events for a given period of time.

non-protocol-specs/0003-api-node.md
peterbarrow: I would say tendermint specific events are not part of this scope. We can look in stage 2 about the best way to do this.

jeremyletang:
@vegaprotocol/core to review so it can be merged soon.

edd (approved):

PR #539, opened 2021-05-18
Add delegation/rewards spec

### LINKED ISSUE #492: Simple staking
**3. Implement simple staking**

* You will still own the tokens that you have delegated Delegated tokens are still valid for voting
* This changes over [https://github.com/vegaprotocol/specs/issues/485](epochs) - pre-defined, multi block periods. It is not clear if Tendermint already supports this

**Note** Putting up stake to become a validator is the first transaction in the process of becoming a validator. Stakes will be evaluated at the end of the epoch and the set will change a result.

# Scope
> * Stake + Delegate txns (do not need to affect TM validator weight but are used for [reward calcs](#493))
> * Lock staked/delegated tokens so they can’t be withdrawn, transferred, or used as margin (i.e. put them in a staked/delegated funds account) until unstaked/undelegated (by staking or delegating zero)
> * Lock staked/delegated tokens can be used for governance voting
> * Calculation and distribution of validator rewards based on staked plus delegated  (#493)
> * NB: much of the rules around when/how you can stake/delegate, max amounts, etc. could be deferred to a future phase I think as they only effect rewards at this point

See #475. Scope set by @barnabee [here](https://github.com/orgs/vegaprotocol/projects/65#card-55923190)

### LINKED ISSUE #493: Simple rewards payout (for validators & delegators)
**3. Payout for validators & delegators**

> This is based on the amount of stake that the validators have

# Scope
> * Calculation and distribution of validator rewards based on staked plus delegated (could be a simpler tactical calc than the final curve if needed or it’s not defined)

See #475. Scope set by @barnabee [here](https://github.com/orgs/vegaprotocol/projects/65#card-55923190)

EVODelavega:


protocol/0053-simple-staking-and-delegating.md
```diff
@@ -0,0 +1,100 @@
...
+
+Once Vega is aware of locked tokens, the users will have an account with the balance reflecting how many tokens were locked. At this point, the user can submit a transaction to stake (delegate) their tokens. The amount they stake must be `<= balance`, naturally. Users can remove stake by submitting an `Undelegate` transaction. The tokens will then be restored back to their token balance. How users can delegate/undelegate is discussed below.
+
+### Mechanics of locking/unlocking stake
+
+At the top level, `Stake_Deposited` simply adds `amount` of tokens to the account of the user associated with the `user`. Likewise, the `Stake_Removed` event subtracts the `amount` of tokens from their account.
+
+- If the `Stake_Removed` amount of tokens is higher than the balance of said user, something went seriously wrong somewhere. This is a good time to panic.
+- If the amount is higher than the amound of undelegated stake, the missing amount must be freed using the undelegate function. There is currently no rule how to choose this; a first heuristic would be to take from the highest delegation first and then go down, e.g.
```
edd, line 29:
```suggestion
- If the amount is higher than the amount of undelegated stake, the missing amount must be freed using the undelegate function. There is currently no rule how to choose this; a first heuristic would be to take from the highest delegation first and then go down, e.g.
```

protocol/0053-simple-staking-and-delegating.md
```diff
@@ -0,0 +1,100 @@
...
+			Amount: reward,
+			Asset:  market.Asset,
+		},
+		MinAmount: reward,
+	})
+}
+
+```
+
+The transfer type informs the collateral engine that the `FromAccount` ought to be the infrastructure fee account, and the `ToAccount` is the general account for the party for the given asset. The delegator can then withdraw the amount much like they would any other asset/balance.
```
edd, line 100: # Epochs
## Acceptance Criteria
- There is a new network parameter that controls epoch length
- An epoch change event is emitted on the event bus every time an epoch change occurs
- As a delegator, at the start of an epoch, if I…
   - …delegate stake…
   - …undelegate stake…
     - …that change is  not true until the epoch changes
       - …then it is actioned 
-  As a delegator, if I delegate stake and immediately undelegate stake, at the next epoch, there is no stake delegated

## Delegator fee payouts
- As a delegator, at the end of an epoch, I earn appropriate fees
..


I suspect there are more but here’s a start

edd (approved):
Looks good. One thing that isn’t clear, but is true across specs, is exactly what is being implemented now and what isn’t. I *think* it’s just the `undelegateNow` and `undelegateInAStrop`. It’s not a blocker, but maybe worth … a mention in acceptance criteria?

protocol/0053-simple-staking-and-delegating.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Payment calculation
+
+At the end of an epoch, payments are calculated. This is done per active validator:
+
+* First, `score_val(stake_val)` calculates the relative weight of the validator given the stake it represents.
+* For each delegator that delegated to that validator, `score_del` is computed: `score_del(stake_del, stake_val)` where `stake_del` is the stake of that delegator, delegated to the validator, and `stake_val` is the stake that validator represents.
+* The fraction of the total available reward a validator gets is then `score_val(stake_val) / total_score` where `total_score` is the sum of all scores achieved by the validators. The fraction a delegator gets is calculated accordingly.
+* Finally, the total reward for a validator is computed, and their delegator fee subtracted and divided among the delegators
+
```
tamlyn10, line 67: who sets the delegator fee?

protocol/0053-simple-staking-and-delegating.md
```diff
@@ -0,0 +1,100 @@
...
+	Validator Val = 2;
+}
+```
+
+Where `Delegate` adds the `Amount` to the delegation of validator `Val` at the biginning of the next epoch (if still available to them), and `Undelegate` subtracts the amount from the delegation of `Val` by the next epoch if available.
+
+_optional (not needed now, but later)_
+
+`UndelegateNow`:
+The user is marked to not receive any reward from the validator in that epoch. The reward should instead go into the insurance pool. The stake is marked as free for the delegator, but is not yet removed from the validator stake (this happens at the end of the epoch).
```
tamlyn10, line 53: Let's put in the treasury for that asset

protocol/0053-simple-staking-and-delegating.md
tamlyn10: network parameter..

gordsport: @tamlyn10 to add in who sets the delegator fee when merging this into your branch

tamlyn10:
Closed as this is incorporated into this PR: https://github.com/vegaprotocol/specs-internal/pull/582

PR #543, opened 2021-05-25
rename directory releases to milestones

gordsport:
rename directory releases to milestones to fit with the new workflow


pennyandrews (approved):
Looks good to me. I may need to re-make my changes to the 3-oregon-trail.md file changing the wording in there but that's cool

PR #550, opened 2021-06-01
Create new_issues_to_protocol_design_project

gordsport:
A Github action to move any new issues raised against this repo to the Protocol Design project board to have a single clear view of work required on protocol design.

### DIFF
--- a/.github/workflows/new_issues_to_protocol_design_project
+++ b/.github/workflows/new_issues_to_protocol_design_project
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Add new issues to org Protocol Design project
     steps:
-      - uses: alex-page/github-project-automation-plus@v0.7.1
+      - uses: alex-page/github-project-automation-plus@50502d399cbb98cefe7ce1f99f93f78c6756562e
         with:
           project: Protocol Design
           column: To do
--- a/.github/workflows/new_issues_to_protocol_design_project
+++ b/.github/workflows/new_issues_to_protocol_design_project
@@ -1,0 +1,16 @@
+name: Auto Assign New Issues to Project(s)
+
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  assign_one_project:
+    runs-on: ubuntu-latest
+    name: Add new issues to org Protocol Design project
+    steps:
+      - uses: alex-page/github-project-automation-plus@v0.7.1
+        with:
+          project: Protocol Design
+          column: To do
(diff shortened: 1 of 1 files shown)


.github/workflows/new_issues_to_protocol_design_project
```diff
@@ -0,0 +1,16 @@
...
+on:
+  issues:
+    types: [opened]
+
+jobs:
+  assign_one_project:
+    runs-on: ubuntu-latest
+    name: Add new issues to org Protocol Design project
+    steps:
+      - uses: alex-page/github-project-automation-plus@v0.7.1
```
fkondej: Let's follow good practices for [Using third-party actions](https://docs.github.com/en/actions/learn-github-actions/security-hardening-for-github-actions#using-third-party-actions), and `Pin actions to a full length commit SHA`.
Currently the latest version is [v0.7.1](https://github.com/alex-page/github-project-automation-plus/releases/tag/v0.7.1) which corresponds to `50502d399cbb98cefe7ce1f99f93f78c6756562e`.
So please change:
```suggestion
      - uses: alex-page/github-project-automation-plus@50502d399cbb98cefe7ce1f99f93f78c6756562e
```

.github/workflows/new_issues_to_protocol_design_project
```diff
@@ -0,0 +1,16 @@
...
+        with:
+          project: Protocol Design
+          column: To do
+          repo-token: ${{ secrets.GH_NEW_CARD_TO_PROJECT }}
```
fkondej: Please add an empty line at the end.

fkondej: Is this GitHub Access Token (`GH_NEW_CARD_TO_PROJECT` secret value) for your account?

fkondej (changes requested):
Minor changes and question.

pennyandrews (commented):
Looks good to me, so if Fryd is happy then all good with me.

.github/workflows/new_issues_to_protocol_design_project
gordsport: Yes this is a GitHub Access Token I created against my account - with ‘repo’ and ‘write:org’ access as tested in the repoplayground repo. For actions like these should we be using a different account for the access token?
The token was then applied as an org secret as the intention was this action would be repeated across a few repos thus could use a common token.

fkondej: Secret value got updated to use bot's access token. You can find it in 1password.

fkondej (approved):

PR #555, opened 2021-06-01
Add GitHub workflow to synchronise the private specs repo with the public one

fkondej:
Part of vegaprotocol/devops-infra#539.


fkondej:
Please don't land it yet. We still need:
- setup secret,
- do the initial sync (script can do it, but it is better to do it manually by someone)

ashleyvega (approved):
LGTM

PR #567, opened 2021-06-07
Create add_research_PR_to_board.yml action

gordsport:
Add new action to move any PRs created for the specs internal repo to the in progress column on the protocol design project board. Its a separate action to the issues in order to be able to turn off in the Github UI if it becomes too noisy on the project board. They may want to be merged to a single action at a later date.


ashleyvega (approved):
LGTM

PR #570, opened 2021-06-08
Staking bridge

edd:
@C0deMunk33 please


gordsport:
@jeremyletang as discussed Ive removed the R2D2 from the spec as we will not need the event queue at this time - added you as reviewer.

campbellssource:
Question (might not strictly be a staking bridge question): Does the Vega network keep separate accounts for vega tokens that came via the Vesting contract Vs the Staking bridge?

Background:
Given users can stake unlocked as well as locked Vega tokens,
I assume that un-staking returns tokens to their origin (Vesting OR ETH wallet),

Also: if my tokens have run their vesting period while staked I can’t just sent them back to an ETH wallet

campbellssource:
The above question we answered by @barnabee  
> No separate accounts. The accounts for staked tokens (from either Ethereum contract) afre really just keeping score. They are a “view” onto the [total] staked amount on Ethereum, as unlike the ERC20 bridge the Vega network cannot transfer these tokens, lock/hold onto them, use them for margin, etc. and they can therefore only be used for things like determining the number of votes someone has in governance actions and counting up the total weight of tokens delegated to each validator. The protocol treats tokens staken from either place in the same way and the fact there are two separate contracts and functions is an implementation detail because of how we implemented the vesting contract and staking.

and in responce to:

> I assume that un-staking returns tokens to their origin,
And that if my tokens have run their vesting period while staked I can’t just sent them back to an ETH wallet

he said

> Yes, this is true, but unstaking is done entirely on Ethereum, and as they are staked in two different places with two different contracts, it’s simply the case that they can’t be unstaked anywhere they shouldn’t be. The account on Vega and the Vega network has no involvement in the unstaking process (so this is TOTAlLY different to withdrawals) and is told when tokens are unstaked when it sees the events on the Ethereum network.

campbellssource:
Question: If I have Staked some Vega tokens Via the Staking Bridge, Are the tokens locked on Ethereum that prevents me sending them to another wallet?

I assume the answer is that they are locked and can not be sent but am doubting myself so thought it is worth adding an Acceptance criteria.

barnabee:
> Question: If I have Staked some Vega tokens Via the Staking Bridge, Are the tokens locked on Ethereum that prevents me sending them to another wallet?
> 
> I assume the answer is that they are locked and can not be sent but am doubting myself so thought it is worth adding an Acceptance criteria.

Definitely requires acceptance criteria. I think they are in fact transferable while staked (though I’m not particularly sure why), but would like @C0deMunk33 to confirm

non-protocol-specs/0004-staking-bridge.md
```diff
@@ -0,0 +1,35 @@
...
+* `Transfer_Stake(uint256 amount, address new_address, bytes32 vega_public_key)` -Transfers staked Vega from the sender to the target address
+
+### Other Implementations
+[ERC20_Vesting.sol](https://github.com/vegaprotocol/Vega_Token_V2/blob/main/contracts/ERC20_Vesting.sol) emits both `Stake_Deposited` and `Stake_Removed` events.
+
+* `stake_tokens(uint256 amount, bytes32 vega_public_key)` emits `Stake_Deposited`
+* `remove_stake(uint256 amount, bytes32 vega_public_key)` emits `Stake_Removed`
+
+## Accounts
+Staked assets will appear in a user's [staking account](../protocol/0013-accounts.md). As the staked tokens will be used for [governance](../protocol/0028-governance.md) in the first mainnet (aka Sweetwater), governance will need to be updated to check for staked balances as well as general account balances.
```
barnabee: Where is the specification for how the Vega network processes the events emitted on Ethereum to adjust amounts recorded in staking accounts? We should link that here (or expand this section if it does not exist).

non-protocol-specs/0004-staking-bridge.md
```diff
@@ -0,0 +1,35 @@
...
+The staking bridge contracts live in [vegaprotocol/staking_bridge](https://github.com/vegaprotocol/Staking_Bridge)
+
+### IStake.sol
+IStake.sol contains the events necessary to track all Ethereum-side transactions using the CQRS design pattern.
+Any contract that wants to be recognized by Vega as valid for staking, it must emit the specified events.
+
+These events include:
+* `event Stake_Deposited(address indexed user, uint256 amount, bytes32 indexed vega_public_key)` - This event is emitted when a stake deposit is made and should be credited to provided vega_public_key
+* `event Stake_Removed(address indexed user, uint256 amount, bytes32 indexed vega_public_key)` - This event is emitted when a user removes stake and should trigger targeted vega_public_key to be decremented
+* `event Stake_Transferred(address indexed from, uint256 amount, address indexed to, bytes32 indexed vega_public_key)` - This event is emitted when a user transfers the given amount of stake between the `from` and `to` wallets. This allows a user to change owing ETH wallet without removing potentially delegated stake
```
barnabee: Is this called manually or automatically when tokens are transferred in the base contract? If the latter we need to spec the changes to the base ERC20 contract that do this. If manually, can users transfer tokens to someone else using the ERC20 while staked, and if so what happens to the staked tokens?

non-protocol-specs/0004-staking-bridge.md
```diff
@@ -0,0 +1,35 @@
...
+
+In order to manage the staking of Vega tokens from mainnet Ethereum to Vega mainnet, events need to be raised on ETH that can then be consumed by Vega.
+Much like [the Ethereum bridge](../protocol/0031-ethereum-bridge-spec.md), a bridge smart contract will be used.
+Unlike the ERC20 or ETH bridges however, the Staking Bridge does not rely on any sort of multisignature control and deposits/withdrawals are entirely at the discretion of the depositor.
+
+
+## Solidity contracts
+The staking bridge contracts live in [vegaprotocol/staking_bridge](https://github.com/vegaprotocol/Staking_Bridge)
+
+### IStake.sol
```
barnabee: This "spec" is mostly a description of some smart contract functionality not a spec for the required behaviour. It doesn't desacribe what should be possible and what is not, what validation is performed, and what should trigger the emission of the events. If @edd and @vega-paul are happy then I think we can leave it for now but seems like it will need much more rework to be a useful spec.

barnabee (changes requested):
Some comments in here for small changes and also a general one that this spec does not, to me, seem to cover the required behaviour and effectively leaves the contracts themselves as the source of truth, which makes me nervous. If we are OK with that I would like @edd and QA (Paul) to explicitly say so.

barnabee:
@C0deMunk33 @edd @davidsiska-vega please review 💥

non-protocol-specs/0004-erc20-governance-token-staking.md
```diff
@@ -0,0 +1,83 @@
...
+In order to manage the staking of Vega tokens from mainnet Ethereum to Vega mainnet, events need to be raised on ETH that can then be consumed by Vega.
+Much like [the Ethereum bridge](../protocol/0031-ethereum-bridge-spec.md), a bridge smart contract will be used.
+Unlike the ERC20 or ETH bridges however, the Staking Bridge does not rely on any sort of multisignature control and deposits/withdrawals are entirely at the discretion of the depositor.
+
+
+## Components of staking
+
+### Vega network
+
+Staked assets will appear in a user's [staking account](../protocol/0013-accounts.md) once the Vega network sees the relevant `Stake_Deposited` event(s) with enough confirmations (defined by a network parameter). As the staked tokens will be used for [governance](../protocol/0028-governance.md), the governance will weight votes based on the staking account balance instead of the general account balance. Delegation functionality will also use the staking account balance as the source of truth for the maximum number of delegated tokens.
```
davidsiska-vega, line 19: "defined by a network parameter" which network parameter? Let's either declare a new one here or make it clear that it's governed by the one used for ERC20 deposits. In my opinion it can be the same as the ERC20 deposit one because it's determining the probability that an Ethereum block is not going to be orphaned.

davidsiska-vega:
I think it may be good to have some stories / scenarios for QA on this. I wouldn't want this to stop the merge but if it is merged let's create an issue so that they get added later.

non-protocol-specs/0004-erc20-governance-token-staking.md
```diff
@@ -0,0 +1,83 @@
+# ERC20 Governance Token Staking
+
+Vega makes uses of a ERC20 token on the Ethereum blockchain as a governance token for delegation to validators and creation and voting of governance proposals. A user's governance tokens must first be recognised against a Vega public key before they can be used on the Vega network for governance and delegation.
+
+Although it would be possible to use the standard ERC20 bridge to deposit governance tokens and put themn in full control of the Vega network, the system will not do this for governance tokens. Instead there will be a separate system that allows governance tokens (only) to be "staked" to a Vega public key (and "unstaked" when done) without any action on the Vega network, and without putting the tokens under the control of the Vega network. This approach has been chosen for two primary reasons:
```
edd:
```suggestion
Although it would be possible to use the standard ERC20 bridge to deposit governance tokens and put them in full control of the Vega network, the system will not do this for governance tokens. Instead there will be a separate system that allows governance tokens (only) to be "staked" to a Vega public key (and "unstaked" when done) without any action on the Vega network, and without putting the tokens under the control of the Vega network. This approach has been chosen for two primary reasons:
```

Typo fix: Otherwise, really like the clarity of this

non-protocol-specs/0004-erc20-governance-token-staking.md
```diff
@@ -0,0 +1,83 @@
...
+
+1. Any attacker who gains control of or is able to exploit the Vega network will be unable to steal staked Vega tokens. This means that even if an attacker was able to take over the network, the tokenholders would remian unchnaged and could fix the issue and relaunch the network by delegating to new validators.
```
edd:
```suggestion
1. Any attacker who gains control of or is able to exploit the Vega network will be unable to steal staked Vega tokens. This means that even if an attacker was able to take over the network, the tokenholders would remain unchanged and could fix the issue and relaunch the network by delegating to new validators.
```

Typo fix

protocol/0013-accounts.md
```diff
@@ -9,9 +9,15 @@ Feature name: accounts
...
--  [ ] When a party withdraws collateral onto Vega, the asset account for that asset will decrease in balance by the same amount. 
+  -  [ ] Only one general asset account exists per party per asset.
+  -  [ ] When a party deposits collateral onto Vega, the asset account will increase in balance by the same amount. 
+  -  [ ] When a party withdraws collateral onto Vega, the asset account for that asset will decrease in balance by the same amount. 
+
+## Trader staking accounts
+- [ ] Every party that deposits staked asset on Vega will have a stake account created for that asset.
+  - [ ] Only one staked asset account exists per party per asset.
+  - [ ] The balance can only be delegated to Validators
+  - [ ] The balance cannot be traded, or used as margin, or transferred, or withdrawn
```
edd, line 22: **thought (blocking)**: Needs an acceptance criteria that dictates what happens to account balances when a user delegates stake. I realise this isn't the delegation PR, so this complaint can be moved to that PR or dealt with here

protocol/0013-accounts.md
```diff
@@ -9,9 +9,15 @@ Feature name: accounts
...
```
edd, line 20:
```suggestion
  - [ ] The balance cannot be traded, or used as margin, or transferred, or withdrawn
  - [ ] Delegated stake remains in the trader's staking account
```

(refers to my other comment: https://github.com/vegaprotocol/specs-internal/pull/570#discussion_r666028935)

protocol/0013-accounts.md
```diff
@@ -9,9 +9,15 @@ Feature name: accounts
...
 ## Trader asset accounts
 - [ ] Every party that deposits an asset on Vega will have an asset account created for that asset.
--  [ ] Only one asset account exists per party per asset.
--  [ ] When a party deposits collateral onto Vega, the asset account will increase in balance by the same amount. 
--  [ ] When a party withdraws collateral onto Vega, the asset account for that asset will decrease in balance by the same amount. 
+  -  [ ] Only one general asset account exists per party per asset.
+  -  [ ] When a party deposits collateral onto Vega, the asset account will increase in balance by the same amount. 
+  -  [ ] When a party withdraws collateral onto Vega, the asset account for that asset will decrease in balance by the same amount. 
+
+## Trader staking accounts
```
davidsiska-vega: Should we change "Trader" to "Party"?

protocol/0028-governance.md
```diff
@@ -59,7 +59,7 @@ In future, governance weighting for some proposal types will be based on alterna
...
 
-Initially the weighting will be based on the amount of the configured governance asset that the user has on the network as determined by their general account balance of this asset. 1 token represents 1 vote (0.0001 tokens represents 0.0001 votes, etc.). A user with a balance of 0 cannot vote or submit a proposal of that type, and ideally this would be enforced in a check _before_ scheduling the voting transaction in a block.
+Initially the weighting will be based on the amount of the configured governance asset that the user has on the network as determined by their general account balance and staking account balance of this asset. 1 token represents 1 vote (0.0001 tokens represents 0.0001 votes, etc.). A user with a balance of 0 cannot vote or submit a proposal of that type, and ideally this would be enforced in a check _before_ scheduling the voting transaction in a block.
```
davidsiska-vega: @barnabee / @edd what's the thinking here: do we 
a) want to simplify and say that from now on only assets on the staking contract count for governance or
b) add general account balance and staking account balance for governance purposes? 

I prefer the simplicity of a)... as it's also how the rewards from staking are determined.

protocol/0013-accounts.md
edd: Yes

non-protocol-specs/0004-erc20-governance-token-staking.md
barnabee: @C0deMunk33 @jeremyletang should tell us whether it's ok to use the same one or we might want to use a different one for some reason (i.e. in case we want extra security on staking or in case we may want to change how we do it for ERC20 bridge, perhaps to choose number of confs dynamically by value)

protocol/0028-governance.md
edd: ~I vote A~ ~No I don't. I like B. That means that the rewards someone gets increases their governance power, which seems like a positive. If we ignored the general balance, you would have to withdraw the assets via the ERC20 bridge then stake them via the staking bridge for them to count for governance which is gross and weird.~

Overruled below. A it is.

non-protocol-specs/0004-staking-bridge.md
barnabee: I expanded on this

protocol/0028-governance.md
barnabee: It should be (a) I think. It's simpler. Also prevents people from trading with a balance they are using for governance, which I like, and prevents them from lazily leaving their governance tokens in the less secure ERC20 bridge, which feels like a good thing.

edd:
> > Question: If I have Staked some Vega tokens Via the Staking Bridge, Are the tokens locked on Ethereum that prevents me sending them to another wallet?
> > I assume the answer is that they are locked and can not be sent but am doubting myself so thought it is worth adding an Acceptance criteria.
> 
> Definitely requires acceptance criteria. I think they are in fact transferable while staked (though I’m not particularly sure why), but would like @C0deMunk33 to confirm

When you stake using the staking bridge, the tokens leave your wallet. The token bridge *does* support a transfer function, which changes who can unstake and receive the tokens. I'll add an acceptance criteria for that.

protocol/0013-accounts.md
barnabee: Changed

qa-scenarios/0004-erc20-governance-token-staking.feature
```diff
@@ -0,0 +1,82 @@
+# Scenario: An ethereum key with a non-zero balance of vested Vega tokens deposits via the staking bridge, and receives the stake on Vega 💧
+    # a stake_deposited event is emitted by the staking bridge contract
+    # and the vega key has a staking account for the VEGA asset
+    # and the vega key's staking account has increased by the appropriate amount
+# Scenario: An ethereum key with a zero balance of vested Vega tokens deposits via the staking bridge, the transaction is rejected 💧
```
barnabee: We have to be careful with "vested" and "unvested" vs "redeemed" and "not-redeemed"

Unvested: 
- are ALWAYS not-redeemed, ALWAYS a balance for their holder in vesting contract and can ONLY be staked via the vesting contract

Vested:
- IF not-redeemed, are a balance for their holder in vesting contract and can be staked there (but can be redeemed IF not staked, then used as below)
- IF/ONCE redeemed, are a balance in their holder's ERC20 wallet and must be staked via the staking bridge (can never go back to the vesting contract)

non-protocol-specs/0004-staking-bridge.md
edd: This is now outdated - it should only check the staked balance, not the general account: https://github.com/vegaprotocol/specs-internal/pull/570/files/aa6e9917826c669dbd8137a73982c60d3e2e6805#r666082918. Updating.

barnabee (changes requested):
One comment re vesting contract states with change needed, otherwise scenarios look good and comprehensive, I added one around redemption being prevented while tokens are staked

qa-scenarios/0004-erc20-governance-token-staking.feature
edd: I'll go through and fix this now

non-protocol-specs/0004-erc20-governance-token-staking.md
edd: Moved to a new issue

PR #572, opened 2021-06-09
Add normalising factor to prob of trading

### LINKED ISSUE #571: Possible spec bug with prob of trading
At the moment in 0034-prob-weighted-liquidity-measure.ipynb we have

image

I'll explain the problem on the `x > S` case but the other case is mutatis mutandis the same. If `x_max` is close to `x` then the probability of trading will be artificially low - but that's bonkers. What we need to do is either of:
a) re-weight by the probability contained in the `x_min` to `x_max` integral i.e. divide by `\sum_{x_min \leq x' \leq x_max} f(x')`
b) ignore the x_max in the sum and replace it by infinity.

davidsiska-vega:
Closes #571 .


witgaw (approved):
LGTM

PR #574, opened 2021-06-09
Target stake and auction extension edit

### LINKED ISSUE #569: Market data API change: target stake
Market data API currently always displays the target stake based on current open interest.

This can lead to markets which are in liquidity monitoring auction, but the market data can display target stake and supplied stake such that `supplied stake` >= `target stake` which may be confusing for users.

I’d like to change it so that whilst in auction `target stake` field contains the theoretical target stake that would be attained if the current indicative volume were to be realised - that way LPs know exactly how much extra stake is needed.

Does that sound sensible?

witgaw:
Updated specs to reflect the facts that:
* when opening auction gets extended by liquidity auction the `auction extension trigger` should reflect that.
* we want market data API to return theoretical target stake during auction.

Also, updated market data API spec to reflect fields currently being returned.

Closes #569

### DIFF
--- a/protocol/0021-market-data-spec.md
+++ b/protocol/0021-market-data-spec.md
@@ -23,20 +23,34 @@ Due to supporting dynamic orders such as pegged orders the main market data fiel
 
 ## Market data
 
-### Continuous trading (order book)
+List of market data fields to be available via the API. All these values can be empty/nothing if there is insufficient relevant data.
 
-All these values can be empty/nothing if there is insufficient relevant data.
-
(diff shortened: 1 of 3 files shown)


witgaw:
As an aside (can be done in a different PR): I'm wondering if we really want to return `best static bid price` and `best bid price` (same for offer).

By construction the two fields will always be the same (well unless there aren't any pegged orders, but that will rarely happen - only if LPs choose to supply all their liquidity via manual orders) so we may wish to either drop one of them or change `best bid price` to `best non-static bid price` (both label and functionality of course).

tamlyn10 (approved):

PR #580, opened 2021-06-11
Write up small weight spec

EVODelavega:


edd:
@barnabee should this merge in with the spec work you are doing on Validators?

barnabee:
> @barnabee should this merge in with the spec work you are doing on Validators?

Not sure- I'd be happy to leave it separate tbh, I don't think it overlaps with any other Sweetwater requirements and as it talks about validator weight relating to rewards I wonder if it's non-Sweetwater (in which case I am not thinking about speccing it right now)

@Vegaklaus @EVODelavega @jeremyletang  is this a requirement for Sweetwater?

edd:
> @Vegaklaus @EVODelavega @jeremyletang is this a requirement for Sweetwater?

It is not (ref: https://github.com/orgs/vegaprotocol/projects/65#card-56795038)

Vegaklaus:
> > @Vegaklaus @EVODelavega @jeremyletang is this a requirement for Sweetwater?
> 
> It is not (ref: https://github.com/orgs/vegaprotocol/projects/65#card-56795038)

If it is what I think it is (setting Validator weights inside Tendermint), then it comes in comparatively late (definitively after sweetwater).

gordsport:
Closing as post Oregon Trail specs will cover this in more detail

PR #581, opened 2021-06-11
Fix normalisation

davidsiska-vega:
Fix the fix to normalisation.

### DIFF
--- a/protocol/0034-prob-weighted-liquidity-measure.ipynb
+++ b/protocol/0034-prob-weighted-liquidity-measure.ipynb
@@ -41,7 +41,7 @@
     "$$\n",
     "Z = \\sum_{S \\leq x' \\leq x_{max}} f(x') \\quad \\text{for $x>S$ and } \\quad Z = \\sum_{x_{min} \\leq x' \\leq S} f(x')   \\quad \\text{for $x<S$}\\,.\n",
     "$$\n",
-    "We adopt the convention that $p(x) = \\frac12$ for x between (and including) best bid and best ask. \n",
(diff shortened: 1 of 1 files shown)


protocol/0034-prob-weighted-liquidity-measure.ipynb
```diff
@@ -39,7 +39,7 @@
     "$$\n",
     "where the normalising factor $Z$ is  \n",
     "$$\n",
-    "Z = \\sum_{x_{min} \\leq x' \\leq x_{max}} f(x')\\,.\n",
+    "Z = \\sum_{S \\leq x' \\leq x_{max}} f(x') \\quad \\text{for $x>S$ and } \\quad Z = \\sum_{x_{min} \\leq x' \\leq S} f(x')   \\quad \\text{for $x<S$}\\,.\n",
     "$$\n",
     "We adopt the convention that $p(x) = \\frac12$ for x between (and including) best bid and best ask. \n",
```
witgaw:
```suggestion
    "We adopt the convention that $p(x) = \\frac12$ for x between (and including) best bid and best ask and we normalise p(x) so that 0 \\leq p(x) \\leq \\frac12 \\forall x. \n",
```

witgaw (approved):
Looks good to me, just a minor suggestion included below.

protocol/0034-prob-weighted-liquidity-measure.ipynb
edd: @witgaw as @davidsiska-vega is away, do you think we should just merge this?

witgaw: I think it can wait till he's back, the change is already implemented so this is more about keeping the specs in line with code.

PR #582, opened 2021-06-14
Staking delegation overview

tamlyn10:


protocol/0000-simple-POS-rewards.md
```diff
@@ -0,0 +1,63 @@
...
+
+
+## Network Parameters
+
+`infra-fee-hold-time` - the length of time between when a price taker infrastructure fee is incurred and when it is paid out to validators.
+
+
+## Payment of rewards
+- [Infrastructure fees](./0029-fees.md) are collected into an infrastructure fee account for the asset
+- These fees are distributed to the general accounts of the validators and delegators after `infra-fee-hold-time` in amounts calculated according to the above calculation.
```
edd: This also needs to mention that fees can come from the on chain treasury, or be less specific and link to the fees doc which needs to expand

protocol/0000-simple-staking-and-delegating.md
```diff
@@ -0,0 +1,185 @@
...
+
+### Staking for the first time
+- To lock tokens, a participant must:
+  - Have some balance of vested or unvested governance asset in an Ethereum wallet. These assets must not be locked to another smart contract (including the [Vega collateral bridge]()).
+  - Have a Vega wallet
+  - Lock the tokens on the [Vega staking bridge contract](../non-protocol-specs/0004-staking-bridge.md)
+- To delegate the locked tokens, a participant must:
+  - Have enough tokens to satisfy the network parameter: "Minimum delegateable stake" 
+  - Delegate the locked tokens to one of the eligible validators (fixed set for Alpha mainnet).
+- These accounts will be created:
```
edd, line 160: This is going to require a additional info in the accounts spec, or here. I’m not totally clear on how the accounts work for a user delegating:
* We have a new ‘staking’ account
* A user may also have a general account in the same governance token (rewards, or deposits when they have some unlocked tokens
* A user may want to withdraw unlocked tokens, which should be fine with the existing ERC20 functionality
* But I’m not sure what happens to tokens in the governance asset that did not come through the staking bridge. Are they delegatable? Or do they have to be withdrawn through the ERC20 bridge and then locked in the staking bridge?

protocol/0000-simple-staking-and-delegating.md
```diff
@@ -0,0 +1,185 @@
...
+and unlocking of stake are initiated by the [Vega staking bridge contract](../non-protocol-specs/0004-staking-bridge.md), not by Vega.
+
+In order to delegate, users require tokens that will be locked in a smart contract (see [Vega staking bridge contract](../non-protocol-specs/0004-staking-bridge.md)). Vega will be made aware of how many tokens a given party has locked through bridge events. When the same tokens are unlocked, a corresponding event will be emitted:
+
+```
+  event Stake_Deposited(address indexed user, uint256 amount, bytes32 vega_public_key);
+  event Stake_Removed(address indexed user, uint256 amount);
+```
+
+This provides the information the core needs to keep track of:
```
edd, line 37: How should this be handled over network resets/downtime?

protocol/0000-simple-POS-rewards.md
```diff
@@ -0,0 +1,63 @@
+# Validator and Staking POS Rewards
+This describes the Alpha Mainnet requirements for calculation and distribution of rewards to delegators and validators. For more information on the overall approach, please see this [research document](https://github.com/vegaprotocol/research-internal/blob/master/validator_rewards/ValPol7.pdf).
```
edd:
```suggestion
This describes the Alpha Mainnet requirements for calculation and distribution of rewards to delegators and validators. For more information on the overall approach, please see the relevant research document.
```

This will be a broken link in the public repo

non-protocol-specs/0000-technical-notes.md
```diff
@@ -0,0 +1,20 @@
+# Technical notes
```
edd: This needs to move out to a separate PR with a new spec for what this backup mode does

protocol/0000-simple-staking-and-delegating.md
edd: These are now all answered in the Staking Bridge specs (and related updates) which are now merged

protocol/0000-simple-POS-rewards.md
```diff
@@ -0,0 +1,63 @@
...
+
+- `min_val`: minimum validators we need (for now, 5)
+- `compLevel`: competitition level we want between validators (1.1)
+- `num_val`: actual number of active validators
+- `a`: The scaling factor; which will be `max(min_val, num_val/compLevel)`. So with `min_val` being 5, if we have 6 validators, `a` will be `max(5, 5.4545...)` or `5.4545...`
+
+Functions:
+
+- `score_val(stake_val)`: `sqrt(a*stake_val/3)-(sqrt(a*stake_val/3)^3)`
+- `score_del(stake_del, stake_val)`: for now, this will just return `stake_del`, but will be replaced with a more complex formula later on, which deserves independent testing.
```
davidsiska-vega, line 25: Do we mean "for now, this will just return `stake_val`"? Otherwise we have a recursive definition with no terminating state ;-).

protocol/0000-simple-POS-rewards.md
```diff
@@ -0,0 +1,63 @@
...
+- `delegator_reward(stake_val)`: `stake_val*.1`. Long term, there will be bonuses in addition to the reward.
```
davidsiska-vega: This looks incomplete to me.

protocol/0000-simple-POS-rewards.md
davidsiska-vega: @Vegaklaus - probably for you.

protocol/0000-simple-POS-rewards.md
davidsiska-vega: @Vegaklaus probably one for you again.

protocol/0000-simple-staking-and-delegating.md
```diff
@@ -0,0 +1,185 @@
...
+### Staking for the first time
+- To lock tokens, a participant must:
+  - Have some balance of vested or unvested governance asset in an Ethereum wallet. These assets must not be locked to another smart contract (including the [Vega collateral bridge]()).
+  - Have a Vega wallet
+  - Lock the tokens on the [Vega staking bridge contract](../non-protocol-specs/0004-staking-bridge.md)
+- To delegate the locked tokens, a participant must:
+  - Have enough tokens to satisfy the network parameter: "Minimum delegateable stake" 
+  - Delegate the locked tokens to one of the eligible validators (fixed set for Alpha mainnet).
+- These accounts will be created:
+  - After locking a staking account denominated in the governance token asset is created
```
campbellssource: Is there an account created for each node that I've delegated stake to?
```
VEGA
- General = 0
- Staked (not delegated) = 10
- Staked: Node A = 90
```

Or does the user see where their stake is with a query that is separate to accounts?
I guess the later will be useful for seeing delegation changes due in the next epoch. But I think the former is possible too.

protocol/0000-simple-POS-rewards.md
Vegaklaus: No, this should be correct; the score that delegators get to determine their reward is linear to the amount of stake they delegated (stake_del), at least for Sweetwater.

protocol/0000-simple-staking-and-delegating.md
campbellssource: This appears to say you do not have an account per node: https://github.com/vegaprotocol/specs-internal/blob/f97ec31babf3d41b17dc8fe8d1077f25d255ae1e/protocol/0013-accounts.md#party-staking-accounts

Leaving this comment here as I think it is good to add a link to `/0013-accounts.md#party-staking-accounts` around here:

protocol/0000-simple-POS-rewards.md
davidsiska-vega: You're right, I was reading `score_del` instead of `stake_del`. My bad.

protocol/0000-simple-POS-rewards.md
barnabee: Yes agree: 
* `.1` here looks like maybe the delegator reward fraction net param?
* this needs to be defined in terms of the fact that the network has a set of *infrastructure fee accounts* per asset contaning some balances `B[asset]` of accrued infra fees. These reward specs together with the fees specs need to tell Vega how to distribute all the money in those accounts between validators. They need to use consistent terminology and "line up" together so they can be read together to understand how this works.
* these functions will ALSO be used to distribute the contents of other accounts for the on chain rewards system, so I imagine here we would calculate `{account, fraction_of_balance}` pairs or something like that...

Vegaklaus: As Barney said. 0.1 Is a governance parameter that snuck in as a constant. For SW, we can probably just set it to some parameter that defines the delegator reward (likely smaller than .1 ?)

barnabee: Cool the spec should reflect this and provide a name for the network parameter then.

I am not going to be able to do any more updates today.

@Vegaklaus can you take this one?

Issue #587, opened 2021-06-24
Add spec containing table of market parameters and their governance category (easy / hard)


tamlyn10:
This will be done alongside finishing the network parameters spec updates, which include better detail on the governance categories for updating (currently thinking "easy / hard").

PR #588, opened 2021-06-24
updating market parameters added to market framework

tamlyn10:
Updating market parameters is a governance action already specified in the governance spec. We have not yet specified which parameters should be updateable through this method.

To do before implementation ready:
- [ ] #587


davidsiska-vega:
Closing because it's stale and will need new proper discussion.

PR #590, opened 2021-06-25
Acceptance criteria for limited network life

### LINKED ISSUE #586: Add acceptance criteria for limited network life
#584 added some new features required to be rolled over at limited network life. It's clear what they are but I want some acceptance criteria, so I am taking it upon myself to add them.

edd:
Adding clearer acceptance criteria. Closes #586


edd:
Question here for @davidsiska-vega, but as he's away @barnabee instead:
- This spec does not list Open Orders. Should it? I think No
- This spec doesn't list market status, only proposals. Should it? I think no - the market will be re-proposed and restart, but won't maintain the book or auction status vs active trading - it will start again as a newly proposed market, effectively.
- Do we need special handling for genesis markets or assets that exist at startup? I think they'll need to skip some validation. Maybe? Unsure. I feel like there's something here.

davidsiska-vega:
> Question here for @davidsiska-vega, but as he's away @barnabee instead:
> 
> * This spec does not list Open Orders. Should it? I think No
> * This spec doesn't list market status, only proposals. Should it? I think no - the market will be re-proposed and restart, but won't maintain the book or auction status vs active trading - it will start again as a newly proposed market, effectively.
> * Do we need special handling for genesis markets or assets that exist at startup? I think they'll need to skip some validation. Maybe? Unsure. I feel like there's something here.

1) Clarified. 
2) No market status or any other market info to be stored, only proposals. 
3) There will be no genesis markets and no genesis assets except perhaps Vega token for staking.

non-protocol-specs/0003-limited-network-life.md
```diff
@@ -22,80 +15,244 @@ This is especially important early on when rapid iteration is desirable, as the
...
+- All [network parameters](../protocol/0054-network-parameters.md)
+- Including those defined [below](#network-parameters).
+- All [asset definitions](../protocol/0040-asset-framework.md#asset-definition). Insurance pool balance from the markets will be summed up per asset and balance per asset stored. 
+- All market proposals ([creation](../protocol/0028-governance.md#1-create-market) and [update](../protocol/0028-governance.md#2-change-market-parameters)) that have been *accepted*. 
+- All [asset proposals](../protocol/0028-governance.md) that have been *accepted*.
 - All delegation info.
-- On chain treasury balances.
-- Balances for all parties per asset: sum of general, margin and LP bond accounts. 
+- On chain treasury balances and on-chain reward functions / parameters (for 💧 Sweetwater this is only the network params that govern [Staking and delegation](0000-reward-functions.md) ).
+- [Account balances](../protocol/0013-accounts.md) for all parties per asset: sum of general, margin and LP bond accounts. See exception below.
```
edd: @davidsiska-vega This won't work if the user has staked assets, as after restore they'd be transferred in to the general account. We probably need two new tests: 

A user has a general account in VEGA with a balance of 100
And that user has a staking account in VEGA with a balance of 100
And that user has a margin account in VEGA with a balance of 100
After the reset,
That user has a general account in VEGA with a balance of 200
And that user has a staking account in VEGA with a balance of 100

Otherwise they can transfer out staked assets.

barnabee: This wording doesn't seem to mention stake account balance so it seems like on restore it'd be zero rather than represented in the wrong place. 

My understanding form @jeremyletang is that we will not save stake account balance in our checkpoints at all but restore these by ingesting all stake deposited/removed events from Ethereum and calculating the net staked to each Vega key accordingly (we have to do this anyway as while network is down there can be more stake deposited/removed events that change the stake account balance).

One way to think of this is that the stake account balance is always the sum of `Stake_Deposited` amounts grouped by key less the sum of `Stake_Removed` amounts (as agreed by the validators) and is not driven by any Vega-originated data.

@jeremyletang please correct me if I got any of this wrong

davidsiska-vega: Once (and if) @jeremyletang confirms I'll do a small edit just to clarify that stake account balances are not to be part of checkpoint.

jeremyletang: @barnabee yes you're right, the staking full amount of assets deposited for staking purposes will be re-computed at network startup using all Stake_Removed and Stake_Deposited events from the stakings bridge(s) which have been configured. 
Also just reading the previous messages, there's no way the balances could be mixed with the general / margin account, these accounts won't even been managed, used in the same places in the core, it will be really impossible to mixes them (but maybe the message was just making that explicit and I did not understand).,

non-protocol-specs/0003-limited-network-life.md
```diff
@@ -22,80 +15,244 @@ This is especially important early on when rapid iteration is desirable, as the
...
+### 💧 Test case 3.7: Suspended markets retain status after restore
+1. A market is proposed by a party that commits a stake. 
+2. Market becomes Active as enactment date reached and vote successful
+3. Traded market falls into Suspended status by either Price monitoring or liquidity monitoring trigger, or product lifecycle trigger
+2. Checkpoint is taken 
+1. The network is shut down. 
+1. The network is restarted with the checkpoint hash from the above checkpoint in genesis. The checkpoint replay transaction is submitted and processed.
+1. Market can be seen to be still Suspended.
+
+## 💧 Test case 4: Party's Margin Account balance is put in to a General Account balance for that asset after a reset
```
edd: @jeremyletang / @barnabee Please fix this to correctly represent what happens to a user's margin account balance between reset

non-protocol-specs/0003-limited-network-life.md
edd: Edit: this one is now addressed. Just the one below to address.

non-protocol-specs/0003-limited-network-life.md
```diff
@@ -22,80 +15,244 @@ This is especially important early on when rapid iteration is desirable, as the
...
-
-3. Replay events from bridged chains from the last event id stored in the checkpoint.
+    1. The network will compare the asset coming from the restore file with the genesis assets, one by one. If there is an exact match on asset id:
+      -  either the rest of the asset definition matches exactly in which case move to next asset coming from restore file. 
+      -  or any of the part of the definition differ, in which case ignore the restore transaction. 
+    2. If the asset coming from the restore file is a new asset (asset id not matching any genesis assets) then ignore the restore transaction.
+4. Load the accepted market proposals.
+    - If the enactment date is in the past then set the enactment date to `now + net_param_min_enact` (so that opening auction can take place) and status to pending. 
+    - In case `now + net_param_min_enact >= trading_terminated` set the status to cancelled. 
+5. Replay events from bridged chains from the last event id stored in the checkpoint.
```
edd: @jeremyletang please fix this as it does not match what you said here: https://github.com/vegaprotocol/specs-internal/pull/590/files#r666242095

non-protocol-specs/0003-limited-network-life.md
edd: And this: https://github.com/vegaprotocol/specs-internal/pull/590/files#r666245647

non-protocol-specs/0003-limited-network-life.md
barnabee: Are we storing the raw events in checkpoints so we don't have to dig deep back in the chain or are we storing state derived from those events at a point in time and only recalculating state for events after the last ID stored in the checkpoint?

Or does this only apply to events on properly bridged chains and not the "staking observation" infrastructure, which might be done differently?

Happy to make changes once we can clarify how this works for both ERC20 bridge and staking.

non-protocol-specs/0003-limited-network-life.md
barnabee: this is already correct

barnabee: although irrelevant for 💧 'cos no markets and no trading => no margin accounts

edd: Fair point on the relevance. The reason I raised this was @jeremyletang’s comment here: https://github.com/vegaprotocol/specs-internal/pull/590/files#r666242095 so @jeremyletang will have to tell me if he’s happy.

edd: @barnabee has caught me up on Slack - what @jeremyletang meant is that it would be impossible for the staking account balance to be summed up the way these ones are, as the staking account balance is reconstituted from a replay, not from assets on the Vega chain

davidsiska-vega: @edd This scenario is correct and doesn't need fixing.

edd:
With the recent changes in, I think this is good do merge, but post-merge needs more acceptance criteria. So that’s what I’m going to do.

PR #598, opened 2021-07-02
Add more detail to Sweetwater milestone definition

gordsport:
The start of linking features to specs, parts of specs or in the exception work items (issues) so that it is easy/easier to understand the scope of the milestone features from the milestone definition.


edd:
@gordsport I've updated this PR now some of the specs are merged, and I've fixed some number and such. On monday I'll go through and make the scope column clearer, and hopefully be done with that for when you're back on Tuesday.

gordsport:
No longer required

PR #599, opened 2021-07-06
Add limits for Sweetwater

barnabee:
Add description of network limits scope for Sweetwater

### DIFF
--- a/non-protocol-specs/0003-limits-aka-training-wheels.md
+++ b/non-protocol-specs/0003-limits-aka-training-wheels.md
@@ -21,9 +21,12 @@ These features:
 
 ### Sweetwater
 
-For 💧 Sweetwater we require only the ability to entirely disable deposits of assets until governance (either via token holders or a quroum of validators) chooses to enable them. Ideally this would also require a code release in the first instance so that it could not be purely voted for initially. Once a certain level of code readiness is achieved it would then be a community governance (token holder vote or valdiator quorum) decision.
+For Sweetwater, we only require the ability to:
 
-It would be advantageous is assets could be proposed and added to the asset framework via governance during Sweetwater, though again this could be delivered after the initial release. This would only be useful if the save and restore of asset framework definitions was also implemtented as per [limited network life](./0003-limited-network-life.md)
+- Prevent the submission of market creation proposals until a validator initiated and agreed change (i.e. a genesis/config/code change required rather than a network parameter)
+- Prevent the submission of asset addition proposals until a validator initiated and agreed change (i.e. a genesis/config/code change required rather than a network parameter)
+- Set a date/time before which no market creation proposal will be enacted as a network parameter (note if the above submission prevention is in place the proposal must still be rejected after this date)
+- Set a date/time before which no asset addition proposal will be enacted as a network parameter (note if the above submission prevention is in place the proposal must still be rejected after this date)
 
 
 ### Oregon Tail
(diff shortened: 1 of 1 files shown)


non-protocol-specs/0003-limits-aka-training-wheels.md
```diff
@@ -18,6 +18,16 @@ These features:
 
 ## Required limits and controls
 
+
+### Sweetwater
+
+For 💧 Sweetwater we require only the ability to entirely disable deposits of assets until governance (either via token holders or a quroum of validators) chooses to enable them. Ideally this would also require a code release in the first instance so that it could not be purely voted for initially. Once a certain level of code readiness is achieved it would then be a community governance (token holder vote or valdiator quorum) decision.
```
edd:
```suggestion
For 💧 Sweetwater we require the ability to entirely disable deposits of assets. Once a certain level of code readiness is achieved it would then be a community governance (token holder vote or valdiator quorum) decision.

1. At first this will be by disabling the asset bridge.
2. Then a new network parameter will be introduced, `enable_deposits` , that will default to `false`, and the asset bridge will be re-enabled. Deposits will not work while `enable_deposits` is set to `false`.
3. When a successful governance proposal changes `enable_deposits`, deposits will work as normal
```

Removing ambiguity. Is this what you want? Thought if so: people will be able to deposit assets in the smart contract and not get them back, if this is what we do. So it might not be the right thing.

non-protocol-specs/0003-limits-aka-training-wheels.md
```diff
@@ -18,6 +18,16 @@ These features:
...
+
+It would be advantageous is assets could be proposed and added to the asset framework via governance during Sweetwater, though again this could be delivered after the initial release. This would only be useful if the save and restore of asset framework definitions was also implemtented as per [limited network life](./0003-limited-network-life.md)
```
edd:
```suggestion
It would be advantageous if assets could be proposed and added to the asset framework via governance during Sweetwater, though again this could be delivered after the initial release. This would only be useful if the save and restore of asset framework definitions was also implemtented as per [limited network life](./0003-limited-network-life.md)
```

This should be more decisive. If limited network life is required (which it is) then presumably we need this.

non-protocol-specs/0003-limits-aka-training-wheels.md
edd: Note for anyone watching: Both the original and my proposed update are incorrect. The asset bridge will still be running.

barnabee:
@edd David and I have reviewed, please merge if you are happy for Sweetwater

non-protocol-specs/0003-limits-aka-training-wheels.md
```diff
@@ -18,6 +18,19 @@ These features:
...
+
+### Sweetwater
+
+For Sweetwater, we only require the ability to:
+
+- Prevent the submission of market creation proposals until a validator initiated and agreed change (i.e. a genesis/config/code change required rather than a network parameter)
+- Prevent the submission of asset addition proposals until a validator initiated and agreed change (i.e. a genesis/config/code change required rather than a network parameter)
+- Set a date/time before which no market creation proposal will be enacted as a network parameter (note if the above submission prevention is in place the proposal must still be rejected after this date)
+- Set a date/time before which no asset addition proposal will be enacted as a network parameter (note if the above submission prevention is in place the proposal must still be rejected after this date)
+
```
edd, line 30: We also need a way to stop deposits of assets. But not withdrawals.

edd: Resolved with a note

edd (approved):

PR #607, opened 2021-07-08
Add Network Parameters section to Epochs spec

edd:
**NOTE** This initial ticket has errors. See comments below.

There are 3 things in this spec that are possibly network parameters. 2 definitely - the minimum delegatable stake, and the length of an epoch:

- I *think* the epoch length is milliseconds. It could be nanoseconds, but that seems unreasonable. It could also be a block count, as blocks are the trigger. I don’t know which it should be. 
- `max_wanted_stake[Validator]` is listed as a ‘parameter’ but seems very different to me to any other network parameter, so I suspect it isn’t a network parameter. Am I right?

When these are sorted, I’ll remove the duplication and add them to the Network Parameters spec


edd:
The initial ticket is wrong.

- Epoch length is in *seconds*
- `max_wanted_stake` is a potential future per-validator value. For now it has been entirely removed, with the closest thing being a network parameter that sets the maximum stake & delegation that any one validator can have

61BB7271-516E-4EB3-A64C-7820DC01402B

PR #614, opened 2021-07-12
LP fees to go into general account plus a few links in fees spec.

davidsiska-vega:


cpuja84 (approved):
The changes looks good for liquidity fees being credited to general account along with fess calculated updates.

PR #615, opened 2021-07-13
Add basic outlines for delegating

edd:
Adds scenario titles for scenarios related to delegation, which is to say assigning stake after it has been put in Vega. Currently draft as it only covers delegation, not undelegation. Draft as I need to:
- Switch to writing acceptance criteria for things we'll get to first
- Check with someone (@cpuja84 / @vega-paul) that these are useful as they are, or if there's a better format


cpuja84:
Hi @edd - Is this still valid ? If yes, then please could you provide more information as to what needs to be done. Thanks !
FYI - @gordsport @witgaw

davidsiska-vega:
Superseded by discussion around https://github.com/vegaprotocol/specs-internal/pull/619 and changes therein.

PR #618, opened 2021-07-14
Limited Network Life / Staking Bridge acceptance

edd:
This covers the intersection of Limited Network Life and the Staking Bridge. But mainly the staking bridge

- Add a missing checkpoint field (event ID of last processed deposit)
- Fix typos
- Introduce typos, then fix them
- Add acceptance criteria for staking bridge event queue thing


non-protocol-specs/0005-limited-network-life.md
```diff
@@ -14,11 +14,13 @@ This is especially important early on when rapid iteration is desirable, as the
 - Bugs, security breaches, or other issues during alpha could either take out the chain OR make it desirable to halt block production. It's important to consider what happens next if this occurs.
 
 # Overview
-There are really two main features:
+There are three main features:
```
davidsiska-vega, line 17: So 4 in fact?

non-protocol-specs/0005-limited-network-life.md
```diff
@@ -238,25 +241,54 @@ If for `network.checkpoint.timeElapsedBetweenCheckpoints` the value is set to `0
...
+1. And A checkpoint is taken
+1. And that party has a staking account balance of 100 VEGA.
+1. And that party has delegated 60 VEGA to validator 1
+1. And that party has delegated 30 VEGA to validator 2
+1. And that party has 10 VEGA undelegated
+1. And the network is shut down.
+1. And the network is restarted with the checkpoint hash from the above checkpoint in genesis. The checkpoint replay transaction is submitted .
+1. When the state is restored, but before the EthereumEventListener replays all events
+1. Then that party has 100 tUSD restored to their tUSD general account
+1. And that party has delegated 60 VEGA to validator 1
```
davidsiska-vega, line 268: It seems to me that we took  a checkpoint, then the party did its delegation and then the network was shutdown / restarted. 

How or why should or would the network know that the delegation happened?

protocol/0059-simple-staking-and-delegating.md
```diff
@@ -14,20 +14,17 @@ Delegation and staking are terms that may be used interchangably, since delegati
 A delegator can lock a token in the [Vega staking bridge contract](../non-protocol-specs/0004-staking-bridge.md), which is then available for
 staking. To this end, an Vega token
 (or a fraction thereof) can be
-- Unlocked: The tokenholder is free to do with the token as they
-	want, but cannot delegate it
-- Locked: The token is locked in the smart contract, and cen be used
-	inside the delegation system
+- Unlocked: The tokenholder is free to do with the token as they want, but cannot delegate it
+- Locked: The token is locked in the smart contract, and cen be used inside the delegation system
```
davidsiska-vega: cen -> can

davidsiska-vega (commented):
I've left a few comments but it looks like an improvement to me. 

One question: why are we including an empty `00005-limited-network-life.feature ` in qa-scenarios? To me most of this is not testable with godog - but maybe I am wrong here and we can do mock shutdown and restores via godog too.

non-protocol-specs/0005-limited-network-life.md
edd: If that happened between checkpoints, it would be lost as delegating doesn’t trigger a checkpoint. I think that’s acceptable though.

PR #619, opened 2021-07-14
Glossary for Staking and Governance

campbellssource:
I've created a new page for this (rather than add to the existing `distributed ledger` glossary) as I thought it was a different topic

Much more that could be added to this but I just wanted to document the outcome of a chat with @barnabee


campbellssource:
Updated:
1. ~Bond~ **Associate** VEGA tokens to a Vega key,  So that
2. You can **Nominate** a validator node, and receive staking rewards.

Nominate rather than Delegate because they don't get to vote for you.
~Bond~ Associate because it sounds more specific than "Link".

Avoiding using staking in a specific sense (e.g. a button, can be page title) so that users don't expect to be receiving staking rewards after "Staking" (they have to Nominate first). Internally we've already use "staking" to refer to both of the two steps. so I propose "Staking" is the combination of both actions. 

`Associate + Nominate = Staking`

Avoid "delegate" because it implies that the node you choose also have voting rights with your VEGA tokens, it doesn't. (although @barnabee did raise the idea that they could, if the owner doesn't vote).

> Bond VEGA tokens to participate in governance, and/or to nominate a validator.

Similar ish terminology used by [Polkadot](https://support.polkadot.network/support/solutions/articles/65000168057-how-do-i-stake-nominate-on-polkadot-).

gordsport:
@campbellssource to update the wording and detail the 'normal' flow in steps.

campbellssource:
Note: the Smart contract, that is already deployed, uses to "Deposit" and "withdraw" instead of Associate and Dissociate.  
We already ~know~ (have discussed that) we'd rather not use these terms in docs etc.

glossaries/staking-and-governance.md
```diff
@@ -0,0 +1,34 @@
+## Staking
+A general term to describe the overall process of [Association](#associate) and [Nomination](#nominate)
+
+## Associate
+This allows a user with Vega Tokens on Ethereum to associate their tokens with a vega key so that the Vega key can participate in [Governance](#Governance) and [Nominate](#Nominate) their Stake to a [validator](distributed-ledger-glossary.md#validators).
+Note: this was previously known as "Staking" also as "Bonding".
+
+## Dissociate
+To opposite of association. Tokens can be un-staked via Ethereum which does not know when an [Epoch](#Epoch) ends meaning the associated Vega key will not be due any rewards from that epoch. Users could instead [De-nominate](#De-nominate) their stake using their Vega key and when the Epoch ends they can dissociate without loosing any due rewards.
```
Vegaklaus: Users are recommended to first [De-nominate](#De-nominate) sufficient  stake using their Vega key. Thus, when the Epoch ends they can dissociate without loosing any due rewards, and don't get the network to decide which tokens to De=nominate to match the balance.

glossaries/staking-and-governance.md
```diff
@@ -0,0 +1,34 @@
...
+Tokens are returned either the Staking bridge or the Vesting contract, depending on how they were [Associated](#associate).
+
```
Vegaklaus, line 18: I wouldn't call that returned, but unlocked - the token are always in the staking bridge, just locked.

glossaries/staking-and-governance.md
```diff
@@ -0,0 +1,34 @@
...
+## Governance
+Using Vega tokens that are [associated](#associate) with a Vega key to propose changes to the network or vote for or against proposed changes.
+
+## Nominate
+The act of securing the Vega Network by staking a Vega validator node. This is done via a Vega key with [associated](#associate) Vega tokens. Users who have nominated a node with their stake will receive a proportional share of the infrastructure fees collected by that validator.
```
Vegaklaus: I'd change this to "one or several validators" to make sure it's understood that users can nominate to different validators.

glossaries/staking-and-governance.md
```diff
@@ -0,0 +1,34 @@
...
+The act of removing your staked Vega tokens from a given node at the end of an Epoch using a Vega key.
+
+## Re-nominate
+The act of removing your stake of Vega tokens from a given node at the end of an Epoch and nominating another for the next, using a Vega key.
+
+## Epoch
+A window of time, In which tokens that are staked on a are due a reward. If tokens are dissociated before an epoch they loose any reward that is due.
+
+## Self Staking 
+That act of putting Vega tokens on a Validator node that you are running, as apposed to having tokens staked on a node via [Nomination](#Nominate).
```
Vegaklaus, line 35: There currently is no practical difference between a node owning tokens and nominating itselfs and self-staking.

gordsport:
Hey @campbellssource,
With staking and governance happening in mainnet very soon - do we still need this? if so, can you give an indication of priority on this?

glossaries/staking-and-governance.md
barnabee: Also: tokens never leave the vesting contract when associated from there

When they’re [already redeemed and] in your wallet, they leave your wallet (to the staking bridge) when you associate them and are returned from the staking bridge to your wallet when you disassociate them

glossaries/staking-and-governance.md
barnabee: Yeah they are exactly the same I think, not just effectively the same

Self staking is the method by which you stake to [hopefully] run a node yourself rather than delegate to a node someone else is running. 

In future when anyone can compete to run a node self staking will be the way you tell the network and the other token holders that they can delegate to you, and I’d expect the system to eventually reject nominations to keys that have not indicated they intend to run a node by self staking.

@Vegaklaus do you agree with this?

PR #620, opened 2021-07-14
Updating fees spesc

cpuja84:
Changing the wording specs for disturbing fees from immediately to end of next block.


witgaw (approved):
LGTM

PR #624, opened 2021-07-16
Link more network parameters

edd:
WIP updating network parameters spec


edd:
- `market.liquidity.targetstake.triggering.ratio`
is not mentioned in any spec

PR #625, opened 2021-07-20
Fees to be positive and charged first before margin

cpuja84:
Fees to be positive and charged first before margin

### DIFF
--- a/protocol/0029-fees.md
+++ b/protocol/0029-fees.md
@@ -3,7 +3,7 @@
 
 Fees are incurred on every trade on Vega. 
 
-An order may cross with more than one other order, creating multiple trades. Each trade incurs a fee which is always postiive.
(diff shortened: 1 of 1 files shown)


protocol/0029-fees.md
```diff
@@ -3,7 +3,7 @@
 
 Fees are incurred on every trade on Vega. 
 
-An order may cross with more than one other order, creating multiple trades. Each trade incurs a fee.
+An order may cross with more than one other order, creating multiple trades. Each trade incurs a fee which is always postiive.
```
witgaw: Shouldn't fees be non-negative? I think they are at the moment and that seems sensible to me (so the fee of 0 is fine too).

cpuja84: Updated the changes as per suggestion

witgaw (approved):
Looks good to me

PR #628, opened 2021-07-22
Mark feature tests as mocks

witgaw:
Marked existing .feature tests as mocks and copied additional one (to be removed later) from vega so that @fkondej can progress with the CI steps. Existing tests were failing either because they are mocks or due to integration test framework refactoring on `vega` side which occurred after these got copied. 

cc @davidsiska-vega @edd  (please add any new pre-feature tests with the .feature.mock extension from now on)


witgaw (commented):
Merging

PR #639, opened 2021-07-29
adding staking delegation prefeature tests

cpuja84:
adding staking delegation prefeature tests

### DIFF
--- a/qa-scenarios/draft-staking-delegation.feature
+++ b/qa-scenarios/draft-staking-delegation.feature
@@ -1,0 +1,176 @@
+Feature: Staking and Delegation
+
+Background:
+    Given the following network parameters are set:
+      | name                                                | value |
+      | validators.epoch.length                             | 10m   |
+      | validators_minimum_number                           | 1     |
+      | epoch                                               | 20s   |
+      | minimum_delegateable_stake                          | 2000  |
+      | max_stake_per_validator                             | 300000 |
+
+    And the average block duration is "1"
+    And a period equal to "1" epochs.
+
+Scenario: A party can delegate to a validator
+Desciption: A party with a balance in the staking account can delegate to a validator
+
+    Given the parties deposit the following token amount:
+    | party      | asset | stake available | 
+    | delegator1 | VEGA  |     100000      |
+    When the party requests to lock the following tokens:
+    | party      | asset | stake deposited | 
+    | delegator1 | VEGA  |     100000      |
+    And the following tokens transfer should be executed:
+    | party      | asset | stake deposited | locked | 
+    | delegator1 | VEGA  |     100000      | 100000 | 
+    Then the party requests following delegation:
+    | from       | to         | asset | delegate amount | 
+    | delegator1 | validator1 | VEGA  |  100000         |
+    And the parties should have following token amount balances:
+    | from       | to         | asset | delegated amount | 
+    | delegator1 | validator1 | VEGA  |  100000          |
+
+Scenario:  A party can undelegate from a validator
+Description: A party can undelegate i.e. remove stake from a validator
+ 
+    Given the parties should have following token amount balances:
+    | from       | to         | asset | amount delegated | 
+    | delegator1 | validator1 | VEGA  |  100000          |
+    When the parties requests following undelegation:
+    | from       | to         | asset | undelegate amount | 
+    | delegator1 | validator1 | VEGA  |  100000           |
+    Then following tokens are unlocked:
+    | from       | to         | asset | stake removed | 
+    | validator1 | delegator1 | VEGA  |  100000       |
+    Then the following tokens transfer should be executed:
+    | from       | to         | asset | undelegated stake | locked |
+    | validator1 | delegator1 | VEGA  |     100000        | 100000 |
+    Then the parties should have following token amount balances:
+    | from       | to         | asset | delegated amount | 
+    | delegator1 | validator1 | VEGA  |       0          |
+
+Scenario: A party can delegate to multiple validators
+Desciption: A party with a balance in the staking account can delegate to multiple validator
+   
+    Given the following tokens transfer should be executed:
+    | party      | asset | stake deposited | locked | 
+    | delegator1 | VEGA  |     200000      | 200000 | 
+    When the party requests following delegation:
+    | from       | to         | asset | delegate amount | 
+    | delegator1 | validator1 | VEGA  |  100000         |
+    | delegator1 | validator2 | VEGA  |  100000         |
+    And the parties should have following token amount balances:
+    | from       | to         | asset | delegated amount | 
+    | delegator1 | validator1 | VEGA  |  100000          |
+    | delegator1 | validator2 | VEGA  |  100000          |
+    
+Scenario: A party can undelegate from multiple validators 
+Description: A party can undelegate i.e. remove stake from multiple validators
+
+    Given the parties should have following token amount balances:
+    | from       | to         | asset | delegated amount | 
+    | delegator1 | validator1 | VEGA  |  100000          |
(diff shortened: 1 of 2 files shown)


qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
+Feature: Staking and Delegation
+
+Background:
+    Given the following network parameters are set:
+      | name                                                | value |
+      | validators.epoch.length                             | 10m   |
+      | validators_minimum_number                           | 1     |
+      | epoch                                               | 20s   |
+      | minimum_delegateable_stake                          | 2000  |
+      | max_stake_per_validator                             | 300000 |
```
ze97286, line 10: this is not a thing anymore - this is calculated as a function of the total delegated amount.

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+    When the party requests to lock the following tokens:
+    | party      | asset | stake deposited | 
+    | delegator1 | VEGA  |     100000      |
+    And the following tokens transfer should be executed:
+    | party      | asset | stake deposited | locked | 
+    | delegator1 | VEGA  |     100000      | 100000 | 
+    Then the party requests following delegation:
+    | from       | to         | asset | delegate amount | 
+    | delegator1 | validator1 | VEGA  |  100000         |
+    And the parties should have following token amount balances:
```
ze97286, line 30: before this step the time must advance to the end of the epoch

ze97286: or otherwise you can add an epoch seq to the table then you can express that on epoch 1 the delegator1 still has 0 balance and on epoch 2 it has (expected) balance of 100000, then when the epoch ends you can verify that the balance for the epoch 2 is confirmed to be 100000

ze97286: there is also a subtle point about max delegation per validator, as this tries to encourage balanced delegation, suppose you have 10 validators each having a total delegation of 1000 and someone wants to delegate 10k, then the max per validator becomes roughly 20000 / 13 and it will only let you delegate 538 rather than 10k.

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+    | delegator1 | VEGA  |     400000      |
+    When the party requests following delegation:
+    | from       | to         | asset | delegate amount | 
+    | delegator1 | validator1 | VEGA  |    400000       |
+    Then the parties should have following token amount balances:
+    | from       | to         | asset | delegated amount | 
+    | delegator1 | validator1 | VEGA  |    300000        |
+    | delegator1 | validator2 | VEGA  |    100000        |
+
+Scenario: A party cannot delegate stake size such that it exceeds maximum amount of stake for a validator
```
ze97286, line 133: this is unfortunately a bit more subtle. this is not known until the time of actual delegation at the end of the epoch

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+Scenario: A party changes delegation from one validator another in the same epoch
+Desciption: A party can change delegatation from one Validator to another
+
+    Given the party requests following delegation:
+    | from       | to         | asset | delegate amount | 
+    | delegator1 | validator1 | VEGA  |    100000       |
+    # Change dleegation in the same epoch / period
+    And the network moves ahead "5" blocks
+    When the parties requests following undelegation:
+    | from       | to         | asset | undelegate amount |
```
ze97286, line 160: I don't think there should be column called undelegate amount. The balance is always in terms of delegated. If you undelegate it subtracts from the balance until it's empty.

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+    Given the party requests following delegation:
+    | from       | to         | asset | delegate amount | 
+    | delegator1 | validator1 | VEGA  |    100000       |
+    # Change dleegation in the same epoch / period
+    And the network moves ahead "5" blocks
+    When the parties requests following undelegation:
+    | from       | to         | asset | undelegate amount | 
+    | delegator1 | validator1 | VEGA  |  100000           |
+    Then following tokens are unlocked:
+    | from       | to         | asset | stake removed |
```
ze97286, line 163: this is not a thing - tokens are never locked. They just sit in some account and can be moved at any point in time, certainly delegation and undelegation don't affect the amount of tokens in the staking account.

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+    Given the parties deposit the following token amount:
+    | party      | asset | stake available | 
+    | delegator1 | VEGA  |     100000      |
+    When the party requests to lock the following tokens:
+    | party      | asset | stake deposited | 
+    | delegator1 | VEGA  |     100000      |
+    And the following tokens transfer should be executed:
+    | party      | asset | stake deposited | locked | 
+    | delegator1 | VEGA  |     100000      | 100000 | 
+    Then the party requests following delegation:
```
ze97286, line 27: maybe something like "the parties submit the following delegation requests:"? (or association or whatever the right term is :))

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+    | from       | to         | asset | delegate amount | 
+    | delegator1 | validator1 | VEGA  |  100000         |
```
ze97286, line 29: I suggest:
from -> party
to -> node id
asset is redundant
delegate amount -> amount

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+    And the parties should have following token amount balances:
+    | from       | to         | asset | delegated amount | 
+    | delegator1 | validator1 | VEGA  |  100000          |
+
+Scenario:  A party can undelegate from a validator
```
ze97286, line 34: I think this makes more sense to go as a final step in the previous scenario otherwise it would require repeating everything from the previous scenario and then add two lines

qa-scenarios/draft-staking-delegation.feature
```diff
@@ -0,0 +1,176 @@
...
+    Then the following tokens transfer should be executed:
+    | from       | to         | asset | undelegated stake | locked |
+    | validator1 | delegator1 | VEGA  |     100000        | 100000 | 
+    | validator2 | delegator1 | VEGA  |     100000        | 100000 |
+    Then the parties should have following token amount balances:
+    | from       | to         | asset | delegated amount | 
+    | delegator1 | validator1 | VEGA  |       0          |
+    | delegator1 | validator2 | VEGA  |       0          |
+
+Scenario: A party cannot delegate less than minimum delegateable stake
```
ze97286, line 93: I think all the tests up to here can be folded to a single one.

qa-scenarios/draft-staking-delegation.feature
witgaw: @ze97286 In reality tokens are never deposited to Vega, right? They are just locked (& thus associated with a given wallet/party on the system) and possibly unlocked. 

Will it be possible to test these actions via integration tests or are we to assume that:
token locked -> governance asset deposited
token unlocked -> governance asset withdrawn
If so, can withdrawal be initiated at any point through the flow (just like tokens can be unlocked at any time)?

Side question: do we have the `undelegateNow` function from the spec or is that not part of Sweetwater in the end?

cpuja84:
No longer needed as the scenarios have been converted and moved to Core repo for the time being.

PR #640, opened 2021-07-29
Updating fees specs with insufficient margin scenarios

cpuja84:
Updating fees specs with insufficient margin scenarios to include mention of maintenance margin levels


witgaw (approved):
LGTM

PR #643, opened 2021-08-02
Added correction for new epoch timings

### LINKED ISSUE #642: Update epoch spec to clarify end of epoch block
The current spec has a condition where we have to place a message in a certain blcok to indicate when a epoch begins. As we have no control over which block a message goes into (until wendy is added) we will remove this requirement for now.

peterbarrow:
Closes #642


gordsport:
Agreed to mereg in specs-sprint meeting

PR #644, opened 2021-08-02
Update staking on the vega side

jeremyletang:
Add new network parameter and explanations to bootstrap the network and reconcile the staking account balances


davidsiska-vega (approved):
LGTM.

PR #655, opened 2021-08-18
Fix the delegator share in example go

### LINKED ISSUE #657: Reward function update for 💧 Sweetwater
After having done a bit more analysis with @Vegaklaus and @MM0819  we've uncovered some properties which will lead to undesirable incentives. 

@Vegaklaus has a fix in the works and in practice it means just he "validator score" and distribution details will change so once it's agreed in specs it will be easy to implement.

davidsiska-vega:
As discussed on Monday.


davidsiska-vega:
Discussed and concluded that this is not the way to go... will be dealt with in #657 .

PR #670, opened 2021-09-03
Tune slack messages

### LINKED ISSUE #660: Jenkins on #protocol-design
Can we please re-tune Jenkins on #protocol-design so that it only prints failed CI steps? Many thanks in advance!

fkondej:
Closes #660.


davidsiska-vega (approved):
Yes, as far as I understand this is what I want: we don't print on success, we only show if something failed.

PR #684, opened 2021-09-14
WIP: Incentivised data node spec

### LINKED ISSUE #685: Incentivised Data Node Spec
In the WIP pull request: https://github.com/vegaprotocol/specs-internal/pull/684 
- @Vegaklaus statistical oddities of hash functions: fill in parameter inputs, decide what are network parameters, give calculation for the expected distance from start block
- @Vegaklaus please check what are the other network parameters arising
- @vegaprotocol/core Please read and flag any questions you have.

davidsiska-vega:
Closes #685 .

### DIFF
--- a/protocol/0000-incentivised-data-node.md
+++ b/protocol/0000-incentivised-data-node.md
@@ -59,19 +59,19 @@ vega blocks.
     {"vega_challenge_datanode", chain_id, epoch, validator_id, data_node_id, challenge}
     
     
-    For the randomization, every datanode has a 'suspicion factor' s (initiallt 1). Ff the hash of the latest block modulo 40000/s equals the datanodes ID (padded with zeros), then
+    For the randomization, every datanode has a 'suspicion factor' s (initially 1). Ff the hash of the latest block modulo 40000/s equals the datanodes ID (padded with zeros), then
     thet block forms a challenge. Similarly, the start bundle S is calculated pseudorandomly from the hash of that block.
     The challenge frequency can be changed by governance vote (which changes the modulus. The value of 40000 will result in
     roughly one challenge per epoch.
-- If datanodes consistently get statistical oddities (e.g., searching 50000 blocks to find the hash rather than the expected 5000 as appropriate for the difficulty/number of zeroes search for), they probably didn't store the full chain. The expected number of blocks is `(1/2) x 10^N`. 
-- A small statistical irregularity occurs if a datanode lies above the expected number of blocks three times in a row, or one measurement exceeds
+- If datanodes consistently get statistical oddities (e.g., searching 50000 data units (i.e., events) to find the hash rather than the expected 5000 as appropriate for the difficulty/number of zeroes search for), they probably didn't store the full dataset. The expected number of untis is `(1/2) x 10^N`. 
+- A small statistical irregularity occurs if a datanode lies above the expected number of dataunits three times in a row, or one measurement exceeds
   105% of the expected value. If a datanode shows a small statiscical irregularity, then its factor s is multiplied by 1.2. Any measurement that
   does not show a small statisticall irregularity multiplies the factor s by 1/1.2, to a minimum of s=0.1.
 - A large statistical irregularity occurs if a datanode lies 800% above the expected value, or lies above the expectation 36 times in a row.
   A datanode that shows a large statistical irregularity is automatically unregistered.
 - The Merkle tree stored by both valdiators and data nodes (described above) is used by the validator to verify that the data returned in response to the challenge is correct, i.e. the answer is the actual data that existed at block B. This does not verify that the node is storing it, hence the expensive search challenge above.
 - This challenge/response to prove data storage can be done fairly irregularly.
-- As the tests are done auotmatically through the system, we do not have an auutomated function that triggers a validator to perform this test;
+- As the tests are done auotmatically through the system, we do not have an automated function that triggers a validator to perform this test;
   rather, the validators get an API call through which they can do a test is they so desire.
 
 The values of 40000 for the testing frequency, 105% for the sensitivite and 3 for the number of allowed failures in a row are network paramaeters (see below). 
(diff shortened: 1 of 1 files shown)


protocol/0000-incentivised-data-node.md
```diff
@@ -0,0 +1,92 @@
+# Incentivised Data Node
+
+On Vega we have validator nodes running Vega core that participate in consensus (produce blocks), non-validator nodes running Vega core  which don't participate in consensus but still have whole blockchain provide all core functionality. Transactions can be submitted to non-validator node and they will gossip with validators. 
+
+Either of these can optionally run data-node. These provide "additional" functionality by storing data emitted by the Vega node and making it accessible through APIs. This additional functionality isn't needed for running the protocol *but* is essential for any meaningful use of the Vega network (Console etc.). 
+
+This spec isn't about what functionality data-nodes should have. It is about how to incentivise other parties to run instances of the data node.
```
edd, line 7: **praise (non-blocking)**: Good intro. Some grammar errors, but good summary 👍

protocol/0000-incentivised-data-node.md
```diff
@@ -0,0 +1,92 @@
...
+
+On Vega we have validator nodes running Vega core that participate in consensus (produce blocks), non-validator nodes running Vega core  which don't participate in consensus but still have whole blockchain provide all core functionality. Transactions can be submitted to non-validator node and they will gossip with validators. 
+
+Either of these can optionally run data-node. These provide "additional" functionality by storing data emitted by the Vega node and making it accessible through APIs. This additional functionality isn't needed for running the protocol *but* is essential for any meaningful use of the Vega network (Console etc.). 
+
+This spec isn't about what functionality data-nodes should have. It is about how to incentivise other parties to run instances of the data node. 
+
+## 
+
+- It is beneficial to the network for the community to run data nodes as a service to users of Vega
```
edd, line 12: **nitpick (non-blocking)**: I think it’s worth re-iterating the above, it’s also **nearly** a requirement than at least **SOMEONE** runs a data-node.

I say nearly - obviously everything can function without a single data-node, but in the early days, that’s so likely to mean no trading, that at least one person *has* to, and the likely cost of only one person doing so is likely to be high so… etc etc.

```suggestion
- Without at least one data node, it is likely that a network will see minimal activity 
- It is beneficial to the network for the community to run data nodes as a service to users of Vega
```

This is an ultra-nitpick and I’m not offended if it’s ignored.

protocol/0000-incentivised-data-node.md
```diff
@@ -0,0 +1,92 @@
...
+A data node's score is the average of all score submitted for that node.
+
+Rewards are paid by multiplying a data node's reward score as a fraction of the sum of all reward scores by the total to be distributed for the epoch.
+
+A data node receiving a zero score for 3 consecutive epochs will be unregistered. 
+
+
+### MVP version requirements
+- only measure API challenge response time
+- only build Merkle tree but don't bother with the rest of the challenge
```
edd, line 136: **question (blocking)**: We can do better than this, but I suspect it’s the stuff above that needs verification, then we can come back to a clearer scope/requirements (see also acceptance criteria below)

protocol/0000-incentivised-data-node.md
```diff
@@ -0,0 +1,92 @@
...
+
+
+### Data bucketing and Merkle tree (commmon, used by both checks below)
+- Data is divided into sequential event bundles which contain all events the included types related to a defined period, e.g. 1 block
+- Only defined event types are included: trades, positions, accounts [TODO: confirm types]
+- Validators and data nodes will build a Merkle tree from event bundles. Validators do not store the data, but data nodes must (and this is what we verify with this algorithm).
+- This merkle tree is exposed by APIs and also allows any user to verify an aPI response from the data node
+
+
+### Is the node carrying the full set of data (NOT needed for MVP / Sweetwater++)
```
edd, line 65: **thought (non-blocking)**: are we inadvertently recreating [LazyLedger/Celestia](https://github.com/celestiaorg) here? I understand the need to ensure that nodes are not malicious/unwittingly bad, but the overheard to the core protocol here seems high.

edd:
Brutal scope chop mode: If only validators can run data nodes, and we build in the proper validators checks that would slash/remove validators that were voted out, we would have a much simpler system than this automatic verification.

protocol/0000-incentivised-data-node.md
```diff
@@ -0,0 +1,92 @@
...
+- Validators can send a datanode a challenge at any time consisting of a random seed value S and a start event bundle number; the datanode then needs to find the first event bundle B in the data set after the start bundle such that HASH(pubkey, S, B) ends with N zeros (e.g. N = 4).
+- The validator request is signed to prevent a DoS
```
Vegaklaus: Better:
A challenge is automatically generated the following way:
In the beginning of each epoch, c=H(last block of the last epoch, blocktime of that block, pubkey) mod 10000 determines when a datanode is challenged. Mod 10000 assures this is in the first part of the epoch, but well spread out between datanodes.
Then, in the epoch, Hash(Block(c), Blocktime(c), pubkey) is used to compute S and B as above.
After computing the result, the datanode posts it on the chain [Which means we'd need to add this to spam protection]. 
Validators evaluate the statistics at the end of the epoch

Future add-on: Other data nodes can test if the data submitted to the chain was correct, and file an accusation in that case. The first accuser gets the $$ from the cheating node if shown right.

davidsiska-vega:
> Brutal scope chop mode: If only validators can run data nodes, and we build in the proper validators checks that would slash/remove validators that were voted out, we would have a much simpler system than this automatic verification.

I think the whole point of this is to allow other community members to run something and get rewarded. 

So having only validators run data nodes solves one problem (we need data nodes) but it doesn't lead to the community engagement plus the resiliency of consoles / bots being able to hit hundreds of data nodes rather than just one of 13. 

But getting rid of the automatic verification could be something worth considering. Could we build a "2nd layer delegation" whereby people can delegate their Vega tokens to data nodes? The vague idea would be to not reward data nodes without delegation and delegators choosing data nodes that provide good service manually. I can see a lot of problems with this... but it's potentially much simpler than "proof of data" + "proof of performance" we're proposing above.

protocol/0000-incentivised-data-node.md
```diff
@@ -0,0 +1,137 @@
+# Incentivised Data Node
+
+On Vega we have validator nodes running Vega core that participate in consensus (produce blocks), non-validator nodes running Vega core  which don't participate in consensus but still have whole blockchain provide all core functionality. Transactions can be submitted to non-validator node and they will gossip with validators. 
+
+Either of these can optionally run data-node. These provide "additional" functionality by storing data emitted by the Vega node and making it accessible through APIs. This additional functionality isn't needed for running the protocol *but* is essential for any meaningful use of the Vega network (Console etc.). 
+
+This spec isn't about what functionality data-nodes should have. It is about how to incentivise other parties to run instances of the data node.
```
davidsiska-vega, line 7: We should also say that there will be a separate spec for how a data-node can "catch-up" is the non-validator node is started from a snapshot.

PR #686, opened 2021-09-14
Net treasury update for 🤠 Oregon Trail

davidsiska-vega:
More details on specific reward types still needed.

### DIFF
--- a/protocol/0056-rewards-overview.md
+++ b/protocol/0056-rewards-overview.md
@@ -66,7 +66,7 @@ Reward schemes are created by a transaction that specifies the information descr
 - Reward type
 - Scope of reward scheme, e.g. a specific market, asset, or network wide (must be a scope compatible with the reward type)
 - Reward scheme parameters (as required for the reward type)
-- Frequency of calculation (specified as interval between payouts in seconds)
+- Frequency of calculation (specified as interval between payouts)
 - Reward scheme start date/time
 - Reward scheme end date/time (blank for never)
 - Payout type and parameters, either:
--- a/protocol/0056-rewards-overview.md
+++ b/protocol/0056-rewards-overview.md
@@ -104,7 +104,7 @@ A Reward Scheme may only be cancelled by a governance proposal. On cancellation,
 
 ## Reward Pool Accounts
 
-A Reward Pool is account is created for a `{reward scheme, asset}` combination when funds are transferred to the reward scheme ID. 
+A Reward Pool is account created for a `{reward scheme, asset}` combination when funds are transferred to the reward scheme ID. 
 
 ## Reward Execution:
 
--- a/protocol/0056-rewards-overview.md
+++ b/protocol/0056-rewards-overview.md
@@ -114,7 +114,7 @@ For each reward pool account for the reward scheme with a non-zero balance:
 
 1. Calculate the `total_payout` either as a fixed fraction of the balance (if payout type = fraction) or as the fraction of the balance resulting from dividing the balance equally by the number of remaining periods (if payout type = balanced).
 1. Calculate the payout per eligible account by: `account_payout = total_payout * (account_scaling_factor / sum(scaling_factors)`
-1. If a per asset max amount per recipient is specified for the asset cap each eligible account's payout to the `max_amount` specified and redistribute any leftover funds between accounts receiving less than the `max_amount`. If all accounts are receiving the maximum allowable, the remaining funds will not be distributed and so remain in reward scheme account.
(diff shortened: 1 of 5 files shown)


edd:
Good update so far :+1:

protocol/0057-reward-functions.md
```diff
@@ -1,33 +1,53 @@
...
+Parameters: A Vega asset or a list of markets to define scope. A string specifying a fee type to be one of: `taker fee paid`, `maker fee received`, `liquidity provision fees received`.
+
+### T.2 - Good risk citizen  (not for 🤠, later)
+
+Principle: anyone who's got a position but hasn't been closed out is rewarded. 
 
 Scope: This reward type can be scoped either to all trading settled in an asset, or to trading in one or more markets which must all settle in the same asset.
 
-Calculation: the scaling factor for a party's rewards is simply the sum of all *taker fees* the party has paid within the defined scope (asset or market(s)) during the period.
+Calculation: From the start of a payout interval collect all parties that have position open during the entire period but haven't had their position closed out. The scaling factor for a party is `1` divided by the total number of parties that qualify within the scope for the period.
```
witgaw, line 24: Shouldn't it be scaled by position size?

protocol/0056-rewards-overview.md
```diff
@@ -100,13 +99,12 @@ A Reward Scheme may be updated by the same method it was created (i.e. governanc
...
 
-A Reward Scheme may be cancelled by the same method it was created (i.e. governance proposal or individual transaction). On cancellation, any undistributed funds in the reward pool account will be transferred to the network treasury for the asset.
+A Reward Scheme may only be cancelled by a governance proposal. On cancellation, any undistributed funds in the reward pool account will be transferred to the network treasury for the asset.
 
 
 ## Reward Pool Accounts
 
-A Reward Pool is account is created for a `{reward scheme, asset}` combination when funds are transferred to the reward scheme ID. Reward Pool Accounts do not need to be tracked by core if they have a zero balance, however APIs may be required to previously used accounts for active reward schemes even if they have a zero balance, as some reward scheme configurations will reach zero balance after each payout and be expected to be topped up, so having the account disappear from APIs/views after payouts would be strange.
-
+A Reward Pool is account is created for a `{reward scheme, asset}` combination when funds are transferred to the reward scheme ID.
```
witgaw:
```suggestion
A Reward Pool is account created for a `{reward scheme, asset}` combination when funds are transferred to the reward scheme ID. 
```

protocol/0056-rewards-overview.md
```diff
@@ -116,8 +114,8 @@ For each reward pool account for the reward scheme with a non-zero balance:
 
 1. Calculate the `total_payout` either as a fixed fraction of the balance (if payout type = fraction) or as the fraction of the balance resulting from dividing the balance equally by the number of remaining periods (if payout type = balanced).
 1. Calculate the payout per eligible account by: `account_payout = total_payout * (account_scaling_factor / sum(scaling_factors)`
-1. If a per asset max amount per recipient is specified for the asset cap each eligible account's payout to the `max_amount` specified and redistribute any leftover funds between accounts receiving less than the `max_amount`. If all accounts are receiving the maximum allowable, the remaining funds will not be distributed.
-1. If a non-zero payout delay is specified, wait for the number of seconds required before continuing to the next step
+1. If a per asset max amount per recipient is specified for the asset cap each eligible account's payout to the `max_amount` specified and redistribute any leftover funds between accounts receiving less than the `max_amount`. If all accounts are receiving the maximum allowable, the remaining funds will not be distributed and so remain in reward scheme account.
```
witgaw: Do we want this? Sees like unnecessary implementation overhead and it is even more "punitive" to the reward whales (the resulting relative reward is even smaller after the redistribution of haircut due to `max_amount` threshold. Why not just leave those funds in the reward scheme?

witgaw (approved):
LGTM, just a few minor comments/suggestions. 

Just a thought really: wondering about all those extra accounts (for direct deposits to treasury or reward schemes). Will these be derived from the same private key (e.g. for eth bridge) that validators already have multisig control over or will there be more overhead in creating them? If it's the latter then wouldn't it make sense to consider prioritising transfers within Vega first so we don't have to create them all in the first place?

protocol/0057-reward-functions.md
davidsiska-vega: That would be another version: have a big position and don't get closed out.

protocol/0056-rewards-overview.md
davidsiska-vega: I agree, I simplified as you suggested.

PR #688, opened 2021-09-14
Web Wallet WIP initial spec for comments

### LINKED ISSUE #702: Create a web wallet spec
In order to ensure we are taking the correct approach and we have non-protocol documentation we need an HSM non-protocol spec based on the Engineering approach

- [ ] Spec updated to detail the feature requirements for development between the core and uidev teams
- [ ] Acceptance criteria so that the feature can be tested (@vega-paul @Ditmir-Vega)
- [ ] Detail any QA scenarios for feature files (if applicable @cpuja84)

See: https://github.com/orgs/vegaprotocol/projects/94

barnabee:


non-protocol-specs/0023-web-wallet.md
```diff
@@ -0,0 +1,25 @@
+# Web Wallet
+
+This file describes the requirements for web based wallet infrastrcuture, including an entirely in-browser wallet library and UI, and a library to standardise conecting to Vega wallets of any type.
```
edd, line 3: Is this an extension, or is this hosted somewhere? If we have local keys, but you need to interact with different sites, then we're going to run in the cross-site data sharing problems. Can you provide some examples of libraries you're thinking of?

non-protocol-specs/0023-web-wallet.md
```diff
@@ -0,0 +1,25 @@
...
+
+
+## In-browser wallet
+
+- User can create an in-browser wallet with new randomly generated keys
+- User can backup an in-browser wallet by taking note of the seed phrase (TBC if this should be possible after initialisation or only during initialisation)
+- User can restore an in-browser wallet from a seed phrase
+- User can use the in browser wallet to interact with the Vega network via dapps such as Console and the Token Frontend (as well as 3rd party dapps built using the same library, see below)
```
edd, line 11: The local go-wallet works around a problem that the browser can't solve: GRPC. I guess the in browser wallet would use REST to submit protobufs it created and signed in the library?

barnabee: Yes I think so. Believe TM has a REST transaction submission endpoint but tbh this WIP spec is a starting point so happy to make changes if it's not possible/practical

non-protocol-specs/0023-web-wallet.md
barnabee: This probably needs investigation. If not built into an extension it seems like it'd either be a per-site wallet or need to be interacting with an API that serves up an encrypted seed phrase/wallet backup, which might not be too bad but worth discussing… the more use cases we can enable and support the better but got to be realistic.

campbellssource:
This may be true by default but as a test case: "A dApp should be able to interact with a web3 provider as well as the Vega web wallet, to enable deposits/withdrawals and association".

Assumption: if Vega wallet was a browser exertion, clicking a "deposit" link couldn't open the metamask extension to sign the Eth transaction. therefor the dApp would have to act as the middleman providing all the ui for the Deposit flow

edd:
Question from @MM0819 : can we make it so that users can import their keys from the local wallet?

I can’t see a reason why not. But please add it to the spec.

gordsport:
see linked issue comment

PR #692, opened 2021-09-16
Snapshot spec

peterbarrow:
Information gathering around the snapshot feature


edd:
Something that isn't clear to me is how this overlaps with https://docs.tendermint.com/master/spec/abci/apps.html#state-sync. I assume this is the code to make state sync possible. Should we make that explicit?.

UPDATE UPON READING FORWARD ONE PARAGRAPH: It's explicit. Can we just add a note in the intro that explains what I said in the paragraph above. I'll add a suggestion.

> This spec describes how Vega will produce the snapshots. Tendermint handles the process of triggering a snapshot's creation and broadcasting snapshots to new nodes. For more information see [the Tendermint state sync documentation](https://docs.tendermint.com/master/spec/abci/apps.html#state-sync)


Otherwise: good spec, even in its current state without the explicit list :+1:

EVODelavega (approved):

PR #693, opened 2021-09-16
Data node discovery via governance and IPFS

edd:
**NOTE** The functionality intended in this spec (node discovery) will be covered in https://github.com/vegaprotocol/specs-internal/pull/684

~Adds a proposal for a new network parameter and how that can be used to bootstrap clients with a data node address~

~This *could* be simplified to just be a governance parameter that contained a JSON list of data-nodes. The advantage of adding IPFS was the public gateways between IPFS and the web. Without IPFS, we have a different bootstrapping problem.~

~This spec is scoped to the problem we have now (client discovery of data nodes) and does not attempt to solve the updating of data node addresses, but it doesn't do anything that precludes solving it either.~

### DIFF
--- a/protocol/006x-data-node-discovery.md
+++ b/protocol/006x-data-node-discovery.md
@@ -1,0 +1,61 @@
+# Summary
+As a user with a Vega client side application, in order to start interacting with the nework I need to be able to fetch a URL for a data node
+
+This spec adds a single network parameter that allows us to leverage our existing Governance system to have a network-wide agreed upon IPFS hash for 
+
+# Guide-level explanation
+However, IPFS doesn't solve everything. Most browsers and platforms can't natively resolve an IPFS hash and grab the content behind it. However:
+- Some forward looking browsers like Brave are building in IPFS support
+- Other browsers have extensions
+- There are publicly accessible gateways
+
+This makes IPFS a reasonable place to put the list of data nodes. It does have the classic bootstrapping problem - if you don't know a data node to get the IPFS hash from, how do you get the list of data nodes? This is a solvable problem that isn't yet solved. Our first limited mainnet will have a set of known validators, any of whom could choose to provide a static, web-accessible link that will resolve the IPFS content.
+
+# Reference-level explanation
+This is broken down by requirements for an initial, known-validator set and slightly beyond. This does not solve for a completely fluid validator set.
+
+## Restricted mainnet
+A new network parameter is added that contains an IPFS hash.
+
+The hash that is set in the genesis file is the hash of a file that has been hand-coded by one of the validators in the initial genesis set of validators.
+
+The file contains a set of URLS for data nodes, as a JSON array. For example, our current testnet data node list would look like:
+
+```
+[
+  "https://n01.testnet.vega.xyz",
+  "https://n02.testnet.vega.xyz",
+]
+```
+
+### Updating the restricted mainnet data node list
+The process for changing this list are:
+- Make a new file containing the updated list of validators
+- Pin the file to IPFS
+- Submit a governance proposal to update `data-node-list-hash` with the new list hash
+- It is then up to users to verify the list and approve or vote down the proposal 
+
+## Client usage
+Now that there is a list of data nodes agreed on by the network, we have an established, provably approved list of data nodes that the Console, token front end or Go wallet could use. The only problem for them is getting hold of it, as currently none of them have IPFS libraries built in.
+
+### Short term
+Initially, we can hardcode the same hash as the genesis file `data-node-list-hash` in to any client, and use the IPFS.io public gateway.
+
+### Long term
+An IPNS can be provided that always points to the IPFS hash of the current network's `data-node-list-hash`.
+
+# Network parameters
+
+- `data-node-list-hash` - A string containing an IPFS hash.
+
+# Acceptance criteria
+
+## Basic
+- There is a new network parameter that contains a string.
+	- That string should look like an IPFS hash
+	- It does not need to be verified as a valid IPFS hash
+	- It does not need to be verified as an available IPFS hash
+	- The contents of the IPFS hash does not need to be verified as a valid address list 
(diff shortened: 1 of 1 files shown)


protocol/006x-data-node-discovery.md
```diff
@@ -0,0 +1,61 @@
+# Summary
+As a user with a Vega client side application, in order to start interacting with the nework I need to be able to fetch a URL for a data node
+
+This spec adds a single network parameter that allows us to leverage our existing Governance system to have a network-wide agreed upon IPFS hash for
```
davidsiska-vega, line 4: Hash for what?

protocol/006x-data-node-discovery.md
```diff
@@ -0,0 +1,61 @@
...
+```
+[
+  "https://n01.testnet.vega.xyz",
+  "https://n02.testnet.vega.xyz",
+]
+```
+
+### Updating the restricted mainnet data node list
+The process for changing this list are:
+- Make a new file containing the updated list of validators
```
davidsiska-vega, line 33: validators or data nodes?

protocol/006x-data-node-discovery.md
```diff
@@ -0,0 +1,61 @@
...
+- Pin the file to IPFS
+- Submit a governance proposal to update `data-node-list-hash` with the new list hash
+- It is then up to users to verify the list and approve or vote down the proposal
```
davidsiska-vega, line 36: If it's just data nodes then the validators could reasonably manage the list themselves. It's not clear that storing it in an IPFS hash does that much for us. 

For the MVP version of incentivised data nodes it is proposed that:
- new data nodes submit a transaction to Vega to register their presence and Vega key for rewards.
- validator nodes (or maybe other data nodes, @Vegaklaus is still working on this) test liveness. One way or another the Vega network should have a picture of what data nodes it has and which ones are live. 

So it should be possible for the nodes to update this list, perhaps on IPFS, perhaps elsewhere regularly and confirm the hash themselves. 

Even if the above isn't a good solution I am not sure that asking users to check the data nodes and vote on the hash will work (even if we make a very good UI for this that will basically test all the connections and tell people how to vote we may struggle to get participation).

protocol/006x-data-node-discovery.md
davidsiska-vega: Hash for the IPFS file listing all the data nodes I guess.

barnabee:
> This spec is **scoped to the problem we have now** (client discovery of data nodes) and does not attempt to solve the updating of data node addresses, but it doesn't do anything that precludes solving it either.

Is this a problem for Restricted Mainnet?

tbh this seems duplicative of the data node registration and incentivisation design aka "incentivised data node" so I am (a) uncomfortable making a decision on this without a decision on IDN; and (b) of the opinion that if we do need data node discovery sooner we should probably try and build that part of the incentivised data node spec. rather than do something else that is temporary. 

All dependent on where we get to with incentivised data node prioritisation and scope though…

protocol/006x-data-node-discovery.md
barnabee: Agree I don't see much value in IPFS for data node reg. 

Maybe IPFS would be useful to store a list of reliable Vega [non-data] nodes to query for data nodes but as I don't believe you can use threshold signatures for IPFS content it still begs the question of how it gets updated without being centralised.

If the majority of clients are connecting to datanodes using this method, then pwning this list/resource == pwning the network because you can just relace the real network with your own.

protocol/006x-data-node-discovery.md
edd: Data nodes

protocol/006x-data-node-discovery.md
edd: > If the majority of clients are connecting to datanodes using this method, then pwning this list/resource == pwning the network because you can just relace the real network with your own.
This isn't an issue if (as they are) IPFS clients are checking the hash of the file they fetch. If it matches the hash in the governance param, then this is a non issue. But there is a trust element in that one person is ultimately responsible for updating the file. The spec doesn't solve that.

edd:
I've closed the ticket, but I'm going to respond to the points.

> Is this a problem for Restricted Mainnet?

Yes. We currently would have to build the list of data nodes in to the console. This would avoid that. We will just have to build them in for now.

> tbh this seems duplicative of the data node registration and incentivisation design aka "incentivised data node" s

Yes, but it is deliverable much quicker than any part of the data node spec

>  of the opinion that if we do need data node discovery sooner we should probably try and build that part of the incentivised data node s

Let's do this :+1: Data node discovery is a more pressing issue than incentivised data nodes.

> ...t still begs the question of how it gets updated without being centralised.

This spec doesn't solve that problem.

> If it's just data nodes then the validators could reasonably manage the list themselves. It's not clear that storing it in an IPFS hash does that much for us.

Yes, with a broader API change than this would entail

barnabee:
> Yes. We currently would have to build the list of data nodes in to the console. This would avoid that. We will just have to build them in for now.

Not sure on the value of avoiding a fixed set of data node hostnames when we are committed to a fixed set of validators, hence why I questioned the need for RM.

> Data node discovery is a more pressing issue than incentivised data nodes.

[ ctitation needed ] 

This is debatable.

edd:
> Not sure on the value of avoiding a fixed set of data node hostnames when we are committed to a fixed set of validators, hence why I questioned the need for RM.

If data nodes go down and are no longer accessible, then having them on the list is no longer useful. Clients will individually have to check if the node is available. A dynamic list could remove data nodes when they are down.

barnabee:
> If data nodes go down and are no longer accessible, then having them on the list is no longer useful. Clients will individually have to check if the node is available. A dynamic list could remove data nodes when they are down.

Yeah makes sense. I was thinking that if each validator agrees to run at least one data node, and to provide a hostname (which could be a load balancer) then it seems like we could potentially get to an agreed set of 13 hostnames, one per validator, for RM or something like that and make this less urgent to solve.

PR #694, opened 2021-09-17
V E G A validator master keys

### LINKED ISSUE #698: Update Spec for the Master Key feature required for the validators in 💦
In order to start development on the 💦 feature of 'Master Key' we need the spec updated to include information for the team to review and start development:

- [ ] Spec updated to detail the feature for development
- [ ] Acceptance criteria so that the feature can be tested (@vega-paul)
- [ ] Detail any QA scenarios for feature files (if applicable @cpuja84)

core ticket: https://github.com/vegaprotocol/vega/issues/4061

davidsiska-vega:
Very very very WIP at the moment.


protocol/0000-validator-vega-master-keys.md
```diff
@@ -0,0 +1,26 @@
+# Master Keys for Validators
+
+This is a safety related feature specifically for validators. To run the Vega network validators need three keys: Ethereum, Tendermint and Vega. This spec is about the Vega key. 
+
+The vega wallet is derived from a master key (bip39 and slip-10 are used), from this master key, new keys are derived deterministicaly. As of now the master key is used only for deriving new keys, and is never used to prove identity or signing anything.
+
+In regards to validators, their identity should become the public key of this master key used to derive wallets.
+When a validator wishes to change the private key used to sign vega transaction emitted by the node (the hot key), they then generate a new hot key pair from the master key and send a transaction to notify the network of the change of public hot key.
```
barnabee: It is not clear whether the network can or will verify that the hot key is derived from the master key? 

I assumed this wasn't possible, in which case it seems to me like they could use a completely different key and are not limited to the master key from which the hot key is derived as their static identity key?

If that verification _is_ possible we should state explicitly that when they submit the new hot key the network will verify that it is derived from the master key and reject it if not, if that's the intention.

gordsport:
Hi @Vegaklaus, 
Please can you add some initial acceptance criteria to this so that @cpuja84 / @vega-paul can review and add to if/as required?
This has been raised as highest priority for the 💦 milestone
Thanks

gordsport:
@jeremyletang @ValentinTrinque - are you OK with this as a spec, if so can we merge?

@cpuja84 @vega-paul - can you add/review acceptance criteria?

ValentinTrinque:
It appears @barnabee commented it. Should we wait for the answer ?

protocol/0000-validator-vega-master-keys.md
Vegaklaus: It's not possible as far as I understand slip-10, which makes this text a little unclear. The new key needs to be authenticated by the master key (one of the points is that not all should be lost if you use a hot key), so for the key update function, we need to sign the new key with the master key (which means the statement that nothing is signed is not entirely true, it's nothing but new keys).

What we can do (though I suggest that's post SW) is to derive 2 keys, the hot key and the 'next update key', and use the later for the next key update; this way, we never need to give anyone a critical public key.

barnabee: Ok it would be good to update this to clarify that the new key does not need to be generated from the same root as the master key but it does need to be signed by the master key.

PR #708, opened 2021-09-29
Bring SW staking and delegation spec in line with the implementation

witgaw:
☝️


protocol/0061-simple-POS-rewards - SweetWater.md
```diff
@@ -9,9 +9,9 @@ At the end of an [epoch](./0050-epochs.md), payments are calculated. This is don
...
 * Finally, the total reward for a validator is computed, and their delegator fee subtracted and divided among the delegators.
-* If the validator (or, the associated key) does not have sufficient stake (at least the network parameter min_own_take), 
-  then the reward is set to zero. The corresponding money is kept by the network, not distributed among the other validators. Note this only applies to the part of the reward attributable directly to such a validator, its delegators should still receive their rewards.
+* If the validator (or, the associated key) does not have sufficient stake (at least the network parameter `min_own_stake`), then the reward (both the validator and self-delegation component) is set to zero. The corresponding amount is kept by the network, not distributed among the other validators. Note this only applies to the part of the reward attributable directly to such a validator, its delegators should still receive their rewards. If a validator delegates amount below `min_own_stake` to a different validator then the reward associated with that delegation will be paid out.
```
Vegaklaus: Shoudnt' the last 'below min_own_stake' be 'above or equal' ? And it's only validators that can save their peers (not team for that matter) ?

protocol/0061-simple-POS-rewards - SweetWater.md
```diff
@@ -9,9 +9,9 @@ At the end of an [epoch](./0050-epochs.md), payments are calculated. This is don
...
 
+**Note**: changes of any network parameters affecting these calculations will take an immediate effect (they aren't delayed until next epoch).
```
Vegaklaus, line 14: Where does that come from ? It might cause some oddities, e.g., a validator not having enough own_stake, and not being able to fix this until the next epoch (as all changes only happen per epoch) ?

protocol/0061-simple-POS-rewards - SweetWater.md
ze97286: I think what @witgaw is trying to say is that regardless how small amount a validator delegates to other nodes, even if those validators have self-stake less than `min_own_stake` they still get rewarded as delegators for that.

protocol/0061-simple-POS-rewards - SweetWater.md
ze97286: while network parameters changes are applied as they occur, I don't think this is relevant as delegations, undelegations are done at the end of the epoch anyways and so is reward calculation.

protocol/0061-simple-POS-rewards - SweetWater.md
witgaw: Yep, so I wanted to highlight the fact that if a validator is to stake below `min_own_stake` they are actually better off delegating to a different validator as in that case they will at least receive some rewards due to delegation (if they self-delegated instead they'd get no rewards). Seems like something we might want to change in the future, not sure.

protocol/0061-simple-POS-rewards - SweetWater.md
witgaw: Yep, it's not a huge deal, but goes somewhat against the concept of the epoch in my mind at least (you'd expect that things stay more or less constant within an epoch), so it would perhaps make sense to delay any changes so that they apply from next epoch onwards (after the rewards have been calculated for current epoch). 

Again, just trying to document the current behaviour in a spec as it took a while for me to figure some of those things out and wanted to flag them in the SW spec in case we think any changes to these behaviours should be scheduled for SW++.

ze97286: what I'm trying to say is that they can change many times during the epoch but as they are only used at the end of the epoch (or even later for reward params), only the last value matters.

witgaw: Sure, what I meant is that it could be considered unfair: people make their delegations for the epoch based on what they look-up to be the current value of the parameters so they'd expect certain outcomes at least within that epoch I think. 

Parameter changes have an enactment time anyway I think, so they can probably check if they are about to change within the epoch so perhaps nothing to worry about at all, just wanted to flag it in the spec.

protocol/0061-simple-POS-rewards - SweetWater.md
Vegaklaus: Ok, get it. Maybe better to add 'just like any other delegator' ?

protocol/0061-simple-POS-rewards - SweetWater.md
Vegaklaus: I was more worried that, for example, we do a change that affects the minimum stake needed to be a validator; the way it's formulated now, a validator could be kicked out immediatelly (though we did have a delay parameter for those changes anyhow, right ?)

witgaw: Currently they never get kicked out, if anything they could stop receiving rewards. I've not used the governance properly (just via tests) in a while, but yeah, from what I remember each proposal has an enactment period (so there's a bit of a gap between when it gets voted it and actually takes effect).

protocol/0061-simple-POS-rewards - SweetWater.md
witgaw: Good point, added now.

protocol/0061-simple-POS-rewards - SweetWater.md
Vegaklaus: Ok, it's fine for SW then, though I do see some issues in the future when network parameter changes have a bigger impact on the chain operation (e.g., changing the validator weights), so some more work needs to go into this to make sure we don't get any odd sideeffects.

Vegaklaus:
Some future work needed for SW+++ on the time of activation for network parameter changes (i.e., does anything explode if we change them mid-epoch once we have more functionallity).

PR #711, opened 2021-09-29
WIP Update 0062-spam-protection.md for post Sweetwater improvements

### LINKED ISSUE #737: Spam protection doesn’t cover orders in Sweetwater, but users can place orders
https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0062-spam-protection.md#spam-protection explicitly excludes order from the Spam Protection, and it is implemented this way. However I think this was written at a time where we thought Order transactions would be disabled. As a result, I think there’s a minor hole, that’s slightly patched up with Rate Limiting.

Rate limiting is currently set to restrict the same key from sending more than 500 txs within 10 blocks. As long as they stay below this number, they can send as many orders as they like to gum up some blocks. Orders will pass our basic `checkTX`, but then be rejected because there won’t be any markets in sweetwater.

Additionally, we’re considering removing rate limiting, which would leave this wide open. Should we start implementing more of spam protection before removing rate limiting?

### LINKED ISSUE #905: Update 0062-SPAM-spam_protection - post Oregon Trail
```
0062-SPAM-spam_protection.md
  Criteria:      0
  Referenced by: 0
```

Vegaklaus:
First throw at the future of spam protection (as a PR on the current version to keep the number of files down, with no intent to merge it before it matures a little more)

### DIFF
--- a/protocol/0062-spam-protection.md
+++ b/protocol/0062-spam-protection.md
@@ -92,7 +92,7 @@ The primary purpose of spam protection is to protect the blockchain from overloa
 also does to some extend protect the core (as transactions that never make it onto the chain
 also don't bother the core), but that is a secondary goal.
 
-Thus, spam protetion defines a set of rules that allow transafctions to be processed by
+Thus, spam protection defines a set of rules that allow transactions to be processed by
 the blockchain (or not). Due to current mempool limitations, the only input the spam
 protection is allowed to use is any data derived from comitted blocks on the chain. Thus,
 if a party is allowed 3 votes on an issue but sends 4 in one block, the fourth vote is 
--- a/protocol/0062-spam-protection.md
+++ b/protocol/0062-spam-protection.md
@@ -85,3 +85,83 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
   banning policy this is not doable from one account, but with a sybil attack it can be done. If this ends up being a
   problem, we can address it by increasing the ban-time.
   
+# Post Sweetwater model
+(For discussion now, not to be merged anytime soon)
+
+The primary purpose of spam protection is to protect the blockchain from overloading. This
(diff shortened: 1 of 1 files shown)


protocol/0062-spam-protection.md
```diff
@@ -85,3 +85,83 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
...
+account has. Thus, an account with one cent worth of assets in it is not allowed to make 
+10.000 trades in one block, while an account with 100 million dollars may well have a good reason
+to do so. To be able to implement this, we need some form of normalized asset values, as the
+asset of an individual account can come in a large number of currencies.
+As the primary currency in the beginning will be ETH, it is probably a good idea to measure
+assets in ETH as well. On the long run, they could be linked to something more stable,
+or even the average of a basket of assets.
+Thus, we assume that spam protection has a function
+
+	float normalized_assets(account)
```
davidsiska-vega, line 134: My one comment is that all the computations should be designed so that they can be performed using ints / decimal types.

protocol/0062-spam-protection.md
```diff
@@ -85,3 +85,83 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
...
   problem, we can address it by increasing the ban-time.
   
+# Post Sweetwater model
+(For discussion now, not to be merged anytime soon)
+
+The primary purpose of spam protection is to protect the blockchain from overloading. This
+also does to some extend protect the core (as transactions that never make it onto the chain
+also don't bother the core), but that is a secondary goal.
+
+Thus, spam protetion defines a set of rules that allow transafctions to be processed by
```
edd:
```suggestion
Thus, spam protection defines a set of rules that allow transactions to be processed by
```

PR #713, opened 2021-10-06
Null blockchain spec proposal

### LINKED ISSUE #724: Specify null blockchain with time fast forwarding
Spec proposal: https://github.com/vegaprotocol/specs-internal/pull/713

witgaw:
The corresponding core ticket is https://github.com/vegaprotocol/vega/issues/4155


non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -0,0 +1,46 @@
...
+This is effectively first step towards a "minimal" scenario runner but it may also be useful for QA to allow them to run tests faster (no need to wait for time to pass).
+
+It must have the following features:
+
+## Time forwarding commands / transactions
+
+The blockchain can be started with time set (different from wall time) so e.g. `1st January 2010 00:00:00`. 
+User can submit a transaction saying: move time forward by `1h:10m` or move time to `1st January 2010 01:10:00`. 
+Alternatively the user can set the blocktime length and move forward by a fixed number of blocks. 
+Upon receiving this transaction Vega must carry out all the protocol actions that happen with passage of time (e.g. ending auctions, closing / enacting governance proposals, updating price monitoring bounds changing due to passage of time, etc.).
```
jeremyletang: Does that means that every time move needs to be trigger by "a call of some sort" to the null blockchain? And no "blocks" would be produce until this call is made? Or would you like the null blockchain to produce block periodically? I don't think both approach would be compatible easily, and the second one clearly not much predictible.

In what case I can imagine "endpoints" (this is to be defined how this is done) enabled when using a null blockchain which would allow the user as mention to move to a given time, or just move 1 block forward, by do soing increasing the time of a configured default duration (expected block time).

non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -0,0 +1,46 @@
...
+## Time forwarding commands / transactions
+
+The blockchain can be started with time set (different from wall time) so e.g. `1st January 2010 00:00:00`. 
+User can submit a transaction saying: move time forward by `1h:10m` or move time to `1st January 2010 01:10:00`. 
+Alternatively the user can set the blocktime length and move forward by a fixed number of blocks. 
+Upon receiving this transaction Vega must carry out all the protocol actions that happen with passage of time (e.g. ending auctions, closing / enacting governance proposals, updating price monitoring bounds changing due to passage of time, etc.). 
+
+## Parties and balances
+
+User can create parties and add / remove balances, mocking withdrawals without having to connect to any Ethereum chain. User can register "associated" staking / governance tokens for purposes of staking and delegation and voting on governance without having to connect any Ethereum chain.
```
jeremyletang: That is possible, but would mean more layer of abstraction unrelated with the tendermint chain itself.

Doing this imply 2 different null blockchain:
- a tendermint null blockchain replacing the tendermint consensus
- a ethereum null blockchain which allow replacing the staking implementation with e.g: a builtin asset

non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -0,0 +1,46 @@
...
+Upon receiving this transaction Vega must carry out all the protocol actions that happen with passage of time (e.g. ending auctions, closing / enacting governance proposals, updating price monitoring bounds changing due to passage of time, etc.). 
+
+## Parties and balances
+
+User can create parties and add / remove balances, mocking withdrawals without having to connect to any Ethereum chain. User can register "associated" staking / governance tokens for purposes of staking and delegation and voting on governance without having to connect any Ethereum chain. 
+
+## Submitting transactions
+
+User can submit transactions without signing or alternatively padding with possibly invalid signature and Vega will assume the transaction signature is valid. 
+Basically, we don't want core to waste CPU cycles on signature verification in case the user wants to run many scenarios fast.
```
jeremyletang: This one is more tricky, At the moment all transaction goes through a unique flow of processing where verifying signature is part of this flow, I'm not sure we would have a good way of avoiding signing / verifying transaction here. I'm also not sure how useful is that? I'd tend to say that outside of replacing the blockchain, everything else should probably not impact / modify the flow of a normal transaction if possible.

non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -0,0 +1,46 @@
...
+It must be possible to record event bus events.
+
+## CLI support
+
+There is a command line tool to launch vega with null blockchain, data node and any other processes needed for this functioning. 
+It will should have the following command line options:
+
+- genesis config file
+- current time (optional, if wanted to be different from wall time)
+- event bus output file
```
jeremyletang: I think the command line should be the normal vega command line, and be able to just specify via a flag the blockchain that needs to be used (e.g vega node --blockchain="tendermint|null").

Also yes we will need support a genesis file, should that be a custom format? or do we want the tendermint format? ultimately we will need to support at least the same app_state format.

Not sure we need the event bus output into a file, no reason you could not connect to the normal stream and save it / process it on your application side right? I mean this can be added if necessary, but if we can avoid adding functionnalities "just" for this in the core, then that would be great

non-protocol-specs/0000-null-blockchain-vega.md
witgaw: Could we somehow define block capacity (say number of transactions - irrespective of type) we fit into a block and block time (difference in timestamp between two consecutive blocks). Then transactions would be packaged into blocks as they come and when the "Move time forward by 1h:10m" is submitted the necessary number of empty blocks (or near empty depending on what sort of transactions get included in a real network that doesn't get any external transactions submitted) such that at least 1h:10m elapsed gets produced before any new transactions get processed)?

non-protocol-specs/0000-null-blockchain-vega.md
witgaw: I think that'd be ideal (having 2 different null blockchains).

non-protocol-specs/0000-null-blockchain-vega.md
witgaw: I think it will be fine to leave it out in the first implementation and only look into it if signing transactions turns out to be a major bottleneck.

non-protocol-specs/0000-null-blockchain-vega.md
witgaw: You're right, this goes beyond this immediate spec and deals more with supporting tools and usage of this functionality - removed this section from the spec.

non-protocol-specs/0000-null-blockchain-vega.md
davidsiska-vega: Or a fake Ethereum event forwarder, that we can submit "transactions" to that then get passed via the null blockchain into Vega.

non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -9,7 +9,8 @@ It must have the following features:
 
 The blockchain can be started with time set (different from wall time) so e.g. `1st January 2010 00:00:00`. 
 User can submit a transaction saying: move time forward by `1h:10m` or move time to `1st January 2010 01:10:00`. 
-Alternatively the user can set the blocktime length and move forward by a fixed number of blocks. 
+The user can submit a command to start a new block and end a current block.
+Alternatively the user can set the block time length and move forward by a fixed number of blocks.
```
witgaw: I don't think it's really an alternative, to me these are two somewhat different things. One deals with block composition, the other with timestamp difference between subsequent blocks. 

I'd just expose:
- block time
- block max capacity

Then we can have blocks generated automatically when capacity is reached or manually by moving a fixed number of blocks forward before submitting new transactions. Block start and end commands seem like too much granularity to me.

peterbarrow: OK, I've updated this to represent what you have suggested.

non-protocol-specs/0000-null-blockchain-vega.md
peterbarrow: During the meeting we decided this was an implementation detail and does not have to be specified here.

non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -0,0 +1,41 @@
...
+This is effectively first step towards a "minimal" scenario runner but it may also be useful for QA to allow them to run tests faster (no need to wait for time to pass).
+
+It must have the following features:
+
+## Time forwarding commands / transactions
+
+The blockchain can be started with time set (different from wall time) so e.g. `1st January 2010 00:00:00`. 
+The length in time of each block is defined at startup (default is 1 second).
+The maximum number of transactions in a block is defined at startup (default 10). The null blockchain system will automatically end a block and create a new one if the transaction count is met/exceeded.
+User can submit a transaction saying: move time forward by `1h:10m` or move time to `1st January 2010 01:10:00`. Moving time will automatically end the current block if it contains transactions and will create all the required empty blocks needed to get us to the new time.
```
witgaw: Nitpick, but might be worth getting it in so we think about it the right way from the start: 
```suggestion
User can submit a transaction saying: move time forward by at least `1h:10m` or move time beyond `1st January 2010 01:10:00`. Moving time will automatically end the current block if it contains transactions and will create all the required empty blocks needed to get us to the new time.
```

witgaw (commented):
Looks good to me, just one tiny comment.

non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -0,0 +1,41 @@
...
+This is effectively first step towards a "minimal" scenario runner but it may also be useful for QA to allow them to run tests faster (no need to wait for time to pass).
+
+It must have the following features:
+
+## Time forwarding commands / transactions
+
+The blockchain can be started with time set (different from wall time) so e.g. `1st January 2010 00:00:00`. 
+The length in time of each block is defined at startup (default is 1 second).
+The maximum number of transactions in a block is defined at startup (default 10). The null blockchain system will automatically end a block and create a new one if the transaction count is met/exceeded.
+User can submit a transaction saying: move time forward by at least `1h:10m` or move time beyond `1st January 2010 01:10:00`. Moving time will automatically end the current block if it contains transactions and will create all the required empty blocks needed to get us to the new time.
```
davidsiska-vega: There perhaps already is but if not, there should be a method to ask Vega with null blockchain what time it is on the null blockchain. Not sure where to best incorporate it @peterbarrow .

non-protocol-specs/0000-null-blockchain-vega.md
```diff
@@ -0,0 +1,41 @@
...
+User can move time forward by a fixed number of blocks. 
+Upon receiving this transaction Vega must carry out all the protocol actions that happen with passage of time (e.g. ending auctions, closing / enacting governance proposals, updating price monitoring bounds changing due to passage of time, etc.). 
+
+## Parties and balances
+
+User can create parties and add / remove balances, mocking withdrawals without having to connect to any Ethereum chain. User can register "associated" staking / governance tokens for purposes of staking and delegation and voting on governance without having to connect any Ethereum chain. 
+
+## Submitting transactions
+
+Transactions are submitted in the same way as for a normal live system, i.e. messages must be signed as usual.
```
davidsiska-vega: Could we have a "fake signature" here so that basically anything is accepted as signed? So that the logic flow in core doesn't have to change, just the "verify signature" function will always return true if running with null blockchain... (thinking how to save CPU time). If that's too much bother then signing for now is ok.

davidsiska-vega (commented):
I've left a few comments but nothing "critical". Feel free to incorporate or ignore and consider this approved from my end, once you've reviewed the comments @peterbarrow .

non-protocol-specs/0000-null-blockchain-vega.md
peterbarrow: The existing API will still work, so you will be able to call GetVegaTime() to get the current chain time. No extra work needed by us.

non-protocol-specs/0000-null-blockchain-vega.md
peterbarrow: It's possible to add code to handle a fake signature but like Jeremy said it's best if we don't touch the core code if we can help it. I don't think processing the signatures will add much to the run time of each scenario so for version 1 we should leave it as it is.

PR #715, opened 2021-10-12
Add target stake calc clarification

witgaw:
Target stake calculation: use indicative uncrossing price whilst in auction.


davidsiska-vega (approved):
LGTM

PR #716, opened 2021-10-12
Add maxPayoutPerParticipant netparam

witgaw:
Add `reward.staking.delegation.maxPayoutPerParticipant` network parameter to the spec and clarify that the cap does not get applied when it's set to `0`.


davidsiska-vega (approved):
LGTM

PR #721, opened 2021-10-15
Draft Key-Risk-and-management

Vegaklaus:
Vega related keys, their risks, and their management

### DIFF
--- a/protocol/0000-002KeyManagement.md
+++ b/protocol/0000-002KeyManagement.md
@@ -5,6 +5,10 @@ The term "key loss" usually refers to both the key becomming unavailable, and th
 by a non-authorized party. To distinguish these two cases, we will use the term 'key loss' for the former
 (the validator looses access to the key), and 'key compromise' for the later.
 
+For all thresholds, we assume n validators, up to t of them may be malicious. At this point, we do not
+take into account validator weights; this will change eventually for the Tendermint key, at which 
+point we will have a total weight of n with up to t weight being tolerated to be corrupted.
+
 ## Keys
 
 ### Ethereum Key [Staking]
@@ -49,9 +53,15 @@ Hence, a resonably frequent key update would constitute good practice.
 Adding and removing an ethereum key from the MultiSig contract incurs gas costs to all the validators,
 not just the one switching the key.(*) 
 
-Disaster management procedures in case this key gets lost are currently in the works.
-
 The local management of the ETH key is done using CLEF. Details of this are specified elsewhere.
+
+## Future Features
+Better disaster management procedures in case this key gets lost are currently in the works. Especially, we're
+currently looking into integrating threshold signatures, which can allow Validators to be removed, added
+or change their keys without any interaction with the smart contract - the main issue here is the asynchrony
+and assuring we generate a Ethereum native signature.
+
+
 
 ### Vega Key [Identity]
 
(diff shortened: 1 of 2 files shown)


davidsiska-vega (changes requested):
I think this should be split into two documents. First one should be readme-keys.md and it should go directly into the https://github.com/vegaprotocol/networks repo.  

It should strictly document the current behaviour (none of the "we may implement" but it can link to the one below saying this is in the works). 

The second document should be a spec for the tendermint key certificate signing stuff and that should stay in the specs repo. 

What do you think?

Also, could you please define `n` and `t` somewhere before they're first used? Thx!

protocol/0000-002KeyManagement.md
```diff
@@ -0,0 +1,234 @@
...
+### Ethereum Key [Staking]
+The staking Ethereum Key is the key to which the Vega ERC20 tokens are tied. This key is no different
+to the key held by any other tokenholder or indeed any participant on the Ethereum blockchain. 
+It is only required to move the vega tokens to another Ethereum address (public key) or to assiciate the tokens 
+to a Vega key. 
+
+Once the association to a Vega key is finished this key is not required and thus can be kept in cold storage. 
+"Changing" this key would require the holder to move the Vega tokes to another public key they control. 
+This is not possible for Vega tokens held by the vesting contract. If this key
+gets lost, the the Vegsa tokens become unmovable. If they have previously been associated to a Vega key
```
jeremyletang, line 23:
```suggestion
gets lost, the Vegas tokens become unmovable. If they have previously been associated to a Vega key 
```

protocol/0000-002KeyManagement.md
```diff
@@ -0,0 +1,234 @@
...
+1) an attacker compromises one validator, asks the HSM to sign a transaction that
+transfers all assets to an address of their choice, leaves that validator. 
+Of course one signature isn't enough to transfer the assets. 
+2) Repeat the same over time with other validators. 
+3) Now they have n-t signatures and at this point they can withdraw all Ethereum assets controlled by the 
+MultiSig contract.
+
+Hence, a resonably frequent key update would constitute good practice. 
+Adding and removing an ethereum key from the MultiSig contract incurs gas costs to all the validators,
+not just the one switching the key.(*)
```
jeremyletang: Well not really, we assume that the validator willing to change it's key would pay the gas cost right? To remove / add a key a single transaction is required (2 actually, one for each action).

protocol/0000-002KeyManagement.md
```diff
@@ -0,0 +1,234 @@
...
+and assuring we generate a Ethereum native signature.
+
+
+
+### Vega Key [Identity]
+
+This key is split into master and hot key as specified in https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0063-validator-vega-master-keys.md . 
+
+It is possible to update the hot key at network restarts. The master key can be kept in cold storage
+and only used to sign the update transaction.
```
jeremyletang, line 69: That is true now, nut once we have snapshots fully working we expect the following flow though:
- validator submit transaction to rotate key
- transaction is accepted and key will be rotated at block N
- validator shutdown node
- validator load new keey
- validator restart, and key is in use from block N

protocol/0000-002KeyManagement.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Vega Key [Event-Forwarder]
+This key (which may be the same of the above one) is used by the event forwarder to authorize events seen on  Vega bridge contracts (at this 
+point, only Ethereum, ERC20, staking and vesting contracts). 
+This key is a hot key, and is constantly used in operations. 
+However, as events signed with this key come in at Ethereum speed, the latency in accessing this key is of little relevance, and it can easily stay in a remote signer or an HSM.
+
+
+
+Compromise of this key is only critical if a significant number of keys are compromised (i.e., 2/3); in this case, it is possible to authorize non-existing events on the Vega chain.
```
jeremyletang: If we are speaking of the ethereum-event-forwarder, the key is really used only between a node and it's eef. But that being compromised, it would lead to the node eeventually emitting wrong event in the vega network.

Also it's probably worth noting that we wish to re-implement the eef as part of the core node, so this key will node be needed anymore.

protocol/0000-002KeyManagement.md
```diff
@@ -0,0 +1,234 @@
...
+   is either a systemic bug, or a cross-validator attack (though of little use to the 
+   attacker if the number is < t+1
+- more than t+1 validators double sign. This is either a critical attack or a critical bug.
+   As the base assumption for the consensus is that less than t+1 parties act malicious, 
+   this should prompt drastic measures, potentially even stopping the chain for an investigation,
+   and at the minimum closing down the MultiSig bridge until the cause is known.
+
+The exact measures and meaning of 'frequently' are still to be done.
+
+The Tenderint key is the only performance critical signing key in the Vega system. This is
```
jeremyletang:
```suggestion
The Tendermint key is the only performance critical signing key in the Vega system. This is
```

protocol/0000-002KeyManagement.md
```diff
@@ -0,0 +1,234 @@
...
+
+- If t+1 validators are outright malicious, they control access to their HSM, and nothing 
+  changes
+- If some validators are compromised, the best case scenario is that they detect this latest
+  on the first double signed message and then cut access to the HSM. If the threshold of
+  malicious+compromised validators is smaller than t+1, this will not lead to a fork; other
+  wise, this could lead to a fork of lenght 2 blocks (with above parameter).
+  Given the severity of that situation (more than a third of the validators compromised), 
+
+As the certitifacion key is 'hot', i.e., needs to be reachable at any time, we also
```
jeremyletang, line 167:
```suggestion
As the certification key is 'hot', i.e., needs to be reachable at any time, we also
```

jeremyletang (approved):

protocol/0000-002KeyManagement.md
Vegaklaus: If I understood Danny right, then all other validators need to remove you, and then add you again. Are these instead single transactions for which I can collect the signatures out of band ?

PR #734, opened 2021-10-28
Change "system" to "network" parameter

candida-d:
Minor change to clarify that what the spec calls a system parameter is a defined network parameter


gordsport:
PR approved and confirmed this is a network param

PR #742, opened 2021-11-08
Remove max delegatable stake and change validator score function

### LINKED ISSUE #740: Current way staking is done is hard to reason about
Current way staking is done is hard to reason about and leads to sub-optimal user experience.  

1) Someone who tries to delegate doesn't know how much stake they ended up putting onto a node until the end of epoch (24h is a long time to wait). Potentially, to delegate the whole associated stake, an unlucky user might have to attempt an unlimited number of days in a row. 

2) We have no way of communicating this in the UI. This is mainly not lack of APIs but the whole design that tries to reconcile staking and un-staking requests only at the end of an epoch. 

A suggested change of approach: process the delegation / un-delegation transactions immediately and on first-come first-served basis (like we do with orders). So we will not be reconciling delegation amounts at epoch start. 

During epoch `n` everything is fixed for calculating rewards at the end of epoch end. If people add stake or remove(*) stake during epoch it won't affect their rewards. 

People can submit staking requests for period `n+1` during epoch `n`. We process those requests to add / remove stake to validators immediately in the order they come in. At the end of period `n` we pay the rewards for `n` as per the amounts fixed at start of `n`. At start of period `n+1` we freeze the end stake of staking and delegation requests that were processed during `n`. These will then determine rewards at the end of `n+1` . And so on. 

(*) There should be an "nuclear" unstake option which will remove stake from a validator if the user has a reason to think the validator is cheating. If this happens remove the stake from the validator during `n` for the purposes of consensus weight (which by the way we don't have implemented) but let whoever delegated still get their rewards. They cannot get any more rewards for this epoch even if the put the removed stake on another validator they can only get rewarded for `n+1`.

davidsiska-vega:
Closes #740


davidsiska-vega:
Btw. you can see here: https://docs.google.com/spreadsheets/d/1Xf7EJJNTBN0MqwCnS2_UwWTfxj7nZDv1CjvDSrrZ-g0/edit?usp=sharing what the function will do in terms of rewards.

davidsiska-vega:
This was reviewed and agreed on the "Research hour" call today. Merging @Vegaklaus , @barnabee .

Vegaklaus (approved):
All ok as discussed on the call.

PR #747, opened 2021-11-11
Updated for free form proposals

### LINKED ISSUE #746: Update governance spec after addition of free form proposals
The new free form proposals have been added to the spec but parts of the original spec no longer make sense due to the the free form proposals not having an enactment and associated governance action. This ticket is to update the spec to make sense for all proposal types.

peterbarrow:
Updated the spec to make things more clear in terms of free form proposals.

Closes #746


wwestgarth:
The spec says for the freeform proposals to not check that the content of the URL matches the hash, which makes sense. Is there any value in validating that the URL atleast _looks_ like a valid URL, even if we don't check that it goes anywhere? 

FWIW, I have no real opinion eitherway.

gordsport:

PR #749, opened 2021-11-16
Document current settlement at expiry oracle behaviour

witgaw:
**MVP behaviour (current)**: If settlement data was received prior to trading termination use the last value received, otherwise use the first value received after trading is terminated 
**Desired behaviour**: Suspend the market if we receive settlement data before trading termination this would require investigation and governance action


gordsport:

PR #752, opened 2021-11-19
Validator performance-based rewards spec

### LINKED ISSUE #654: Rewarding / punishing validator performance / lack of
At the moment if a party is in the validator set it is treated as a validator for the purposes of simple POS rewards. This means that they are eligible for rewards (and those who delegate to them likewise) whether they are participating in running the protocol, producing blocks ...  or not. 

Soon after 💧 Sweetwater we want ways to avoid blatant freeloading.

Vegaklaus:
PM1 is pretty simple and should be implementable like this. (it's also the most burning), PM2-6 need some more discussion.


protocol/000.003.performancemeasurements
```diff
@@ -0,0 +1,56 @@
+
+Perfomance Measurement 1 (PM1): Offline Validator
+Goal: Detect how long a validator is offline and punish them 
+Detection: Validator does not act as a leader
```
davidsiska-vega: @ze97286 do we get the count how many times a node was supposed to be a leader and wasn't directly from Tendermint? 

It's the `f` below.

protocol/000.003.performancemeasurements
```diff
@@ -0,0 +1,56 @@
...
+was elected leader (l).
+
+During the payout, the paymenty is reduced by multipling it with (l-f)/l
+This is done before the delegator part is substracted, i.e., delegators are punished for bad validator behavior.
+
+
+##Future Stuff (in here for discussion purposes, not yet to be implemented)
+# Non Linear punishment
+Currently, PM1 does a linear reduction of payment - maybe an s-curve would be better so that small violations are punished less, 
+while a validator that is offline half the time gets more than 50% substraction. For example, 1/(1+2^((3*x-1)*10)) would punish
```
davidsiska-vega: Maybe something that doesn't involve floating point arithmetic.

davidsiska-vega (commented):
Just the two comments really.

protocol/000.003.performancemeasurements
ze97286: so from what I see, tm emits an event for every round of proposal at a given height, so we know who the proposer should have been. it also emits an event with t timeout for the height round so we can tell that the node wasn't available. 

In addition to what's emitted from the state machine as events there's a bunch of interesting metrics available through the metrics RPC (https://docs.tendermint.com/master/nodes/metrics.html, http://141.94.162.118:26660/metrics). For example, there's amount of blocks missed by a validator, number of validators who did not sign, number of validators who tried to double sign, etc.

I'll play with it later this week or next week.

davidsiska-vega: That's cool. Maybe we can use today's protocol design meeting to look at these and @Vegaklaus may want to refine the calc to include other info? But perhaps we just decide on what the simplest viable metric is and go with that.

witgaw (approved):
LGTM

PR #754, opened 2021-11-23
Update 0010-margin-orchestration.md

Jiajia-Cui:
updated the information when margin is recalculated

### DIFF
--- a/protocol/0010-margin-orchestration.md
+++ b/protocol/0010-margin-orchestration.md
@@ -47,15 +47,15 @@ If risk factors have been updated, the margin levels for all market participants
 
 Recalculate all margins for when any of the following are met:
 
-1. After a new trade
-2. upon leaving an auction
+1. when market observable used in the [margin calculation](0019-margin-calculator.md) changes,
+2. when risk factors are [updated](0065-floating-point-consensus.md).
 
 #### 3. Updating margins when positions have changed
 
 If already re-calculating all margins, don’t need to check for this. Otherwise, recalculate one participant's margins if any of the conditions below are met:
 
 1. If a trader’s net open position changes
-1. If either or both of a traders net potential long or short position (e.g. the sum of all buy order volume and sum of all sell order volume) has changed
+2. If either or both of a traders net potential long or short position (e.g. the sum of all buy order volume and sum of all sell order volume) has changed
 
 
 ### **Utilising margins to evaluate solvency**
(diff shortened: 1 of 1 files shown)


protocol/0010-margin-orchestration.md
```diff
@@ -47,9 +47,8 @@ If risk factors have been updated, the margin levels for all market participants
 
 Recalculate all margins for when any of the following are met:
 
-1. Order book changes (subject to a threshold being exceeded)
-1. If mark price changes
-1. When market observable used in the [margin calculation](./0019-margin-calculator.md) changes
+1. After a new trade
```
davidsiska-vega: Should this be mark price change?

protocol/0010-margin-orchestration.md
```diff
@@ -47,9 +47,8 @@ If risk factors have been updated, the margin levels for all market participants
 
 Recalculate all margins for when any of the following are met:
 
-1. Order book changes (subject to a threshold being exceeded)
-1. If mark price changes
-1. When market observable used in the [margin calculation](./0019-margin-calculator.md) changes
```
davidsiska-vega, line 52: This one should still be there (for the future, for our only product now it doesn't matter).

protocol/0010-margin-orchestration.md
```diff
@@ -47,9 +47,8 @@ If risk factors have been updated, the margin levels for all market participants
...
+1. After a new trade
+2. upon leaving an auction
```
davidsiska-vega: I think leaving an auction updates the mark price so that would be covered by "mark price change".

davidsiska-vega (changes requested):
Please look at my comments and perhaps edit.

protocol/0010-margin-orchestration.md
Jiajia-Cui: No, it should be "after a new trade" because the new mark price can be the same as the old mark price, and margin for all will still be recalculated.

protocol/0010-margin-orchestration.md
Jiajia-Cui: yes, makes sense

protocol/0010-margin-orchestration.md
Jiajia-Cui: Is the default mark price is "0" during an auction?

protocol/0010-margin-orchestration.md
witgaw: New trade could still have the same price as the previous, "if the mark price changes" sounds correct to me.

protocol/0010-margin-orchestration.md
witgaw: Mark price is undetermined during opening auction and it's just the last mark price prior to going into the auction for other auction types.

witgaw (approved):
LGTM

PR #755, opened 2021-11-24
adjust rewards due to reducing the number of iterations. in this case…

ze97286:
… after the first iteration there are 2e-18 more to give but we're stopping


peterbarrow (approved):

davidsiska-vega (commented):
LGTM

PR #761, opened 2021-12-07
Floating-point consensus spec

witgaw:
A spec outlining necessary code changes to assure floating point determinism (so that we don't get a consensus failure due to floating-point computation discrepancies across different platforms).


protocol/0065-floating-point-consensus.md
```diff
@@ -0,0 +1,76 @@
...
+
+If at least one value differs then:
+
+1) Emit an event announcing this (if there are many such events this indicates either an unstable risk calculation or malicious nodes).
+1) A node is chosen at random with probability equal to tendermint weight to propose a bundle of values, submit as candidate and to be voted upon by other nodes. Each node should vote with their tendermint weight. A node will vote `yes` to the proposed bundle if all the values are within the prescribed tolerance and `no` otherwise. 
+The bundle is accepted if `(2/3)` of nodes by tendermint weight vote to accept the bundle. 
+If the bundle is rejected go back to selecting a node at random by tendermint weight and repeat.
+This is repeated until a value has been accepted.
+1) This may continue indefinitely. It will only be terminated when a new event arrives asking for a calculation with new inputs. If we got here emit an event annoucing this (we either have a really badly unstable calculation or malicious nodes).
+1) If update hasn't been achieved after `3` update events an appropriate event should be emitted to indicate that market is operating with stale state variables.
```
jeremyletang, line 58: Should we also start an auction of some sort at this point, alike a market in opning auction which should not start until the value have been received at least once?

protocol/0065-floating-point-consensus.md
```diff
@@ -0,0 +1,76 @@
...
+- [`PriceRange(price, yearFraction, probability num.Decimal) (minPrice, maxPrice num.Decimal)`](https://github.com/vegaprotocol/vega/blob/4be994751b0012b0904e37ad2b0d1540d24abb5e/risk/model.go#L25) - calculates risk minimum and maximum price at specified probability level and projection horizon given current price, called from [price montiroing engine](https://github.com/vegaprotocol/vega/blob/4be994751b0012b0904e37ad2b0d1540d24abb5e/monitor/price/pricemonitoring.go#L80) each time bounds are updated.
+- [`ProbabilityOfTrading(currentP, orderP *num.Uint, minP, maxP, yFrac num.Decimal, isBid, applyMinMax bool) num.Decimal`](https://github.com/vegaprotocol/vega/blob/4be994751b0012b0904e37ad2b0d1540d24abb5e/risk/model.go#L26) - calculates probability of trading at a specified projection horizon of the specified order given current price and min/max bracket, called from [supplied liquidity engine](https://github.com/vegaprotocol/vega/blob/4be994751b0012b0904e37ad2b0d1540d24abb5e/liquidity/engine.go#L34) each time new liquidity provision orders are deployed.
```
jeremyletang, line 82: I'm not sur these 2 rely anymore on actual floats, all calculation on the core have been turned to use decimals if not mistaken.

protocol/0065-floating-point-consensus.md
```diff
@@ -0,0 +1,76 @@
...
+If all candidate values for a given variable are equal to each other then just accept that value. 
+If all values in a bundle are accepted then the whole bundle is accepted and the values in core are updated with the appropriate decimal represenations.
+
+If at least one value differs then:
+
+1) Emit an event announcing this (if there are many such events this indicates either an unstable risk calculation or malicious nodes).
+1) A node is chosen at random with probability equal to tendermint weight to propose a bundle of values, submit as candidate and to be voted upon by other nodes. Each node should vote with their tendermint weight. A node will vote `yes` to the proposed bundle if all the values are within the prescribed tolerance and `no` otherwise. 
+The bundle is accepted if `(2/3)` of nodes by tendermint weight vote to accept the bundle. 
+If the bundle is rejected go back to selecting a node at random by tendermint weight and repeat.
+This is repeated until a value has been accepted.
```
jeremyletang, line 56: Not related to the specs here but to this mecanism. As of now we already have code which have a similar behaviour to verify that resources exists in other blockchains (Witness), the only diffrerence is that it uses 1 vote per node, and they are not weighted. It might makes sense to apply tendermint weight in here as well so this code could be reuse.

jeremyletang:
Sounds good to me, this matches what was discussed in Edinburgh.

jeremyletang:
Should the delay in between 2 calculation of these paramters be configured through a network paramtre of some sort?

protocol/0065-floating-point-consensus.md
davidsiska-vega: We were considering saying an auction should start. But fundamentally working with "stale" parameters is a step better than working with defaults (which is why we don't allow opening auction to end if nothing has been seen but defaults). 

If we have the event emitted it will be an easy change to start an auction if we ever want to change our mind.

protocol/0065-floating-point-consensus.md
davidsiska-vega: Yes, the core is "clean" now but the underlying quant calculations are floats inside out.

protocol/0065-floating-point-consensus.md
davidsiska-vega: 100% 

We probably should apply tendermint weights in Witness too and then see how much code can be reused.

davidsiska-vega:
> Should the delay in between 2 calculation of these paramters be configured through a network paramtre of some sort?

I'll add this network parameter in a minute.

davidsiska-vega:
I've deliberately removed the update on risk parameter change to reduce complexity. 

Everything will get updated every `network.floatingPointUpdates.delay` anyway so after the change is enacted there will be at most that delay. 

I think that's a price worth paying to reduce complexity of implementation and testing.

PR #763, opened 2021-12-09
Data node v2 spec

davidsiska-vega:


gordsport:
Need to add the FE team requirements: https://www.notion.so/vegaprotocol/Core-API-design-ideas-data-node-Requirements-e8d00fbd0c96425687bfa0bc00baf8eb#915a0966aec742eba64aaa0125816c8e

gordsport:
@campbellssource - please can you review this and add the FE requirements to this as its part of Oregon Trail

gordsport:
@pscott31 and @Vegaklaus - to discuss the db choice and how this can work with hashing and update spec requirements if needed

non-protocol-specs/0000-data-node.md
```diff
@@ -0,0 +1,135 @@
...
+  string from_account = 1;
+  // One or more accounts to transfer to
+  string to_account = 2;
+  // An amount to transfer
+  string amount = 3;
+  // A reference for auditing purposes
+  string reference = 4;
+  // Type of ledger entry
+  string type = 5;
+  // Timestamp for the time the ledger entry was created, in nanoseconds since the epoch
```
pscott31: I have a couple of questions about this actually

- What does it mean precisely? Looking in engine.go it looks like the wall time on which ever validator was processing the block.
- Do we want this timestamp *as well* as the block time. And if so which of those makes the most sense to use when asking "Show me ledger entries between x and y"

non-protocol-specs/0000-data-node.md
```diff
@@ -0,0 +1,135 @@
...
+
+Vega core node (consensus and non-consensus) nodes run the core protocol and only keep information required to do so. 
+
+Users of the protcol often need various data (price history / delegation history / transfers etc.). The core node doesn't store these but only *emits events* when things change.
+
+The job of the data node is to collect and store the events and make those available. Since storing "everything forever" will take up too much data it must be possible to configure (and change at runtime) what the data node stores and for how long (retention policy). 
+
+
+## Working with events
+Each chunk of data should contain the eventID that created it and the block from which the event was created.
```
pscott31: It's not a big deal and it can certainly be included, but what is the purpose of storing the eventID? (my only consideration is that it's adding a few tens of bytes to *every single* row, which adds up). The current prototype stores only the timestamp of the block (which is a key into a table of blocks with hashes & heights)

non-protocol-specs/0000-data-node.md
```diff
@@ -0,0 +1,135 @@
...
+- Position mark-to-market events
+
+
+### Candle data
+
+Whatever the candle data are, store them at the resolution of every blockchain time change and build up lower resolution series from that as you see fit. 
+
+### Orders
+
+Store the orders at the configured resolution.
```
pscott31: Is it part of this spec to think about how the data is to be retrieved retrieved as well? Perhaps that's a separate thing to think about

pscott31: There are a few things that we have talked about at various points which aren't in here; such as proof of storage, bootstrapping another node without playing the whole chain etc.. Again maybe they are topics for another day or another spec. Something else to consider is what behaviour to expect when there is restore from a checkpoint (delete everything?) or a snapshot (delete everything with a block height > the snapshot and continue?)

Vegaklaus:
> @pscott31 and @Vegaklaus - to discuss the db choice and how this can work with hashing and update spec requirements if needed

Results: The database will be designed so it can (deterministically) map a random number into a table, and then (also deterministically) be able to go sequentially through the contents of this table (e.g., by timestamp) starting at a point determined by another random number.
The database can also (again deterministically) convert records into an appropriate serialisation that can be fed into a hash function.

@pscott31 does that sum it up correctly ?

pscott31:
Yeah, that's about it Klaus :)

non-protocol-specs/0000-data-node.md
```diff
@@ -0,0 +1,135 @@
+# Data Node Spec
+
+Vega core node (consensus and non-consensus) nodes run the core protocol and only keep information required to do so. 
+
```
Vegaklaus: Is it possible that a core node also is a data node ? Or does it mean 'core protocol', which cannot store more information unless altered ?

non-protocol-specs/0000-data-node.md
```diff
@@ -0,0 +1,135 @@
...
+*Never* use the wall time for anything.
+
+## Retention policies
+
+It should be possible to configure to store only "current state" and no history of anything (in particular the order book). 
+
+It should be possible to configure the data node so that all data older than any time period (e.g. `1m`, `1h`, `1h:22m:32s`, `1 months`) is deleted. 
+
+It should be possible to configure the data node so that all data of certain type is deleted upon an event (and configurably with a delay) e.g. event: MarketID `xyz` settled + `1 week`. 
+
```
Vegaklaus: Is it possible to query the configuration of a data node externally ?
Dependency: The node reward calculation needs to be aware of what configuration a node uses, and there should be different payment models for those three.

Vegaklaus (commented):
Added two questions for clarification.

davidsiska-vega:
The data node v2 must support the following workflow:
1) a vega network is running
2) a core node joins (as non-validator) by restoring state from a snaphot
3) a new data node can be spun up that obtains full history from other data nodes (get addresses from the non-validator node and then download) and finally start receiving events from "its own" non validator node.

campbellssource:
@gordsport forgive this PR into a PR https://github.com/vegaprotocol/specs-internal/pull/886
I thing this spec is too general to add all of the FE requirements to.
As mentioned, we need to figure out a process for specifying each FE requirement on data node. At the moment we have this list in [notion](https://www.notion.so/vegaprotocol/05242ea50fdf42e4921199987a8d3b68?v=83de91b31d324740bef80214ed0eef45) under the `new from front end` column. Either the Q2 objectives get agreed and we write a mini-spec for each Data node augmentation or we just have data node capacity to work in real time with Front end team (as well as other data node users)

davidsiska-vega:
If we get this working https://github.com/vegaprotocol/quant/issues/22 then there may be a way of using the `quant` libraries we have (or with small extension) on client side.

non-protocol-specs/0000-data-node.md
gordsport: Datanode can be run independently

non-protocol-specs/0000-data-node.md
gordsport: @jeremyletang - can you follow up on this?

non-protocol-specs/0000-data-node.md
davidsiska-vega: Let's make it clear that it must be possible to query the data retention policy of a data node and also to what extent it has full set of data (subject to its retention policy).

non-protocol-specs/0000-data-node.md
davidsiska-vega:
```
// Timestamp for the time the ledger entry was created
``` 
This must be the block time, everything is block time. So there is only one timestamp which is the block time.

non-protocol-specs/0000-data-node.md
jeremyletang: @pscott31 I guess the eventID is not much useful, what we will want is what's called the traceID I think which basically is the transaction hash. By doing so, we could query from the database all data / events resulting from a given transaction.

pscott31: Cool - we're already storing the traceid/hash in the 'block' table

```sql
create table blocks
(
    vega_time     TIMESTAMP WITH TIME ZONE NOT NULL PRIMARY KEY,
    height        BIGINT                   NOT NULL,
    hash          BYTEA                    NOT NULL
);
```

Every row in every table joins onto this via `vega_time`

non-protocol-specs/0000-data-node.md
pscott31: I mean yes, the database knows what it's retention policy is and we can add some API for querying that :)

pscott31 (dismissed):

gordsport:
Hi @jeremyletang @pscott31 
I have renamed this file and created a ticked to review this spec with the aim to add acceptance criteria:
- https://github.com/vegaprotocol/specs-internal/issues/999

If you are OK with the naming please can you give it a review and merge?

gordsport (approved):

PR #766, opened 2021-12-10
Validators joining and leaving

davidsiska-vega:
Joint work with Jeremy; good use of research offsite time ;-).


gordsport:
Hey @Vegaklaus  @vegaprotocol/core @C0deMunk33 - please can you review this spec as this will be part of Oregon Trail work.

@vega-paul - ca you please review and review the acceptance criteria

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+
+Basic vega chain liveness criteria is covered in their [performance score](0064-validator-performance-based-rewards.md). 
+
+Verifying Ethereum (and later other chain) integration
+1) They will be the first node to forward a subsequently accepted ethereum event at least `validator.minimumEthereumEventsForNewValidator` with a default of `3`. 
+1) They are the first one to vote for any ethereum event at least `validator.minimumEthereumEventsForNewValidator` times. 
+
+## Multisig weight updates WIP
+
+This section might make more sense in 0061-simple-POS-rewards-SweetWater.md - TBD.
```
edd: **Thought (non-blocking)**: Should this go before merging it? I’m indifferent - it’s easy to fix.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+Thus a validator who is not there but should be has incentive to pay gas to update the multisig. Moreover a validator who's score has gone up substantially will want to do so as well. 
+
+Note that this could become obsolete if we go to threshold signatures. 
+
+
+# Acceptance criteria
+
+## Joining / leaving VEGA chain
+1) A running non-validator node can submit a transaction to become a validator. 
+2) Their perfomance score will be calculated.
```
edd: **Nitpick (blocking)**: Link this to 0064 where the calculation is defined

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+3) If they meet the Ethereum verification criteria and have enough stake they will become part of the validator set at the start of next epoch.
```
edd: **Nitpick (non-blocking)**: Also worth adding a heading on line 49 and linking this acceptance criteria to it

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+Note that this could become obsolete if we go to threshold signatures. 
+
+
+# Acceptance criteria
+
+## Joining / leaving VEGA chain
+1) A running non-validator node can submit a transaction to become a validator. 
+2) Their perfomance score will be calculated.
+3) If they meet the Ethereum verification criteria and have enough stake they will become part of the validator set at the start of next epoch. 
+4) The node that got "pushed out" will no longer be a validator node for Tendermint.
```
edd: **Nitpick (non-blocking)**: I know it’s in point three above that this is happening at the next epoch change, but this would be clearer if point 4 started with ‘After the epoch changeover is complete…’

```suggestion
4) After the epoch change is complete, the node that got "pushed out" will no longer be a validator node for Tendermint. 
```

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
+# Validators chosen by stake
+
+New network parameter `network.numberOfValidators`. 
+
+At a high level a participant that wishes to become a validator will:
+1) start a Vega node as non-validating node + associated infra 
+1) submit a transaction, see below for details, with their keys, saying they want to validate.
+1) self-stake to their validator Vega key at least `reward.staking.delegation.minimumValidatorStake`.
```
ze97286: they can't really do that before they are accepted as validators can they?

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+## Joining / leaving VEGA chain
+1) A running non-validator node can submit a transaction to become a validator. 
+2) Their perfomance score will be calculated.
+3) If they meet the Ethereum verification criteria and have enough stake they will become part of the validator set at the start of next epoch. 
+4) The node that got "pushed out" will no longer be a validator node for Tendermint. 
+
+## Multisig update
+1) Vega network receives the ethereum events updating the weights and stores them (`key`,`value`). 
+2) For validators up to `number_multig_signers` the `validator_score` is capped by the value on `ethereum`, if available and it's `0` for those who should have value on Ethereum but don't (they are one of the top `number_multig_signers` by `validator_score` on VEGA). 
+3) It is possible to submit a transaction to update the weights.
```
edd: **Question (blocking)**: What happens in the situation where the validators set changes (as per 70) but this transaction has not been sent? Does the epoch changeover still happen?

**Question (blocking)**: Related to the above, what happens in in a validators set of 13, 8 (or any 2/3+1 number) new validators propose themselves and have stake to replace existing validators? Would that stop withdrawals being signed? For how long?

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
+# Validators chosen by stake
+
+New network parameter `network.numberOfValidators`. 
+
+At a high level a participant that wishes to become a validator will:
+1) start a Vega node as non-validating node + associated infra
```
ze97286: at which point do they restart as validators?

protocol/0000-validators-chosen-by-stake.md
barnabee: this should be possible at any time but they wouldn't be paid any rewards (nor would anyone else staking to them) unless they are also in the validator set

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+At a high level a participant that wishes to become a validator will:
+1) start a Vega node as non-validating node + associated infra 
+1) submit a transaction, see below for details, with their keys, saying they want to validate.
+1) self-stake to their validator Vega key at least `reward.staking.delegation.minimumValidatorStake`. 
+1) wait for others to delegate to them. 
+
+At the end of each epoch Vega will choose the validators with the `validator_score`, see [rewards spec](0061-simple-POS-rewards-SweetWater.md) up to `network.numberOfValidators`. Note that this number combines own + delegated stake together with `performance_score` which measures basic node performance. A completely dead node will have `performance_score = 0` and will thus get automaticaly excluded, regardless of their stake.
+
+These will be the validating nodes for the next epoch. Note that to be eligible certain criteria need to be met: 
+1) Own stake >= `reward.staking.delegation.minimumValidatorStake`.
```
ze97286: does this apply for validators in genesis as well?

protocol/0000-validators-chosen-by-stake.md
barnabee: would they need to do this or could Tendermint switch "on the fly" once they are accepted?

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+
+New network parameter `network.numberOfValidators`. 
+
+At a high level a participant that wishes to become a validator will:
+1) start a Vega node as non-validating node + associated infra 
+1) submit a transaction, see below for details, with their keys, saying they want to validate.
+1) self-stake to their validator Vega key at least `reward.staking.delegation.minimumValidatorStake`. 
+1) wait for others to delegate to them. 
+
+At the end of each epoch Vega will choose the validators with the `validator_score`, see [rewards spec](0061-simple-POS-rewards-SweetWater.md) up to `network.numberOfValidators`. Note that this number combines own + delegated stake together with `performance_score` which measures basic node performance. A completely dead node will have `performance_score = 0` and will thus get automaticaly excluded, regardless of their stake.
```
barnabee: I assume this should be the `network.numberOfValidators` validators with the _highest_ `validator_score`?

ze97286: do validators that are not in the current validator set (as calculated in the previous epoch's end) get rewards?

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+
+From now we assume that the transaction has been submitted and the node started. 
+
+## Minimum performance criteria
+
+Basic vega chain liveness criteria is covered in their [performance score](0064-validator-performance-based-rewards.md). 
+
+Verifying Ethereum (and later other chain) integration
+1) They will be the first node to forward a subsequently accepted ethereum event at least `validator.minimumEthereumEventsForNewValidator` with a default of `3`. 
+1) They are the first one to vote for any ethereum event at least `validator.minimumEthereumEventsForNewValidator` times.
```
barnabee: What does this mean in practice? Is the idea that if the node is too slow they will find it hard to win this "lottery"?

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+1) They will be the first node to forward a subsequently accepted ethereum event at least `validator.minimumEthereumEventsForNewValidator` with a default of `3`. 
+1) They are the first one to vote for any ethereum event at least `validator.minimumEthereumEventsForNewValidator` times. 
+
+## Multisig weight updates WIP
+
+This section might make more sense in 0061-simple-POS-rewards-SweetWater.md - TBD.
+
+Once (if) the ethereum multisig contract supports validator weights the vega node will watch for Ethereum events announcing the weight changing. Thus for each validator that is on the multisig contract it will know the validator score (weight) the ethereum multisig is using. 
+
+We will have `number_multig_signers` represented on the multisig (currently `13`) but this could change.
```
barnabee: Does this need to clarify that it will be the `number_multisig_signers` with the highest `validator_score` (or perhaps `ethereum_multisig_val_score`) or something?

Also typo: "multig" in several of these param/variable names (here and below)

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+
+Once (if) the ethereum multisig contract supports validator weights the vega node will watch for Ethereum events announcing the weight changing. Thus for each validator that is on the multisig contract it will know the validator score (weight) the ethereum multisig is using. 
+
+We will have `number_multig_signers` represented on the multisig (currently `13`) but this could change. 
+
+In the reward calculation for the top `number_multig_signers` by `validator_score` (as seen on VEGA) use `min(validator_score, ethereum_multisig_val_score)` when calculating the final reward with `0` for those who are in the top `number_multig_signers` by score but *not* on the multisig contract. 
+
+Thus a validator who is not there but should be has incentive to pay gas to update the multisig. Moreover a validator who's score has gone up substantially will want to do so as well. 
+
+Note that this could become obsolete if we go to threshold signatures.
```
barnabee: "we go to threshold signatures" → "a future version of the protocol implements threshold signatures or another method that allows all validators to approve Ethereum actions"

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+## Running a candidate non-validator node
+Start [node as a validator node](https://github.com/vegaprotocol/networks/blob/master/README.md).
+
+From now we assume that the transaction has been submitted and the node started. 
+
+## Minimum performance criteria
+
+Basic vega chain liveness criteria is covered in their [performance score](0064-validator-performance-based-rewards.md).
```
ze97286: performance score here is a bit tricky here - if a node is not in the current tm validator set their performance score is by definition 0. In that case do we prioritise a node that is not in the validator set currently over a node that is and has 0 performance score?

protocol/0000-validators-chosen-by-stake.md
barnabee: Re: second question I think this is a good point.

I think we discussed in the past something like a network param for the maximum number validators that can change (or leave the set) in a given epoch, so that for example only the highest scoring _n_ (where _n_ may be as low as 1 or 2) newly eligible validators would actually become a validator in each epoch.

This would seem sensible for this and to prevent weird "flash" type attacks on the validator set...

barnabee (dismissed):

protocol/0000-validators-chosen-by-stake.md
jeremyletang: yes.

pscott31:
I am sure you guys have thought about this much harder than I have, but here's my tuppence' worth anyway :)

What's the motivation behind allowing people to join just by staking `minimumValidatorStake` and behaving, rather than by vote? 

The reason I ask is because (assuming I have understood this correctly) doesn't that make it relatively easy to take over the network? If I can afford to start and stake 2n new nodes (n being number of existing validators), I now own the network.

I guess it all depends on how high minimumValidatorStake is :)

barnabee:
> I am sure you guys have thought about this much harder than I have, but here's my tuppence' worth anyway :)
> 
> What's the motivation behind allowing people to join just by staking `minimumValidatorStake` and behaving, rather than by vote?
> 
> The reason I ask is because (assuming I have understood this correctly) doesn't that make it relatively easy to take over the network? If I can afford to start and stake 2n new nodes (n being number of existing validators), I now own the network.
> 
> I guess it all depends on how high minimumValidatorStake is :)

You can't join by having `>= minimumValidatorStake`, only "apply" to join. 

You then need to be one of the _n_ validators with the most total stake (or actually _n_ highest scores, where score includes stake plus some modifiers), where _n_ is `network.numberOfValidators`, so you either need a lot of stake or a lot of others with stake to have staked it on you (this is the delegated in delegated proof of stake). So it is kind of a vote, though it's a continuous one rather than a governance proposal.

You can't also control/take over the network just by controlling more nodes (that is a Sybil attack and one of the key attacks that BFT systems need to be designed to mitigate). Each node does not have the same voting power, again it is their score that determines the node's voting power and that's mostly a function of stake. So even if noone else had more stake/higher scores and you end up running 2n nodes at `minimumValidatorStake` you'd have a tiny fraction of the votes and still not be able to control/corrupt the network (unless this allowed for some liveness attack, which it might, assuming noone else wanted to run nodes and it was cheap enough to acquire enough slots).

protocol/0064-validator-performance-based-rewards.md
```diff
@@ -16,7 +16,16 @@ was elected leader `l` and carried out the job.
 To be more precise: Tendermint emits an event for every round of proposal at a given height, so we know who the proposer should have been. It also emits an event with a timeout for the height of the round so we can tell that the node wasn't available. 
 This counts need to be stored, per validator node, in vega core. 
 
-Define the performance score to be `performance_score := (l-f)/l` if `l > 0` and `performance_score = 0` if `l = 0`. 
+For validators participating in consensus define the performance score to be `performance_score := (l-f)/l` if `l > 0` and `performance_score = 0` if `l = 0`.
```
barnabee: As @ze97286 pointed out to me, I think we may need to make `performance_score := 1` for potential new validators (unless they have a recent, previous lower score?) so that they can be ranked properly when deciding if any of them are to join the validator set

pscott31:
Thanks Barney, that was educational :)

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,79 @@
...
+
+Verifying Ethereum (and later other chain) integration
+1) They will be the first node to forward a subsequently accepted ethereum event at least `validator.minimumEthereumEventsForNewValidator` with a default of `3`. 
+1) They are the first one to vote for any ethereum event at least `validator.minimumEthereumEventsForNewValidator` times. 
+
+## Multisig weight updates WIP
+
+This section might make more sense in 0061-simple-POS-rewards-SweetWater.md - TBD.
+
+Once (if) the ethereum multisig contract supports validator weights the vega node will watch for Ethereum events announcing the weight changing. Thus for each validator that is on the multisig contract it will know the validator score (weight) the ethereum multisig is using.
```
ze97286: how would that work with zk rollups?

protocol/0000-validators-chosen-by-stake.md
wwestgarth: Its probably quite unlikely, but for completeness, in the case where there is a tie for which validator has the `nth` highest score (i.e one will make it into the set and one won't) is the validator to chosen pseudo-randomly, or is there some preference towards introducing a new validator to the set?

And I guess in general, is the desireable outcome that over time we get a constant rotation of different validators in the set each epoch, or are we expecting it to remain mostly static? or do we simply not care :p

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: Yes. I'll make the edit to clarify.

protocol/0064-validator-performance-based-rewards.md
davidsiska-vega: I am not sure whether you're making this comment having read to the end of this file or not. We specifically have a score for those wishing to join defined below. If you think there's a better / more robust way let's put that down instead.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: Indeed, they can self stake and start accepting delegations but unless they become a validator no reward will go to them and those who delegated to them.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: I was hoping they won't have to restart (because of the associated delay + a chance for a screw up).

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: Yes, I'll fix that.

davidsiska-vega: Those not in the validator set get nothing.

davidsiska-vega: I'll fix that. In case of a tie we favour the incumbent.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: Yes, the fact they were in genesis shouldn't matter. Though there is a little thinking needed about how the network can get restarted in this case.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: We cover that in 0064 for those who are not in the validator set.

davidsiska-vega: Add comments there if you think this is unworkable or if there is a better way to go around this.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: I hope I'll know more after Monday 10th research hour.

protocol/0000-validators-chosen-by-stake.md
ze97286: they won't if they start as vega as a validator (by which I mean if they start with flag indicating they want to be a validator, not that they are actually selected to be ones)

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: The point of this section is to incentivise those who became validators to update the multisig contract (pay gas) so that they're included and the score used for their reward is the correct one.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: 👍

protocol/0000-validators-chosen-by-stake.md
ze97286: is self staking an ongoing requirement? or just a one off check? I'm asking because currently validators don't have to self stake. If they don't, they don't get rewarded but their delegators do.

protocol/0000-validators-chosen-by-stake.md
ze97286: but I guess their stake is counted for the total stake i.e. used for calculating validator score?

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: ok

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: ok

protocol/0000-validators-chosen-by-stake.md
ze97286: I don't think 0064 (the non future part) talks about validators not in the validator set, maybe I missed it?

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: I'd leave it... but then I am slack.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: It's on this branch from line 21 (though @barnabee also seemed to miss it so maybe I messed up with git):

> For validators who [have submitted a transaction to become validators](0000-validators-chosen-by-stake.md) the `performance_score` is defined as follows: during each epoch
Every `1000` blocks the candidate validator node is to send a hash of block number `b` separetely signed by all the three keys and submitted; the network will verify this to confirm that the validator owns the keys. 
Here `b` is defined as:
First time it is the the block number in which the joining transaction was included. Then it's incremented by `1000`. 
The network will keep track of the last `10` times this was supposed to happen and the `performance_score` is the number of times this has been verified divided by `10`.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: Reply to 1st question (the easy one): if they've not submitted a transaction saying they want to validate their score isn't being calculated (so it's 0) and they cannot, by definition, push any other validator out.

protocol/0064-validator-performance-based-rewards.md
```diff
@@ -16,7 +16,16 @@ was elected leader `l` and carried out the job.
 To be more precise: Tendermint emits an event for every round of proposal at a given height, so we know who the proposer should have been. It also emits an event with a timeout for the height of the round so we can tell that the node wasn't available. 
 This counts need to be stored, per validator node, in vega core. 
 
-Define the performance score to be `performance_score := (l-f)/l` if `l > 0` and `performance_score = 0` if `l = 0`. 
+For validators participating in consensus define the performance score to be `performance_score := (l-f)/l` if `l > 0` and `performance_score = 0` if `l = 0`. 
+
+For validators who [have submitted a transaction to become validators](0000-validators-chosen-by-stake.md) the `performance_score` is defined as follows: during each epoch
+Every `1000` blocks the candidate validator node is to send a hash of block number `b` separetely signed by all the three keys and submitted; the network will verify this to confirm that the validator owns the keys.
```
wwestgarth: Is there any plan for this `1000` to be configurable? 

Also I wondering in reality what this hoop-jumping actually shows. I'm guessing the intention is to prove some sort of activity on the network and thus investment in being part of the network, but in reality isn't sending in all these transactions just going to be someone hitting play on a script a couple of hours before an epoch ends?

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: For restarts:
- at each checkpoint we include node IDs of validators and their scores (meaning all the ones participating in consensus and those who submitted a transaction to become a validator) 
- when initiating the restart all the nodes participating have the same Tendermint weight in genesis and this is used until the LNL file has finished processing
- when loading LNL file we have to run the same algorithm that selects the "correct" validators; after this is done Tendermint weights are updated.

protocol/0064-validator-performance-based-rewards.md
davidsiska-vega: We discussed seetting `performace_score = 1` if a node never gets asked to lead.

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: DIscussion in research hour: The main thing is making sure that after a validator is removed from the validating set they may no longer run the node (or become malicious in the worst case). If that happens they won't sign the transaction to update the multisig signer set. So if we remove too many (more than 1/3 at the moment) validators without updating the multisig contract we may lose control there. 

A possible solution is e.g. that:
- we only allow one validator change per epoch *and*
- anyone wishing to become a validator must post Ethereum for gas somewhere so that the network can run the multisig update (so it's guaranteed to happen) when the change is mage.  

But this hasn't been discussed and I am hoping for other, better, ideas.

protocol/0064-validator-performance-based-rewards.md
davidsiska-vega: I agree. You could of course do that. But why? It should be simpler to just run the vega node at this point. If you cheat here with a script and then get selected for the next epoch and you cannot propose any tendermint blocks you'll be kicked out at the end of the epoch. The people who staked their tokens to you will be pissed off and won't do it again.

It's just meant as a basic sanity check more to help honest people not to make mistakes rather than prevent outright cheating.

davidsiska-vega: We can make the `1000` a network parameter; but bear in mind that at the moment it's easier to change the code and give the update to the validators than to run successful governance vote.

wwestgarth: Ah right, I completely missed the bit where it said "candidate validator node" sends the transaction, so the idea is that code will be added in core to do this.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,94 @@
...
+1) submit a transaction, see below for details, with their keys, saying they want to validate.
+1) self-stake to their validator Vega key at least `reward.staking.delegation.minimumValidatorStake`. 
+1) wait for others to delegate to them. 
+
+Note that to be eligible as a potential validator certain criteria need to be met: 
+1) Own stake >= `reward.staking.delegation.minimumValidatorStake`. 
+1) Network has verified key ownership (see below).
+
+
+At the end of each epoch Vega will calculate `validator_score`, see [rewards spec].
```
davidsiska-vega: Clarify here that this is the un-normalised score.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,94 @@
...
+
+
+At the end of each epoch Vega will calculate `validator_score`, see [rewards spec]. 
+For validators currently in the Vega validator set it will scale the `validator_score` by `(1+see [rewards spec])`. 
+Note that this number combines own + delegated stake together with `performance_score` which measures basic node performance togther whether the multisig contract carries the correct information [multisig](0030-multisig_control_spec.md); more on this later.
+
+Vega will sort all current validators as `[v_1, ..., v_n]` with `v_1` with the highest and `v_n` with the lowest score. 
+If `v_l = v_m` then we place higher the one who's been validator for longer.
+Vega will sort all those who submitted a transaction wishing to be validators using `validator_score` as `[w_1, ..., w_k]`. 
+If `w_1>v_n` (i.e. the highest scored potential validator has more than the lowest score incumbent validator) then in the new epoch `w_1` becomes a Tendermint validator. If `w_l = w_m` then we resolve this by giving priority to the one who submitted the transaction to become validator earlier.
```
davidsiska-vega: Clarify that if there are empty slots then more than one can join in the epoch.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,94 @@
+# Validators chosen by stake
+
+New network parameter `network.numberOfValidators`. 
+New network parameter `network.validatorIncumbentBonus`.
+New network parameter `network.numberMultisigSigners`
```
davidsiska-vega: Clarify `network.numberMultisigSigners \leq network.numberOfValidators`.

protocol/0061-simple-POS-rewards-SweetWater.md
```diff
@@ -18,11 +18,16 @@ This applies both the rewards coming from the [on-chain-treasury](0055-on-chain-
 - `reward.staking.delegation.maxPayoutPerParticipant` - the maximum (applies only to on-chain tresaury rewards in the form of thestaking and governance asset) that each participant may receive as a payout from single epoch. 
 - `reward.staking.delegation.maxPayoutPerEpoch` - the maximum (applies only to on-chain tresaury rewards in the form of thestaking and governance asset) that may be paid out for a given epoch.
 
+
 **Note**: changes of any network parameters affecting these calculations will take an immediate effect (they aren't delayed until next epoch).
 
 # Calculation
 This applies to the on-chain-treasury for each asset as well as network infrastructure fee pool for each asset. 
 
+As step *zero*: if the number of validators on the Vega chain is lower than the number of signers on the [multisig ethereum contract](0030-multisig_control_spec.md) then the reward is zero.
```
davidsiska-vega: This actually needs to check that all the validators that are on multisig are vega validators and if not, give everyone zero.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Verifying Ethereum (and later other chain) integration
+1) They will be the first node to forward a subsequently accepted ethereum event at least `validator.minimumEthereumEventsForNewValidator` with a default of `3`. 
+1) They are the first one to vote for any ethereum event at least `validator.minimumEthereumEventsForNewValidator` times. 
+
+## Multisig updates (and multisig weight updates if those are used)
+
+Once (if) the ethereum multisig contract supports validator weights the vega node will watch for Ethereum events announcing the weight changing. 
+Thus for each validator that is on the multisig contract it will know the validator score (weight) the ethereum multisig is using. 
+Vega node will watch for multisig signer changes.
```
davidsiska-vega: @C0deMunk33 Danny, can you check whether the current multisig contract emits all the events we need for this to work?

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -92,3 +92,34 @@ Note that this could become obsolete if a future version of the protocol impleme
...
+#Performance
+Ersatzvalidators are required to monitor the primary chain and keep an upded state of
+Vega at all times. Any performance measurements that relate to Validators
+being required to keep an accurate stake also apply to Ersatzvalidators.
+
+As Ersatzvalidators are not part of the Tendermint chain, their network performance
+on that chain cannot be measured. For this reason, Ersatzvalidators are supposed to
+participate in a separate chain (an exact copy of Vega with run only by the Ersatzvalidators
+that does not contain any trading), which generates the applicble performance
+numbers.
```
davidsiska-vega: We have a way of measuring performance of validators not running the tendermint consensus (but are non-validator nodes); I think that may be easier than what you're proposing here.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -92,3 +92,34 @@ Note that this could become obsolete if a future version of the protocol impleme
 1) Vega network receives the ethereum events updating the weights and stores them (`key`,`value`). 
 2) For validators up to `number_multig_signers` the `validator_score` is capped by the value on `ethereum`, if available and it's `0` for those who should have value on Ethereum but don't (they are one of the top `number_multig_signers` by `validator_score` on VEGA). 
 3) It is possible to submit a transaction to update the weights. 
+
+## Ersatzvalidators
+New Network Parameter: NumberOfErsatzValidators
+New Network Parameter: MinmumStakeForErsatzValidator
```
davidsiska-vega: I wouldn't introduce a new parameter; we want this to be the same as for real validators so they are really ready to jump in.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -92,3 +92,34 @@ Note that this could become obsolete if a future version of the protocol impleme
...
+on that chain cannot be measured. For this reason, Ersatzvalidators are supposed to
+participate in a separate chain (an exact copy of Vega with run only by the Ersatzvalidators
+that does not contain any trading), which generates the applicble performance
+numbers.
+
+#Payment
+Ersatzvalidators are paied through the same formulars as normal validators.
+
+#Multisig
+At this point, Ersatzvalidators are not part of the Multisig.
```
davidsiska-vega: I'd say they'll never be part of multisig - simpler.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -92,3 +92,34 @@ Note that this could become obsolete if a future version of the protocol impleme
...
+being required to keep an accurate stake also apply to Ersatzvalidators.
+
+As Ersatzvalidators are not part of the Tendermint chain, their network performance
+on that chain cannot be measured. For this reason, Ersatzvalidators are supposed to
+participate in a separate chain (an exact copy of Vega with run only by the Ersatzvalidators
+that does not contain any trading), which generates the applicble performance
+numbers.
+
+#Payment
+Ersatzvalidators are paied through the same formulars as normal validators.
```
davidsiska-vega: Please add reference. Also "formulars" -> "formulas" and "paied" -> "paid". Also should "Payment" be "Rewards" we call it rewards elsewhere.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -92,3 +92,34 @@ Note that this could become obsolete if a future version of the protocol impleme
 1) Vega network receives the ethereum events updating the weights and stores them (`key`,`value`). 
 2) For validators up to `number_multig_signers` the `validator_score` is capped by the value on `ethereum`, if available and it's `0` for those who should have value on Ethereum but don't (they are one of the top `number_multig_signers` by `validator_score` on VEGA). 
 3) It is possible to submit a transaction to update the weights. 
+
+## Ersatzvalidators
```
davidsiska-vega: Should we call them something like "potential validators" or "spare validators"? Trying to stick to english in the spec.

protocol/0000-validators-chosen-by-stake.md
jeremyletang: I'm gonna say it, there's no way we build a separate chain now just to evaluate "validators to be" performances. Let's try to deliver 1 chain which works perfectly first.

protocol/0000-validators-chosen-by-stake.md
gordsport: suggestion to change `numberMultisigSigners` to `numberOfEthMultisigSigners`?

protocol/0000-validators-chosen-by-stake.md
gordsport: Add space between words  in the title

protocol/0000-validators-chosen-by-stake.md
gordsport: Should be `MaximumNumberOfErsatzValidators`

davidsiska-vega: Let's change NumberOfErzatzValidators to "MultipleOfTendermintValidators" with suggested value of `1` and anything between `0.5` to `2` probably making sense. Se we will calculate on the fly 
```
maxNumberOfErzatzValidators := ceil(MultipleOfTendermintValidators x NumberOfTendermintValidators)
```

protocol/0000-validators-chosen-by-stake.md
C0deMunk33: it does emit `signer added` `signer removed` events


I don't know how ethereum_multisig_val_score gets calculated, seems to me this is delegated $VEGA which is a Vega side thing

other than that I believe this is all possible with the current setup of signature bundles and add/remove signer

davidsiska-vega: So with the emitted `signer added` and `signer removed` Vega can let the ethereum forwarder to replay the entire history and know who's keys are there at any time. That's ok. 

If (when) we add weights we will need to emit events for all the weight changes.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,143 @@
...
+
+Vega will sort all current Tendermint validators as `[v_1, ..., v_n]` with `v_1` with the highest and `v_n` with the lowest score. 
+If `v_l = v_m` then we place higher the one who's been validator for longer.
+Vega will sort all those who submitted a transaction wishing to be validators using `validator_score` as `[w_1, ..., w_k]`. 
+These may be ersatz validators (ie getting rewards) or others who just submitted the transaction to join.
+If `empty_slots := network.numberOfTendermintValidators - n > 0` (we have empty tendermint validator slots) then the top `empty_slots` from `[w_1, ..., w_k]` are promoted to tendermint validators. 
+If `w_1>v_n` (i.e. the highest scored potential validator has more than the lowest score incumbent validator) then in the new epoch `w_1` becomes a Tendermint validator, and the lowest scoring incubent becomes an ersatz validator. If `w_l = w_m` then we resolve this by giving priority to the one who submitted the transaction to become validator earlier.  Note that we only do this check once per epoch so at most one validator can be changed per epoch in case `empty_slots == 0`.
```
davidsiska-vega: What is w_l? it's the `l` that's not clear.

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,143 @@
...
+A completely dead node that's proposing to become a validator will have `performance_score = 0` and will thus get automaticaly excluded, regardless of their stake.
+
+The same way, if there are free slots for ersatz validators and nodes that have submitted the transaction to join and satisfy all joining conditions, they are added as ersatzvalidators in the next round.
+If a node that submitted the transaction to join and satisfies all other conditions and there and has a higher score than the lowest scoring ersatz validator (scaled up by the incubent factor), then (assuming it did not just become a Tendermint validator), it becomes an ersatz validator and the lowest scoring ersatz validator is kicked out. The 'transaction to join' of a validator kicked out this way remains active until the delegated stake drops below the required minimum. As the nodes have not have the opportunity to get a performance record, their performance valued as the average of the performance scores of all ersatzvalidators.
```
davidsiska-vega: [Comment (kku): I'm not happy with this, as it meansthat someone with a near zero performance can make the request to join, get in, perform poorly, be kicked out, wait an epoch, get in again etc. This still seems a better way than doing something really complicated here. One other scenario is that a validator stops operating, but keeps having some stake; this one would then linger around as a zombie (and potentially be promoted from time to time) forever. Thus, it might be better to expire the 'transaction to join' at some point (?)]

[Comment (kku): Didn't we want the optional mechanism that one can become ersatzvalidator based on a minimum amount of stake and have an open number ?]

protocol/0061-simple-POS-rewards-SweetWater.md
```diff
@@ -69,4 +86,20 @@ function validatorScore(valStake) {
   return linearScore
 ```
 
+For ersatz validators, the formular changes slightly:
+```
+ linearScore = (valStake)/s_total
+ linearScore = Math.min(1.0, Math.max(0.0,linearScore))
+```
+i.e., there is no anti-whaling function applied here (the penaltys are removed)
```
davidsiska-vega:
```suggestion
i.e., there is no anti-whaling function applied here (the penalties are removed)
```

protocol/0064-validator-performance-based-rewards.md
```diff
@@ -16,7 +16,17 @@ was elected leader `l` and carried out the job.
...
+For validators participating in consensus define the performance score to be `performance_score := max[(l-f)/l, 0.05]` if `l > 0` and `performance_score = 1` if `l = 0`. 
+Flooring the score at `0.05` is there to make sure that every validator with non-zero own+delegated gets a chance to be a leader at least ocassionaly, even if they were poorly performing recently. 
+
+For validators who [have submitted a transaction to become validators](0000-validators-chosen-by-stake.md) the `performance_score` is defined as follows: during each epoch
+Every `1000` blocks the candidate validator node is to send a hash of block number `b` separetely signed by all the three keys and submitted; the network will verify this to confirm that the validator owns the keys. 
+Here `b` is defined as:
+First time it is the the block number in which the joining transaction was included. Then it's incremented by `1000`. 
+The network will keep track of the last `10` times this was supposed to happen and the `performance_score` is the number of times this has been verified divided by `10`.  
+
```
davidsiska-vega:
```suggestion
Initially the performance score is set to `0`.
```

protocol/0000-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,139 @@
...
+        "id": "126751c5830b50d39eb85412fb2964f46338cce6946ff455b73f1b1be3f5e8cc",
+        "vega_pub_key": "a6e6f7daf8610f9242ab6ab46b394f6fb79cf9533d48051ca7a2f142b8b700a8",
+        "ethereum_address": "0x14174f3c9443EdC12685578FE4d165be5f57fBd3",
+        "tm_pub_key": "0ShTSZ9Ss8AFHuDz1nIGMrGucjXhFdJyUTT7Eqibjq8=",
+        "info_url": "https://www.greenfield.one",
+        "country": "Germany",
+        "name": "Greenfield One",
+        "avatar_url": "https://www.greenfield.one/avatar.png"
+```
+
```
davidsiska-vega:
```suggestion
## Removing non-performing candidate validators

- Any party with fewer than `reward.staking.delegation.minimumValidatorStake` at the start of an epoch is removed. (the aim is to remove parties who don't delegate to self, however we achieve it is not too important)
- Any party with performance score == `0` at the end of an epoch is removed. 



```

protocol/0000-validators-chosen-by-stake.md
davidsiska-vega: Updated

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
```diff
@@ -18,15 +19,32 @@ This applies both the rewards coming from the [on-chain-treasury](./0055-TREA-on
...
 
+## Tendermint Validators and Ersatz validators
+ In Vega, we have two sets of Validtors, the primary validators (which run tendermint) and the [ersatz validators](0000-validators-chosen-by-stake.md) (which are running a non-validator node and can be promoted to a validator node by the protocol if they meet the right criteria). 
+ Both these validators get rewards following the method above:
+ 1. The reward pool is split into two parts, propotional to the total own+delegated stake the primary- and ersatzvalidators have. 
+ Thus, if `s_t = network.ersatzvalidators.reward.factor x s_e + s_t` is the total amount of own+delegated stake to both sets (with ersatz scaling taken into account), `s_p` the total stake delegated to the primary / tendermint validators and `s_e x network.ersatzvalidators.reward.factor` the total stake delegated to the ersatz validators (scaled appropriately), then the primary / tendermint pool has a fraction of `s_p/s_t` of the total reward, while the ersatz pool has `network.ersatzvalidators.reward.factor x s_e / s_t` (both rounded down appropriately).
```
ze97286: I reckon there's a typo here. should be: `s_t = network.ersatzvalidators.reward.factor x s_e + s_p`

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
```diff
@@ -69,4 +87,20 @@ function validatorScore(valStake) {
...
+```
+ linearScore = (valStake)/s_total
+ linearScore = Math.min(1.0, Math.max(0.0,linearScore))
+```
+i.e., there is no anti-whaling function applied here (the penalties are removed)
+
+
+# Acceptance criteria
 
+## Spare key on multisig
```
ze97286: not sure where does the logic that the acceptance criteria refers to is defined? the other direction is defined, i.e. if you are in the top multisig signer and not registered with the multisig contract you'll get a score of 0 and not get a reward. But this is saying that if there's a key on the multisig that doesn't belong to any one of the tm validators all get 0. 

EDIT - I see that in line 29. ignore me.

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
davidsiska-vega: Oh yes, definitely don't want a recursive formula!

davidsiska-vega:
```suggestion
 Thus, if `s_t = network.ersatzvalidators.reward.factor x s_e + s_p` is the total amount of own+delegated stake to both sets (with ersatz scaling taken into account), `s_p` the total stake delegated to the primary / tendermint validators and `s_e x network.ersatzvalidators.reward.factor` the total stake delegated to the ersatz validators (scaled appropriately), then the primary / tendermint pool has a fraction of `s_p/s_t` of the total reward, while the ersatz pool has `network.ersatzvalidators.reward.factor x s_e / s_t` (both rounded down appropriately).
```

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
davidsiska-vega: 👍

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
davidsiska-vega: Fixed.

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
```diff
@@ -18,15 +19,32 @@ This applies both the rewards coming from the [on-chain-treasury](./0055-TREA-on
...
+ Both these validators get rewards following the method above:
+ 1. The reward pool is split into two parts, propotional to the total own+delegated stake the primary- and ersatzvalidators have. 
+ Thus, if `s_t = network.ersatzvalidators.reward.factor x s_e + s_p` is the total amount of own+delegated stake to both sets (with ersatz scaling taken into account), `s_p` the total stake delegated to the primary / tendermint validators and `s_e x network.ersatzvalidators.reward.factor` the total stake delegated to the ersatz validators (scaled appropriately), then the primary / tendermint pool has a fraction of `s_p/s_t` of the total reward, while the ersatz pool has `network.ersatzvalidators.reward.factor x s_e / s_t` (both rounded down appropriately).
+
+ The following formulas then apply to both primary and ersatz validators, where 'total available reward' and 'total delegation', total_stake and 'number_of_validators' or `s_total` refer to the corresponding reward pool and the total own+delegated corresponding set of validators (i.e., `s_p` or `s_e`, respectively).
```
ze97286: does this only apply to staking and delegation reward or does it also hold for infrastructure fees given to validators? i.e. do ersatz validators get infra fees rewards?

protocol/0064-VALP-validator_performance_based_rewards.md
```diff
@@ -16,7 +16,17 @@ was elected leader `l` and carried out the job.
...
 This counts need to be stored, per validator node, in vega core. 
 
-Define the performance score to be `performance_score := (l-f)/l` if `l > 0` and `performance_score = 0` if `l = 0`. 
+For validators participating in consensus define the performance score to be `performance_score := max[(l-f)/l, 0.05]` if `l > 0` and `performance_score = 1` if `l = 0`. 
+Flooring the score at `0.05` is there to make sure that every validator with non-zero own+delegated gets a chance to be a leader at least ocassionaly, even if they were poorly performing recently. 
+
+For validators who [have submitted a transaction to become validators](0000-validators-chosen-by-stake.md) the `performance_score` is defined as follows: during each epoch
+Every `1000` blocks the candidate validator node is to send a hash of block number `b` separetely signed by all the three keys and submitted; the network will verify this to confirm that the validator owns the keys. 
+Here `b` is defined as:
+First time it is the the block number in which the joining transaction was included. Then it's incremented by `1000`.
```
ze97286, line 25: 2 somewhat technical questions:
1.  What if they miss it by a block, what if they're transaction meant to be included in block 1000 but was included in block 1001? in the current implementation this will not be counted towards their performance because it will be expected to be received at block 1000. 
2. I'm assuming the validator will always send this message every 1000 blocks but it will count for their performance score if they are not tendermint validators. Otherwise they need to check if they are tm or ersatz or waiting list.

protocol/0069-VCBS-validators-chosen-by-stake.md
```diff
@@ -0,0 +1,147 @@
+# Validators chosen by stake
+
+New network parameter `network.numberOfTendermintValidators`. 
+New network parameter `network.validatorIncumbentBonus`.
+New network parameter `network.numberEthMultisigSigners`.
+New network parameter `network.ersatzvalidators.reward.factor`.
```
ze97286: nitpick - these 4 lines don't get rendered nicely as they do here.

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
davidsiska-vega: They should also get fee rewards; will clarify.

protocol/0061-REWP-simple_pos_rewards_sweetwater.md
```diff
@@ -18,15 +19,32 @@ This applies both the rewards coming from the [on-chain-treasury](./0055-TREA-on
...
+This is to ensure that validators (all validators) have incentive to pay Etherum gas to update the multisig signer list.  
+
 At the end of an [epoch](./0050-EPOC-epochs.md), payments are calculated. First we determine the amount to pay out during that epoch: 
+
 1. multiply the amount in the reward pool by `reward.staking.delegation.payoutFraction`; this is the amount going into next step, call it `stakingRewardAmtForEpoch`.
 1. If the reward pool in question is the on-chain treasury for the staking and governance asset then `stakingRewardAmtForEpoch` is updated to `min(stakingRewardAmtForEpoch, reward.staking.delegation.maxPayoutPerEpoch)`. 
 
+## Tendermint Validators and Ersatz validators
+ In Vega, we have two sets of Validtors, the primary validators (which run tendermint) and the [ersatz validators](0000-validators-chosen-by-stake.md) (which are running a non-validator node and can be promoted to a validator node by the protocol if they meet the right criteria). 
+ Both these validators get rewards following the method above:
```
davidsiska-vega:
```suggestion
 Both tendermint validators and ersatz get rewards (both from fees and additional from on chain treasury) following the method above:
```

protocol/0064-VALP-validator_performance_based_rewards.md
davidsiska-vega: Clarified, please have a look.

protocol/0069-VCBS-validators-chosen-by-stake.md
davidsiska-vega: 👍 fixed

edd:
There's some follow up in #817 but otherwise this is approved

edd (approved):

PR #769, opened 2021-12-14
Starting work on matching engine spec

### LINKED ISSUE #767: Write specification of the Vega matching engine
We have 
- https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0014-order-types.md
- https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0025-order-submission.md 
- https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0026-auctions.md (matching happens in auctions)
- https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0033-cancel-orders.md 
- https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0037-pegged-orders.md 
- https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0038-liquidity-provision-order-type.md 

*but* we don't have a spec for the matching engine. 

@barnabee would be the best person to quickly write something down; I am happy to add some acceptance criteria.

peterbarrow:
Adding a spec to define functionality required from our matching engine.

Closes #767


edd:
I've made a lot of (insignificant) changes so far:
- `master` has been merged,
- `0022-matching-engine.md` is now `0068-MATC-matching_engine.md` because:
  - `auth` is 0022 already
  - There was no code for the filename
  - Underscores separate words in the descriptive part of the title
With those renames, hooray! `make` passes its checks, so we can use `0061-MATC-001` and onwards to code ACs. Then, I
- Added links to the order status/validity as neccessary
- Added a section on Self-trading then removed and linked to the wash trading part of Order Status 
  
and have some questions:
- [x] `Entering auction is possible if the volume on either side of the book is empty` What does possible mean? Is it 'will' or 'must'?
- [x] `In auction mode, FOK/IOC/GFN orders are rejected.` How are they rejected? Is their status set to REJECTED or are rejected like spam and never acknowledged ever again?
- [x] When self-trading, it's only the Aggressive order that's stopped - the passive order remains open and neither trade any volume?
- [x] What `parked` is isn't always isn't clear here (because of the order the specs were written, I reckon). Basically it's only defined in the Pegged Order spec, and we're saying more than just pegged order scan be parked. SO. I'll make a ticket to cover parks better. But for now - for 'When a market moves in to an auction', I can see GFN is cancelled, and pegs are parked, what about the rest? I'm not clear if GFN is an umbrella Order Validity or if this spec just doesn't make it clear what happens to a GTT order. I think GTT is different to GFN and GTTs will be parked. How would I check if a GTT becomes parked - status?
- [x]  You say that self-trading is allowed when uncrossing an auction (or at least, a minor edit I made leads it to look like that) - that's not clear in the auction spec. IS IT TRUE? If it is, I'll make a note to add it to the auction spec
- [x] You say any persistent order can be deleted - I've changed it to any ACTIVE persistent order can be CANCELLED. If deleted is real, I don't know about it and it's not documented

edd:
Oh, for a future ticket: time priority, and operations that result in a change in that (amends eg)

edd:
@peterbarrow This feels like it's in a state where we could merge it - shall we un-WIP and merge? Or are there specific things to add after your testing with @vega-paul

edd (approved):

PR #771, opened 2022-01-03
Zk Multisig Control

### LINKED ISSUE #969: Validate ZK replacement for multisig
Validate ZK replacement for multisig:
- how it integrates with every aspect
- does it solve it all of our problems (stranded withdrawals)

C0deMunk33:
ZK rollup version of multisig control.

close vegaprotocol/specs#969 .


davidsiska-vega:
Can we clarify in this spec (even at just high level): 
- the process of updating validator weight (what has to happen on Vega chain on Ethereum),
- the process of adding and removing a validator (what has to happen on Vega chain on Ethereum),
- what happens if a withdrawal is generated, not actioned on ethereum and signer weights / set changes? 

Some of these may also make decent acceptance criteria (stories QA can work through).

gordsport:
In referencing to auditing I have found 2 auditors that have experience in zksnarks/circom:

https://leastauthority.com/static/publications/LeastAuthority_Ethereum_Foundation_Zkopru_zk-SNARK_Circuits_Smart_Contracts_Final_Audit_Report.pdf

https://github.com/trailofbits/publications/blob/master/reviews/hermez.pdf

If we have indicative timings of when we may need this to be audited I can get this booked in

PR #779, opened 2022-01-11
Transfers spec

### LINKED ISSUE #759: Create transfers spec
Write spec allowing internal transfers among Vega assets:

It’s probably relatively straightforward
```
transfer(destination vega address, asset, amount)
```

Question: should be a small percentage fee (imagine a situation where nobody uses vega for trading but they keep assets on the bridge and transfer them around. Unlikely - but why would validators / token holders want run the chain in that case. @barnabee

jeremyletang:
Add a first draft of a transfer specs. Missed 2 section at least for now:
- spam
- fees

close #759


protocol/0067-transfers.md
```diff
@@ -0,0 +1,70 @@
+Transfers
+=========
+
+We introduce here a mechanism to transfer fund from an account to another, and this iniated explicitly by an user of the vega network.
```
edd:
```suggestion
This spec introduces a mechanism to transfer funds from one account to another, initiated explicitly by a user of the Vega network.
```

protocol/0067-transfers.md
```diff
@@ -0,0 +1,70 @@
...
+These transfers are not to be confused with the internal concept of transfers which results from event happening inside the protocol.
```
edd: **Question (non-blocking)**: Presumably it’s the same underlying mechanism? If so, I’m fine with continuing this spec as it is, but we have no spec for Transfers so in a later PR I would expand it to cover ALL transfers. This doesn’t need to change anything in the current draft.

protocol/0067-transfers.md
```diff
@@ -0,0 +1,70 @@
...
+
+This feature would allow a couple of scenario:
```
edd:
```suggestion
Allowing users to initiate transfers allows for the following capabilities:
```

protocol/0067-transfers.md
```diff
@@ -0,0 +1,70 @@
...
+- A user could transfer funds from a public key A to a public key B.
+- A user could transfer funds back to a reward pool
+- A user could transfer funds from and to a locked account used for staking.
+- etc..
+
+### Limits
+Transfer can only be iniated by a user using their own funds from their own account, directed to a limited set of accounts types.
```
edd:
```suggestion
Transfer can only be initiated by a party using their own funds from [accounts](./0013-accounts.md) that they are in control of:
```

protocol/0067-transfers.md
```diff
@@ -0,0 +1,70 @@
...
+
+Here's the list of accounts types from which a user send funds from:
+- GENERAL
+- LOCKED_FOR_STAKING (note: this is an account to be introduce in the future, where the funds would be locked, and will be accounted for staking during the epoch).
```
edd:
```suggestion
- LOCKED_FOR_STAKING (note: this account does not yet exist, but will hold funds that are locked for staking during an epoch).
```
I’ve edited this to make sense, but I’d drop it and then add it when we update the accounts spec. Your choice.

protocol/0067-transfers.md
```diff
@@ -0,0 +1,70 @@
...
+The system should be able to delay transfer. Such feature would be useful in the context of distributing token related to incentives for example.
+In order to do this the request for transfer should contain a field indicating when the destination account should be credited. The funds should be taken straigh away from the
+origin account, but distributed to the destination only once the time is reached.
+
+### Spam protection
+
+### Fees
+
+### Proposed command
+This new functionnality requires the introduction of a new command in the transaction API. Here's a propose payloda for it, using protobuf syntax:
```
edd:
```suggestion
This new functionality requires the introduction of a new command in the transaction API. The payload is as follows:
```

protocol/0067-transfers.md
```diff
@@ -0,0 +1,85 @@
+# Transfers
+
+This spec introduces a mechanism to transfer funds from one account to another, initiated explicitly by a user of the Vega network.
+These transfers are not to be confused with the internal concept of transfers which results from event happening inside the protocol.
+
+Allowing users to initiate transfers allows for the following capabilities:
+- A user could transfer funds from a public key A to a public key B.
+- A user could transfer funds back to a reward pool
```
davidsiska-vega:
```suggestion
- A user could transfer funds to a reward pool
```

protocol/0067-transfers.md
```diff
@@ -0,0 +1,85 @@
...
+Transfer can only be initiated by a party using their own funds from [accounts](./0013-accounts.md) that they are in control of:
+
+Here's the list of accounts types from which a user send funds from:
+- [GENERAL](./0013-accounts.md)
+- [LOCKED_FOR_STAKING](./0059-simple-staking-and-delegating.md)
+
+Here's the list of accounts types into which funds can be sent:
+- [GENERAL](./0013-accounts.md)
+- [LOCKED_FOR_STAKING](./0059-simple-staking-and-delegating.md)
+- [REWARD_POOL](./0056-rewards-overview.md)
```
davidsiska-vega:
```suggestion
- [REWARD_POOL](./0056-rewards-overview.md)
- [ON_CHAIN_TREASURY](./0055-on-chain-treasury.md)
```

protocol/0067-transfers.md
```diff
@@ -0,0 +1,85 @@
+# Transfers
+
+This spec introduces a mechanism to transfer funds from one account to another, initiated explicitly by a user of the Vega network.
+These transfers are not to be confused with the internal concept of transfers which results from event happening inside the protocol.
+
+Allowing users to initiate transfers allows for the following capabilities:
+- A user could transfer funds from a public key A to a public key B.
+- A user could transfer funds back to a reward pool
+- A user could transfer funds from and to a locked account used for staking.
```
jeremyletang: This is actually a new account specifically created so user can lock funds deposited on the erc20 bridge so they are accounted in yheir staking balance @davidsiska-vega . Not the existing account. Yhat would allow them to re stake straight away their rewards. I need to push a spec for it too.

protocol/0067-transfers.md
```diff
@@ -0,0 +1,85 @@
...
+- [GENERAL](./0013-accounts.md)
+- [LOCKED_FOR_STAKING](./0059-simple-staking-and-delegating.md)
+- [REWARD_POOL](./0056-rewards-overview.md)
+
+## Delayed transfer
+The system should be able to delay transfer. Such feature would be useful in the context of distributing token related to incentives for example.
+In order to do this the request for transfer should contain a field indicating when the destination account should be credited. The funds should be taken straight away from the origin account, but distributed to the destination only once the time is reached.
+
+## Spam protection
+In order to prevent the abuse of user-initiated transfers as spam attack, the system will be configurabled with a [network parameter](#network-parameters) that will limit the number of transfers that a user can initiate within a set period of time.
```
davidsiska-vega: Choose a name for the network parameter.

davidsiska-vega: Should this be per epoch? Per block? What is a sensible default? It should be high enough for a batch transfer of e.g. incentivised games rewards.

We could also have a minimum amount per asset (and then the fee will make sure these make sense for our validators).

protocol/0067-transfers.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Fees
+A fee is taken from all transfers, and paid out to validators in a similar manner to the existing [infrastructure fees](./0059-simple-POS-rewards.md).
+
+The fee is set by a [network parameter](#network-parameter) that defines the proportion of each transfer taken as a fee. The fee is taken from the transfer initiator's account immediately on execution, and is taken on top of the total amount transferred. It is [paid in to the infrastructure fee pool](./0029-fees.md#collecting-and-distributing-fees). Fees are charged in the asset that is being transferred.
```
davidsiska-vega: Please let's choose a name e.g. `tranfets.fee.factor` ...

protocol/0067-transfers.md
```diff
@@ -0,0 +1,85 @@
...
+  int64 deliver_on = 6;
+  // an arbitrary reference, 100 chars max
+  string reference = 7;
+}
+```
+
+## Network Parameters
+| Property         | Type   | Example value | Description |
+|------------------|--------| ------------|--------------|
+| `spam.protection.maxUserTransfersPerMinute`       | String (integer) |  `"1"`        | The most transfers a use can initiate per minute |
```
davidsiska-vega: Ah I see there's a name. Would it make sense to go per epoch rather than per minute? That way we could potentially allow a larger number "in one go" (thinking again about incentive rewards).

jeremyletang: Per epoch would make more senses to me too. It would would also match thr behaviour of our other resources spams protected.

protocol/0067-transfers.md
davidsiska-vega: For spam protection let's also introduce that the minimum transfer amount should be the asset `quantum` which will be introduced here: https://github.com/vegaprotocol/specs-internal/issues/797 .

davidsiska-vega:
Btw with @barnabee we are thinking that there need to be the possibility of setting up periodic transfers.

barnabee:
> Btw with @barnabee we are thinking that there need to be the possibility of setting up periodic transfers.

Yeah, these are summarised in [this ticket ](https://github.com/vegaprotocol/specs-internal/issues/800)and will replace [periodic allocation to reward pool accounts](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0055-TREA-on_chain_treasury.md#periodic-automated-allocation-to-reward-pool-account) assuming we go with that proposal.

protocol/0067-transfers.md
edd: This is incorrect, @edd  - collateral transfers are already covered in 005-COLL-collateral.md

protocol/0067-transfers.md
```diff
@@ -0,0 +1,85 @@
+# Transfers
+
+This spec introduces a mechanism to transfer funds from one account to another, initiated explicitly by a user of the Vega network.
+These transfers are not to be confused with the internal concept of transfers which results from event happening inside the protocol.
```
edd:
```suggestion
These transfers are not to be confused with the internal concept of transfers which results from event happening inside the protocol, which are covered in spec [0005-COLL](./0005-COLL-collateral.md).
```

davidsiska-vega:
From https://github.com/vegaprotocol/specs-internal/issues/800: periodic transfers should be saved in LNL checkpoints so they resume after network reset.

protocol/0067-transfers.md
```diff
@@ -0,0 +1,167 @@
...
+  - [ ] Until the epoch is reached not transfers are executed
+  - [ ] Once I reach the start epoch transfers happens.
+  - [ ] The same amount is transfered every epoch
+  - [ ] After I reach the epoch after the `end epoch`, no transfers are executed anymore
+- [ ] As a user I can create a recurring transfer _that decreases over time_
+  - [ ] I specify a start and end epoch, and a factor of `0.7`
+  - [ ] Until the epoch is reached not transfers are executed
+  - [ ] Once I reach the start epoch transfers happens.
+  - [ ] The amount transfered every epoch decreases
+  - [ ] After I reach the epoch `?`, no transfers are executed anymore
```
jeremyletang:
```suggestion
  - [ ] After I reach the end epoch, no transfers are executed anymore
```

edd: What happens if you don’t specify an end - at some point we can’t reduce it by 0.7 every time. @barnabee

barnabee: We could do one of these two things when there's no end in sight:
- If the next and previous amounts are the same or next == 0 it ends (this will mean some go on for a very long time)
- If the transfer amount drops below `quantum` for the asset (or some multiple thereof, specified by a network parameter), it ends

barnabee: cc @davidsiska-vega

protocol/0067-transfers.md
davidsiska-vega: Cool

protocol/0067-transfers.md
davidsiska-vega: We should have `quantum_mult_for_min_transfer_amt` and the period transfer should stop once it falls below this. We can also use the same multiplier for spam protection.

davidsiska-vega:
How do we handle a perpetual transfer when the source account is empty? Do we
a) cancel the perp transfer or
b) it hangs in the background waiting for the account to fill?

@jeremyletang / @barnabee

barnabee:
> How do we handle a perpetual transfer when the source account is empty? Do we a) cancel the perp transfer or b) it hangs in the background waiting for the account to fill?
> 
> @jeremyletang / @barnabee

@davidsiska-vega My initial thought would have been transfer as much of the requested amount as possible (which may be 0) each time the transfer is supposed to happen but neither hang waiting for the rest nor cancel the transfer. The owner can always cancel it if they need to OR transfer extra at manually any time to make up for it.

Not sure if there are any downsides to this I haven't seen though…

davidsiska-vega:
> > How do we handle a perpetual transfer when the source account is empty? Do we a) cancel the perp transfer or b) it hangs in the background waiting for the account to fill?
> > @jeremyletang / @barnabee
> 
> @davidsiska-vega My initial thought would have been transfer as much of the requested amount as possible (which may be 0) each time the transfer is supposed to happen but neither hang waiting for the rest nor cancel the transfer. The owner can always cancel it if they need to OR transfer extra at manually any time to make up for it.
> 
> Not sure if there are any downsides to this I haven't seen though…

That seems fine and also the easiest to spec implement; always transfer `max(specified amount, balance in source account)`.

davidsiska-vega (approved):
LGTM

PR #794, opened 2022-01-12
add sequence number to multi sig control

### LINKED ISSUE #795: [SPEC ONLY] Review multsig contract
We still have the issue of stranded withdrawals if signer set changes sufficiently. 

This issue is to write / update the specs for the multisig contract to  store a history of the valid signatures signers together with an update sequence number? That way a withdrawal that contains the signatures and sequence number can be validated even after more than a third of the signers changed. This would fix the problem of a party generating a withdrawal bundle, not submitting it to Ethereum and validators changing.

see: https://vegaprotocol.slack.com/archives/CMVDT0ZU6/p1641995034007500

jeremyletang:


jeremyletang:
@edd @barnabee @davidsiska-vega

davidsiska-vega:
Is there any reason why not to assume we store the weights corresponding to the sequence number? Maybe optional weights so the spec works whether the multisig uses those or not.

davidsiska-vega:
Pinged @C0deMunk33 about the question above ☝️, I think it got missed in github's crazy notification system.

C0deMunk33:
> Is there any reason why not to assume we store the weights corresponding to the sequence number? Maybe optional weights so the spec works whether the multisig uses those or not.

I cannot include weights on-chain because I WILL run out of gas calling out for balances OR accepting a big payload to cycle through

PR #796, opened 2022-01-13
Written spec for market decimal places

### LINKED ISSUE #776: Decimal place conversions
In order to start work on the decimal place conversions we require and issue or spec detailing the direction of the functionality required. This should also include acceptance criteria for the feature.

Open question on this is how will we account for fees etc. with differing decimal places?

EVODelavega:
I've tried to be as explicit/clear as possible. If things need to be clarified further, or other changes are needed, feel free to comment


non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,77 @@
+# Market decimal places
+
+This document aims to outline how we are to handle the decimal places of a given market, if said market is configured with fewer decimal places than its settlement asset. For example: a market settling in ETH can be configured to use only 9 decimal places (GWei) compared to ETH's 18. A market cannot specify _more_ decimal places than its settlement asset supports. This means that a market that settles in GBP can have anywhere between 0 and 2 decimal places, but no more. A market settling in Japanese yen can have no decimal places. Common sense.
```
barnabee: Why can a market not specify more decimals that it's settlement asset? We _could_ enforce this now but it is not something we would want to require in the long term so I am wondering if it's necessary now.

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Terminology
+
+* Settlement asset: the asset in which transactions for a given market are made (margin balances, fees, settlements, etc...).
```
barnabee, line 7: NB: in future there will be multiple settlement assets for some markets, which could have different numbers of decimal places

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,77 @@
...
+* Market precision: the number of decimal places a market uses (as mentioned previously, a market where the smallest unit of ETH is a GWei has a 9 decimal places, so the market precision is 9)
```
barnabee: The "market precision" is really the price/quote precision, it is the number of decimal places used on the order book. Price and quotes on the order book are NOT guaranteed to be the same units as the settlement asset. For instance a price might be in USD to 4 decimal plades but settlement is in ETH to 18 places.

It is essential that we understand that these are NOT always going to be just the same types with different decimals and write our designs clearly.

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,77 @@
...
+* Asset precision: the number of decimal places for a given asset. Again, a market with precision 9 that settles in ETH will have a market precision of 9, whereas the asset precision is 18.
```
barnabee, line 9:
```suggestion
* Asset precision: the number of decimal places for a given asset. Again, a market with precision 9 that settles in ETH will have a market precision of 9, whereas the asset precision is 18. Note that the asset precision is fixed for a given asset in the [asset framework](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0040-ASSF-asset_framework.md).
```

edd:
[Ironically this comment is WIP, but useful now so I’m posting it partially done]

The spec is nice and clear! What we need to do is add acceptance criteria in lots of different specs. I'm willing to do *that*, if we as a group can add more in this spec.

# New _governance_ ACs…
**Note**: These belong somewhere in 0001-MKTF and…
- When creating a market, a precision must be specified
- If a precision is specified that is more than the precision of the underlying, the market proposal must be `REJECTED`
- If a precision is specified that is equal to or less than the precision of the underlying, all events and orders, market data etc must have their monetary values returned in this precision, not the precision of the underlying

# New _margin_/insurance/account ACs…
**Note**: These belong in 0011-MARA, 0015-INSR, 0013-ACCT at least
- Margin calculations must be in the same precision as the market to which it relates.
- Insurance pool transfers must be in the same precision as the market to which it relates.
- General accounts are in the precision of the asset to which they relate

# New _order_ ACs…
**Note**: These belong in 0014-ORDT - pegs & market orders are unaffected
- Limit order prices must be specified in the correct precision for the market they are placed in. Clients must handle rounding or padding
- Trade values (price) are expressed in the precision of the market.

jeremyletang:
> [Ironically this comment is WIP, but useful now so I’m posting it partially done]
> 
> The spec is nice and clear! What wE need to do is add acceptance criteria in lots of different specs. I'm willing to do _that_, if we as a group can add more in this spec.
> # New governance ACs need to cover
> 
> **Note**: These belong somewhere in 0001-MKTF and…
> 
>     * When creating a market, a precision must be specified
>       
>       * Alternatively: If not specified, defaults to the precision of the underlying
>         
>         * In this case, the precision of the underlying should be returned when requesting the precision for the market
> 
>     * If a precision is specified that is more than the precision of the underlying, the market proposal must be `REJECTED`
> 
>     * If a precision is specified that is equal to or less than the precision of the underlying, all events and orders, market data etc must have their monetary values returned in this precision, not the precision of the underlying

I don't know if these are example or other, but I would make the decimalPlace mandatory for the new market proposal.

EVODelavega:
Yeah, this would make decimal place a mandatory field, as it's a crucial part of the market definition after all

edd:
> Yeah, this would make decimal place a mandatory field, as it's a crucial part of the market definition after all

Alright, I’ve edited my comment to remove the option / defaulting

edd:
This will likely have merge conflicts with #799 when adding acceptance criteria. Give me a shout and I will tidy it up.

edd:
Merged master to deal with some conflicts, now reviewing 👀

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,79 @@
+# Market decimal places
+
+This document aims to outline how we are to handle the decimal places of a given market, if said market is configured with fewer decimal places than its settlement asset. For example: a market settling in ETH can be configured to use only 9 decimal places (GWei) compared to ETH's 18. A market cannot specify _more_ decimal places than its settlement asset supports. This means that a market that settles in GBP can have anywhere between 0 and 2 decimal places, but no more. A market settling in Japanese yen can have no decimal places. Common sense.
```
witgaw: I'd get rid of the last sentence. There are USD denominated markets that have more than 2 decimal places, see: https://www.sec.gov/divisions/marketreg/subpenny612faq.htm#:~:text=In%20general%2C%20the%20Rule%20prohibits,greater%20than%20%241.00%20per%20share.

So, in a sense, sure we can't have more precision in the market than we have in the representation of the asset that Vega uses, but that representation need not be tied to any real properties of the asset, say physical representation of USD where the lowest multiple is 1 cent. 

```suggestion
This document aims to outline how we are to handle the decimal places of a given market, if said market is configured with fewer decimal places than its settlement asset. For example: a market settling in ETH can be configured to use only 9 decimal places (GWei) compared to ETH's 18. A market cannot specify _more_ decimal places than its settlement asset supports.
```

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Terminology
+
+* Settlement asset: the asset in which transactions for a given market are made (margin balances, fees, settlements, etc...).
+* Market precision: the number of decimal places a market uses (as mentioned previously, a market where the smallest unit of ETH is a GWei has a 9 decimal places, so the market precision is 9)
+* Asset precision: the number of decimal places for a given asset. Again, a market with precision 9 that settles in ETH will have a market precision of 9, whereas the asset precision is 18.
+
+## Mechanics
+
+It is possible to configure a market where orders can only be priced in increments of a specific size. This is done by specifying a different (smaller) number of decimal places than its settlement asset supports. Simply put: a market that settles in GBP can be configured to have 0 decimal places, in which case the price levels on the orderbook will be at least separated by £1, rather than the default penny (as GBP is an asset with 2 decimal places).
```
witgaw: See previous comment:

```suggestion
It is possible to configure a market where orders can only be priced in increments of a specific size. This is done by specifying a different (smaller) number of decimal places than its settlement asset supports. Simply put: a market that settles in GBP can be configured to have 0 decimal places, in which case the price levels on the orderbook will be at least separated by £1, rather than the default penny.
```

protocol/0001-MKTF-market_framework.md
```diff
@@ -40,7 +40,7 @@ Data:
...
   - **Tick size**: (size of an increment in price in terms of the quote unit)
-  - **Quoted Decimal places**: number of decimals places for quote unit, e.g. if quote unit is USD and decimal places is 2 then prices are quoted in integer numbers of cents.
+  - **Quoted Decimal places**: number of decimals places for quote unit, e.g. if quote unit is USD and decimal places is 2 then prices are quoted in integer numbers of cents. This value is mandatory (though 0 values are accepted). A market cannot be valid if the quoted decimal places > decimal places of the settlement asset. If the market settles in ETH, then the max valid quoted decimal places is 18.Events related to this market (trades, orders, market data) will be expressed with the market decimal places. Transfers are expressed in asset decimal places. A market settling in ETH, but with 10 decimal places will represent 1 ETH as `10,000,000,000`, but transfers resulting from things like MTM settlement will represent 1 ETH as `1,000,000,000,000,000,000`.
```
witgaw, line 43: Could we not have some simple data structure that knows the number of decimals for each quantity? Might make it easier to stay sane when debugging or if those things make it to events in unscaled form?

witgaw (dismissed):
Looks good to me overall, just some comments:

Is there any particular reason why we're not calling "market precision" a "tick size"? If there's one to one relationship we can at least mention that in the spec as later is more commonly used term I think. 

Am I right in thinking that `10**-(asset_precision - market_precision)` basically gives us the lowest order fraction that we can support (for when we add fractional orders)?

Other comments in markdowns:

EVODelavega:
@witgaw The reason why I'm not calling it tick size here, is mainly to differentiate between the `TickSize` field that currently exists, but yeah, 10**(adp - mdp) is the tick size, and the smallest supported fraction for that market.

protocol/0001-MKTF-market_framework.md
EVODelavega: That would mean we'd have to update the protos for orders, trades, market data, etc... We have a lot of tests that would fail if we didn't adjust the prices as and where we need them to be adjusted. We discussed what to do with these events before this spec was written, and general consensus was that order/trade events should use market ticks, transfers and balances (like P&L in positions) should be expressed in the smallest units for the relevant asset.

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,79 @@
...
+
+```
+priceExponent = 10**(asset_precision - market_precision)
+// for GBP markets with 0 decimal places this is:
+priceExponent = 10 ** (2 - 0) == 100
+// for ETH markets with 9 decimal places, this is:
+priceExponent = 10 ** (18 -9) == 1,000,000,000
+```
+
+When an order is submitted, amended, or otherwise updated, the core emits an event for the data-node (and any other interested parties). The price in this order event should still be represented as a value in market precision. Updating the price on the order internally, for reasonse we shall elaborate on later, should not effect the market data on the event bus. To clarify:
```
davidsiska-vega:
```suggestion
When an order is submitted, amended, or otherwise updated, the core emits an event for the data-node (and any other interested parties). The price in this order event should still be represented as a value in market precision. Updating the price on the order internally, for reasons we shall elaborate on later, should not effect the market data on the event bus. To clarify:
```

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,79 @@
...
+
+order.Price *= market.priceExponent // ie multiply by 100
+broker.Send(events.NewOrderEvent(ctx, order)) // should still emit the event where the order is priced at 1
+```
+
+In short, market related events should specify prices in the _"unit"_ the market uses.
+
+### Benefits
+
+By allowing markets to specify their own precision, the price levels can be more closely controlled, and any changes in the mark price could be made as _"significant"_ as we want. By converting the prices internally to asset precision, we are able to accurately calculate fees and margin levels based on the asset precision. Fees and margins are, after all, a percentage of a value. By operating a market with a lower precision than the asset itself, fees and margin requirements can be calculated with a higher level of precision.
```
davidsiska-vega:
```suggestion
By allowing markets to specify their own precision, the price levels can be more closely controlled, and any changes in the mark price could be made as _"significant"_ as we want. By converting the prices internally to asset precision, we are able to accurately calculate fees and margin levels based on the asset precision. By operating a market with a lower precision than the asset itself, fees and margin requirements can be calculated with a higher level of precision.
```

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,79 @@
...
+Order submissions and amendments are received, the submitted price (aka market or original price) is copied to a private field, so the events can be constructed in an accurate way. Before submitting the order to the book, or amending it, the price we received from the chain is multiplied by the price exponent. From that point on, the core will operate on the value that has the asset precision. Calculating fees and margin requirements, updating/creating the price levels, etc... all happens in asset precision. Any order events the core sends out will look exactly the same as before.
+
+### Pegged orders
+
+When (re-)pricing pegged orders, the offset values are multiplied by the same price factor before we add/subtract them from the reference price. The offsets themselves (as in: the field on the order object) is not updated. Events that contain this data, therefore, will still look exactly the same as they do now.
+
+
+### Liquidity provisions
+
+Orders created for an LP work pretty much exactly the same as pegged orders. The offsets will, again, be multiplied when the price is calculated, but the LP shape object is not updated.
```
davidsiska-vega:
```suggestion
Orders created for an LP work pretty much exactly the same as pegged orders. The offsets will, again, be multiplied by the price exponent when the price is calculated, but the LP shape object is not updated.
```

non-protocol-specs/0009-market-decimal-places.md
```diff
@@ -0,0 +1,79 @@
...
+When repricing LP orders, we ensure the price of the orders fall inside the upper and lower price bounds. These values are going to be calculated based on the prices used internally (in asset precision). So as to not create orders at a price point that is more precise than the market is configured to support, we floor the max price bound, and ceil the minimum price, as those values will effectively be the max and min prices that are allowed without trades resulting in an auction being triggered.
+
+### Market data
+
+The market data returns the same min/max prices mentioned above. As the name implies, _"MarketData"_ is clearly market related data... The min/max price values we return from this call should therefore be floored (max) and ceiled (min) in the same way.
+
+
+### Trades
+
+Trades of course result in transfers. The amounts transferred (for the trade as well as the MTM) happen at asset precision. The trade events the core sends out, however, are once again market related data. The prices on these trade events will be represented as a value in market precision.
```
davidsiska-vega, line 70: I wonder whether we should send both precisions out. It may be difficult to reconcile certain histories if I don't see the full precision events. Imagine an asset with 1 decimal and market with 0 decimals. For some reason party gets the 10 x transfer of `0.1`. If I am watching the trade events then I see 10x transfer of `0` (at market precision) but at the end of this the party has balance increased by `10x0.1=1` but I wouldn't know. Or am I missing somerthing?

EVODelavega: The transfers are never going to be sent out at market precision. The transfers are moving an asset between accounts, and as such is always going to show up with the nr of decimals of the assets. You'll never see a transfer of 0.

non-protocol-specs/0009-market-decimal-places.md
witgaw: I'd get rid of it the bit in parentheses.

witgaw (dismissed):
Looks good to me, I'd just delete the bit about 2 decimal places for GBP (as suggested below), can go in either way though.

EVODelavega:
Removed bit in parenthesis as requested, clarified that market decimals essentially are market ticks.

gordsport:
@EVODelavega @witgaw is there anything else needed on this before it can be merged?

Can the acceptance criteria be reviewed and IDs added?

EVODelavega:
Conflicts in other spec files, so I just created a new PR with this file, based on the current master: https://github.com/vegaprotocol/specs-internal/pull/934

barnabee (commented):
I realise this is submitted after the fact (I fucked up, forgot to complete the review so they were left pending), but completing it now (and approving) so I can link to these comments.

non-protocol-specs/0009-market-decimal-places.md
gordsport: @ze97286 - this seems different to the comment here:
https://github.com/vegaprotocol/vega/issues/5190#issuecomment-1089287212
(This could be my misunderstanding but wanted to make sure)

PR #798, opened 2022-01-14
Introduce asset quantum amount instead of min_Lp_stake

### LINKED ISSUE #797: Replace asset parameter min_Lp_stake with quantum
In many places we need to have information about what a minimum economically meaningful amount of an asset is. Often it's not 1 because for an asset with 18 decimals this would be 1e^{-18}. 

Thus we introduce "quantum" as "the minimum economically meaningful amount of an asset". Min_Lp_stake will be removed and replaced by `quantum x min_LP_stake_quantum_multiple`, where `min_LP_stake_quantum_multiple` is a new network parameter.

davidsiska-vega:
Closes #797 .


davidsiska-vega:
One comment from @jeremyletang on slack was that this will be a bit of pain with respect to LNL restore when we change the parameter name inside the asset; it the LNL restore code will have to handle it specifically I guess.

edd:
I'm going to deal with the merge conflict after merging #799

**Update**: Done

jeremyletang (dismissed):

ze97286 (approved):

PR #801, opened 2022-01-19
Add Acceptance Criteria codes to a feature

edd:
This is intended as a discussion for how we map the new Acceptance Criteria codes (#781, #799) on to feature files. It's not set in stone - let's decide if we like it, then expand the usage elsewhere.

# What's changed
1. If a `Scenario` in a `.feature` file appears to test _exactly one_ Acceptance Criteria, the AC code for that criteria is _appended to the scenario name_, in parenthesis
  ```Scenario: Example One``` becomes
  ```Scenario: Example One (0000-EXMP-0000)```
3.. If a `Scenario` in a `.feature` file appears to test _more than one_ Acceptance Criteria, the AC codes for that criteria is _appended to the scenario name_, in parenthesis, separated by commas
  ```Scenario: Example One``` becomes
  ```Scenario: Example One (0000-EXMP-0000, 0000-EXMP-0001)```
2. If a `Scenario` in a `.feature` file appears to test valuable things that are not explicitly Acceptance Criteria, the AC code `0000-EXMP-unmapped` is appended to the scenario name, in parenthesis
  ```Scenario: Example One``` becomes
  ```Scenario: Example One (0000-EXMP-additional-tests)```

Which is based on the following rules:
1. It is valid for a scenario to test things that aren't explicitly Acceptance Criteria in order to verify the system is operating correctly
2. Exhaustively mapping every scenario back in to a spec file is not a good use of time
3. It is valid for a scenario to touch multiple different Acceptance Criteria
4. Every Acceptance Criteria that has a code **should** have a test, but exactly where is not yet decided.
5. Every Scenario in a feature file **must** be updated to have one of the three things above - explicitly unmapped, 1 or more linked ACs
6. No scenario can have `0000-EXMP-unmapped` *and* AC codes. Unmapped should be removed when there are more than 0 ACs to link to.

# Notes
All of the above is open for change. Also, we may disagree on if a feature tests an AC. That's fine too! As long as we all agree what we want at the end of this PR. Also I think I missed some scenarios in this PR, so... I've already broken the rules.

### DIFF
--- a/qa-scenarios/0061-delegation.feature
+++ b/qa-scenarios/0061-delegation.feature
@@ -60,7 +60,7 @@ Feature: Staking & Delegation
     #complete the first epoch for the self delegation to take effect
     Then the network moves ahead "7" blocks
 
-  Scenario: A party can delegate to a validator and undelegate at the end of an epoch
+  Scenario: A party can delegate to a validator and undelegate at the end of an epoch (0059-STKG-002, 0059-STKG-013, 0059-STKG-014)
     Description: A party with a balance in the staking account can delegate to a validator
 
     When the parties submit the following delegations:
@@ -115,7 +115,7 @@ Feature: Staking & Delegation
     | party1 |  node2   |   0    |       
     | party1 |  node3   |   0    |   
 
-  Scenario: A party cannot delegate less than minimum delegateable stake  
+  Scenario: A party cannot delegate less than minimum delegateable stake (0059-STKG-001)
     Description: A party attempts to delegate less than minimum delegateable stake from its staking account to a validator minimum delegateable stake
 
     When the parties submit the following delegations:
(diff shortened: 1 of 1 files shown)


cpuja84:
@edd @davidsiska-vega @witgaw @vega-paul - 
I'd suggest to go with `Additional-tests` over `unmapped` so that we can live with them if we choose to at a later period. And these feature tests would be made open source at some point. And `unmapped` sounds work is still not finished as opposed to `Additional-tests` which puts us in good light or may be this is just me !

edd:
@cpuja84 I've changed `additional_tests` to `additional-tests`, and as far as I am concerned this file is ready to merge.

If you want to review if my mappings from feature tests to acceptance criteria is correct, then now is the time! Otherwise, please merge.

cpuja84 (approved):
Looks good to me.

PR #806, opened 2022-01-21
0050-EPOC Acceptance criteria

### LINKED ISSUE #805: 0050-EPOC-epochs needs acceptance criteria
There are currently 0 acceptance criteria for how epochs transition and how to test various events around it. This task is for someone to add some clarity as a starting point.

edd:
This spec currently has 0 acceptance criteria. It's a nice clear spec, so I'm working through to add some
- Add ACs for when epochs change
- Add clarity about what happens with multiple epochs passing in a single block
- Add ACs for delegating
- Add ACs for undelegating
- Add ACs for undelegating now
- Add link to staking and delegation spec, as it covers how tokens are locked and unlocked

See last comment for some things I don't think are covered well.

Closes #805

### DIFF
--- a/protocol/0050-EPOC-epochs.md
+++ b/protocol/0050-EPOC-epochs.md
@@ -198,3 +198,9 @@ All parameters that are changed through a governance vote are valid starting the
      - `party 1`s staking balanced is reduced immediately upon execution of the transaction (<a name="0050-EPOC-005" href="#0050-EPOC-005">0050-EPOC-005</a>)
      - `validator 1`s staked balance is not increased in epoch 1 (<a name="0050-EPOC-006" href="#0050-EPOC-006">0050-EPOC-006</a>)
      - `validator 1`s staked balance is increased in the first block of epoch 2 (<a name="0050-EPOC-007" href="#0050-EPOC-007">0050-EPOC-007</a>)
+4. Delegation can be 'undone' by undelegating within the same epoch
+   - During epoch 1, `party 1` stakes any valid amount to `validator 1`
+   - Subsequently, but during the same epoch, `party 1` unstakes the same amount from `validator 1`
+     - `validator 1`s staked balance is not increased in epoch 2 (<a name="0050-EPOC-008" href="#0050-EPOC-008">0050-EPOC-008</a>)
+     - `party 1`s staked-then-unstaked balance is returned in epoch 2 (<a name="0050-EPOC-009" href="#0050-EPOC-009">0050-EPOC-009</a>)
+4
--- a/protocol/0050-EPOC-epochs.md
+++ b/protocol/0050-EPOC-epochs.md
@@ -188,28 +188,25 @@ All parameters that are changed through a governance vote are valid starting the
 
 ## Acceptance Criteria
 
-1. Epochs change at the start of the first block that is after the epoch length has passed
+1. Epochs change at the end of the first block that is after the epoch expiry has passed
    - Given an epoch length of `1d`, with a block time of `1h`, at block 1 the current epoch is `1` (<a name="0050-EPOC-001" href="#0050-EPOC-001">0050-EPOC-001</a>)
    - Given an epoch length of `1d`, with a block time of `1h`, at end of block 23 the current epoch is `1` (<a name="0050-EPOC-002" href="#0050-EPOC-002">0050-EPOC-002</a>)
    - Given an epoch length of `1d`, with a block time of `1h`, at end of block 24 the current epoch is `2` (<a name="0050-EPOC-003" href="#0050-EPOC-003">0050-EPOC-003</a>)
 2. Edge case: Multiple epochs can pass within the same block (<a name="0050-EPOC-004" href="#0050-EPOC-004">0050-EPOC-004</a>) 
    - Given an epoch length of `1s`, with a block time of `1m`, at end of block 1 the current epoch is `1`
    - Given an epoch length of `1s`, with a block time of `1m`, at end of block 61 the current epoch is `61`
-3. Delegation takes effect at epoch changeover
-   - During epoch 1, `party 1` stakes any valid amount to `validator 1`
-     - `party 1`s staking balanced is reduced immediately upon execution of the transaction (<a name="0050-EPOC-005" href="#0050-EPOC-005">0050-EPOC-005</a>)
-     - `validator 1`s staked balance is not increased in epoch 1 (<a name="0050-EPOC-006" href="#0050-EPOC-006">0050-EPOC-006</a>)
-     - `validator 1`s staked balance is increased in the first block of epoch 2 (<a name="0050-EPOC-007" href="#0050-EPOC-007">0050-EPOC-007</a>)
-4. Delegation can be 'undone' by undelegating within the same epoch
-   - During epoch 1, `party 1` stakes any valid amount to `validator 1`
-   - Subsequently, but during the same epoch, `party 1` unstakes the same amount from `validator 1`
-     - `validator 1`s staked balance is not increased in epoch 2 (<a name="0050-EPOC-008" href="#0050-EPOC-008">0050-EPOC-008</a>)
-     - `party 1`s staked-then-unstaked balance is returned in epoch 2 (<a name="0050-EPOC-009" href="#0050-EPOC-009">0050-EPOC-009</a>)
+3. Nomination takes effect at epoch changeover
+   - During epoch 1, `party 1` nominates any valid amount to `validator 1`
+     - `party 1`s staking balanced is reduced immediately upon execution of the transaction (<a name="0050-EPOC-005" href="#0050-EPOC-005">0050-EPOC-005</a>) (note: this can be tested by trying to delegate again, which will be rejected)
+     - `validator 1`s nominated balance is not increased in epoch 1 (<a name="0050-EPOC-006" href="#0050-EPOC-006">0050-EPOC-006</a>)
+     - `validator 1`s nominated balance is increased in the first block of epoch 2 (<a name="0050-EPOC-007" href="#0050-EPOC-007">0050-EPOC-007</a>)
+4. Delegation can be 'undone' by undelegating within the same epoch (<a name="0050-EPOC-008" href="#0050-EPOC-008">0050-EPOC-008</a>)
+   - If this occurs in the same epoch, rewards for the validator and delegator will not change due to the delegation (<a name="0050-EPOC-009" href="#0050-EPOC-009">0050-EPOC-009</a>)
 5. [Undelegate now](#undelegate-now) does not return the balance to the delegator immediately
    - During epoch 1, `party 1` stakes any valid amount to `validator 1`
    - At the end of epoch 2, `validator 1` earns rewards based on a delegation count that includes `party 1`s stake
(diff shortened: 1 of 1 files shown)


edd:
I've marked this as ready for review, but I think there are some missing parts that could be covered by someone who knows this part of the system better:
- `Undelegate in anger` is not described in sufficient detail - but it's also not required now, so I've left it
- `Undelegation of locked stake` I'm not sure if it's adequately covered
- **Wanted Stake** is mentioned, but not adequately described. I'll make that a future ticket.

protocol/0050-EPOC-epochs.md
```diff
@@ -176,9 +179,43 @@ In mainnet alpha this is sufficient as the chain dies relatively quickly anyhow.
...
+| `validators.epoch.length`       | String (period) |  `"1h"` | The length of each Epoch. The block after this time will be the first block of the next epoch  |
 
 See the [network parameters spec](./0054-NETP-network_parameters.md#current-network-parameters) for a full list of parameters.
 
 ## Parameter changes
 All parameters that are changed through a governance vote are valid starting the epoch following the one the block is in that finalized the vote.
+
+## Acceptance Criteria
+
+1. Epochs change at the start of the first block that is after the epoch length has passed
```
ze97286: the definition is not accurate. An epoch is ended when the first block whose start time exceeds the expiry of the current epoch ends. To be even more precise - epochs start and end on the beginning of a block, not at its end, so when the first block whose start time exceeds the epoch expiration, we'd mark the epoch for completion, then when the new block begins we would end the prior epoch and start the new one.  

Here's an example:
epoch duration is 24h, we're in epoch 10 which started at 07-Feb 10:00:00.0 and expires on the 08-Feb 10:00:00.0. Now the next day comes, the time is 10:00:00 and a block begins - as the time is not AFTER the epoch expiration, nothing happens and we're still in epoch 10. The next block starts at 10:00:01 - now it's already after the expiration so when this block ends - we mark epoch as ready to end but not end it yet. When the new block begins lets say at 10:00:03 we end epoch 10 and begin epoch 11. This way epochs are continuous and there's never a gap between the end of epoch x and the start of epoch x+1.

protocol/0050-EPOC-epochs.md
```diff
@@ -176,9 +179,43 @@ In mainnet alpha this is sufficient as the chain dies relatively quickly anyhow.
...
+   - Given an epoch length of `1d`, with a block time of `1h`, at block 1 the current epoch is `1` (<a name="0050-EPOC-001" href="#0050-EPOC-001">0050-EPOC-001</a>)
+   - Given an epoch length of `1d`, with a block time of `1h`, at end of block 23 the current epoch is `1` (<a name="0050-EPOC-002" href="#0050-EPOC-002">0050-EPOC-002</a>)
+   - Given an epoch length of `1d`, with a block time of `1h`, at end of block 24 the current epoch is `2` (<a name="0050-EPOC-003" href="#0050-EPOC-003">0050-EPOC-003</a>)
+2. Edge case: Multiple epochs can pass within the same block (<a name="0050-EPOC-004" href="#0050-EPOC-004">0050-EPOC-004</a>) 
+   - Given an epoch length of `1s`, with a block time of `1m`, at end of block 1 the current epoch is `1`
+   - Given an epoch length of `1s`, with a block time of `1m`, at end of block 61 the current epoch is `61`
+3. Delegation takes effect at epoch changeover
+   - During epoch 1, `party 1` stakes any valid amount to `validator 1`
+     - `party 1`s staking balanced is reduced immediately upon execution of the transaction (<a name="0050-EPOC-005" href="#0050-EPOC-005">0050-EPOC-005</a>)
```
ze97286: not sure what you mean here.

protocol/0050-EPOC-epochs.md
```diff
@@ -176,9 +179,43 @@ In mainnet alpha this is sufficient as the chain dies relatively quickly anyhow.
...
+     - `validator 1`s staked balance is not increased in epoch 1 (<a name="0050-EPOC-006" href="#0050-EPOC-006">0050-EPOC-006</a>)
+     - `validator 1`s staked balance is increased in the first block of epoch 2 (<a name="0050-EPOC-007" href="#0050-EPOC-007">0050-EPOC-007</a>)
+4. Delegation can be 'undone' by undelegating within the same epoch
```
ze97286: it can, but that's what undelegate now is which you describe in 5. I don't understand anything in this criteria :)

protocol/0050-EPOC-epochs.md
```diff
@@ -176,9 +179,43 @@ In mainnet alpha this is sufficient as the chain dies relatively quickly anyhow.
...
+## Acceptance Criteria
+
+1. Epochs change at the start of the first block that is after the epoch length has passed
+   - Given an epoch length of `1d`, with a block time of `1h`, at block 1 the current epoch is `1` (<a name="0050-EPOC-001" href="#0050-EPOC-001">0050-EPOC-001</a>)
+   - Given an epoch length of `1d`, with a block time of `1h`, at end of block 23 the current epoch is `1` (<a name="0050-EPOC-002" href="#0050-EPOC-002">0050-EPOC-002</a>)
+   - Given an epoch length of `1d`, with a block time of `1h`, at end of block 24 the current epoch is `2` (<a name="0050-EPOC-003" href="#0050-EPOC-003">0050-EPOC-003</a>)
+2. Edge case: Multiple epochs can pass within the same block (<a name="0050-EPOC-004" href="#0050-EPOC-004">0050-EPOC-004</a>) 
+   - Given an epoch length of `1s`, with a block time of `1m`, at end of block 1 the current epoch is `1`
+   - Given an epoch length of `1s`, with a block time of `1m`, at end of block 61 the current epoch is `61`
+3. Delegation takes effect at epoch changeover
```
ze97286: in general I think it would be less confusing to use the association/nomination terminology rather than staking.

protocol/0050-EPOC-epochs.md
```diff
@@ -176,9 +179,43 @@ In mainnet alpha this is sufficient as the chain dies relatively quickly anyhow.
...
+5. [Undelegate now](#undelegate-now) does not return the balance to the delegator immediately
+   - During epoch 1, `party 1` stakes any valid amount to `validator 1`
+   - At the end of epoch 2, `validator 1` earns rewards based on a delegation count that includes `party 1`s stake
+   - During epoch 3, `party 1` uses Undelegate Now
+   - During epoch 3, `party 1` does not receive their staked amount back in their unstaked balance (<a name="0050-EPOC-010" href="#0050-EPOC-010">0050-EPOC-010</a>)
+   - During epoch 3, `validator 1`s voting weight remains unchanged (<a name="0050-EPOC-010" href="#0050-EPOC-010">0050-EPOC-010</a>)
+   - At the end of epoch 3, `validator 1` earns rewards based on a delegation count that *does not include* `party 1`s stake (<a name="0050-EPOC-012" href="#0050-EPOC-012">0050-EPOC-012</a>)
+   - At the end of epoch 3, `party 1` does not earn rewards for the stake that was undelegated (<a name="0050-EPOC-013" href="#0050-EPOC-013">0050-EPOC-013</a>)
+   - At the start of epoch 4, `party 1`s unstaked balance includes the undelegated amount (<a name="0050-EPOC-014" href="#0050-EPOC-014">0050-EPOC-014</a>)
```
ze97286, line 213: if they "undeledateNow" their free-to-delegate balance (which I assume is what you mean by "unstaked balance" is updated immediately as the transaction is executed.

protocol/0050-EPOC-epochs.md
```diff
@@ -176,9 +179,43 @@ In mainnet alpha this is sufficient as the chain dies relatively quickly anyhow.
...
+5. [Undelegate in anger](#undelegate-in-anger) acceptance TBC
```
ze97286, line 214: this is basically when they withdraw sake from the bridge, i.e. un-associate without first un-nominate.

protocol/0050-EPOC-epochs.md
edd: As soon as the transaction that requests that I delegate X to `validator 1` is executed, the unstaked balance is reduced. Or to be more specific, using the following query:

```graphql
{
  party(id:"9ba9def5e752c72cef91e0515b1c9f33b604e7475a36f4c049b74b2b0c5fad8c") {
    id
    delegations {
      amount
    }
    stake {
      currentStakeAvailable
    }
  }
}
```

Before the TX, you'd see: 
```json
{
  "data": {
    "party": {
      "id": "9ba9def5e752c72cef91e0515b1c9f33b604e7475a36f4c049b74b2b0c5fad8c",
      "delegations": null,
      "stake": {
        "currentStakeAvailable": "100"
      }
    }
  }
}
```

and after you'd see: 
```json
{
  "data": {
    "party": {
      "id": "9ba9def5e752c72cef91e0515b1c9f33b604e7475a36f4c049b74b2b0c5fad8c",
      "delegations": [{ "amount": "100" }],
      "stake": {
        "currentStakeAvailable": "0"
      }
    }
  }
}
```

protocol/0050-EPOC-epochs.md
edd: Yeah I'll squish these together/remove one

protocol/0050-EPOC-epochs.md
edd: Will do...

protocol/0050-EPOC-epochs.md
ze97286: hmm, I don't think that's correct. IIRC currentStakeAvailable just tells you how much you've associated, it doesn't do the maths to check how much is associated vs how much is nominated.

edd: Alright. Is there any valid way to test that the balance is reduced when a validator has been nominated, but that that the delegation has not yet occurred due to the fact the Epoch has not yet changed?

ze97286: the best way is to try to double spend it and fail. Once your delegation transaction is processed you can query your delegation balance for the next epoch and it should reflect this balance.

edd: I'll put that instead. It feels like there might be a data node API that can tidy this up a bit.

protocol/0050-EPOC-epochs.md
edd: Thanks, updated

protocol/0050-EPOC-epochs.md
edd: I've simplified this one. The more detailed criteria were trying to describe that if I send a delegation, then reverse it with undelegation, at epoch changeover nothing changes for anyone. But reading the criteria back, they were *more* confusing than just saying that if delegate and undelegate happens in the same epoch, the rewards amount will be unchanged for the validator.

protocol/0050-EPOC-epochs.md
edd: It reads from the title above that `Undelegate Now`  and `Undelegate in anger` are both txs on Vega, so I don't understand what you mean - or do you mean that the effect is the same whether they unassociate on Ethereum OR if they use `Undelegate in anger`?

Either way I think this can be a follow up ticket - is that OK?

PR #816, opened 2022-01-31
Rewards and Net Treasury simplification for oregon trail and beyond

### LINKED ISSUE #800: Rewards and network treasury simplification
## Overview

To be ready for trading as quickly as possible we want to simplify the design of the reward system, and the network treasury accounts and transfers that support it. This ticket is for discussion of a proposed revised scope and also serves as the ticket to implement the agreed changes (once agreed) in the specs. 

It should only be closed when engineering (including @vegaprotocol/core and @vegaprotocol/qa) are happy with the updated specs. 

The below proposal affects/supersedes specs `0055-TREA`, `0056-REWA`, `0057-REWF`, and impacts WIP spec `0067-transfers` ([PR](https://github.com/vegaprotocol/specs-internal/pull/779)).

## Principles and approach

In order to simplify the design as far as possible, we have looked to drop entire features if possible and otherwise remove transactions, configurability, and a calculations. The following is a summary of the simplifications in this proposal:

- We no longer create or control rewards with transactions and there are no per-reward-scheme-instance parameters
- All rewards therefore now implicitly exist for all markets/assets
- We now only ever have one reward scheme per reward type + settlement asset + reward payout asset combination
- All rewards are now calculated once per epoch with the same epochs as validators / staking
- We always pay out the FULL reward pool each epoch
- All new rewards being added are now based on a simple sum of already available data
- We longer have governance controlled transfers in scope. It will therefore initially be impossible to pay out of network treasury accounts, and they exist ONLY as the place that the insurance pool balance goes to when a market closes). Note: this means we will no longer top up validator rewards to 0x0 network treasury but instead send them to a Vega pubkey and use transfer transactions to move from the pubkey to the validator reward pool

## Summary of new scope / requirements

- **Network treasury account** (per asset) exists and can be paid INTO by depositing the asset to 0x0 as well as via transfers and from insurance pools at market closing. At this point funds can’t be paid OUT of the network treasury for any asset.
- **Transfers** from a Vega pubkey general account ONLY
    - To: another (different) pubkey or the network treasury or any valid reward pool account
    - Created by transactions to the network going through consensus (in future governance proposals will be used to create transfers from network treasury or reward pool accounts to other accounts, for now that is not possible)
    - One-off transfers
        - Amount
        - From/to accounts
        - Date/time (amount is taken immediately and paid out on date)
    - Regular transfers (maximum of 1 per from and to account pair)
        - Start epoch
        - End epoch
        - Start amount
        - From/to accounts
        - Factor
        - Before start epoch, after end epoch 0 is transferred at beginning of epoch (cancelled after end epoch)
        - In between, amount is paid at the start of each epoch where: amount = start amount * factor ^ (current epoch - start epoch)
        - Change by sending a new regular transfer for same from and to accounts
        - Cancel by sending one that sets the amount to 0
    - Transaction initiated/controlled (from = pubkey)
- **Rewards general framework**
    - Reward fund account (per reward + settlement asset + reward payout asset combo except for validator rewards which is just one per reward payout asset)
    - Count metric:
        - value of X at end of period (e.g. validator score)
        - sum of X over period (i.e. fees)
    - Period = 1 epoch
    - Distribute full reward pool account per 1 epoch pro rata 
- **Rewards needed for MVP**
    - Value of validator score (already exists)
    - Sum of fees paid (accross all markets for an asset)
    - Sum of maker/taker fees received (accross all markets for an asset)
    - Sum of LP fees received (accross all markets for an asset)
    - For any epoch if lifetime sum of trade value for fee purposes for any markets goes over network_param * quantum FOR THE FIRST TIME, the original proposers of those markets share the reward pool equally (if none, no payout). No market ever gets paid out more than once. (Across all markets for an asset.)

## Next steps

We need to validate these requirements, assess the complexity and impact to build and QA them, and then get specs updated (including acceptance criteria) and review by @vegaprotocol/qa and @vegaprotocol/core.

davidsiska-vega:
This should implement the proposal in this issue https://github.com/vegaprotocol/specs-internal/issues/800 and closes #800 when done. It assumes that https://github.com/vegaprotocol/specs-internal/pull/779 will cover transfers and periodic transfers.


protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,161 +1,165 @@
...
 
-A reward type may have one or more parameters that can be defined for each reward scheme of that type. These parameters may be used by the calculation logic to control how the split of rewards is decided between eligible participants.
+There will be *one* reward account per every Vega asset (settlement asset) and per every reward metric per every Vega asset (reward asset). 
+Any asset on Vega can be either settlement asset or reward asset or both. 
 
+It must be possible for any party to run a one off [transfer](????-????-transfers.md) or create a [periodic transfer](????-????-transfers.md) to any of these reward accounts. 
+Note that saying "per every Vega asset" twice above isn't a typo. We want to be able to pay rewards e.g. in $VEGA for markets settling in e.g. $USDT. 
 
-### Calculation method
+Reward account balances are to be summed per reward asset and added to the network treasury account for the asset when a [LNL checkpoint](../non-protocol-specs/0005-limited-network-life.md) is made.
```
ze97286: That’s not true. Reward accounts (both network treasury for the asset as well as the other types of reward accounts for the asset are actually included in the LNL checkpoint so nothing is being transferred to the network treasury at the time of taking a checkpoint. This is only done for market insurance accounts,

ze97286 (approved):

PR #833, opened 2022-02-02
Add acceptance criteria for limits

### LINKED ISSUE #831: 0003-limits-aka-training-wheels.md needs acceptance criteria
As training wheels is on the critical path for 🤠 we need to have ACs added to the follwoing spec:

[Non-protocol spec](https://github.com/vegaprotocol/specs-internal/blob/master/non-protocol-specs/0003-limits-aka-training-wheels.md)

Other considerations that would be good when adding ACs:

- Adding spec CODES to the non-protocol specs (except LNL as that may move to /protocol)
- Updating scripts so we get coverage from protocol and non-protocol (if not there already)

edd:
Added Acceptance Criteria for how the testnet limits work

- Added 10 ACs
  - 2 relate to the core functionality (this feature intentionally put most of the work on to the smart contract) - they control when Assets and Markets can be created
  - 8 relate to smart contract functionality, which is already built

Slightly weird edge cases: this is the first 'non protocol' spec that we have deemed requires testing. I wrestled with moving it in to the protocol, and keeping it out of the protocol. I went with the latter. It means that the AC codes in *this* file are actually codes from *another* file, which is weird. But it does mean that the coverage metrics can track this, which is the goal for now, so that's why I went with this. 
 
Closes #831


edd:
This is no longer WIP. It's pretty uncontroversial. @cpuja84 could you check if you're happy with how I put Acceptance Criteria in a non protocol spec? It's a bit of a weird case - it's sort of the balance of "Should this just be a spec in core" and "should it stay in this other folder". I went with the latter option, but the two Core specific ACs fit nicely in to the existing Governance spec, and the smart contract tests are... well I don't know.

But see what you think

cpuja84:
@edd - I'm happy with the approach you have taken with ACs and they look great.
For Smart Contract Acceptance Criteria, may I suggest the choice of following ACs so that we atleast are able to track them. What are your thoughts ?

0003-limits-smartcontracts-0005
0003-limits-0005
0003-LATW-smartcontracts-0005
0003-LATW-0005

protocol/0028-GOVE-governance.md
```diff
@@ -363,9 +363,13 @@ APIs should also exist for clients to:
 
 
 ## Governance proposal types
+### New Asset proposals
+- [ ] New asset proposals cannot be created before [`governance.proposal.asset.createFrom`](../non-protocol-specs/0003-limits-aka-training-wheels.md) is in the past (<a name="0028-GOVE-025" href="#0028-GOVE-025">0028-GOVE-025</a>)
```
cpuja84: Can we fix the create form links in both places, its broken currently.

cpuja84 (changes requested):
Smart Contract tests are not numbered, rest everything looks good

protocol/0028-GOVE-governance.md
edd: Done

edd:
I've added Acceptance Criteria tags (0003-NP-LIMI-001 onwards) to the smart contract ACs. Thanks for taking the time to remind me what we had agreed.

protocol/0028-GOVE-governance.md
cpuja84: Link still returns 404

cpuja84 (approved):
All sorted, happy to approve and merge it

PR #838, opened 2022-02-04
changes to support the change in core that validator with no self sta…

ze97286:
…ke gets nothing nor its delegators


gordsport:
@ze97286 - Reviewed in the protocol design meeting, should this actually be put into the `verified` directory in the vega repo?

ze97286:
> @ze97286 - Reviewed in the protocol design meeting, should this actually be put into the `verified` directory in the vega repo?

This should just be closed. A copy of this already exists in the verified in vega.

Issue #34, opened 2020-03-27
As a trader, I enter my name as party id to differentiate my trades from other people's trades


ashleyvega:
In GitLab by @tamlyn10 on Jul 12, 2018, 16:14

null

ashleyvega:
In GitLab by @tamlyn10 on Jul 12, 2018, 16:17

closed

ashleyvega:
In GitLab by @tamlyn10 on Jul 12, 2018, 16:17

reopened

ashleyvega:
In GitLab by @edd on Aug 17, 2018, 11:35

closed

PR #842, opened 2022-02-07
Add testable acceptance criteria to 0034-PROB

Jiajia-Cui:
The acceptance criteria we had before were impossible to test for. 

This PR takes a feature test from `vega`, moves it to `qa-scenarios` and makes it into the one and only acceptance criteria for this spec.


qa-scenarios/verify-complex-shape-order-sizes.feature
```diff
@@ -0,0 +1,112 @@
...
+      | party3 | ETH/DEC19 | buy  | 50     | 12000005 | 0                | TYPE_LIMIT | TIF_GTC | party3-9  |
+      | party3 | ETH/DEC19 | buy  | 50     | 12000004 | 0                | TYPE_LIMIT | TIF_GTC | party3-10 |
+      | party3 | ETH/DEC19 | buy  | 50     | 12000003 | 0                | TYPE_LIMIT | TIF_GTC | party3-11 |
+      | party3 | ETH/DEC19 | buy  | 50     | 12000002 | 0                | TYPE_LIMIT | TIF_GTC | party3-12 |
+      | party3 | ETH/DEC19 | buy  | 50     | 12000001 | 0                | TYPE_LIMIT | TIF_GTC | party3-13 |
+      | party3 | ETH/DEC19 | buy  | 167    | 12000000 | 0                | TYPE_LIMIT | TIF_GTC | party3-14 |
+
+    Then the order book should have the following volumes for market "ETH/DEC19":
+      | side | price    | volume |
+      | sell | 12000020 | 185    |
```
witgaw: @Jiajia-Cui do you understand how we end up with volume of `185` at price `12000020`? Is see `17` from `party-lp-1` and `167` from `party3`, so not sure where the extra 1 is coming from.

witgaw: ah fine, there's `1` from line 24 (`party2`)

witgaw (changes requested):
I don't think this should be tested without price monitoring bounds. 

I'm also not able to make intuitive sense of the numbers, not sure why buy LP orders have higher volumes (compared to sells) even though there's more of them.and their are closer to mid. 

Would be nice to be able to observe that:
1. If there's more orders on one side their individual sizes are larger (probably need a big difference, like 2 as many buys vs sells).
2. The order sizes increase as the offset increases (larger offset steps might be needed)

davidsiska-vega:
> not sure why buy LP orders have higher volumes (compared to sells) even though there's more of them and they are are closer to mid.

I think it's because:
```
volume = ceiling(liquidity_obligation x liquidity-normalised-proportion / probability_of_trading / price)
```
So (all other things being equal) sell side -> higher price -> less volume (and rounding of course exacerbates this).

qa-scenarios/verify-complex-shape-order-sizes.feature
```diff
@@ -0,0 +1,359 @@
...
+    And the order book should have the following volumes for market "ETH2/MAR22":
+      | side | price | volume |
+      | sell | 1359  | 217206866  |
+      | sell | 1109  | 1  |
+      | sell | 909   | 0  |
+      | buy  | 1000  | 0  |
+      | buy  | 728   | 48675135  |
+      | buy  | 900   | 1  |
+
+Scenario:  Test LP Volume being pushed by limit of Probability of Trading (capped at 1e-8);
```
witgaw: Capped -> floored I think. If you agree then please also correct that in comments below.

qa-scenarios/volume-near-price-mon-bounds-2.feature
```diff
@@ -0,0 +1,120 @@
+# Test volume and margin when LP volume is pushed inside price monitoring bounds
```
witgaw: I think this whole file can be deleted, seems like those scenarios have already been copied / covered above.

qa-scenarios/volume-near-price-mon-bounds.feature
```diff
@@ -0,0 +1,203 @@
+# Test volume and margin when LP volume is pushed inside price monitoring bounds
```
witgaw: I think this whole file can be deleted, seems like those scenarios have already been copied / covered above.

qa-scenarios/verify-complex-shape-order-sizes.feature
```diff
@@ -0,0 +1,359 @@
...
+
+    Then the liquidity provisions should have the following states:
+      | id  | party       | market    | commitment amount | status       |
+      | lp1 | party-lp-1 | ETH/DEC19 | 1000000000        | STATUS_ACTIVE |
+
+    And the trading mode should be "TRADING_MODE_CONTINUOUS" for the market "ETH/DEC19"
+
+    And the orders should have the following states:
+      | party      | market id | side | volume | price    | status        |
+      | party-lp-1 | ETH/DEC19 | sell | 17     | 12000014 | STATUS_ACTIVE |
```
davidsiska-vega: Can we choose bigger offsets here so that we see the volume increasing as we go away from best bid / best ask?

protocol/0034-PROB-prob_weighted_liquidity_measure.ipynb
```diff
@@ -69,48 +74,39 @@
...
-    "3. More details please see features tests: integration/features/liquidity-provisin \n",
     "\n"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "\" href=\"#0034-PROB-004\">0034-PROB-004</a>)\n"
+    "## Acceptance Criteria\n",
```
davidsiska-vega, line 84: Can all of these have their own label so that this spec has more than one acceptance criteria.

davidsiska-vega (dismissed):
Two small things.

qa-scenarios/verify-complex-shape-order-sizes.feature
witgaw: Did you try to modify prices or offsets to see volumes increase as orders go further away from the mid? That's the main reason for "probability of trading" really so would be good to know if that's not the case or it's difficult to observe that in sensible markets.

witgaw (changes requested):

protocol/0034-PROB-prob_weighted_liquidity_measure.ipynb
```diff
@@ -69,48 +74,39 @@
...
+    "\n",
+    "1. Order from liquidity provision and from normal order submission are correctly cumulated in order book's total size(<a name=\"0034-PROB-001\" href=\"#0034-PROB-001\">0034-PROB-001</a>);\n",
+    "\n",
+    "2. LP pegged volume is pushed inside price monitoring bounds(<a name=\"0034-PROB-002\" href=\"#0034-PROB-002\">0034-PROB-002</a>);\n",
+    "\n",
+    "3. LP pegged volume is pushed by Price Monitoring lower bound(<a name=\"0034-PROB-003\" href=\"#0034-PROB-003\">0034-PROB-003</a>);\n",
+    "\n",
+    "4. LP Volume being pushed by limit of Probability of Trading (capped at 1e-8)(<a name=\"0034-PROB-004\" href=\"#0034-PROB-004\">0034-PROB-004</a>).\n",
+    "\n"
    ]
```
witgaw, line 100: "5. Probability of trading decreases away from the mid-price (<a name=\"0034-PROB-005\" href=\"#0034-PROB-005\">0034-PROB-005</a>).\n",
    "\n"

witgaw (changes requested):
If you could just uncomment the remaining scenarios in `lp-measure-test.feature` and add acceptance criteria tags (e.g. see how it was done here: https://github.com/vegaprotocol/specs-internal/blob/7b7bc7a15c1f907787dec0b0ea13507fbb2fbfcc/qa-scenarios/0059-delegation.feature#L60 
that will be ready to go in I think.

qa-scenarios/lp-measure-test.feature
```diff
@@ -0,0 +1,323 @@
+Feature: Verify the order size is correctly cumulated (0034-PROB-001); Probability of trading decreases away from the mid-price (0034-PROB-005)
```
witgaw: The description from here should probably go to line 13 (the scenario), this is the description of entire feature file, so probably something along the lines of "Tests confirming probability of trading acceptance criteria" or whatever else you think makes sense.

witgaw (changes requested):

protocol/0034-PROB-prob_weighted_liquidity_measure.ipynb
Jiajia-Cui: comments been addressed

qa-scenarios/verify-complex-shape-order-sizes.feature
Jiajia-Cui: comments been addressed

witgaw (approved):
LGTM, thanks a lot for all the edits - sorry if I made it seem more tedious then necessary but I think it looks great now and we won't have to revisit it for a good while (if at all), so hopefully less work in the long run.

PR #847, opened 2022-02-08
Pegged order update

### LINKED ISSUE #846: Further update to IOC/FOK orders
Due to the restriction of offsets being positive only, any PEGGED orders placed with an IOC or FOK TIF will be stopped as it is impossible for them to cross. This ticket is to update the spec to make it clear we will not currently support this order combo but it can be brought back at a later time when negative offsets are re-introduced.

peterbarrow:
Update for pegged orders

Closes #846


edd:
Fix conflict, reviewing...

edd (approved):

PR #860, opened 2022-02-10
Add ‘parked’ to order status, dedupe some ACs

edd:
- Add (short) section on parked orders to Order Status spec 
- Move Acceptance Criteria to the bottom of the file, which is the more common place
- Deduplicate ACs that are the same as the ones in the matching engine spec
- Add percentages to the reference script

Closes #858

PR #863, opened 2022-02-11
Updated the acceptance criterias for 0002-STTL-settlement

### LINKED ISSUE #822: 0002-STTL-settlement needs acceptance criteria
In order to get QA test coverage [this spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0002-STTL-settlement.md) needs to have clear acceptance criteria defines with acceptance IDs.

cpuja84:
Updated the acceptance criterias for 0002-STTL-settlement spec


edd (approved):
Great criteria. 

**thought (non-blocking)** I would like to do another sweep through at some point and more heavily link specs together. I did it a bit over in #860 but (for example) you say 'as per collection and distribution methods above' and 'transfers for collateral', I feel like there will be a value in linking those to where the thing is defined. It's not worth doing here yet (it would just make this PR harder to read) but eventually it would let us map relationships between specs:

spec-relationships

Which might be a useful indicator for 'also check you haven't broken this spec when you update this other one'

It also might not. Hence the non-blocking tag!

Issue #75, opened 2020-03-27
UX workshop preparation


ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:15

Next week we have a UX designer come in, and we’re specifically going to be looking at either onboarding new users, or product discovery and selection. Whichever of these gets the most attention, we have a couple of tasks to make the front-end architecture ready for some aspects of this

# Tasks
- Remove hardcoded references to workspace[0] in GraphQL queries
- Styling consistency/tidyup
- Persisting layout changes between browser reloads

ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:16

marked this issue as related to client#130

ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:16

marked this issue as related to client#129

ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:17

marked this issue as related to client#126

ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:17

changed the description

ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:17

assigned to @eddhannay and @mattrussell36

ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:18

marked this issue as related to client#99

ashleyvega:
In GitLab by @edd on Nov 27, 2018, 10:18

changed the description

ashleyvega:
In GitLab by @edd on Dec 3, 2018, 12:16

closed

PR #874, opened 2022-02-14
add base protocol upgrade spec

### LINKED ISSUE #1160: Write/update the protocol upgrade spec
When starting the new protocol upgrade spec we had the following document:
 - https://github.com/vegaprotocol/specs/pull/874
 
 The details in this have now been superseded with the latest design/implementation
 
 This issue is to cover documenting the current process in a spec

jeremyletang:
Base specs for protocol upgrades without LNL restarts.

### DIFF
--- a/protocol/0070-PLUP_protocol_upgrades.md
+++ b/protocol/0070-PLUP_protocol_upgrades.md
@@ -12,33 +12,33 @@ Using LNL have the following unconvenient side effect:
 
 ### How different protocol proceed
 
-Other protocol e.g Ethereum proceed to updates in an asynchronous manner. Usually a new version of the protocol is made available, an hardcoded block height is set
-at which a new code path will be enabled. If enough nodes runner have deployed the updated code, then the network will continue with the new code path. Others nodes
+Other protocol e.g Ethereum carry out updates in an asynchronous manner. Usually a new version of the protocol is made available, a hardcoded block height is set
+at which a new code path will be enabled. If enough node runners have deployed the updated code, then the network will continue with the new code path. Others nodes
 which haven't updated will then fork the blockchain.
 
 ### Prior example of updgrading the nodes asynchronously
 
 Back in December 2020, vega proceeded to a LNL restore. Unfortunately, a bug in the code prevented the dispatch of the network parameters after the restore. This left
-the network in a semi invalid state where the network parameters where defined by a the ones from the genesis block instead of the one from the checkpoint from the previous network.
+the network in a semi invalid state where the network parameters were defined by the ones from the genesis block instead of the one from the checkpoint from the previous network.
 
 The solution employed at the time was to:
 - implement a hotfix
-- keep the hotfix behing a guard until a given time, once the time is reachd network parameters would be dispatched.
+- keep the hotfix behind a guard until a given time, once the time is reached network parameters would be dispatched.
 - distribute the code to the validator so they can test it.
 - then decide of an actual date which would give a week for the validator to asynchronously update their node.
 
 This upgrade went quite smoothly without any incident, although it was possible as the fix was located in a single place, and wasn't impacting much of the state
-(meaning no changes in the protobug were required for example.)
+(meaning no changes in the protobuf files were required for example.)
 
 ## Asynchronous upgrades mecanism
 
 One of the main challenge with upgrading the network both asynchronously and without a restart / downtime is
-the management of the state, and possible uncompatible changes between the state of two version of the protocol.
+the management of the state, and possible incompatible changes between the state of two version of the protocol.
 
-In order to prevent any state incompatiblity introduce a new framework to propose upgrades of the protocol, this will allow the node to be stopped,
-and restart from the block it stopped using the snapshots or to replay the whole chain. When being restarted
+In order to prevent any state incompatibility, this specification introduces a new framework to propose upgrades of the protocol, which will allow the node to be stopped,
+and restart from the block it stopped at using the snapshots or by replaying the whole chain. When being restarted
 the new node will contain both the previous version of the protocol in use, but also a new version that the
-network should start using at a given block. Upong restart, the node will try to send a transaction to
+network should start using at a given block. Upon restart, the node will try to send a transaction to
 indicate the rest of the network that it's ready to switch to the new version of the protocol. When the block
 defined for the upgrade is reached and if enough nodes have send the transaction notifying of their intent of
 migrating to the new version of the protocol, then the upgrade is applied. If not enough transaction have been
@@ -49,19 +49,19 @@ The existing network parameter `"validators.vote.required"` is used to defined t
 to proceed with the update.
 
 
-The upgrade requires to be done in an atomic way. To do so, the core node will need to have been compiler with
+The upgrade **must** be done in an atomic way. To do so, the core node will need to have been compiled with
 the current and next version of the protocol. Once the block height for the upgrade is reached the currently
 running version of the protocol will stop executing any transaction (e.g: at the ABCI Commit call), then the
-whole state will be serialized, all the engine shut down, and the new version of the protocol will be
-instantiated, and will have to load the state previously serialized by the old version. The new version of
-the protocol will be in charged to adapt / convert the previous state in order to make it compatible with
+whole state will be serialised, all of the engines will shut down, and the new version of the protocol will be
+instantiated, and will have to load the state previously serialised by the old version. The new version of
+the protocol will be then to convert the previous state in order to make it compatible with
 any changes in the protocol.
 
-The most main difficulty here will be the implementation of the serialisation of the whole state of the node.
-Fortunately we already have such feature which should work out of the box with minimal changes: the snapshots.
+The main difficulty here will be the implementation of the serialisation of the whole state of the node.
+Fortunately we already have such feature which should work out of the box with minimal changes: [snapshots](../non-protocol-specs/0009-NP-SNAP-snapshots.md).
 
 
-## Versionning of the protocol
+## Versioning of the protocol
 
 From now on we will need to version the protocol, this can be the current version of the node, and the previous
 version. This should probably use a versioning system like semantic versioning (this will be useful in a later
@@ -84,25 +84,25 @@ Each of those upgrade will go through a different process.
 
 ### Patch upgrade
 
-This upgrade should represent very small change in the protocol, eventually a bug fix, or QOL improvement.
-These upgrade require no involvement from any actors of the network apart from the validator to update their node
+This upgrade should represent very small change in the protocol, for example a bug fix or a quality-of-life improvement.
+These upgrade require no involvement from any actors of the network apart from the validators updating their nodes
 with a new supported version of the protocol.
 
 
 ### Minor upgrade
 
-These are upgrade which brings more features to the protocol, eventual fixes as well, and changes in the state
-produce by the protocol. They may require more works from the validators and will not be applied automatically.
+These are upgrade which bring more features to the protocol, or larger fixes, and changes in the state
+produced by the protocol. They may require more work from the validators and will not be applied automatically.
 A new command will be introduce in the vega toolchain so validator can explicitly notify the network of their
 support into upgrading to the new version.
 
 ### Major upgrade
 
 These are important changes in the protocol, possibly breaking the previously known behaviours of the protocol
-and bringing also new important feature. Due to the nature of the upgrade, a governance vote will be required
+but also implementing new important features. Due to the nature of the upgrade, a governance vote will be required
 so token holders show of their support for the upgrade.
 
-> note: the major upgrade will need refinment, and are not planned at first.
+> note: the definition of a major upgrade will need refinement, and the current version of the specification does not cover it.
 
 ## Framework / data structures
 
@@ -127,7 +127,7 @@ message UpgradeAccepted {
 
 ## Acceptance criteria
 - [ ] A new patch upgrade is made available
-  - [ ] Enough validator have upgraded their node
+  - [ ] Enough validators have upgraded their node
   - [ ] At least `validators.vote.required` message have been sent by the validators
   - [ ] The block time for the upgrade is reached
   - [ ] The network is switching to the new version of the protocol (can be checked with APIs)
--- a/protocol/0070-PLUP_protocol_upgrades.md
+++ b/protocol/0070-PLUP_protocol_upgrades.md
@@ -102,6 +102,7 @@ These are important changes in the protocol, possibly breaking the previously kn
 and bringing also new important feature. Due to the nature of the upgrade, a governance vote will be required
 so token holders show of their support for the upgrade.
 
+> note: the major upgrade will need refinment, and are not planned at first.
 
 ## Framework / data structures
 
@@ -125,3 +126,35 @@ message UpgradeAccepted {
 ```
 
 ## Acceptance criteria
+- [ ] A new patch upgrade is made available
+  - [ ] Enough validator have upgraded their node
+  - [ ] At least `validators.vote.required` message have been sent by the validators
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network is switching to the new version of the protocol (can be checked with APIs)
+- [ ] A new patch upgrade is made available
+  - [ ] Not enough validator have upgraded their node
+  - [ ] At least `validators.vote.required` message have been sent by the validators nodes
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network haven't change version (can be checked with APIs)
+- [ ] A new minor upgrade is made available
+  - [ ] Enough validator have upgraded their node
+  - [ ] At least `validators.vote.required` validators sent a command to accept the upgrade
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network is switching to the new version of the protocol (can be checked with APIs)
+- [ ] A new minor upgrade is made available
+  - [ ] Enough validator have upgraded their node
+  - [ ] Less then `validators.vote.required` validators sent a command to accept the upgrade
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network haven't change the protocol version (can be checked with APIs)
+- [ ] A new minor upgrade is made available
+  - [ ] Enough validator have upgraded their node
+  - [ ] A governance proposal have been created to propose the upgrade
+  - [ ] The proposal have been accepted
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network is switching to the new version of the protocol (can be checked with APIs)
+- [ ] A new minor upgrade is made available
+  - [ ] Enough validator have upgraded their node
+  - [ ] A governance proposal have been created to propose the upgrade
+  - [ ] The proposal have been rejected
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network haven't changed the protocol version (can be checked with APIs)
--- a/protocol/0070-PLUP_protocol_upgrades.md
+++ b/protocol/0070-PLUP_protocol_upgrades.md
@@ -31,7 +31,7 @@ The solution employed at the time was to:
 This upgrade went quite smoothly without any incident, although it was possible as the fix was located in a single place, and wasn't impacting much of the state
 (meaning no changes in the protobuf files were required for example.)
 
-## Asynchronous upgrades mecanism
+## Asynchronous upgrades mechanism
 
 One of the main challenge with upgrading the network both asynchronously and without a restart / downtime is
 the management of the state, and possible incompatible changes between the state of two version of the protocol.
@@ -67,6 +67,17 @@ Fortunately we already have such feature which should work out of the box with m
 From now on we will need to version the protocol, this can be the current version of the node, and the previous
 version. This should probably use a versioning system like semantic versioning (this will be useful in a later
 part of this document).
+
+### Edge cases
+With the upgrade process outlined below, it's possible for an upgrade to be proposed and rejected by validators. This happens independently of the *protocol development* and could lead to a situation in which an upgrade is rejected, and subsequent versions need to be reworked to take this in to account.
+
+For example:
+- In a network of 3 nodes, all running v`0.1.1`
+- v`0.2.0` is released with block height of 100
+- 0 nodes update before block 100, so the new version is not deployed on the network
+- Meanwhile v`0.3.0` is released
+
+The network is now in a situation where `v0.3.0` expects the network to be upgrading from v`0.2.0`, while it would actually be upgrading from v`0.1.0`. This is expected, and upgrades must take this in to account. 
 
 
 ## Types of upgrade
@@ -108,7 +119,7 @@ so token holders show of their support for the upgrade.
 ## Framework / data structures
 
 As protobuf:
-```
+```go
 message Upgrade {
 	enum Type {
 		Patch = 1;
@@ -127,35 +138,50 @@ message UpgradeAccepted {
 ```
 
 ## Acceptance criteria
-- [ ] A new patch upgrade is made available
+### Patch version updates (`0.0.x`)
+- [ ] A new patch upgrade is made available, and is successfully deployed (<a name="0070-PLUP-001" href="#0070-PLUP-001">0070-PLUP-001</a>)
   - [ ] Enough validators have upgraded their node
   - [ ] At least `validators.vote.required` message have been sent by the validators
   - [ ] The block time for the upgrade is reached
-  - [ ] The network is switching to the new version of the protocol (can be checked with APIs)
-- [ ] A new patch upgrade is made available
+  - [ ] The network changes to the new version of the protocol (can be checked with APIs)
+- [ ] A new patch upgrade is made available, but not enough validators update before the deadline (<a name="0070-PLUP-002" href="#0070-PLUP-002">0070-PLUP-002</a>)
   - [ ] Not enough validator have upgraded their node
   - [ ] At least `validators.vote.required` message have been sent by the validators nodes
   - [ ] The block time for the upgrade is reached
-  - [ ] The network haven't change version (can be checked with APIs)
-- [ ] A new minor upgrade is made available
-  - [ ] Enough validator have upgraded their node
+  - [ ] The network does not change version (can be checked with APIs)
+ 
+### Minor version updates (`0.x.x`)
+- [ ] A new minor upgrade is made available, validators vote to update, and the update is succesfully deployed (<a name="0070-PLUP-003" href="#0070-PLUP-003">0070-PLUP-003</a>)
+  - [ ] Enough validators have upgraded their node
   - [ ] At least `validators.vote.required` validators sent a command to accept the upgrade
   - [ ] The block time for the upgrade is reached
-  - [ ] The network is switching to the new version of the protocol (can be checked with APIs)
-- [ ] A new minor upgrade is made available
-  - [ ] Enough validator have upgraded their node
+  - [ ] The network has switched to the new version of the protocol (can be checked with APIs)
+- [ ] A new minor upgrade is made available, validators update but not enough votes to upgrade are received (<a name="0070-PLUP-004" href="#0070-PLUP-004">0070-PLUP-004</a>)
+  - [ ] Enough validators have upgraded their node
   - [ ] Less then `validators.vote.required` validators sent a command to accept the upgrade
   - [ ] The block time for the upgrade is reached
-  - [ ] The network haven't change the protocol version (can be checked with APIs)
-- [ ] A new minor upgrade is made available
-  - [ ] Enough validator have upgraded their node
+  - [ ] The network does not change version (can be checked with APIs)
+- [ ] A new minor upgrade is made available, but not enough validators update by the block height (<a name="0070-PLUP-005" href="#0070-PLUP-005">0070-PLUP-005</a>)
+  - [ ] Not enough validators have upgraded their node
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network does not change version (can be checked with APIs)
+ 
+### Major version updates (`x.x.x`)
+ > Note: These acceptance criteria are *incomplete*. There will likely be a governance vote from users, as well as the validator vote. For now these Acceptance Criteria test a major version in the same was a minor version upgrade.
+- [ ] A new major upgrade is made available, validators vote to update, and the update is successfully deployed (<a name="0070-PLUP-006" href="#0070-PLUP-006">0070-PLUP-006</a>)
+  - [ ] Enough validators have upgraded their node
   - [ ] A governance proposal have been created to propose the upgrade
   - [ ] The proposal have been accepted
   - [ ] The block time for the upgrade is reached
-  - [ ] The network is switching to the new version of the protocol (can be checked with APIs)
-- [ ] A new minor upgrade is made available
-  - [ ] Enough validator have upgraded their node
+  - [ ] The network has switched to the new version of the protocol (can be checked with APIs)
+- [ ] A new major upgrade is made available, but the validator vote does not pass (<a name="0070-PLUP-007" href="#0070-PLUP-007">0070-PLUP-007</a>)
+  - [ ] Enough validators have upgraded their node
   - [ ] A governance proposal have been created to propose the upgrade
   - [ ] The proposal have been rejected
   - [ ] The block time for the upgrade is reached
-  - [ ] The network haven't changed the protocol version (can be checked with APIs)
+  - [ ] The network does not change version (can be checked with APIs)
+- [ ] A new major upgrade is made available, but not enough validators update by the block height (<a name="0070-PLUP-008" href="#0070-PLUP-008">0070-PLUP-008</a>)
+  - [ ] Not enough validators have upgraded their node
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network does not change version (can be checked with APIs)
+ 
--- a/protocol/0070-PLUP_protocol_upgrades.md
+++ b/protocol/0070-PLUP_protocol_upgrades.md
@@ -17,9 +17,13 @@ Other protocol e.g Ethereum carry out updates in an asynchronous manner. Usually
 at which a new code path will be enabled. If enough node runners have deployed the updated code, then the network will continue with the new code path. Others nodes
 which haven't updated will then fork the blockchain.
 
+Cosmos is using a small program called cosmovisor. The cosmovisor is listening to proposals and when a proposal for upgrade is accepted, it is preparing the binary for a new release and handling the upgrade - https://docs.cosmos.network/main/run-node/cosmovisor.html.
+
+The idea in this spec is to draw inspiration from the design of cosmovisor and build vega-visor to manage protocol upgrades. 
+
 ### Prior example of updgrading the nodes asynchronously
 
-Back in December 2020, vega proceeded to a LNL restore. Unfortunately, a bug in the code prevented the dispatch of the network parameters after the restore. This left
+Back in December 2021, vega proceeded to a LNL restore. Unfortunately, a bug in the code prevented the dispatch of the network parameters after the restore. This left
 the network in a semi invalid state where the network parameters were defined by the ones from the genesis block instead of the one from the checkpoint from the previous network.
 
 The solution employed at the time was to:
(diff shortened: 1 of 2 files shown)


protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+the network in a semi invalid state where the network parameters where defined by a the ones from the genesis block instead of the one from the checkpoint from the previous network.
+
+The solution employed at the time was to:
+- implement a hotfix
+- keep the hotfix behing a guard until a given time, once the time is reachd network parameters would be dispatched.
+- distribute the code to the validator so they can test it.
+- then decide of an actual date which would give a week for the validator to asynchronously update their node.
+
+This upgrade went quite smoothly without any incident, although it was possible as the fix was located in a single place, and wasn't impacting much of the state
+(meaning no changes in the protobug were required for example.)
```
edd:
```suggestion
(meaning no changes in the protobuf files were required for example.)
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+which haven't updated will then fork the blockchain.
+
+### Prior example of updgrading the nodes asynchronously
+
+Back in December 2020, vega proceeded to a LNL restore. Unfortunately, a bug in the code prevented the dispatch of the network parameters after the restore. This left
+the network in a semi invalid state where the network parameters where defined by a the ones from the genesis block instead of the one from the checkpoint from the previous network.
+
+The solution employed at the time was to:
+- implement a hotfix
+- keep the hotfix behing a guard until a given time, once the time is reachd network parameters would be dispatched.
```
edd:
```suggestion
- keep the hotfix behind a guard until a given time, once the time is reached network parameters would be dispatched.
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+### How different protocol proceed
+
+Other protocol e.g Ethereum proceed to updates in an asynchronous manner. Usually a new version of the protocol is made available, an hardcoded block height is set
+at which a new code path will be enabled. If enough nodes runner have deployed the updated code, then the network will continue with the new code path. Others nodes
+which haven't updated will then fork the blockchain.
+
+### Prior example of updgrading the nodes asynchronously
+
+Back in December 2020, vega proceeded to a LNL restore. Unfortunately, a bug in the code prevented the dispatch of the network parameters after the restore. This left
+the network in a semi invalid state where the network parameters where defined by a the ones from the genesis block instead of the one from the checkpoint from the previous network.
```
edd:
```suggestion
the network in a semi invalid state where the network parameters were defined by the ones from the genesis block instead of the one from the checkpoint from the previous network.
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+### Current state of upgrading the vega network
+
+As of today, upgrading the protocol is near impossible when a major changes to the step are possible without proceeding with an LNL restore.
+Using LNL have the following unconvenient side effect:
+- A synchronous restart is required
+- All node need to be restarted in a very short time so all state can be restore from ethereum, and the network can start properly with a checkpoint.
+
+### How different protocol proceed
+
+Other protocol e.g Ethereum proceed to updates in an asynchronous manner. Usually a new version of the protocol is made available, an hardcoded block height is set
```
edd:
```suggestion
Other protocol e.g Ethereum carry out updates in an asynchronous manner. Usually a new version of the protocol is made available, a hardcoded block height is set
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+at which a new code path will be enabled. If enough nodes runner have deployed the updated code, then the network will continue with the new code path. Others nodes
```
edd:
```suggestion
at which a new code path will be enabled. If enough node runners have deployed the updated code, then the network will continue with the new code path. Others nodes
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+- distribute the code to the validator so they can test it.
+- then decide of an actual date which would give a week for the validator to asynchronously update their node.
+
+This upgrade went quite smoothly without any incident, although it was possible as the fix was located in a single place, and wasn't impacting much of the state
+(meaning no changes in the protobug were required for example.)
+
+## Asynchronous upgrades mecanism
+
+One of the main challenge with upgrading the network both asynchronously and without a restart / downtime is
+the management of the state, and possible uncompatible changes between the state of two version of the protocol.
```
edd:
```suggestion
the management of the state, and possible incompatible changes between the state of two version of the protocol.
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+
+In order to prevent any state incompatiblity introduce a new framework to propose upgrades of the protocol, this will allow the node to be stopped,
```
edd:
```suggestion
In order to prevent any state incompatibility, this specification introduces a new framework to propose upgrades of the protocol, which will allow the node to be stopped,
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+and restart from the block it stopped using the snapshots or to replay the whole chain. When being restarted
```
edd:
```suggestion
and restart from the block it stopped at using the snapshots or by replaying the whole chain. When being restarted
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+the new node will contain both the previous version of the protocol in use, but also a new version that the
+network should start using at a given block. Upong restart, the node will try to send a transaction to
```
edd:
```suggestion
network should start using at a given block. Upon restart, the node will try to send a transaction to
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+defined for the upgrade is reached and if enough nodes have send the transaction notifying of their intent of
+migrating to the new version of the protocol, then the upgrade is applied. If not enough transaction have been
+sent by the validators, the protocol will keep running the current version.
+
+
+The existing network parameter `"validators.vote.required"` is used to defined the amount of require validators
+to proceed with the update.
+
+
+The upgrade requires to be done in an atomic way. To do so, the core node will need to have been compiler with
```
edd:
```suggestion
The upgrade **must** be done in an atomic way. To do so, the core node will need to have been compiled with
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+the current and next version of the protocol. Once the block height for the upgrade is reached the currently
+running version of the protocol will stop executing any transaction (e.g: at the ABCI Commit call), then the
+whole state will be serialized, all the engine shut down, and the new version of the protocol will be
```
edd:
```suggestion
whole state will be serialised, all of the engines will shut down, and the new version of the protocol will be
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+instantiated, and will have to load the state previously serialized by the old version. The new version of
```
edd:
```suggestion
instantiated, and will have to load the state previously serialised by the old version. The new version of
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+the protocol will be in charged to adapt / convert the previous state in order to make it compatible with
```
edd:
```suggestion
the protocol will be then to convert the previous state in order to make it compatible with
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+any changes in the protocol.
+
+The most main difficulty here will be the implementation of the serialisation of the whole state of the node.
```
edd:
```suggestion
The main difficulty here will be the implementation of the serialisation of the whole state of the node.
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+Fortunately we already have such feature which should work out of the box with minimal changes: the snapshots.
```
edd:
```suggestion
Fortunately we already have such feature which should work out of the box with minimal changes: [snapshots](../non-protocol-specs/0009-NP-SNAP-snapshots.md).
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+
+
+## Versionning of the protocol
```
edd:
```suggestion
## Versioning of the protocol
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+For the rest of this assume that upgrade are using a semantic version, where for a given M.m.P version which changes either of:
+- M is a major upgrade
+- m is a minor upgrade
+- P is a patch upgrade.
+
+Each of those upgrade will go through a different process.
+
+### Patch upgrade
+
+This upgrade should represent very small change in the protocol, eventually a bug fix, or QOL improvement.
```
edd:
```suggestion
This upgrade should represent very small change in the protocol, for example a bug fix or a quality-of-life improvement.
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+These upgrade require no involvement from any actors of the network apart from the validator to update their node
```
edd:
```suggestion
These upgrade require no involvement from any actors of the network apart from the validators updating their nodes
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+with a new supported version of the protocol.
+
+
+### Minor upgrade
+
+These are upgrade which brings more features to the protocol, eventual fixes as well, and changes in the state
```
edd:
```suggestion
These are upgrade which bring more features to the protocol, or larger fixes, and changes in the state
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+produce by the protocol. They may require more works from the validators and will not be applied automatically.
```
edd:
```suggestion
produced by the protocol. They may require more work from the validators and will not be applied automatically.
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+A new command will be introduce in the vega toolchain so validator can explicitly notify the network of their
+support into upgrading to the new version.
+
+### Major upgrade
+
+These are important changes in the protocol, possibly breaking the previously known behaviours of the protocol
+and bringing also new important feature. Due to the nature of the upgrade, a governance vote will be required
```
edd:
```suggestion
but also implementing new important features. Due to the nature of the upgrade, a governance vote will be required
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+so token holders show of their support for the upgrade.
+
+> note: the major upgrade will need refinment, and are not planned at first.
```
edd:
```suggestion
> note: the definition of a major upgrade will need refinement, and the current version of the specification does not cover it.
```

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,160 @@
...
+
+// a message sent by the node to notify that a node is ready to do an upgrade
+message UpgradeAccepted {
+	string version = 1;
+}
+```
+
+## Acceptance criteria
+- [ ] A new patch upgrade is made available
+  - [ ] Enough validator have upgraded their node
```
edd:
```suggestion
  - [ ] Enough validators have upgraded their node
```

edd: Also this one could be clearer what 'enough' means (e.g. 2/3+1 have signalled with a X event)

edd (commented):
Good spec so far! I've gone through and fixed some basic typos, then I'll go back and add some more feedback

edd:
@jeremyletang looks good to me. I've fixed typos and grammatical mistakes, tidied up and coded the acceptance criteria, and added a section that basically says 'If the network rejects an upgrade, subsequent upgrades need to take that in to account'. That last one might be up for debate, but I think this is ready for review by @vegaprotocol/research

edd:
@barnabee In this spec, @jeremyletang added the requirement that the block height is built in to the code, rather than specified at a point where nodes upgrade their software. It feels brittle to me, and specific to a single network rather than letting us do the same upgrade across devnet/testnet/networks running on ethereum with the same upgrade, but at different points. Could you talk about why you want it compiled in?

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+}
+```
+
+## Acceptance criteria
+### Patch version updates (`0.0.x`)
+- [ ] A new patch upgrade is made available, and is successfully deployed (<a name="0070-PLUP-001" href="#0070-PLUP-001">0070-PLUP-001</a>)
+  - [ ] Enough validators have upgraded their node
+  - [ ] At least `validators.vote.required` message have been sent by the validators
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network changes to the new version of the protocol (can be checked with APIs)
```
wwestgarth: Are we going to entertain the idea that if a patch does not contain changes the break consensus, then a set of validators could be running with the same minor version, but will different patch version? Or are we just enforcing that all validators need to be running identical software, even down to the patch number?

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+
+
+The existing network parameter `"validators.vote.required"` is used to defined the amount of require validators
+to proceed with the update.
+
+
+The upgrade **must** be done in an atomic way. To do so, the core node will need to have been compiled with
+the current and next version of the protocol. Once the block height for the upgrade is reached the currently
+running version of the protocol will stop executing any transaction (e.g: at the ABCI Commit call), then the
+whole state will be serialised, all of the engines will shut down, and the new version of the protocol will be
```
wwestgarth: This is just an observation really. If there is a bug in the code that serialising the whole state of vega (snapshots), for example a vital peice of state is not being captured, then we're screwed right? Because we won't be able to migrate that value into the new vega version.

Unless we replay the whole chain with a new version that has *just* the snapshot fix in. Maybe.

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+- All node need to be restarted in a very short time so all state can be restore from ethereum, and the network can start properly with a checkpoint.
+
+Limited Network Life is not the end goal. This spec outlines how the protocol evolves from LNL checkpoints to rolling software updates, controlled by a reasonable set of governance and user controls.
+
+### How other protocols proceed
+
+Other protocol e.g Ethereum carry out updates in an asynchronous manner. Usually a new version of the protocol is made available, a hardcoded block height is set
+at which a new code path will be enabled. If enough node runners have deployed the updated code, then the network will continue with the new code path. Others nodes
+which haven't updated will then fork the blockchain.
+
```
Vegaklaus: Eth may not be the best example, as it has a different behaviour (mainly, if their upgrade causes disagreement., they forc; if ours causes disagreement, the chain stops until sufficient weight supports either version. Maybe refer to how cosmos does it (https://docs.cosmos.network/master/core/upgrade.html) ?

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+instantiated, and will have to load the state previously serialised by the old version. The new version of
+the protocol will be then to convert the previous state in order to make it compatible with
+any changes in the protocol.
+
+The main difficulty here will be the implementation of the serialisation of the whole state of the node.
+Fortunately we already have such feature which should work out of the box with minimal changes: [snapshots](../non-protocol-specs/0009-NP-SNAP-snapshots.md).
+
+
+## Versioning of the protocol
+
```
Vegaklaus: Since we're (at least in theory) not the only provider of code, the version should probably also contain a hash to make sure it's uniquely defined. Otherwise, it would be possible that validators think they agree on the version (as the numbers match), but run code from different providers. Another way would be that Vega still signs the code, and thus gives their version a unique vendor ID (I'm not sure how far we do that now).
The latter might be a good idea anyhow, as that would create a well defined authentication of update code, and we can implement 4-eye security and such.

Vegaklaus: Another point that should be clarified with versioning is the downgrade option - is it ok to downgrade if a new version has a problem, or should someone in that case provide a new version number with the older code (i.e., the code would refuse to update to a lower version number) ? I suppose given that an update code has both versions of the protocol, a downgrade would require a new code blob from us ?
(It might be good if the validators could switch back if needed; however, this would mean that there old version can parse the serialisation of the new one, which would need a extra layer of code and not be overly feasible.

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+In order to prevent any state incompatibility, this specification introduces a new framework to propose upgrades of the protocol, which will allow the node to be stopped,
+and restart from the block it stopped at using the snapshots or by replaying the whole chain. When being restarted
+the new node will contain both the previous version of the protocol in use, but also a new version that the
+network should start using at a given block. Upon restart, the node will try to send a transaction to
+indicate the rest of the network that it's ready to switch to the new version of the protocol. When the block
+defined for the upgrade is reached and if enough nodes have send the transaction notifying of their intent of
+migrating to the new version of the protocol, then the upgrade is applied. If not enough transaction have been
+sent by the validators, the protocol will keep running the current version.
+
```
Vegaklaus: Those transactions are sent through the chain as well to make sure we have consensus ? And, what is enough ? 2/3 of the validator weight (which allows to run, but no longer for failures) or more ?

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+With the upgrade process outlined below, it's possible for an upgrade to be proposed and rejected by validators. This happens independently of the *protocol development* and could lead to a situation in which an upgrade is rejected, and subsequent versions need to be reworked to take this in to account.
+
+For example:
+- In a network of 3 nodes, all running v`0.1.1`
+- v`0.2.0` is released with block height of 100
+- 0 nodes update before block 100, so the new version is not deployed on the network
+- Meanwhile v`0.3.0` is released
+
+The network is now in a situation where `v0.3.0` expects the network to be upgrading from v`0.2.0`, while it would actually be upgrading from v`0.1.0`. This is expected, and upgrades must take this in to account. 
+
```
Vegaklaus: I'm not sure that's compatible with what's described in line 53; given the code has both the old and the new version, and the code needs to be able to parse and transfer the state, jumping an update should not be possible (unless I misunderstood something). 
Maybe it would be an idea to also add a serial number to upgrades or add to every upgrade which version it expects to upgrade from(so they have to come in order), and it is our job to keep track of what the network is doing and add the right serial number/code

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+- [ ] A new minor upgrade is made available, but not enough validators update by the block height (<a name="0070-PLUP-005" href="#0070-PLUP-005">0070-PLUP-005</a>)
+  - [ ] Not enough validators have upgraded their node
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network does not change version (can be checked with APIs)
+ 
+### Major version updates (`x.x.x`)
+ > Note: These acceptance criteria are *incomplete*. There will likely be a governance vote from users, as well as the validator vote. For now these Acceptance Criteria test a major version in the same was a minor version upgrade.
+- [ ] A new major upgrade is made available, validators vote to update, and the update is successfully deployed (<a name="0070-PLUP-006" href="#0070-PLUP-006">0070-PLUP-006</a>)
+  - [ ] Enough validators have upgraded their node
+  - [ ] A governance proposal have been created to propose the upgrade
```
Vegaklaus: Would the governance proposal be hard-coded ? I.e., if all validators upgrade anyhow, would the code stop them ?
If it is, can the validators use the upgrade on testnet anyhow ? Do they have a way to simulate a governance proposal there easily that's compatible with the enforcement?

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+  - [ ] The proposal have been accepted
+  - [ ] The block time for the upgrade is reached
+  - [ ] The network has switched to the new version of the protocol (can be checked with APIs)
+- [ ] A new major upgrade is made available, but the validator vote does not pass (<a name="0070-PLUP-007" href="#0070-PLUP-007">0070-PLUP-007</a>)
+  - [ ] Enough validators have upgraded their node
+  - [ ] A governance proposal have been created to propose the upgrade
+  - [ ] The proposal have been rejected
+  - [ ] The block time for the upgrade is reached
```
Vegaklaus: We used block height before. Both works, but needs to be clear which one we use.

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: Yes, I was using them as an example of keeping the chain running while processing upgrade, not specifically how they do them. The cosmos way is how we will do our upgrade indeed.

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: Yea, we could decide to do that. That would be a massive difficulty, is something was to change so much that you cannot use it in a previous version. I would instead prefer to have to upgrade all the time, and for us to provide a fix version that the validators would use to upgrade to.

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: Yea that should be 2/3 of the validators weight I suppose.

protocol/0070-PLUP_protocol_upgrades.md
Vegaklaus: Agree here; the only issue is that if we need a fast rollback due to an issue, we still have to be provide the functionality for the rollback version (now with the higher version number) to read the serialisation of the current version, so a fast rollback isn't always possible.

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: Well, I'm not sure about this. I'd imagine that we would want everyone to update to the last version just in case.
Do you think it would be too much of a hassle? I imagine that a patch could actually be something that fix a potential consensus issue in a future block too.

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: yes that's correct, we would miss the data, an replaying the wholke chain would be needed, and applying the fix only when we reach the point in the chain where the upgrade is supposed to be applied, then the state would be save from this point.

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: Well I think that ideally we don't want to vendor lock vega as the only provider of the code. We version the protocol here, and fixes to the protocol, so I would expect that version N of the protocol should behave the same accros different implementation. The vega node itself though, is version separately.

jeremyletang: Well, I'd imagine a rollback would be us providing a new version, where the upgrade is the previous version and the legacy version is the upgrade, with a new layer in between to convert back the data to the previous state?

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: well, a code release will be compatible only for upgrade from a version to another. Now if a version is not accepted in time by validators, then it will be for us to provide code again, with the same legacy version than in production, and change the upgrade path from 1 -> 2 to 1 -> 3 (meaning the convertion code has to be re-written etc, version 2 is discarded completely).

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: I'm not sure we are speaking of a governance proposal here. It's more of a framework internal to the protocol, just for the validators to accept or not the release manually

protocol/0070-PLUP_protocol_upgrades.md
jeremyletang: I don't mind as well, what's the most practical / speaking better to users?

protocol/0070-PLUP_protocol_upgrades.md
wwestgarth: I only asks because its something tendermint supports in that is separates overall version number from the "protocol version". The easiest and safest thing certainly would be to have everyone running identical code.

protocol/0070-PLUP_protocol_upgrades.md
gordsport: Need a way to inform the validators that they cannot upgrade from version 1 to 3 and they need to update to version 2 first then 3. This also needs to account for less simple cases where version numbers are not sequential (i.e. semver + serial number).

gordsport: @gordsport - add risk to risk log of complex rollback if indeed possible.

@jeremyletang Can we have, in the future, an activate/deactivate process to mitigate this risk? (new spec on feature toggling?)

protocol/0070-PLUP_protocol_upgrades.md
gordsport: Do not expect this will be used right now - still needs to be specified

protocol/0070-PLUP_protocol_upgrades.md
gordsport: Should we use both i.e. block height for patch and block time for major?

Do we also need an option in case we never reach the height or time due to the chain having stopped?

gordsport:
Requirement to add to the spec: 
https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1657541187157559?thread_ts=1656933690.212839&cid=CAHA5EX0F

To help non-validators auto-update (if enabled)

protocol/0070-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,187 @@
...
+
+### How other protocols proceed
+
+Other protocol e.g Ethereum carry out updates in an asynchronous manner. Usually a new version of the protocol is made available, a hardcoded block height is set
+at which a new code path will be enabled. If enough node runners have deployed the updated code, then the network will continue with the new code path. Others nodes
+which haven't updated will then fork the blockchain.
+
+### Prior example of updgrading the nodes asynchronously
+
+Back in December 2020, vega proceeded to a LNL restore. Unfortunately, a bug in the code prevented the dispatch of the network parameters after the restore. This left
```
fkondej: I think it was 2021, not 2020.

jgsbennett:
###  Network Parameters and other upgrade-time-config

config.toml - manually changed only, not automated - not supplied with binary

network params - stored within binary - no use case at present

When the network is restarted with the new version, what configuration might we want to change _at exactly that time_? And what mechanisms are there to do it? @karlem mentioned there would be vega parameters , but also there might be new or existing network parameters that validators would wish to set or change at the same block the new version takes effect. Currently, Network Param proposals take effect at a given _time_, not a given block height, so I believe this spec does not allow for setting new values for new versions, and enforces us to hard-code validator chosen defaults, validators to rebuild themselves, or else for the validators to begin the network with values they are not able to control, until such a time as they can arrange for proposals to succeed.
If we choose to add extra functionality for this, we'll need more ACs to cover cases around this.



### Extra ACs / Test cases / Questions based on existing spec:

VISOR
- Vega visor automatically downloads the next version for install
- Vega visor uses the correct version, which has been pre-downloaded into the directory

Negative cases
- Propose an upgrade block in the past - should be rejected / failed check - sign via wallet / check wallet
- Propose a negative upgrade block
- Propose a 0 upgrade block
- Propose (string/other upgrade block)
- Do not supply an upgrade block in the proposal (if at all possible)

Positive
- Propose an upgrade block which ought to be the end of an epoch - 
- Propose an upgrade block which should result in a new network in the same epoch

- Max length vega version string in upgrade proposal - enforced

- AC 1 refers to "invalid tag". What does this mean, in practice? [semmer](https://semver.org/)

- Propose and enact a downgrade - not supported - should be rejected

Need pipelines...
- Upgrade from the latest supported version, to the current version under test (more relevant for CI, once this feature is in the wild).
- Upgrade from the current version under test, to a new version (can be the same code, but which is configured to report a new protocol version. This is the best we can test for forwards compatibility ahead of actually building new versions).



- Propose with a validator which is moved to Ersatz by the time the upgrade is enacted. (What should happen here? If there are 5 validators, 3 vote yes, 2 vote no: One of the yes voters is kicked in favour of a new one, leaving the vote at 2-2. Does the upgrade still happen? - counting votes happens at block upgrade




- Ersatz validator, promoted to validator, votes to propose upgrade as deciding vote (2-2 > 3-2). Upgrade happens - counting votes happens at block upgrade

- Validator X proposes upgrade at block N, upgrade version V. Validator X _then_ proposes upgrade at block N+1, version V +/-n (some other version), other validators approve. Second upgrade succeeds. - void secondary votes

- 5 validators propose upgrade at block N. 5 propose at block N+1.  1 validator who originally proposed block N now proposes block N+1 instead. Block N upgrade is rejected. Upgrade takes place at block N+1 - event produced to show rejection

- Spam? How many transactions can be made per second from a validator? Other wallets? (raise ticket for discussion/not required at this time)


- Non-validator wallet attempts to propose upgrade. (Rejected transaction?  check tx)
- Any difference between non-validator wallet and any random vega wallet? (Not even a non-validator, just some random party). (Rejected transaction? -check tx)


- Ersatz attempts to propose upgrade. (Rejected transaction? 

- Upgrade which upgrades data node _and_ vega version. (not possible)
- Upgrade which upgrades only vega.
- Upgrade which upgrades only data node (not possible)

- Upgrade which attempts to set both tags to the current version. (not Possible)



- 3 validators propose version V at block N. 2 validators propose version V +/-n (some other version) at block N. Version V upgrade takes place. (Is it possible to propose two different versions on the same block?) - possible to vote for different tagged version across network)

No voters will stop producing blocks

- Proposal remembered in checkpoints? Proposal is scheduled for upgrade at block N. Checkpoint is taken and network is stopped at some block prior to N. Restart network from checkpoint, proposal should be remembered, upgrade takes place at block N - not applicable - votes are not carried across checkpoints

shapshot - post load - should be able to propose a vote to upgrade
Node starting from snapshot which has a proposal at a given block, ensure during replay when the block height is reached a new version is loaded and also post load an upgrade takes place at target block.


- As above, but when you restart from checkpoint, upgrade to the new version (is this valid?), such that the network is loaded with a pending proposal to the current version. What should happen? - (not applicable)


- Upgrade takes place, but insufficient validators are restored for 1, 5, 10, minutes. Validators which are restored immediately patiently wait for consensus to be achieved, and then blocks continue  - can test this.


- Upgrade takes place at block N. Minimum number of validators -1 restart with correct version. Deliberately cause one validator to restart with previous version. Consensus is not achieved. Start another validator with the correct version, consensus is achieved - no blocks produced - valid

- 5 validator network. Upgrade takes places at block N. Start 3 validators immediately. Allow several blocks to pass. Start two remaining validators. (All validators continue to work) - blocks producing if 3 have enough weight.

- 5 validator network. Upgrade takes places at block N. Start 3 validators immediately. Allow several blocks to pass. Start two remaining validators. (All validators continue to work) - no blocks producing if 3 do not have enough weight - need 70% weight to produce blocks

Test that weighting threshold (API TM) is not reached after restart, network does not produce blocks

RomanLeca:
One more scenario:

- there are 5 validators
- 1 validator submits an incorrect proposal (bad tag, etc)
- 3 validators submit a correct one, same version, having majority'

will this proposal pass? Yes

vega-paul:
Another scenario to consider to add 

Proposal upgraded requested by other validator, another validator submits multiple proposals to upgrade Y followed by N. Subsequent votes should always override previous until majority has been reached by network.

doesnt make sense - no such thing as No vote.

vega-paul:
ACs updated - needs review

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,167 @@
...
+   - Non-validator attempts to propose upgrade
+   - Ersatz validator attempts to propose upgrade
+ 
+### Block height validation (<a name="0075-PLUP-002" href="#0075-PLUP-002">0075-PLUP-002</a>)
+   Proposal will not be accepted as valid if validator
+   -  Proposes an upgrade block in the past - should be rejected / failed check - sign via wallet / check wallet
+   -  Proposes a negative upgrade block
+   -  Proposes a 0 upgrade block
+   -  Proposes (string/other upgrade block)
+   -  Proposes with supply a block height
```
jgsbennett: Is this supposed to be "without supplying" ?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,167 @@
...
+   - Propose and enact a version downgrade 
+   - Non-validator attempts to propose upgrade
+   - Ersatz validator attempts to propose upgrade
+ 
+### Block height validation (<a name="0075-PLUP-002" href="#0075-PLUP-002">0075-PLUP-002</a>)
+   Proposal will not be accepted as valid if validator
+   -  Proposes an upgrade block in the past - should be rejected / failed check - sign via wallet / check wallet
+   -  Proposes a negative upgrade block
+   -  Proposes a 0 upgrade block
+   -  Proposes (string/other upgrade block)
```
jgsbennett: @jeremyletang I think the python/protos code I've used so far tries to enforce an integer type. To what extent is it possible to send an incorrect type over the wire? i.e: By hacking together funky transactions, like with the work Muthu is doing.

jeremyletang: yea well just try to build transaction yourself would be what you can do really.

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,167 @@
...
+   -  Proposes with supply a block height
+
+### VISOR (<a name="0075-PLUP-005" href="#0075-PLUP-005">0075-PLUP-005</a>) 
+   - Can be seen to automatically download the latest version for install when available at the source location when file meets the format criteria defined
+   - Visor automatically upgrades validators to proposed version if requiredMajority has been reached
+   
+### Epochs (<a name="0075-PLUP-006" href="#0075-PLUP-006">0075-PLUP-006</a>) 
+   - Proposing an upgrade block which ought to be the end of an epoch
+   - Propose an upgrade block which should result in a new network in the same epoch
```
jgsbennett: These are maybe 3 different test cases, which one do we mark with the AC? Similar deal for several of the above... Our current system seems to assume one python test case per AC... For _some_ of these I could theoretically cram them into one case, but for some of the others it doesn't necessarily make sense.

protocol/0075-PLUP_protocol_upgrades.md
vega-paul: yes

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+   - (<a name="0075-PLUP-003" href="#0075-PLUP-003">0075-PLUP-003</a>) Propose and enact a version downgrade
+   - (<a name="0075-PLUP-004" href="#0075-PLUP-004">0075-PLUP-004</a>) Non-validator attempts to propose upgrade
+   - (<a name="0075-PLUP-005" href="#0075-PLUP-005">0075-PLUP-005</a>) Ersatz validator attempts to propose upgrade
+ 
+### Block height validation
+   Proposal will not be accepted as valid if validator
+   -  (<a name="0075-PLUP-006" href="#0075-PLUP-006">0075-PLUP-006</a>) Proposes a negative upgrade block
+   -  (<a name="0075-PLUP-007" href="#0075-PLUP-007">0075-PLUP-007</a>) Proposes a 0 upgrade block
+   -  (<a name="0075-PLUP-008" href="#0075-PLUP-008">0075-PLUP-008</a>) Proposes (string/other upgrade block)
+   -  (<a name="0075-PLUP-009" href="#0075-PLUP-009">0075-PLUP-009</a>) Proposes without supplying a block height
```
davidsiska-vega, line 114: Should we add an AC that the block number has to be strictly greater than the current block?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+
+### VISOR 
+   - (<a name="0075-PLUP-010" href="#0075-PLUP-010">0075-PLUP-010</a>) Can be seen to automatically download the latest version for install when available at the source location when file meets the format criteria defined
```
davidsiska-vega: The latest  version and the one in an accepted proposal may be different. Should we add an AC that the vegavisor downloads all the versions? Or all the ones in active proposal? Or is that not needed / intended behavior?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+   - (<a name="0075-PLUP-011" href="#0075-PLUP-011">0075-PLUP-011</a>) Visor automatically upgrades validators to proposed version if requiredMajority has been reached
+   
+### Epochs 
+   - (<a name="0075-PLUP-012" href="#0075-PLUP-012">0075-PLUP-012</a>) Proposing an upgrade block which ought to be the end of an epoch
```
davidsiska-vega: ... should have what result? Is it just we want to test that if an upgrade block and epoch end coincide then all goes well?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Protocol upgrades mechanism
+
+One of the main challenge with upgrading the network both asynchronously and without a restart / downtime is
+the management of the state, and possible incompatible changes between the state of two version of the protocol.
+
+The following describes the general workflow of upgrades: 
+1. Vega continuesly make new releases available 
+2. Validators can suggest to the network an upgrade with a given release tag and a block height for the upgrade to take place
+3. If a proposal gets enough votes (`protocol.upgrade.proposal.requiredMajority` of the validators, not stake) at the given height vega will take a snapshot automatically and stop processing further blocks until restarted by the vega-visor, the vega process manager.
```
davidsiska-vega: I am guessing this was discussed - but does it make sense to count validators not something like "own + delegated stake" or "tendermint weight"? Because the purpose is for the chain to continue post upgrade - so I'd think we need not only majority but the "right ones". Thoughts?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+   - (<a name="0075-PLUP-018" href="#0075-PLUP-018">0075-PLUP-018</a>) When majority reached during the process of upgrading, those validators which didnt propose will stop producing blocks
+   - (<a name="0075-PLUP-019" href="#0075-PLUP-019">0075-PLUP-019</a>) Proposals for multiple versions at same block height will be rejected if majority has not been reached, network continues with the current running version 
+   - (<a name="0075-PLUP-020" href="#0075-PLUP-020">0075-PLUP-020</a>) Propose with a validator which is moved to Ersatz by the time the upgrade is enacted. If there are 5 validators, 3 vote yes, 2 vote no: One of the yes voters is kicked in favour of a new one, leaving the vote at 2-2 so the upgrade should not happen as counting votes happens at block height only
+
+
+### Multiple proposals (<a name="0075-PLUP-021" href="#0075-PLUP-021">0075-PLUP-021</a>)
+   - If multiple proposals are submitted from a validator before the block heights are reached then only the last proposal is considered
+
+## Spam (<a name="0075-PLUP-022" href="#0075-PLUP-022">0075-PLUP-022</a>)
```
davidsiska-vega:
```suggestion
## Spam (<a name="0075-COSMICELEVATOR-022" href="#0075-COSMICELEVATOR-022">0075-COSMICELEVATOR-022</a>) 
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+
+
+### Multiple proposals (<a name="0075-PLUP-021" href="#0075-PLUP-021">0075-PLUP-021</a>)
+   - If multiple proposals are submitted from a validator before the block heights are reached then only the last proposal is considered
+
+## Spam (<a name="0075-PLUP-022" href="#0075-PLUP-022">0075-PLUP-022</a>) 
+   - Excessive numbers of proposals from a single validator within an epoch should be detected and rejected - (Future requirement)
+   
+## Snapshots
+   - (<a name="0075-PLUP-023" href="#0075-PLUP-023">0075-PLUP-023</a>) Post validator joining they should be immediately allowed to propose and included in the overall total count
```
davidsiska-vega: "should be immediately allowed to propose" what does propose refer to here? Propose blocks? Propose new upgrades?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+   - (<a name="0075-PLUP-024" href="#0075-PLUP-024">0075-PLUP-024</a>) Validators attempting to join during upgrade will be rejected if protocol versions differ
```
davidsiska-vega: Can we define "during" more precisely? Does during mean "proposal to update at block 1000 has been submitted, we are now at block 500 so no-one can join"? 

Or does "during" mean really we got to block 1000 the chain stopped while nodes restart and when it restarts it will again accept transactions and we create block 1001?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+   - (<a name="0075-PLUP-025" href="#0075-PLUP-025">0075-PLUP-025</a>) Ensure that required majority is not met when enough validators join between validator proposals and target block
```
davidsiska-vega: What does this mean?

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+   - (<a name="0075-PLUP-026" href="#0075-PLUP-026">0075-PLUP-026</a>) Node starting from snapshot which has a proposal at a given block, ensure during replay when the block height is reached a new version is loaded and also post load an upgrade takes place at target block.
+
+## Checkpoints
```
davidsiska-vega:
```suggestion
## LNL Checkpoints
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+### Successful upgrade  (<a name="0075-PLUP-030" href="#0075-PLUP-030">0075-PLUP-030</a>) 
+  - A new release is made available, and is successfully deployed
+  - Setup a network with 5 validators running version x
+  - Have 4 validator submit request to upgrade to release >x at block height 1000
+  - At the end of block height 1000 a snapshot is taken and vega is stopped by the vegavisor
+  - All nodes are starting from the snapshot of block 1000 and the network resumes with version >x
+ 
+ 
+ ### Failing consensus (<a name="0075-PLUP-031" href="#0075-PLUP-031">0075-PLUP-031</a>)
+  - Upgrade takes place at block N. Minimum number of validators -1 restart with correct version. One validator restarts with previous version. Consensus is not achieved. Start another validator with the correct version, consensus is achieved
```
davidsiska-vega: Regardless of what own stake + delegations they have?

davidsiska-vega (changes requested):
Left a lot of comments - some of them may stem from my ignorance of the design - if that's the case say so and mark as resolved ;-).

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: We should come back up in the new epoch, and rewards should have been paid out.

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: I think the latter.

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: Again, I think the latter.

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: I don't know for sure what was originally intended, but I might imagine:
1: 5 validators, 3 vote for upgrade to version V on block B.
2: Prior to that, 2 more validators join, giving total 7 validators. Block B passes without upgrade (if new validators do not vote).
(Intention is that required majority is met in step 1, but not in step 2. I haven't played with that field yet, may also require votes from the new validators to achieve this case.)

protocol/0075-PLUP_protocol_upgrades.md
gordsport: @barnabee - please can you review this and confirm if we should leave as is, or factor in the tendermint weight?

gordsport: @jgsbennett - once we have a decision here can you please add an AC accordingly?

protocol/0075-PLUP_protocol_upgrades.md
gordsport: @jgsbennett - please can you add this as an AC?

protocol/0075-PLUP_protocol_upgrades.md
gordsport: Decision in spec review meeting: We will remain as is and use count and not factor in score/weights for this milestone

protocol/0075-PLUP_protocol_upgrades.md
gordsport:
```suggestion
   - (<a name="0075-PLUP-010" href="#0075-PLUP-010">0075-PLUP-010</a>) Can be seen to automatically download the tagged version proposed for install when available at the source location when file meets the format criteria defined
```

protocol/0075-PLUP_protocol_upgrades.md
gordsport: Try to test this with null blockchain. If this proves to be very complex we can consider testing this later: ```suggestion
   - (<a name="0075-COSMICELEVATOR-012" href="#0075-COSMICELEVATOR-012">0075-COSMICELEVATOR-012</a>) Proposing an upgrade block which ought to be the end of an epoch
```

File: protocol/0075-PLUP_protocol_upgrades.md [hunk 46]
```diff
@@ -0,0 +1,166 @@
...
+
+
+### Multiple proposals (<a name="0075-PLUP-021" href="#0075-PLUP-021">0075-PLUP-021</a>)
+   - If multiple proposals are submitted from a validator before the block heights are reached then only the last proposal is considered
+
+## Spam (<a name="0075-COSMICELEVATOR-022" href="#0075-COSMICELEVATOR-022">0075-COSMICELEVATOR-022</a>) 
+   - Excessive numbers of proposals from a single validator within an epoch should be detected and rejected - (Future requirement)
+   
+## Snapshots
+   - (<a name="0075-PLUP-023" href="#0075-PLUP-023">0075-PLUP-023</a>) Post validator joining they should be immediately allowed to propose and included in the overall total count
```
gordsport:
```suggestion
   - (<a name="0075-PLUP-023" href="#0075-PLUP-023">0075-PLUP-023</a>) Post a validator becoming a tendermint validator they should be immediately allowed to propose an upgrade and be included in the overall total count
```

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: Drop this AC. We think it's just another specific case of the "validator joins with the wrong version" scenario.

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: Clarify AC.

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: Step 1:
Restart the network with a number of validators whose voting power is equal to less than two thirds. Consensus is not achieved.
Step 2:
Restart one of the validators (whose voting power would take the total voting power over the two thirds threshold) with the wrong version. Consensus should still not be achieved.
Step 3:
Restart validator from step 2 with the correct version. Consensus is achieved.

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+  - Setup a network with 5 validators running version x
+  - Have 4 validator submit request to upgrade to release >x at block height 1000
+  - At the end of block height 1000 a snapshot is taken and vega is stopped by the vegavisor
+  - All nodes are starting from the snapshot of block 1000 and the network resumes with version >x
+ 
+ 
+ ### Failing consensus (<a name="0075-PLUP-031" href="#0075-PLUP-031">0075-PLUP-031</a>)
+  - Upgrade takes place at block N. Minimum number of validators -1 restart with correct version. One validator restarts with previous version. Consensus is not achieved. Start another validator with the correct version, consensus is achieved
+  - 5 validator network. Upgrade takes places at block N. Start 3 validators immediately. Allow several blocks to pass. Start two remaining validators. (All validators continue to work) - no blocks producing as 3 validators do not have enough weight - need 70% weight to produce blocks
+  - Weighting threshold (API TM) is not reached after restart, network does not produce blocks
```
jgsbennett: Drop

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,166 @@
...
+  - Upgrade takes place, but insufficient validators are restored for 1, 5, 10, minutes. Validators which are restored immediately patiently wait for consensus to be achieved, and then blocks continue  - concensus acheived
```
jgsbennett: Add an AC for the mainnet test case. https://github.com/vegaprotocol/devops-infra/issues/1522

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: Net param is actually validators.vote.required . Rename here and in any ACs that mention.

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: Add an AC for "submit a proposal with version == current version and a valid block height retracts any existing proposals"
Add an AC for "submit an invalid proposal does not overwrite any existing proposals" i.e: Set up an upgrade set to happen at block height H. One validator (whose changing vote would cause the upgrade to not happen) submits an invalid upgrade proposal (i.e: invalid version). Proposal rejected. Upgrade still happens.

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,174 @@
...
+   -  (<a name="0075-PLUP-008" href="#0075-PLUP-008">0075-PLUP-008</a>) Proposes (string/other upgrade block)
+   -  (<a name="0075-PLUP-009" href="#0075-PLUP-009">0075-PLUP-009</a>) Proposes without supplying a block height
+
+### VISOR 
+   - (<a name="0075-PLUP-010" href="#0075-PLUP-010">0075-PLUP-010</a>) Can be seen to automatically download the tagged version proposed for install when available at the source location when file meets the format criteria defined
+   - (<a name="0075-PLUP-011" href="#0075-PLUP-011">0075-PLUP-011</a>) Visor automatically upgrades validators to proposed version if required majority has been reached
+   
+### Epochs 
+   - (<a name="(0075-COSMICELEVATOR-012)" href="#(0075-COSMICELEVATOR-012)">(0075-COSMICELEVATOR-012)</a>) Proposing an upgrade block which ought to be the end of an epoch. After upgrade takes place, confirm rewards are distributed, any pending delegations take effect, and validator joining/leaving takes effect.
+   - (<a name="0075-PLUP-013" href="#0075-PLUP-013">0075-PLUP-013</a>) Propose an upgrade block which should result in a new network in the same epoch
```
davidsiska-vega:
```suggestion
   - (<a name="0075-PLUP-013" href="#0075-PLUP-013">0075-PLUP-013</a>) Propose an upgrade block which should result in a network running new code version in the same epoch.
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,174 @@
...
+   - (<a name="0075-PLUP-014" href="#0075-PLUP-014">0075-PLUP-014</a>) Ensure end of epoch processes still run after restore e.g reward calculation and distributions
+
+### Required Majority
```
davidsiska-vega, line 125:
```suggestion
### Required Majority
For the purposes of protocol upgrade each validator that participates in consensus has one vote. Required majority is set by `validators.vote.required network parameter`. 
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,174 @@
...
+   - (<a name="0075-PLUP-015" href="#0075-PLUP-015">0075-PLUP-015</a>) Counting proposal votes to check if required majority has been reached occurs when any proposed target block has been reached
+   - (<a name="0075-PLUP-016" href="#0075-PLUP-016">0075-PLUP-016</a>) Only active network validators proposals are counted when any proposed target block has been reached
```
davidsiska-vega:
```suggestion
   - (<a name="0075-PLUP-016" href="#0075-PLUP-016">0075-PLUP-016</a>) Only proposals from validators participating in consensus are counted when any proposed target block has been reached. 
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,174 @@
...
+   - (<a name="0075-PLUP-017" href="#0075-PLUP-017">0075-PLUP-017</a>) Events are emitted for all proposals which fail to reach required majority when target block is reached
+   - (<a name="0075-PLUP-018" href="#0075-PLUP-018">0075-PLUP-018</a>) When majority reached during the process of upgrading, those validators which didnt propose will stop producing blocks
+   - (<a name="0075-PLUP-019" href="#0075-PLUP-019">0075-PLUP-019</a>) Proposals for multiple versions at same block height will be rejected if majority has not been reached, network continues with the current running version 
+   - (<a name="0075-PLUP-020" href="#0075-PLUP-020">0075-PLUP-020</a>) Propose with a validator which is moved to Ersatz by the time the upgrade is enacted. If there are 5 validators, 3 vote yes, 2 vote no: One of the yes voters is kicked in favour of a new one, leaving the vote at 2-2 so the upgrade should not happen as counting votes happens at block height only
+   - (<a name="0075-PLUP-036" href="#0075-PLUP-036">0075-PLUP-036</a>) Changing validators.vote.required network parameter to a value above two thirds is respected.
```
davidsiska-vega:
```suggestion
   - (<a name="0075-PLUP-036" href="#0075-PLUP-036">0075-PLUP-036</a>) Changing `validators.vote.required` network parameter to a value above two thirds is respected.
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,174 @@
...
+   - (<a name="0075-PLUP-037" href="#0075-PLUP-037">0075-PLUP-037</a>) The value of validators.vote.required is checked at upgrade block, i.e: vote on a proposal with all validators, then change the validators.vote.required net param before upgrade block, to a higher value, which would cause the upgrade to be rejected. Upgrade fails.
```
davidsiska-vega:
```suggestion
   - (<a name="0075-PLUP-037" href="#0075-PLUP-037">0075-PLUP-037</a>) The value of `validators.vote.required` is checked at upgrade block, i.e: vote on a proposal with all validators, then change the `validators.vote.required` net param before upgrade block, to a higher value, which would cause the upgrade to be rejected. Upgrade fails.
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+### Multiple proposals (<a name="0075-PLUP-021" href="#0075-PLUP-021">0075-PLUP-021</a>)
+   - If multiple proposals are submitted from a validator before the block heights are reached then only the last proposal is considered
+
+## Spam (<a name="0075-COSMICELEVATOR-022" href="#0075-COSMICELEVATOR-022">0075-COSMICELEVATOR-022</a>) 
+   - Excessive numbers of proposals from a single validator within an epoch should be detected and rejected - (Future requirement)
+   
+## Snapshots
+   - (<a name="0075-PLUP-023" href="#0075-PLUP-023">0075-PLUP-023</a>) Post a validator becoming a tendermint validator they should be immediately allowed to propose an upgrade and be included in the overall total count
```
davidsiska-vega:
```suggestion
   - (<a name="0075-PLUP-023" href="#0075-PLUP-023">0075-PLUP-023</a>) Post a validator becoming a consensus-participating validator they should be immediately allowed to propose an upgrade and be included in the overall total count
```

protocol/0075-PLUP_protocol_upgrades.md
```diff
@@ -0,0 +1,174 @@
...
+  - A new release is made available, and is successfully deployed
+  - Setup a network with 5 validators running version x
+  - Have 4 validator submit request to upgrade to release >x at block height 1000
+  - At the end of block height 1000 a snapshot is taken and vega is stopped by the vegavisor
+  - All nodes are starting from the snapshot of block 1000 and the network resumes with version >x
+ 
+ 
+### Failing consensus
+  - (<a name="0075-PLUP-030" href="#0075-PLUP-030">0075-PLUP-030</a>) Upgrade takes place at block N. Restart with a number of validators whose voting power is <= two thirds. Restart one more validator whose voting power would take the total voting power >= two thirds, with an incorrect version. Consensus is not achieved. Now restart that validator with the correct version. Consensus is achieved.
+  - (<a name="0075-PLUP-031" href="#0075-PLUP-031">0075-PLUP-031</a>) 5 validator network. Upgrade takes places at block N. Start 3 validators immediately. Allow several blocks to pass. Start two remaining validators. (All validators continue to work) - no blocks producing as 3 validators do not have enough weight - need 70% weight to produce blocks
```
davidsiska-vega: Allow several blocks to pass. ... no blocks producing as 3 validators do not have enough weight  

This is confusing. Do we mean allow *time* to pass first.

davidsiska-vega (changes requested):
One clarification - then done.

protocol/0075-PLUP_protocol_upgrades.md
jgsbennett: Yes, I never saw that. I'll correct it.

davidsiska-vega (approved):

PR #885, opened 2022-02-15
Market lifecycle acceptance criteria

### LINKED ISSUE #821: 0043-MKTL-market_lifecycle needs acceptance criteria
In order to get QA test coverage [this spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0043-MKTL-market_lifecycle.md) needs to have clear acceptance criteria defines with acceptance IDs.

davidsiska-vega:
The market lifecycle spec is fancy but basically only two things should happen with a market:

1) market is proposed but rejected.
1) market is proposed, accepted, trades (going in and out of auctions), trading terminated oracle rings, trading is terminated and finally a settlement prices comes through leading to final cashflows.

So these are the two that were added. Closes #821 .

### DIFF
--- a/protocol/0043-MKTL-market_lifecycle.md
+++ b/protocol/0043-MKTL-market_lifecycle.md
@@ -281,6 +281,6 @@ Parties with open positions still have non-zero margin account balances.
 The market state is `trading terminated`.
 1. The settlement price oracle transaction is sent and it is  *not* equal to `p`. 
 Parties that had open positions see settlement cash-flows happen. 
-Maring account balances are transferred to the general account. 
+Margin account balances are transferred to the general account. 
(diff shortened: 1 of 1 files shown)


edd (dismissed):
I think there are more we could add - but I also think they tend to overlap with other specs:

- What happens at Settlement -> Settlement spec
- How a market moves to suspended -> just auction mode?

This isn't a blocker. I think it's more of a todo to link together specs better, which there's no current strategy for. The two you've added are good - I'll let @jgsbennett decide if they're the right 'level' or if they should be brkoen up or something. They seem big to me, but if this works, I think the two tests are great.

jgsbennett (approved):

PR #890, opened 2022-02-15
LP mechanics ACs

### LINKED ISSUE #823: 0044-LIQM-lp_mechanics needs acceptance criteria
In order to get QA test coverage [this spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0044-LIQM-lp_mechanics.md) needs to have clear acceptance criteria defines with acceptance IDs.

edd:
Pushing so I can hand over to @davidsiska-vega, as I can't parse enough ACs out about how bond account / margin account orchestration happens in precise enough detail to populate adequate ACs.

What I want to add is some labelled acceptance criteria that would let us test that the way penalties happen for bond account providers is correct.
- If an LP has enough in their general account to top up a margin account, but not enough to top up their bond account, then … ? I think the LP commitment is reduced
- A scenario that correctly tests the slashing amount is correctly calculated using the network parameters
- A scenario that tests the transfers that happen between bond & margin accounts is correct
- A scenario that tests that slashed bonds are transferred to the insurance pool

Closes #823

### DIFF
--- a/protocol/0013-ACCT-accounts.md
+++ b/protocol/0013-ACCT-accounts.md
@@ -93,19 +93,19 @@ When a trader no longer has collateral requirements for a  market (because they 
 
 If there is a positive balance in an account that is being deleted, that balance should be transferred to the account specified in the transfer request (which for margin accounts will typically be the insurance pool of the market).
 
+## Bond accounts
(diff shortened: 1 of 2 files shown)


davidsiska-vega:
Added reference to a feature file that tests the above.

davidsiska-vega (approved):

PR #913, opened 2022-02-16
Removing feature tests after moving them to vega verified folder

cpuja84:
Feature tests which have been important to Protocol Design team used to reside on specs internal have been moved to verified folder in Vega Repo so deleting them here along with few mock files.

Closes #914

PR #915, opened 2022-02-16
Remove feature tests run as follow up to #913

edd:
After #913, I think I saw an error in the build due to this PR. Now it shouldn’t.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -41,16 +41,6 @@ pipeline {
 
         stage('Run checks') {
             parallel {
-                stage('Run qa-scenarios') {
-                    steps {
-                        retry(3) {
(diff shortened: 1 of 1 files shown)


cpuja84:
Hi @edd - I already raised this issue with @fkondej who would be looking at it tomorrow ( 17/02/2022). Thanks !

edd:
@cpuja84 🤞 I got it right and have saved him some work. But I did leave a single test running in parallel so I doubt I got it exactly right. You’ve assigned @fkondej to it now so I’ll wait for his review 😄

fkondej (approved):
More cleanup is required. But this should unblock other PR checks.

PR #927, opened 2022-02-21
Removed negative offset values and made it clear how they work now.

### LINKED ISSUE #926: LP order type spec - updates
https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0038-OLIQ-liquidity_provision_order_type.md#how-they-are-submitted

- The LP order type spec still includes references to negative offsets, such as in the example buy shape. 
- It's also missing an example sell shape.

peterbarrow:
Updated the spec to clarify that pegged order offsets are always positive and that we must supply and buy an sell shape when submitting a LP submission.

Close #926


witgaw (dismissed):
LGTM

davidsiska-vega (approved):
LGTM - on protocol design call.

PR #931, opened 2022-02-23
Remove acs that reference deployment to Ropsten from 0031

davidsiska-vega:


C0deMunk33 (approved):
LGTM, having the network in the specs seems silly on review. Thank you

PR #945, opened 2022-03-01
Add null blockchain acceptance criteria

edd:
I’ve added some fairly uncontroversial acceptance criteria to test that what’s delivered matches. I think the unit tests that are already in the folder cover a lot of these already, so it’s not so much a prompt for a whole new set of tests, as it is a third party verifying that they understand the spec and the implementation correctly.

I left the two old ‘criteria’ in there, but mostly ignored them. They’re sort of … aspirations rather than criteria, but hey - I thought they were still useful

ACs before: 0
ACs after: 6
Remaining files without ACs after merge: 13

### DIFF
--- a/non-protocol-specs/0008-NP-NULB-null_blockchain_vega.md
+++ b/non-protocol-specs/0008-NP-NULB-null_blockchain_vega.md
@@ -42,3 +42,9 @@ An extra API is added to allow the user to control the time and block start/end 
 - Time can only move forward
   - I submit `{ forward: "-10s" }` to the time fast forward endpoint. Time does not move and I receive an error. (<a name="0008-NP-NULB-001" href="#0008-NP-NULB-001">0008-NP-NULB-001</a>)
   - I submit `{ forward: "<a date in the past>" }` to the time fast forward endpoint. Time does not move and I receive an error. (<a name="0008-NP-NULB-002" href="#0008-NP-NULB-002">0008-NP-NULB-002</a>)
+- The null blockchain block time does not control the automatic creation of blocks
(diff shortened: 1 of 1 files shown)


non-protocol-specs/0008-NP-NULB-null_blockchain_vega.md
```diff
@@ -34,8 +34,19 @@ Event bus data can be recorded either in the test app via a subscription to the
 Null blockchain Vega must support all the APIs that Vega core supports with all the standard language bindings plus the additional ones outlined above. 
 An extra API is added to allow the user to control the time and block start/end times without changing the standard API endpoints.
 
-## Acceptance criteria
+# Acceptance criteria
 
 - It is possible to take a functioning sample API script, change the config so it points at a null blockchain vega node, add some time passage commands and run it.
```
davidsiska-vega: With what you wrote @edd maybe this is redundant now.

non-protocol-specs/0008-NP-NULB-null_blockchain_vega.md
```diff
@@ -34,8 +34,19 @@ Event bus data can be recorded either in the test app via a subscription to the
...
-
 - There is golang example which creates three parties, gives them assets, one party proposes a market and acts as LP, the remaining two parties trade, placing one trade per day for "365 days".  The LP party submits a trading terminated transaction after "365 days" and a subsequent settlement price transaction and the market settles. All of this executes within 0.1 second on a Raspberry PI.
```
davidsiska-vega: I think this exists somewhere, so we can give it a label too. Maybe we should remove the silly comment on Raspberry PI though.

davidsiska-vega (commented):
Few nitpicking comments.

wwestgarth (dismissed):

PR #949, opened 2022-03-02
Propose removal of Market Depth ac

edd:
@peterbarrow you added this Acceptance Criteria when you were updating the Market Depth updates. I get that it was a design constraint, but do we need to keep it as an Acceptance Criteria, and if so, how do you propose we test it?

Thanks!


peterbarrow:
> @peterbarrow you added this Acceptance Criteria when you were updating the Market Depth updates. I get that it was a design constraint, but do we need to keep it as an Acceptance Criteria, and if so, how do you propose we test it?
> 
> Thanks!

That AC was added because we discussed possible ways to send out the updates and we originally batched them up and decided that was wrong. The market should see updates as soon as they are generated and not have to wait for the end of the block. A far as testing goes, I have no idea. We use the block time to stamp every outgoing message so we cannot tell if the update was batched or not. I guess this means we should remove the AC but still leave something about it in the spec.

peterbarrow (dismissed):

edd:
Thanks @peterbarrow - I’ll find somewhere useful to put it in the text, then merge

PR #953, opened 2022-03-03
Update 0062-SPAM-spam_protection.md

### LINKED ISSUE #905: Update 0062-SPAM-spam_protection - post Oregon Trail

### LINKED ISSUE #1022: Define expected behaviour after a change to: spam.protection.proposal.min.tokens
In order to have a test to be sure that a change to the `spam.protection.proposal.min.tokens` network parameter has the expected resulting behaviour we need to define:

- Values that the parameter should change from / to
- the expected behaviour when the values are changed

Vegaklaus:
Added spam detection for trading & acceptance criteria

### DIFF
--- a/protocol/0062-SPAM-spam_protection.md
+++ b/protocol/0062-SPAM-spam_protection.md
@@ -51,7 +51,8 @@ For Sweetwater, the policies we enforce are relatively simple:
 <num_votes> = 3
 <min_voting_tokens>  = 1
 <num_proposals> = 3
-<min_proposing_tokens> = 200000
+<min_proposing_tokens> = 100000
+(see https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0054-NETP-network_parameters.md for the final word on these values)
 
 
 - Any tokenholder with more than <min_voting_tokens> tokens has <num_votes> voting attempts per epoch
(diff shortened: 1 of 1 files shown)


davidsiska-vega:
There seems to be a section of duplicate text, is that the way it should be?

protocol/0062-SPAM-spam_protection.md
```diff
@@ -69,7 +70,54 @@ For Sweetwater, the policies we enforce are relatively simple:
    There also is a separate parameter to the same end that is enforced in the core. For SW, both these parameters have the same value. 
    In the future, we can set the spam protection value lower, as the amplification effect of a proposal (i.e., a proposal resulting in
    a very large number of votes) would also be covered by the core then.
-   
+
+### Minimum Functionallity for Trading
```
davidsiska-vega, line 75: Is this new functionality? Or this documenting / clarifying stuff that's already implemented?

protocol/0062-SPAM-spam_protection.md
```diff
@@ -88,3 +136,71 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
...
+	d = 10.000
+	e = 0
+	f = 1
+	
+<pow_factor> is ignored for now as including a proof_of_work is not minimum finctiuonallity)
+	    
+If a trader exceeds their quota of messages, they cannot send any transactions anymore in the given epoch.
+If a trader exceeds their quota for 4 consecutive epochs, they are banned for 10 further epochs.
+	   
+### Note: We need to verify that the numbers make sense.
```
davidsiska-vega: Verify with who? Or how?

protocol/0062-SPAM-spam_protection.md
```diff
@@ -88,3 +136,71 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
...
+	    
+### Acceptance Criteria
+	
+- A spam attack using votes/governance proposals is detected and the votes transactions are rejected, i.e.,
```
davidsiska-vega: Could you please add details with exact sequence of events / numbers so that QA can pick it up and get going?

davidsiska-vega (commented):
Thanks for this Klaus, I've left a few comments.

protocol/0062-SPAM-spam_protection.md
Vegaklaus: This is new; the trading spam protection was stuck in a PR to discuss options, the goal of this is to make something that's implementable on the short term. As far as I know, we don't have anything on spam for trading related transactions at this point, only on governance spam.

protocol/0062-SPAM-spam_protection.md
```diff
@@ -88,3 +221,26 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
...
+  a party that issues too many votes/governance proposals gets the follow on transactions rejected. This means
+  (given the current parameters)
+  - More than 360 delegation changes in one epoch
+  - Delegating while having less than one vega (10^18 of our smallest unit)
+  - Making a proposal when having less than 100.000 vega
+  - Making more than 3 proposals in one epoch
+  - Voting with less than 100 vega
```
ValentinTrinque: Since parties with equity-like share can vote on a market update proposal, I think we should rework this criteria to accept equity-like share.

Vegaklaus: That would mean the implementation is ahead of the spec and we need a spec update first. Options I see:
- market proposal/updates are different from governance and get their own rules
- the other assets are counted for all governance votes as well
- market proposals/updates are seen as other market transactions (as newly specified), which take into account assets (not related to that market, but overall).

My first instinct would be to do the first, and handle this similar to governance votes but with a different asset definition

protocol/0062-SPAM-spam_protection.md
```diff
@@ -69,7 +71,138 @@ For Sweetwater, the policies we enforce are relatively simple:
...
+
+#### Linking Value to allowed transactions
+
+##### Minimum Functionallity
+All trade related transactions are regarded equal. We have five network parameters b a1,a2,r1,r2. The number
+of transaction a trader is allowed per epoch is b+ (asset_score * a1)^a2 + (revenue_score * r1)^r2.
+
+The values of the variables that make sense need to still be determined based on fairground data; they also may be different between 
+trading with training wheels and real rtrading, as the former isn't generating the kind of revenue needed to authorize transactions.
+
```
Vegaklaus, line 169: There is a possibility that the spam protection will perform less than stellar during the training-wheels phase, as traders will all operate on limited assets and it's thus very affordable for an attacker to maintain 100 traders that have the maximum possible assets & revenue. 
Options is see:
- We accept this
- We put in an additional training wheels mechanism that implements a limit on trader per account if the network is congested

davidsiska-vega: I think Ethereum gas prices are our friend in this case.

protocol/0062-SPAM-spam_protection.md
davidsiska-vega: Could you please write down exactly against which network parameters are these being compared and what values are you assuming these network parameters have? The system tests aren't always written with the "default" parameters (and anyway, there isn't default as mainnet can be changed and we also have fairground etc whey they can be different).

protocol/0062-SPAM-spam_protection.md
```diff
@@ -69,7 +71,138 @@ For Sweetwater, the policies we enforce are relatively simple:
...
+### Minimum Functionallity for Trading
+
+
+As for governance, the primary purpose for spam protection is to protect the network from overloading; this also protects
+the core from processing unnecessary transactions, though that's the secondary goal.
+
+Spam protection goes in parallel with other functionallities that can reject transactions. The order of these tests is
+1. spam protection
+2. replay protection
+3. signature verification
```
jeremyletang, line 84: This might actually not be true, if not mistaken we first check replay, then spam, then signature.

protocol/0062-SPAM-spam_protection.md
```diff
@@ -69,7 +71,138 @@ For Sweetwater, the policies we enforce are relatively simple:
...
+ 	passed to the application layer. This mechanism allows for more fine-grained policies 
+ 	than the previous one, but at the price that the offending transaction has already 
+ 	taken up space in the blockchain. This primarily happens if the spam happens all in one
+	block, and thus cannot be discovered by post-block-reject
+
+
+For trading, the first check is if the user has any ressources in the first place:
+
+- If the ressources available to the account (at this point only ETH) are zero, then all trading transactions are considered spam and rejected.
+- If possible given the time, this value should change from zero to a governance parameter defining the threshold in a normalized unit.
```
jeremyletang, line 113: We do not have ETH at all for now, only erc20 tokens. Maybe we should precise that we reject the transaction if the party do not have a balance for the settlement asset of the market it tries to submit an order for.

protocol/0062-SPAM-spam_protection.md
```diff
@@ -69,7 +71,138 @@ For Sweetwater, the policies we enforce are relatively simple:
...
+ account has. Thus, an account with one cent worth of assets in it is not allowed to make 
+ 10.000 trades in one block, while an account with 100 million dollars may well have a good reason
+ to do so. 
+To this end, we need a normalisation function that can compute the revenue/ressources of all assets a user
+has as well of generated revenue. At this point, the only asset is ETH, so this is comparatively easy, though 
+it would be better (if possible) to define a more stable unit. To avoid floating point calculations where possible,
+assetds are calculated in 1/10.000 of an ETH, with trailing gwei cut of; this gives sufficient precision for
+the purpose of determining the value of an account and avoids both floating point and overly large numbers.
+
+*Minimum: All assets and revenues are in ETH until we have added other bridges*
```
jeremyletang, line 128: Still not in ETH but in ERC20 token.

protocol/0062-SPAM-spam_protection.md
```diff
@@ -69,7 +71,138 @@ For Sweetwater, the policies we enforce are relatively simple:
...
+*End-Goal: There is an internal way to determine relative asset values. Vega provides (similar to what Libra intended)
```
jeremyletang, line 129: let's not speak about libra pleeeease. no traces of facebook should even touch our things.

jeremyletang (changes requested):
As suggested by Valentin, the specs needs update to cover spam protection with UpdateMarket proposals.

protocol/0062-SPAM-spam_protection.md
Vegaklaus: How about 'A basket of currencies roughly representing their promotional assets on Vega ?

protocol/0062-SPAM-spam_protection.md
Vegaklaus: So the settlement in the first trades is in ERC20 token and not ETH ?

protocol/0062-SPAM-spam_protection.md
Vegaklaus: This will only be needed once trading has started, does that still hold then ?

protocol/0062-SPAM-spam_protection.md
ze97286: as for what we're _doing_, it's signature verification, then replay then spam.

Vegaklaus: So in the current implementation, someone could use spam to fill the replay protection buffer ?

protocol/0062-SPAM-spam_protection.md
Vegaklaus: Mostly, but there's gaps left - given medium resources, one can create a spam account that is indistinguishable (if looking at assets/revenue generated) from someone who is honestly trading intensively

protocol/0062-SPAM-spam_protection.md
ze97286: how would they do that?

protocol/0062-SPAM-spam_protection.md
Vegaklaus: We might want a temporary rule along 'If 0% of accounts caused 90% of traffic, they're slowed down (if we have tendermint prioritie queues, they could be put on lower priority)

protocol/0062-SPAM-spam_protection.md
gordsport: Can we move this to the draft spec for future spam protection

protocol/0062-SPAM-spam_protection.md
gordsport: Can we move this to the draft spec for future spam protection

protocol/0062-SPAM-spam_protection.md
gordsport: Can we move this to the draft spec for future spam protection

protocol/0062-SPAM-spam_protection.md
```diff
@@ -69,7 +71,138 @@ For Sweetwater, the policies we enforce are relatively simple:
...
+ 	
+ <spam_multiplier>: If the vega network is suubject to overload, all the results from the
+ 	previous variables are multiplied with this. This can happen repeatedly, until
+ <spam_maximum>   :  ... the combined multiplier reaches this value.
+
+
+
+
+
+
```
gordsport, line 173: Can we move this to the draft spec for future spam protection

davidsiska-vega (dismissed):

gordsport:
Closing as this is is covered in current specs 

Will refer back as also contains future spec details

PR #959, opened 2022-03-04
Clarify how withdrawals can be blocked by validators

edd:
Prior to this commit, it was said to be required that a transaction could be cancelled, but it wasn't clear how. This PR adds a section that describes that the `nonce` field is the mechanism to do this, and clarifies the Acceptance Criteria


edd:
@jeremyletang mentioned this on a call earlier today, I volunteered to clarify in the spec https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1646413345127439

If @barnabee / @jeremyletang agree that this is clear, is that point now testable? It's the only criteria not currently covered by a test. Would it have to be a system test or can it be done just in the contract tests?

edd:
@barnabee / @jeremyletang this requires your input

edd:
This doesn't work

PR #968, opened 2022-03-07
Rewrite 0020

### LINKED ISSUE #894: 0020-APIS-core_api needs an update and acceptance criteria
Blocked: https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1642784754192400

edd:
The spec, as written, was a long way from what was actually implemented, and given that what was implemented
was fine, we agreed to rewrite it. So this is that rewrite.

- Remove fields that were not added (primarily: things that have since been deemed non-core
- Add network params, governance and assets that were not originally covered
- Add acceptance criteria

It would be possible to add Acceptance Criteria codes through the spec and remove the bespoke section, but having
the Acceptance Criteria section is *consistent* and *less noisy* than adding codes throughout

Closes #894

### DIFF
--- a/protocol/0020-APIS-core_api.md
+++ b/protocol/0020-APIS-core_api.md
@@ -1,96 +1,75 @@
 # Core API
 
-## Background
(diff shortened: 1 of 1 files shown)


davidsiska-vega (approved):
Nice!

PR #972, opened 2022-03-08
Create 0072-spam-protection-PoW

Vegaklaus:
First draft of PoW spam protection


protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,64 @@
...
+ 
+ The validators verify that
+   H(b) is the correct hash of a past block
+   That block is no more than <spam_PoW_number_of_past_blocks> in the past.
+        This check is primarily done by the leader (i.e., block creator). On agreeing
+        on a new block, all parties check their mempool for now outdated transactions and purge them.
+   The hash is computed correctly and begins with <spam_PoW_difficulty> zeroes
+   
+Furthermore, the validators check than
+  The same identifyer has not been used for another transaction
```
ze97286: that can only be for transactions that were included in blocks

ze97286: or otherwise if send to the same node.

protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,64 @@
...
+  The same block has not been used for more than <spam_PoW_number_of_tx_per_block> transactions
```
ze97286: that's not true if increase_difficulty is > 0

protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,64 @@
...
+Notes: 
+- We do not require to feed the hash of the actual transaction into to hash function;
+this allows users to pre-compute the PoW and thus allows them to perform low
+latency transactions.
+- As for replay protection, there is a danger that a trader communicates with a slow validator,
+  and thus gets a wrong block number. The safest is to check validators worth > 1/3 of the 
+  weight and take the highest block hash.
+
+
+If increasing difficulty is set to 1, then more trnasactions can be tired to one block by increasing the difficulty;
```
ze97286: is increasing difficulty a boolean or an int, i.e. is it used to increase the difficulty by i or just to say that difficulty needs to be increased by 1 for each additional tx?

protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,64 @@
+
+
+Parameters: <spam_PoW_number_of_past_blocks>
```
ze97286: are these parameters hard coded? if not how does the client (e.g. wallet) get them? query vega?

protocol/0000-spam-protection-PoW
ze97286: but tbh I don't think the concept of increase difficulty is useful. I don't know if the order of transactions in deliverTx has anything to do with the time they were received so lets say spam_PoW_number_of_tx_per_block is 5 and someone sent 5 transactions to node 1 and then 2 transactions to node 2 with the minimum difficulty - and lets say all would be added to the block. When we check for spam in deliverTx we don't know which 2 transactions should be checked for increased difficulty so unless I'm missing the point here, I don't think we should bother.

protocol/0000-spam-protection-PoW
Vegaklaus: Good point. I was assuming it can be changed, but that would require the wallet to be able to query some configuration information from the datan/nodes/validators (they already query block height, would that add a lot of effort) ?

protocol/0000-spam-protection-PoW
Vegaklaus: We could check that if there's, say, 10 transaction linked to one block, at least 5 of them are with the increased difficulty.
The idea of this was to make it possible for heavy traders to compute more and thus trade more, rather than having a fixed number of transactions that are possible, thus avoiding a scenario where honest traders can't trade anymore.

protocol/0000-spam-protection-PoW
ze97286: adding it to the height request would be cheap but also kind of an odd place to put it. putting it anywhere else would require the client to run two queries to vega just for a single transaction. So it's a choice between bad options.

protocol/0000-spam-protection-PoW
ze97286: I don't know what the baseline difficulty is going to be, and what value spam_PoW_number_of_past_blocks would have, so I don't know realistically if they can prepare too many transactions for a block number that's going to go out of scope. More likely they'd prefer to have it in the next block than to invest in increasing difficulty, but I may be wrong. On the other hand I just don't see how we verify it correctly, in the sense that we're verifying one transaction at the time, what difficulty to we verify it with? in your example, the default for the first 5 and the +1 for each further transaction, and if they fail? btw, the spec also mentions banning them if they have too many transactions.

protocol/0000-spam-protection-PoW
ze97286: but I think I can live with it, i.e. have network parameters on vega side for those and return their values in the blockheight/hash query.

protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,68 @@
...
+            <spam_PoW_difficulty>
+            <spam_PoW_hash_version>
+            <spam_PoW_number_of_tx_per_block>
+            <spam_PoW_increasing difficulty>
```
davidsiska-vega: Can we clearly call these network parameters. 

Can we give valid ranges plus sensible defaults.

protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,68 @@
...
+Hash function:
+The hash-function used is SHA3 <pending verification>. To allow for a more fine-grained control over the difficulty
+of the Pow (the number of zeros only allows halving/doubling), the parameter <spam_PoW_hash_function> allows to 
+increase the number of rounds of the hash function (currently 24), e.g., <spam_PoW_hash_function> = "sha3_36_rounds".
+The parameter can in th future also be used to replace the SHA-3 through a governance vote (assuming other functions
+have been made available by then) should this prove necessary.
+
+
+
+Acceptance Criteria:
```
davidsiska-vega: Can we label following the usual format.

protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,68 @@
...
+
+
+If increasing difficulty is set to 1 (seen as a boolean flag), then more transactions can be tired to one block by increasing the difficulty. 
+This happens in batches of <transactions_per_block>, 
+e.g., if transactions_per_block is 5 and difficulty is 7, then the 6th to 10th transaction can be tied to the same block
+with difficulty 8, the 11th would have difficulty 9, etc.
+
+
+Hash function:
+The hash-function used is SHA3 <pending verification>. To allow for a more fine-grained control over the difficulty
```
Vegaklaus: This may change, as SHA-3 may be too close for hash functions with existing specialised mining rigs.

emilbayes:
I have generated test vectors that we can include and have a few details on the exact method of PoW that was implemented in Go. The exact method is as follows:

```
PoW = hash("Vega_SPAM_PoW" || Hex(H(b)) || Hex(tid) || BEU64(x))
```

Where the prefix `Vega_SPAM_PoW` is an ASCII byte string, `Hex(.)` is the hex encoding of the byte string in ASCII and `BEU64` is the Big Endian encoding as an unsigned 64-bit integer. `||` denotes concatenation. It may also be worth noting that the hash function SHA-3 with 24 rounds is SHA-3-256 proper per spec (which does 24 rounds of the keccak permutation), which I first interpreted as being 24 iterations of SHA-3. (granted this may change per @Vegaklaus comment). Surveying the most popular implementations Rust implementations (for web wallet), it is not possible to customise the rounds or bit-rate, which is worth considering. Another option for more fine-grained control over difficulty may be a popcount of the first N bits.

Example in Rust:

```rust
use tiny_keccak::{Hasher, Sha3};

// Test vector for difficulty >= 8
let mut sha3 = Sha3::v256();
let mut digest = [0u8; 32];

let nonce: u64 = 402;
sha3.update(b"Vega_SPAM_PoW");
sha3.update(b"ffb67ea4111d466d363a5c8f355bf81e2e3504563af273f5de81a005a6247e14"); // H(b)
sha3.update(b"c40de04280ce8c40ee41b5005c23a1358b4fbf31f6dcb675e8246b174458274e"); // tid
sha3.update(&nonce.to_be_bytes());
sha3.finalize(&mut digest); // 0053ea7687bd7652803af4300a7e17868267c32e4fb7f09375c46c367fd7646b
```

Test vectors:

```json
[
  {
    "difficulty": 4,
    "nonce": 0,
    "blockHash": "792ca202b84226c739f9923046a0f4e7b5ff9e6f1b5636d8e26a8e2c5dec70ac",
    "tid": "3b8399cdffee2686d75d1a96d22cd49cd11f62c93da20e72239895bfdaf4b772",
    "proof": "03f9f7d9911d3ca37c3356f10cd04273e788d1f57a9bc2396e7b5aa2e8d74557"
  },
  {
    "difficulty": 8,
    "nonce": 402,
    "blockHash": "ffb67ea4111d466d363a5c8f355bf81e2e3504563af273f5de81a005a6247e14",
    "tid": "c40de04280ce8c40ee41b5005c23a1358b4fbf31f6dcb675e8246b174458274e",
    "proof": "0053ea7687bd7652803af4300a7e17868267c32e4fb7f09375c46c367fd7646b"
  },
  {
    "difficulty": 12,
    "nonce": 2560,
    "blockHash": "d9ae00ce4c4fc96d8e72bb18f6990b833cc7724ad70322604c572f6e194d777f",
    "tid": "fcbbb4cc8dcd402a07af050bb809a04bd82f9c95b6e5a56768d3724a4abb09f0",
    "proof": "0008bbe071959bfe7fc426c4f378fcdb9540b3f931f4a0b09469f5bf0fddcb86"
  },
  {
    "difficulty": 16,
    "nonce": 23845,
    "blockHash": "dc4b61de2138856406acdabcc502be708bff7c945857ea032011a8b4b0cf54f4",
    "tid": "3954a15b2e1ec457ae100c56e2aa43786b4612644926403d59fd8cdcb29d825f",
    "proof": "00000fd8f55699845ac3192af013928916050eab088437943708b83b27490862"
  },
  {
    "difficulty": 20,
    "nonce": 85863,
    "blockHash": "8890702af457ddcda01fba579a126adcecae954781500acb546fef9c8087a239",
    "tid": "74030ee7dc931be9d9cc5f2c9d44ac174b4144b377ef07a7bb1781856921dd43",
    "proof": "000007542dcb39d1471fd6c7424a547b9039382e055ceed10c839f2b76f88c0d"
  }
]
```

protocol/0000-spam-protection-PoW
```diff
@@ -0,0 +1,86 @@
+
+
+Parameters: <spam_PoW_number_of_past_blocks>                = 100
+            <spam_PoW_difficulty>                           = 20   <should correspond to ca. 15 seconds on a PC>
```
ze97286: as discussed offline, this should be 17 at most.

ze97286 (dismissed):

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -0,0 +1,65 @@
...
+- In the original spam protection, we want to do anti-spam before verifying signatures; this order, however, cannot be done if the consequence of spam is banishment. Thus, here the order is:
+  1. check if the account is banished and (if so) ignore the transaction
+  2. check if the basic PoW with lowest difficulty is done properly
+  3. verify the signatures
+  4. put the transaction on the blockchain
+  5. if the signed transactions violate the conditions, issue the banishment 
+  6. if the signed transactions in a block violate the conditions, remove the offending ones from the block before calling vega [May need discussion]
+- Depending on how things pan out, we may have an issue with the timing; to make sure traders have sufficient time to get the block height needs us to have a large parameter of `<PoW_number_of_past_blocks>`, which may allow too many transactions. There are ways to fix this (e.g., the block height needs to end with the same bit as the validator ID), but for now we assume this doesn't cause an issue.
```
barnabee: I am not sure I understand this. What would we consider "a large parameter of `<PoW_number_of_past_blocks>`"?

It would seem reasonable that I can query a data node for the latest block and get a response in less than 1 block at say 1–1.5 second blocks, so it would seem unlikely that I am using a hash more than 5 blocks old? I could presumably query regularly or even listen to a stream of this data, in which case 5 blocks seems a wide buffer beyond worst case?

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -0,0 +1,65 @@
+# Spam protection Proof of Work
+
+```
+Parameters: `<spam_PoW_number_of_past_blocks>`                = 100  (range: 10-500)
```
davidsiska-vega: Could we label these the same as core, now that this has been implemented?

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -0,0 +1,65 @@
+# Spam protection Proof of Work
+
```
davidsiska-vega, line 5:
```suggestion
Vega does not charge fees on transactions. Fees are only charged on trades that execute. The main reason for not charging transaction fees on transactions like limit order submissions or governance votes is that this is an activity that should be encouraged (limit orders provide price information and liquidity, governance votes are essential for smooth running of the system). Because of this Vega has a novel client-side-proof-of-work mechanism to prevent transaction spam. 
```

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -0,0 +1,65 @@
...
+   
+Furthermore, the validators check that:
+- The same identifier has not been used for another transaction from a previously committed block. If the same identifier is used for 
+  several transactions in the same block, these transactions need to be removed during post-processing, and the initiator blocked as a 
+  spammer.
+- The same block has not been used for more than `<spam_PoW_number_of_tx_per_block>` transactions per spam difficulty level 
+  (i.e., if `increase_difficulty` is `> 1`, the same block can be used for more transactions if the PoW accordingly increases in difficulty).
+ 
+ Violations of the latter rules cannot lead to a transaction being removed, as different validators have a different view on 
+ this; however, they can be verified post-agreement, and the offending vega-key can be banished for 4 epochs.
```
davidsiska-vega, line 36: Are we hard-coding 4 epochs here?

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -0,0 +1,65 @@
...
+latency transactions.
+- As for replay protection, there is a danger that a trader communicates with a slow validator, and thus gets a wrong block number. The safest is to check validators worth > 1/3 of the  weight and take the highest block hash.
+- Due to Tendermint constraints, a decision if a transaction is to be rejected or not can only be done based on information that is either synchronized through the chain or contained in the transaction itself, but not based on any other transactions in the mempool. Thus, if a client ties too many transactions to the same block or does not execute the increased difficulty properly, we can not stop this pre-agreement, only detect it post-agreement. This is the reason why some violations are punished with banishment rather than prevented.
+- As some violations could come through missconfigurations, we may consider a less strict way of banishment, e.g., only to a long term banishment for repeat/high volume offenders.
+- In the original spam protection, we want to do anti-spam before verifying signatures; this order, however, cannot be done if the consequence of spam is banishment. Thus, here the order is:
```
davidsiska-vega: "Original spam protection" will be meaningless for most people reading this. Can we have a link to the relevant spec please?

davidsiska-vega (changes requested):
Apart from the specific comments (and one suggestion); this spec now talks about things that have been implemented and so should as much as possible unambiguously set out what things are. We should thus remove the "could do this", "might do that" bits as much as possible.

protocol/0072-SPPW-spam-protection-PoW.md
gordsport: added link: https://github.com/vegaprotocol/specs-internal/pull/972/commits/f0543b3a59f341ad9a45960b03a352b5c026585a

protocol/0072-SPPW-spam-protection-PoW.md
Vegaklaus: We can also turn this into a network parameter

protocol/0072-SPPW-spam-protection-PoW.md
Vegaklaus: There may be an unfortunate combination here (I'm talking to a data-node/validator that is slow, have slow software myself, etc; we'd have to see how this actually would pan out in practice

protocol/0072-SPPW-spam-protection-PoW.md
gordsport: Hey @Vegaklaus 

I have updated the parameters in the spec to align with the network parameter names:

SpamPoWNumberOfPastBlocks = `spam.pow.numberOfPastBlocks`
SpamPoWDifficulty = `spam.pow.difficulty`
SpamPoWHashFunction = `spam.pow.hashFunction`
SpamPoWNumberOfTxPerBlock = `spam.pow.numberOfTxPerBlock`
SpamPoWIncreasingDifficulty = `spam.pow.increaseDifficulty`

in this commit: https://github.com/vegaprotocol/specs-internal/pull/972/commits/dd65fa9e8bb37d1fe524e845bec8fc433497798c

Please can you review and let me know if these are OK? (if not I can revert) 
(cc @davidsiska-vega )

davidsiska-vega: That's awesome, thanks @gordsport .

protocol/0072-SPPW-spam-protection-PoW.md
davidsiska-vega: @ze97286 says hard coded; let's keep it for now because we have other stuff hard-coded in places.

gordsport: @ze97286 - is this banishment period (4 epochs) a network param?

protocol/0000-spam-protection-PoW
gordsport: ACs are now captured with the AC codes as per other specs

davidsiska-vega (approved):

PR #977, opened 2022-03-09
Add staking account details to 0013

### LINKED ISSUE #854: Clarify how staking accounts work
https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0013-ACCT-accounts.md#reference-level-explanation

The spec has Acceptance Criteria for how staking account balances work, but doesn’t actually describe in detail that they’re not really accounts in the same was as the other accounts, or how/when transfers occur etc. I think some work could be done to link this to other specs (e.g. epochs, staking bridge) and generally improve it.

The target audience for this update is external users

edd:
Staking accounts aren't really accounts in the same way as other
accounts, but they are referred to as accounts so I added a section that
clarifies this, and better linked together 0013-ACCT and 0059-STKG.

- Move 0013-ACCT acceptance criteria to the bottom
- Add See More section to 0059-STKG
- Add Staking Accounts section to 0013-ACCT

Closes #854

### DIFF
--- a/protocol/0013-ACCT-accounts.md
+++ b/protocol/0013-ACCT-accounts.md
@@ -1,44 +1,5 @@
 Feature name: accounts
 
-# Acceptance Criteria
-
-## All accounts
-
-- [ ] Double entry accounting is maintained at all points. (<a name="0013-ACCT-001" href="#0013-ACCT-001">0013-ACCT-001</a>)
-- [ ] Only transfer requests move money between accounts. (<a name="0013-ACCT-002" href="#0013-ACCT-002">0013-ACCT-002</a>)
-
-## Party asset accounts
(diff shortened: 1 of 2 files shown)


protocol/0013-ACCT-accounts.md
```diff
@@ -108,4 +67,48 @@ When a [market launches](./0043-MKTL_market_livecycle.md), an insurance pool acc
 
 When a market is finalised / closed remaining funds are distributed to the on chain treasury.  This occurs using ledger entries to preserve double entry accounting records within the collateral engine.
```
edd, line 69: This diff looks bigger than it really is. Lines 70 to 76 are really the bulk of the changes in this file.

davidsiska-vega (approved):

PR #978, opened 2022-03-09
fix: modified acceptance code criterias sequence

cpuja84:
Amended the acceptance code criterias sequence for auction spec


edd (approved):
👍 Luckily these appear to be untested so no problem with re-sequencing. I’m sure you checked too!

PR #981, opened 2022-03-18
Trading terminated on no open positions

davidsiska-vega:


witgaw:
Looks good to me, I also don't have a strong opinion on waiting for settlement price.

For: 
- We don't create a special case, less branching in the code.
- Things work meatier in data node (that market still has settlement price info available)

Against:
- Chances are it's some sort of a weird market anyway, so maybe that price will never arrive? (though that eventuality is probably covered already so maybe not a problem)

davidsiska-vega:
So for me it means that markets proposed with internal time trigger for settlement eventually clean up after themselves if no trading happens on them. 

If there are issues about the oracle (lost keys, dodgy data) then it gives the participants on the market a chance to self-organise, close all positions and the market will then clean up - they won't need a governance vote to provide settlement price. 

As for data node data: if the market had no positions the settlement price sent through can be `0` or `+Inf`  without consequence so potentially rather misleading - perhaps better to have no data.

witgaw (approved):
LGTM

barnabee:
Real concern is what this looks like when we have custom products. Would we be comfortable stating that for every product if the state is `trading terminated` and there are no positions the state is automatically changed to `settled` regardless of any logic that exists for either state? 

My suspicion is that there will be edge cases where this is not desirable.

There's also the data and other processing side of things to consider:
- A market in a given product might always be expected to have a settlement price (for example) when in settled status. If the price is missing then this can't be relied upon.
- Similarly, I should [eventually] be able to query the oracle data on the data node, and this will be silently discarded if we implement this change.
- Other actions that don't require open positions but do require a settlement price (or some other oracle data/trigger — remember that in the general case there can be various triggers for the state change, too) may be tied to the the data source or logic that triggers the state change, but if the market is settled and has been discarded by the core, these would not be processed. (An example here could be something like payment of collected fees to an oracle provider, which may be dependent on verification of the data received, and therefore not possible at an earlier stage.)

However, we can potentially say that for cash settled futures these are not concerns and so I'd be happy to add the shortcut at a product level, but I don't think it would  make sense across all settlement.

So, I think if we do this it should be a change to the `protocol/0016-PFUT-product_builtin_future.md` spec rather than the settlement spec. (On the receipt handler for the trading terminated event, we add a check for open `interest == 0` to select which state to set.)

davidsiska-vega:
> what this looks like when we have custom products. Would we be comfortable stating that for every product if the state is trading terminated and there are no positions the state is automatically changed to settled regardless of any logic that exists for either state?

I really can't think there could be anything to settle if no-one has a position. That should be universal across all derivatives and all markets.

davidsiska-vega:
> A market in a given product might always be expected to have a settlement price (for example) when in settled status. If the price is missing then this can't be relied upon.

But a settlement price that was submitted to a market with no positions should not be relied upon because it's submission had no consequences. I suggest it's better to have it not available than misleading.

davidsiska-vega:
After a long discussion (not here) it was decided that this specific shortcut isn't worth the hassle. 

So even a market in opening auction that receives trading terminated will wait for settlement input before releasing margin / bond balances.

PR #992, opened 2022-03-23
Use retained oracle input if trading terminated is after settlement data arrived.

### LINKED ISSUE #991: Inconsistency in ACs between futures and settlement specs
We've agreed with @barnabee that if settlement data comes before trading terminated then the core is to hold onto the data and settle with the most recent price data *after* the trading terminated is triggered. 

The reason for this is that there may not be another data point from the settlement oracle for whatever reason.

davidsiska-vega:
Closes #991

### DIFF
--- a/protocol/0002-STTL-settlement.md
+++ b/protocol/0002-STTL-settlement.md
@@ -113,12 +113,14 @@ The [market lifecycle spec](./0043-MKTL-market_lifecycle.md) provides detail on 
 
 1. Market has a status of ACTIVE and is trading in default trading mode (continuous trading)
 1. An [oracle event occurs](./0045-DSRC-data_sourcing.md) that is eligible to settle the market, as defined on the [Product](./0001-MKTF-market_framework.md) (see also [cash settled futures spec](./0016-PFUT-product_builtin_future.md)). In other words the settlement price is submitted to the market before trading is terminated. 
-This oracle input is ignored and market is in the default trading mode (continous trading).
-1. At least one party places an order that triggers a trade (just to prove that we can).
+This oracle input retained and market is in the default trading mode (continous trading).
+1. At least one party places an order that triggers a trade (just to prove that we can). 
+1. An [oracle event occurs *again*](./0045-DSRC-data_sourcing.md) that is eligible to settle the market, as defined on the [Product](./0001-MKTF-market_framework.md) (see also [cash settled futures spec](./0016-PFUT-product_builtin_future.md)). In other words the settlement price is submitted to the market before trading is terminated. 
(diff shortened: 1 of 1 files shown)


protocol/0002-STTL-settlement.md
```diff
@@ -113,12 +113,14 @@ The [market lifecycle spec](./0043-MKTL-market_lifecycle.md) provides detail on
...
+1. At least one party places an order that triggers a trade (just to prove that we can again). 
 1. The product's [trading terminated trigger is hit](./0016-PFUT-product_builtin_future.md#41-termination-of-trading)
 1. The market's status is set to [TRADING TERMINATED](./0043-MKTL-market_lifecycle.md) and accepts no trading but retains the positions and margin balances that were in place after processing the trading terminated trigger. No margin recalculations or mark-to-market settlement occurs.
-1. An [oracle event occurs](./0045-DSRC-data_sourcing.md) that is eligible to settle the market, as defined on the [Product](./0001-MKTF-market_framework.md) (see also [cash settled futures spec](./0016-PFUT-product_builtin_future.md))
-1. Final cashflow is calculated according to the valuation formula defined on the product (see [cash settled direct futures product](./0016-PFUT-product_builtin_future.md#42-final-settlement-expiry))
+1. Final cashflow is calculated according to the valuation formula defined on the product (see [cash settled direct futures product](./0016-PFUT-product_builtin_future.md#42-final-settlement-expiry)) using the *most recent* retained settlement price input.
```
barnabee: Note that this must happen during processing of the same transaction that triggered the `trading terminated` status.

barnabee:
It feels like it's going to cause a bunch of trouble in future that acceptance criteria for product logic sit in the settlement spec but I understand the reasons.

Request that we either add a note that the a lot of these criteria are product specific to built in cash-settled futures to the spec or create a sub-heading for these product specific ACs.

barnabee (changes requested):
One change to an AC requested plus addition of note as per https://github.com/vegaprotocol/specs-internal/pull/992#issuecomment-1076483139

protocol/0002-STTL-settlement.md
davidsiska-vega: updated to take this into account

davidsiska-vega:
> It feels like it's going to cause a bunch of trouble in future that acceptance criteria for product logic sit in the settlement spec but I understand the reasons.
> 
> Request that we either add a note that the a lot of these criteria are product specific to built in cash-settled futures to the spec or create a sub-heading for these product specific ACs.

I've added specifically that it's for cash settled future into the AC.

barnabee (approved):

PR #1001, opened 2022-03-30
fix: in the case there is no stake in the network all get voting powe…

ze97286:
…r of 10


MuthuVega:
@ze97286 - Thanks Zohar , please can you check the AC for 0066-VALW-003 as well please ?

Below is what I see in test: (I see 10 instead of 1)

Screenshot 2022-03-30 at 16 31 26 (2)

davidsiska-vega (approved):

PR #1015, opened 2022-04-07
Fix equity like share spec bug

### LINKED ISSUE #1014: The market value proxy in the spec is broken
At the moment we are calculating market value proxy as follows.
```
total_stake = sum of all LP stakes
active_time_window = [max(t-t_market_value_window_length,0), t]
active_window_length = t - max(t-t_market_value_window_length,0)

if (active_window_length > 0)
    factor =  t_market_value_window_length / active_window_length
    traded_value_over_window = total trade value for fee purposes of all trades executed on a given market during the active_time_window
    market_value_proxy = max(total_stake, factor x traded_value_over_window)
else
    market_value_proxy = total_stake
```

By setting the `t_market_value_window_length` longer or shorter I can change the market value proxy arbitrarily. The number is meaningless. 

Change is needed to calculate equity like share without going to makret value proxy and instead rely only on market growth.

davidsiska-vega:
Closes #1014 . 

See also https://docs.google.com/spreadsheets/d/14AgZwa6gXVBUFBUUOmB7Y9PsG8D4zmYN/edit?usp=sharing&ouid=103355453259573428114&rtpof=true&sd=true

### DIFF
--- a/protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
+++ b/protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
@@ -60,7 +60,6 @@ The guiding principle of this section is that by committing stake a liquidity pr
 
 ### Calculating liquidity provider equity-like share
 
-
 The parameter which determines the period over which market value and hence growth is `t_market_value_window_length` which could be e.g. a week. 
 From the end of the opening auction, which we will refer to as `t0` until `t0+t_market_value_window_length` is the `0th` or "bootstrap period". Then from  `t0+t_market_value_window_length` until `t0 + 2 x t_market_value_window_length` is the `1st` period and so on. 
 For each LP we track the stake they have and also their virtual stake. 
@@ -80,16 +79,20 @@ LP i virtual stake <- LP i virtual stake x (LP i stake + delta)/(LP i stake).
 
 Independently of the above we also update all virtual stakes at start of each new period. 
 To that end "total value for fee purposes" is cumulated over the period. For a period `n` call this `T(n)`. 
-It is possible that there have been no trades. To avoid division by `0`, at start of each period add `trade value for fee purposes of size 1` to `T(n)` to avoid possible divisions by `0`.
-For `T(0)` i.e. set `T(0) = trade value for fee purposes of resolving opening auction`.  
+Set `T(0) = trade value for fee purposes of resolving opening auction`.  
+From this we calculate the running average trade value for fee purposes:
+```
+A(n) <- A(n-1) x (n-1)/n + T(n)/n.
+```
 The g`r`owth of the market is then 
 ```
-r =  (T(n)-T(n-1))/T(n-1)
+r =  (A(n)-A(n-1))/A(n-1)
 ```
 Thus at the end of period `n` update
 ```
-LP i virtual stake <- (1 + r) x (LP i virtual stake).
+LP i virtual stake <- max(LP i physical stake, (1 + r) x (LP i virtual stake)).
 ```
+Thus the virtual stake of an LP will always be at least their physical stake.
 
 The equity like share for each LP is then
 ```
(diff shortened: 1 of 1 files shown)


protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -54,124 +54,53 @@ At time of call:
...
 
-At any time let's say we have `market_value_proxy` calculated above and existing liquidity providers as below
-```
-[LP 1 stake, LP 1 avg_entry_valuation]
-[LP 2 stake, LP 2 avg_entry_valuation]
-...
-[LP N stake, LP N avg_entry_valuation]
-```
+The parameter which determines the period over which market value and hence growth is `t_market_value_window_length` which could be e.g. a week. 
+From the end of the opening auction, which we will refer to as `t0` until `t0+t_market_value_window_length` is the `0th` or "bootsrap period". Then from  `t0+t_market_value_window_length` until `t0 + 2 x t_market_value_window_length` is the `1st` period and so on.
```
witgaw:
```suggestion
From the end of the opening auction, which we will refer to as `t0` until `t0+t_market_value_window_length` is the `0th` or "bootstrap period". Then from  `t0+t_market_value_window_length` until `t0 + 2 x t_market_value_window_length` is the `1st` period and so on. 
```

witgaw (dismissed):
LGTM

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -54,124 +54,53 @@ At time of call:
...
-LP 1 stake = 200, LP 1 avg_entry_valuation = 100
+Say an `LP i` wants to decrease their commitment by `delta < 0`. Then we update 
 ```
+LP i virtual stake <- LP i virtual stake x (LP i stake + delta)/(LP i stake).
+``` 
 
-Say now LP 2 wishes to enter and submits a "liquidity-provision-order-type" with `new_stake=200`. We have `n=1` and implicitly `(LP 2 stake)(1) == 0` is `True` and so we set `(LP 2 stake)(2) = 200` and `(LP 2 avg_entry_valuation)(2) = market_value_proxy(1) = 200`. After the update, at `n = 2` we record the state as
+Independently of the above we also update all virtual stakes at start of each new period. 
+To that end "total value for fee purposes" is cumulated over the period. For a period `n` call this `T(n)`. 
+It is possible that there have been no trades. To avoid division by `0`, at start of each period add `trade value for fee purposes of size 1` to `T(n)` to avoid possible divisions by `0`.
```
barnabee: is this 1 in position decimal places (i.e. if P. DP is 2 then `100`) or the absolute value of integer `1`?

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -54,124 +54,53 @@ At time of call:
...
+r =  (T(n)-T(n-1))/T(n-1)
 ```
-
-Another "liquidity-provision-order-type" type comes in saying that LP 1 wants `new_stake = 300`. We have `market_value_proxy(2) = 400` and `(LP 1 equity)(2) = (LP 1 stake)(2) x market_value_proxy(2) / (LP 1 avg_entry_valuation)(2) = 200 x 400 / 100 = 800`.
-We will be in the case `new_stake = 300 > (LP 1 stake)(2) = 200`. So then `delta = 100` and then `(LP i avg_entry_valuation)(1) = (800 x 100 + 100 x 400) / (800 + 200) = 120000 / 1000 = 120`.  After the update, at `n = 3` we record the state as
+Thus at the end of period `n` update
 ```
-LP 1 stake = 300, LP 1 avg_entry_valuation = 120
-LP 2 stake = 200, LP 2 avg_entry_valuation = 200
+LP i virtual stake <- (1 + r) x (LP i virtual stake).
```
barnabee: Should this be `max((1 + r) x (LP i virtual stake), actual_stake)` so that new participants comitting the same stake as me couldn't be better off? True I could remove and re-add my liquidity butt only if there's enough. Also realise that this could have been a problem before but it seems like its maybe more obviously bad now we are talknig "virtual stake" 🤔

barnabee (commented):
Some questions

barnabee (approved):

PR #1044, opened 2022-04-15
Anyone can exempt themselves from deposit limits 

davidsiska-vega:
This is to remove a potential point of centralisation.

### DIFF
--- a/non-protocol-specs/0003-NP-LIMI-limits_aka_training_wheels.md
+++ b/non-protocol-specs/0003-NP-LIMI-limits_aka_training_wheels.md
@@ -68,11 +68,11 @@ This allows the stoppage of all deposits and withdrawals after the discovery of 
 
 ### Deposit Limit Exemptions
 
-This allows for the listing of specific Ethereum addresses to be able to deposit more than the lifetime limit for the asset. This is primarily for liquidity providers and rewards.
+This allows for the listing of specific Ethereum addresses to be able to deposit more than the lifetime limit for the asset also ignoring the limit introduced per vega key. This is primarily for liquidity providers and other sophisticated participants and for depositing rewards.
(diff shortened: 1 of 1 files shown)


non-protocol-specs/0003-NP-LIMI-limits_aka_training_wheels.md
```diff
@@ -129,14 +130,11 @@ This allows for the listing of specific Ethereum addresses to be able to deposit
          - This is true even if both TXs target different [Vega public keys](./../protocol/0017-PART-party.md)
      - Withdrawing all funds after the first transaction, then placing a valid second deposit transaction that causes total lifetime deposits to exceed `max lifetime deposit` is still rejected
      - A single deposit transaction that is more than `max lifetime deposit` rejected 
-     - `lifetime deposit` is tracked across [checkpoints](./0005-NP-LIMN-limited_network_life.md)
```
barnabee, line 132: Why is this removed?

barnabee (commented):
One question about a removed bullet point, otherwise OK to approve

non-protocol-specs/0003-NP-LIMI-limits_aka_training_wheels.md
davidsiska-vega: All of this is on the smart contract side; tracking lifetime deposit happens on core side. Seems unrelated. 

Also, the bridge contract has to track lifetime deposit per Vega key and per Ethereum key to enforce the limits so this seems redundant.

barnabee: Ok makes sense given the Ethereum contract is unaware of checkpoints.

PR #1045, opened 2022-04-19
Add recurring transfer LNL / rewards criteria

gordsport:
We recently implemented a new reward system (using recurring transfers to the pool) which I believe is tested as a feature via system tests already.

Now we want to make sure that this feature is already covered over an LNL restarts.

Acceptance as per the system tests issue:

- https://github.com/vegaprotocol/system-tests/issues/687


gordsport:
Hi @davidsiska-vega @jeremyletang 
Can this be reviewed so that the tests that Jake is working on now can have the AC codes mapped?

davidsiska-vega (approved):

PR #1046, opened 2022-04-19
add AC for change of network parameter in 0042-LIQF-setting fees and rewards

Jiajia-Cui:


davidsiska-vega:
The market value proxy calculation is deprecated; start from the latest master.

PR #1048, opened 2022-04-20
Modify asset proposals

### LINKED ISSUE #1043: Governance to change assets
It looks like we don't have a way of modifying assets via governance. This is needed for all sorts of reasons including limits in https://github.com/vegaprotocol/specs-internal/blob/master/non-protocol-specs/0003-NP-LIMI-limits_aka_training_wheels.md and just to be able to fix incorrectly proposed assets.

davidsiska-vega:
Closes #1043

### DIFF
--- a/protocol/0027-ASSP-asset_proposal.md
+++ b/protocol/0027-ASSP-asset_proposal.md
@@ -40,7 +40,8 @@ Once this has been done, the new asset is ready to be used in the vega network t
 
 ## Modifying an existing asset
 
-If an asset modification that went through [governance](./0028-GOVE-governance.md) is enacted and it changes on of: `maximumLifetimeDeposit`, `withdrawalDelayPeriod` and `withdrawalDelayThreshold` then a signed payload for the appropriate bridge is emmited so that anyone can submit this to the bridge contract and cause the changes to be appropriately reflected there. 
+If an asset modification that went through [governance](./0028-GOVE-governance.md) is enacted and it changes on of: `maximumLifetimeDeposit`, `withdrawalDelayPeriod` and `withdrawalDelayThreshold` then a signed payload for the appropriate bridge is emmited (and that's all that happens) so that anyone can submit this to the bridge contract and cause the changes to be appropriately reflected there. 
+Vega will then update it's internal asset definition once the events are emmitted and confirmed the correct number of times by the bridge chain.
 
 # Pseudo-code / Examples
 
--- a/protocol/0027-ASSP-asset_proposal.md
+++ b/protocol/0027-ASSP-asset_proposal.md
@@ -47,7 +47,10 @@ If an asset modification that went through [governance](./0028-GOVE-governance.m
 ### Bridged chain part
 If it changes one of: `maximumLifetimeDeposit`, `withdrawalDelayPeriod` and `withdrawalDelayThreshold` then a signed payload for the appropriate bridge is emmited. 
 Anyone willing to pay the transaction fee (gas) can submit this to the bridge contract via multisig control and cause the changes to be appropriately reflected there. 
-Vega will then update it's internal asset definition once the events are emmitted and confirmed the correct number of times by the bridge chain.
+Vega will then update it's internal asset definition once the events are emmitted and confirmed the correct number of times by the bridge chain.  
+
+**Note on asset bundles produced but not submitted to the bridge.** If an asset update `A` is produced and never submitted to bridged chain bridge contract and subsequently an asset update `B` is produced then out of order use is a possibility (someone can submit `A` after `B` has been submitted). 
+The onus is on the creator of proposal `B` to submit (and pay the gas for) for proposal `A` before their proposal `B`. (this means that `A` cannot be submitted again).
 
 ### Vega chain part
 If it changes `quantum` then this new value becomes used immediately on enactement. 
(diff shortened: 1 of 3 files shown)


protocol/0027-ASSP-asset_proposal.md
```diff
@@ -38,6 +38,10 @@ Once the asset as been whitelisted, the network will receive from the event queu
 
 Once this has been done, the new asset is ready to be used in the vega network to create new markets.
 
+## Modifying an existing asset
+
+If an asset modification that went through [governance](./0028-GOVE-governance.md) is enacted and it changes on of: `maximumLifetimeDeposit`, `withdrawalDelayPeriod` and `withdrawalDelayThreshold` then a signed payload for the appropriate bridge is emmited (and that's all that happens) so that anyone can submit this to the bridge contract and cause the changes to be appropriately reflected there.
```
jeremyletang: Alright, these 3 are separate signatures. Should we allow all of them to be update at the same time?
What do we do to enforce the bundles to be submitted to ethereum?

jeremyletang (commented):
Seems like there's updates to te LP mecanics in this PR as well, is that expected?

davidsiska-vega:
@jeremyletang the liquidity stuff came when I did merge master (I think). Ignore that.

protocol/0027-ASSP-asset_proposal.md
barnabee: Assuming you mean three signatures for multisig control, if it is possible **without changing the Ethereum smart contracts** to sign a compound/bundled transaction that does all three in one, that would be preferred. However, if that requires contract changes and reaudit then lets avoid that and keep three separate ones. 

From a Vega perspective, the proposal to modify any number of the three should be a compound update, like:

```
changes {
	assetLimits {
		 maximumLifetimeDeposit = 3
		 withdrawalDelayPeriod = 7d
		 withdrawalDelayThreshold = 4
	}
}
```

Any unmodified values could be ignored.

It looks like the only way to prevent out-of-order execution (i.e. a change A gets approved, not submitted on Ethereum, a new change B gets approved and submitted, A is later submitted, reverting B which was approved later) would be for validators to refuse to sign a new update until the previous update is actually submitted and its nonce marked as used. Is that correct?

In that case I guess we either need to do this or we need to accept that out of order use is a possibility and put the onus on the creator of proposal B to submit (and pay the gas for) proposal A before their proposal is enacted in such a case. What do you think @jeremyletang @davidsiska-vega??

barnabee (commented):
Comments and pushed significant changes, pls review @davidsiska-vega @jeremyletang

protocol/0027-ASSP-asset_proposal.md
jeremyletang: I think that's a lot more complexity with overseeing before...

> It looks like the only way to prevent out-of-order execution (i.e. a change A gets approved, not submitted on Ethereum, a new change B gets approved and submitted, A is later submitted, reverting B which was approved later) would be for validators to refuse to sign a new update until the previous update is actually submitted and its nonce marked as used. Is that correct?

That's correct.

> In that case I guess we either need to do this or we need to accept that out of order use is a possibility and put the onus on the creator of proposal B to submit (and pay the gas for) proposal A before their proposal is enacted in such a case. What do you think

Well I'd rather have the least complexity as possible on the core side, but can understand adding the complexity too...

protocol/0027-ASSP-asset_proposal.md
```diff
@@ -1,44 +1,72 @@
...
+Anyone willing to pay the transaction fee (gas) can submit this to the bridge contract via multisig control and cause the changes to be appropriately reflected there. 
+Vega will then update it's internal asset definition once the events are emmitted and confirmed the correct number of times by the bridge chain.
+
+### Vega chain part
+If it changes `quantum` then this new value becomes used immediately on enactement. 
+
+# Changes initiated on chain
+
+In addition to changes initiated by governance on Vega, an asset's particulars can change on its originating chain. 
+The details that are sourced from the originating chain may vary by blockchain and aasset standard.
```
jeremyletang:
```suggestion
The details that are sourced from the originating chain may vary by blockchain and asset standard. 
```

protocol/0027-ASSP-asset_proposal.md
```diff
@@ -1,44 +1,72 @@
...
+The details that are sourced from the originating chain may vary by blockchain and aasset standard. 
+For ERC20 on Ethereum, this would be the asset's `name`, `symbol`, and `totalSupply`.
+
+Vega nodes will run nodes for all bridged chains. They will either listen to the relevant events or poll the current value of these data reguarly in order to ensure that the asset data on Vega reflects the current value (after the configured number of confirmations). 
+In the case of Ethereum ERC20 assets, in will be necessary to poll the contract's "read" functions (name, symbol, totalSupply) as specified in the ERC20 standard, because no events are standardised for changes to these values.
+This polling would ideally be done with each new Ethereum block, but if this is too expensive (in computational costs — there is no gas for read functions) then as long as it occurs at least once per epoch this is acceptable. 
+
+**Note on `decimals`.** The Vega ERC20 bridge does not support assets with a changing number of deicmals, and is unlikely ever to support such assets (due to both the added complexity and the lack of demonstrable use cases for this).
```
jeremyletang:
```suggestion
**Note on `decimals`.** The Vega ERC20 bridge does not support assets with a changing number of decimals, and is unlikely ever to support such assets (due to both the added complexity and the lack of demonstrable use cases for this). 
```

protocol/0027-ASSP-asset_proposal.md
davidsiska-vega: > accept that out of order use is a possibility and put the onus on the creator of proposal B to submit (and pay the gas for) proposal A before their proposal is enacted in such a case. 

I'd go with that for now for simplicity.

barnabee (approved):

PR #1051, opened 2022-04-20
Add details on oracle spec referring internal oracles

### LINKED ISSUE #1086: Rework the Oracle spec
This spec needs to be rework to be easier to understand and match the vocabulary we use in the core.

ValentinTrinque:
This has been implemented but not documented.


protocol/0048-DSRI-data_source_internal.md
```diff
@@ -51,6 +51,31 @@ on: {
 
 ```
 
+## Implementation
+
+Usage of internal oracle data are specified using properties prefixed with `vegaprotocol.builtin`, on the oracle spec.
+
+Today, only the time-triggered one is implemented through the property name `vegaprotocol.builtin.timestamp`.
```
barnabee: We need to addjust this text, this is not a time trigger so much as a built in timestamp data source. We should give it a separate section for built in data sources is it behaves functionally differently. I can do this update.

barnabee (changes requested):
Need to update this still. I will do this.

barnabee (dismissed):
> We need to addjust this text, this is not a time trigger so much as a built in timestamp data source. We should give it a separate section for built in data sources is it behaves functionally differently. I can do this update. 

I've made these updates, please rereview

AnExsomnis (dismissed):

AnExsomnis (approved):

PR #1060, opened 2022-04-22
Add ACs for network parameters in 0059-STKG-simple_staking_and_delega…

### LINKED ISSUE #1017: Define expected behaviour after a change to: reward.staking.delegation.maxPayoutPerParticipant
In order to have a test to be sure that a change to the `reward.staking.delegation.maxPayoutPerParticipant` network parameter has the expected resulting behaviour we need to define:

- Values that the parameter should change from / to
- the expected behaviour when the values are changed

### LINKED ISSUE #1030: Define expected behaviour after a change to: reward.staking.delegation.maxPayoutPerEpoch
In order to have a test to be sure that a change to the `reward.staking.delegation.maxPayoutPerEpoch` network parameter has the expected resulting behaviour we need to define:

- Values that the parameter should change from / to
- the expected behaviour when the values are changed

### LINKED ISSUE #1031: Define expected behaviour after a change to: reward.staking.delegation.competitionLevel
In order to have a test to be sure that a change to the `reward.staking.delegation.competitionLevel` network parameter has the expected resulting behaviour we need to define:

- Values that the parameter should change from / to
- the expected behaviour when the values are changed

Jiajia-Cui:
…ting

close with #1030 
close with #1031


protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,14 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
...
-  
+
+### Changing network parameters
+- Change of network paramter `validators.delegation.minAmount` will change the smallest fraction of the governance token that can be delegated to a validator at the beginning of the next epoch; delegation with tokens less than this parameter will fail the verification at the beginning (when the command is issued and before it is put on the chain). (<a name="0059-STKG-031" href="#0059-STKG-031">0059-STKG-031</a>)
+- Change of network paramter `reward.staking.delegation.delegatorShare` will change the fraction of a validator’s rewards is distributed between its delegators in a new delegation and at the end of an epoch (Default value 0.883. Valid range is between 0 and 1 inclusive). (<a name="0059-STKG-033" href="#0059-STKG-033">0059-STKG-033</a>)
+- Change of network paramter `reward.staking.delegation.minValidators` will change the minimum number of validators in a new delegation. (Default value 5. Valid values are integers greater than or equal to 1.). (<a name="0059-STKG-034" href="#0059-STKG-034">0059-STKG-034</a>)
```
Vegaklaus: This value also affects the stakescore, so it needs to be tested that the next reward as well as the next tendermint weight is computed correctly (the name is a bit unfortunate now, as we settled for a fixed number of validators, so this really is only affecting the validator score.
For the range check, it should be verified that this value is not bigger than the actual number of validators.

Vegaklaus: This value also affects the stakescore, so it needs to be tested that the next reward as well as the next tendermint weight is computed correctly (the name is a bit unfortunate now, as we settled for a fixed number of validators, so this really is only affecting the validator score.
For the range check, it should be verified that this value is not bigger than the actual number of validators.

protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,14 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
...
+- Change of network paramter `reward.staking.delegation.maxPayoutPerParticipant` will change the maximum amount that can be distributed to a single participant from the reward pool at the end of a new epoch. Number must be including the correct padding with zeros instead of possible decimal places. (<a name="0059-STKG-036" href="#0059-STKG-036">0059-STKG-036</a>)
+- Change of network paramter `reward.staking.delegation.minimumValidatorStake` will change the minimum amount of staking asset that a validator needs to self-delegate to be eligible for validator rewards at a new delegation. Number must be including the correct padding with zeros instead of possible decimal places. (<a name="0059-STKG-037" href="#0059-STKG-037">0059-STKG-037</a>)
+
```
Vegaklaus, line 220: Is there a specification on what happens if that value is increased and thus existing validators do no longer meet the criteria, i.e., if there's a grace period in which the delegators can acquire the new stake before their rewards are scrapped ? According to spec (0069), all validators that have less than reward.staking.minimumValidatorStake will be removed, though it's not specified how at this point - we want to avoid that too many validators disappear at the same time.

In either case we need ACs of the form "

- If the value is changed so that one or more validators do not have sufficient own stake anymore, they are removed (according to the removal policy). Also (this is an issue in 0069, but I add this here for now) it is not defined if an Ersatzvalidator needs to have sufficient ownstake.

- if the value is changed so that a validator that obtains sufficient own stake through the parameter change (and did not have it before) qualifies to become a validator/ersatzvalidator, they are promoted accordingly in the next epoch.

Vegaklaus, line 220: Is there a specification on what happens if that value is increased and thus existing validators do no longer meet the criteria, i.e., if there's a grace period in which the delegators can acquire the new stake before their rewards are scrapped ? According to spec (0069), all validators that have less than reward.staking.minimumValidatorStake will be removed, though it's not specified how at this point - we want to avoid that too many validators disappear at the same time.

In either case we need ACs of the form "

- If the value is changed so that one or more validators do not have sufficient own stake anymore, they are removed (according to the removal policy). Also (this is an issue in 0069, but I add this here for now) it is not defined if an Ersatzvalidator needs to have sufficient ownstake.

- if the value is changed so that a validator that obtains sufficient own stake through the parameter change (and did not have it before) qualifies to become a validator/ersatzvalidator, they are promoted accordingly in the next epoch.

Jiajia-Cui: maybe we can put a cap on how much reward.staking.delegation.minimumValidatorStake can be reduced?

protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,19 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
   - Announcing removal of stake for current validator
   - Staking on the new validator, as per normal [function: Stake](../non-protocol-specs/0006-NP-STAK-erc20_governance_token_staking.md)
   - These can happen concurrently, so that at the next epoch, the stake is removed from the current validator and staked on the new validator 
-  
+
+### Changing network parameters
+- Change of network paramter `validators.delegation.minAmount` will change the smallest fraction of the governance token that can be delegated to a validator at the beginning of the next epoch; delegation with tokens less than this parameter will fail the verification at the beginning (when the command is issued and before it is put on the chain). (<a name="0059-STKG-031" href="#0059-STKG-031">0059-STKG-031</a>)
```
davidsiska-vega: This one doesn't say what should happen to those who have current delegation that falls below a new, lower, minimum.

protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,19 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
...
+- Change of network paramter `reward.staking.delegation.delegatorShare` will change the fraction of a validator’s rewards is distributed between its delegators in a new delegation and at the end of an epoch (Default value 0.883. Valid range is between 0 and 1 inclusive). (<a name="0059-STKG-033" href="#0059-STKG-033">0059-STKG-033</a>)
```
davidsiska-vega: Yes, but does impact immediately the epoch in which it's approved (as I think it should), or the next one.

protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,19 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
...
+- Change of network paramter `reward.staking.delegation.minValidators` will change the minimum number of validators in a new delegation. (Default value 5. Valid values are integers greater than or equal to 1.); It will also affects the stake_score on validators, so it needs to be tested that the next reward as well as the next tendermint weight is computed correctly (the name is a bit unfortunate now, as we settled for a fixed number of validators, so this really is only affecting the validator_score. (<a name="0059-STKG-038" href="#0059-STKG-038">0059-STKG-038</a>)
```
davidsiska-vega, line 211: This makes no sense; this parameter impacts how the validator score is calculated. The only question is as above: does impact immediately the epoch in which it's approved (as I think it should), or the next one.

protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,19 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
...
+- For the range check, it should be verified that this value is not bigger than the actual number of validators. (<a name="0059-STKG-034" href="#0059-STKG-034">0059-STKG-034</a>)
```
davidsiska-vega: I'm not sure why there should be the range check - it's just a constant that goes into a formula for rewards - maybe governance wants to change it in anticipation of more validators joining or whatever.

protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,19 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
...
+- Change of network paramter `reward.staking.delegation.maxPayoutPerParticipant` will change the maximum amount that can be distributed to a single participant from the reward pool at the end of a new epoch. Number must be including the correct padding with zeros instead of possible decimal places. (<a name="0059-STKG-036" href="#0059-STKG-036">0059-STKG-036</a>)
```
davidsiska-vega, line 212: I don't think there is `maxPayoutPerParticipant` anymore. 

Update: it got removed from the spec but shouldn't have been. PR correcting this: https://github.com/vegaprotocol/specs-internal/pull/1109

protocol/0059-STKG-simple_staking_and_delegating.md
```diff
@@ -202,7 +202,19 @@ See the [network paramters spec](./0054-NETP-network_parameters.md#current-netwo
...
+- Change of network paramter `reward.staking.delegation.minimumValidatorStake` will change the minimum amount of staking asset that a validator needs to self-delegate to be eligible for validator rewards at a new delegation. Number must be including the correct padding with zeros instead of possible decimal places； 
+
+If the value is changed so that one or more validators do not have sufficient own stake anymore, they are removed (according to the removal policy). Also (this is an issue in 0069, but I add this here for now) it is not defined if an Ersatzvalidator needs to have sufficient ownstake.
```
davidsiska-vega: Yes, ersatz validators need to meet the minimum stake requirement. Please add it to 0069 to clarify as part of this PR.

davidsiska-vega: I'd suggest we clarify that "they are removed (according to the removal policy)" means one per epoch, starting with the one with lowest own stake.

davidsiska-vega (changes requested):

protocol/0059-STKG-simple_staking_and_delegating.md
Vegaklaus: I'd say the best way is to not touch them, and apply this rule only to future delegations, or does anyone see an issue with that ?

protocol/0059-STKG-simple_staking_and_delegating.md
Vegaklaus: Min Validators is a parameter used in the anti-whaling, and thus should also enter somewhere in the tender mint weight calculation. I'll check that.

protocol/0059-STKG-simple_staking_and_delegating.md
Jiajia-Cui:
```suggestion
If the value is changed so that one or more validators do not have sufficient own stake anymore, they are removed (according to the removal policy as defined in 0069) one per epoch, starting with the one with lowest own stake.

File: protocol/0059-STKG-simple_staking_and_delegating.md (hunk 6 above)
Comment (Jiajia-Cui) [reply protocol/0059-STKG-simple_staking_and_delegating.md]:
```suggestion

```

protocol/0059-STKG-simple_staking_and_delegating.md
Jiajia-Cui:
```suggestion
- Change of network paramter `validators.delegation.minAmount` will change the smallest fraction of the governance token that can be delegated to a validator at the beginning of the next epoch; we only apply the change to new delegations. 
New delegations with tokens less than this parameter will fail the verification at the beginning (when the command is issued and before it is put on the chain). Old delegations will not be impacted by this change of parameter. 
(<a name="0059-STKG-031" href="#0059-STKG-031">0059-STKG-031</a>)
```

protocol/0059-STKG-simple_staking_and_delegating.md
Jiajia-Cui:
```suggestion
- Change of network paramter `reward.staking.delegation.delegatorShare` will change the fraction of a validator’s rewards is distributed between its delegators in a new delegation and at the end of an epoch (Default value 0.883. Valid range is between 0 and 1 inclusive). The change will impact immediately the epoch in which it's approved.(<a name="0059-STKG-033" href="#0059-STKG-033">0059-STKG-033</a>)
```

protocol/0059-STKG-simple_staking_and_delegating.md
Vegaklaus: Can you elaborate why it makes no sense ?

gordsport:
Closing as covered in a PR for post enable trading

PR #1063, opened 2022-04-25
Split market proposal and LP commitment

### LINKED ISSUE #1047: Separate initial liquidity from market proposals
It may be better to separate market proposal from the initial liquidity provision since the parties that hold governance token may not be the same as the parties that hold the settlement asset. 

After this change it may be possible to propose a market and have it accepted without any liquidity committed; such a market will sit in an opening auction (extended possibly indefinitely due to lack of liquidity) until some is committed.

davidsiska-vega:
Closes #1047

### DIFF
--- a/protocol/0043-MKTL-market_lifecycle.md
+++ b/protocol/0043-MKTL-market_lifecycle.md
@@ -10,8 +10,10 @@ Markets on Vega are permissionlessly proposed using the [governance mechanism](.
 - whether the market is open for trading
 - status of settlement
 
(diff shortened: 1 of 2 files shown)


jeremyletang (dismissed):

protocol/0043-MKTL-market_lifecycle.md
```diff
@@ -10,8 +10,10 @@ Markets on Vega are permissionlessly proposed using the [governance mechanism](.
...
+Markets proposed via [governance proposals](./0028-GOVE-governance.md#1-create-market) undergo certain additional validations. Note the distinctions between a proposal that is `valid` or `accepted` and a proposal that is `sucessful`. A `valid` proposal has passed or will pass validation checks; an `accepted` proposal has been received in a Vega transaction and passed validation checks; and a `successful` proposal has been voted for and won. The proposal becomes `enacted` when the action specified (i.e. for the purposes of this spec, market creation/update/close). 
 
+A market can be proposed without any [liquidity commitment](./0038-OLIQ-liquidity_provision_order_type.md). If the proposal is successful the market will go into opening auction at least until the proposed `enactment` date.
```
ValentinTrinque: > A market can be proposed without any [liquidity commitment

Does it means we have to support proposing a market with and without liquidity commitment ? Or should we just support, submit a market and, then submit liquidity ?

davidsiska-vega: Good point. Just without, I'll clarify.

ValentinTrinque (approved):

PR #1074, opened 2022-04-27
fix: rewrite reward spec to match the new design

### LINKED ISSUE #1061: Update 0056-REWA spec and ACs
The 0056-REWA spec needs to be updated in line with the new implementation:

- https://github.com/vegaprotocol/vega/issues/5273

> Now each market has it’s own reward account, they no longer compete and the reward is never split between any markets

and the following discussion:

https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1650635067593759?thread_ts=1650561120.374139&cid=CAHA5EX0F

If any addition ACs are required, or existing needing amending, please also do so under this issue.

ze97286:


protocol/0056-REWA-rewards_overview.md
```diff
@@ -58,66 +57,174 @@ That is if we have reward account balance `R`
...
+
 Metrics will be calculated using the [decimal precision of the settlement asset](????-????-market-decimal-places.md).
 
-### For market creation metrics
+### For market creation metrics
 
-In any epoch when for any asset and for any markets settling in the asset, if market total [trade value for fee purposes](0029-FEES-fees.md) since market creation exceeds `quantum x rewards.marketCreationQuantumMultiple` then split the reward pool (for market creation in said asset) equally between all markets. 
+In the *first* epoch during which the market total [trade value for fee purposes](0029-FEES-fees.md) since market creation exceeds `quantum x rewards.marketCreationQuantumMultiple` - if there is balance in the reward account for the corresponding reward type for the market, the proposer of the market gets paid in the reward payout asset. 
 
+If at the end of the epoch during the market total goes over the threshold there is no balance in the market creation reward account - the proposer will not get the reward - the window of opportunity for this reward requires that there is balance in the reward account at the epoch during which the threshold of trading has been crossed.
```
barnabee, line 68: We need to decide if this should even exist in this form. I'm not sure if it would get used.

It would be better if all markets that crossed the threshold since a reward was last paid out in that currency were considered. Otherwise I doubt this will be useful.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -58,66 +57,174 @@ That is if we have reward account balance `R`
...
 
+If at the end of the epoch during the market total goes over the threshold there is no balance in the market creation reward account - the proposer will not get the reward - the window of opportunity for this reward requires that there is balance in the reward account at the epoch during which the threshold of trading has been crossed. 
 
-## Acceptance criteria
+Similarly to the other metrics market creation may be paid in more than one asset if someone funds a reward account with the corresponding metric type and market in scope with an arbitrary payout asset. 
+
+NB: if a market is being recreated from checkpoint, and trading in it is restarted - the tracking of trading value restarts from 0, regardless of how much trading was done before and if a proposer bonus has been given before the checkpoint was taken. Therefore after the checkpoint reload, the proposer may be eligible to the market proposer bonus again once the market trading value goes above the threshold for the first time.
```
barnabee, line 72: This makes this reward even less useful. I am not sure whether we are going to change it or just accept it may not get used at this point…

protocol/0056-REWA-rewards_overview.md
```diff
@@ -58,66 +57,174 @@ That is if we have reward account balance `R`
 then calculate `M:= m_1+m_2+...+m_n` and transfer `R x m_i / M` to party `p_i` at the end of each epoch. 
 If `M = 0` (no-one incurred / received fees for a given reward account)  then nothing is paid out of the reward account and the balance rolls into next epoch. 
 
+Metrics are reset at the end of the epoch, so in the case where there is no reward account or no reward balance at the end of the epoch for the given market in scope, the participants contributing to the relevant metric will not be compensated for their contribution. Their contribution to fees is not being carried over to the next epoch. 
+
 Metrics will be calculated using the [decimal precision of the settlement asset](????-????-market-decimal-places.md).
```
```suggestion
Metrics will be calculated using the [decimal precision of the settlement asset](0070-MKTD-market-decimal-places.md).
```

protocol/0056-REWA-rewards_overview.md
```diff
@@ -58,66 +57,174 @@ That is if we have reward account balance `R`
...
+#### Setup
+There are 3 assets configured on the Vega chain: $VEGA, USDT, USDC. 
+There are no markets.
+
+* `transfer.fee.factor` = 0
+* `maker_fee` = 0.0001
+* `infrastructure_fee` = 0.0002
+*  `ETHUSD-MAR22` market which settles in USDT is launched anytime in epoch 1 by `party_0`
+* `party_0` and `party_1` provide auction orders so there is a trade to leave the opening auction and the remaining best bid = 2700 and and best offer = 2800 are supplied by party_0 each with volume 10. 
+* Moreover `party_0` provides liquidiity with `liquidity_fee` = 0.0003 and offset + 10 (so their LP volume lands on 2690 and 2810).
```
```suggestion
* Moreover `party_0` provides liquidity with `liquidity_fee` = 0.0003 and offset + 10 (so their LP volume lands on 2690 and 2810).
```

protocol/0056-REWA-rewards_overview.md
```diff
@@ -58,66 +57,174 @@ That is if we have reward account balance `R`
...
+* During epoch `2` we have `party_1` make one buy market order with volume `2`.
+* During epoch `2` we have `party_2` make one sell market order each with notional `1`.
 
-A `party_0` makes a transfer of `90` `$VEGA` to `USDT | Sum of fees paid | VEGA` in epoch `2`.
-During epoch `2` we have `party_1` make one buy market order with volume `2` USDT. 
-During epoch `2` we have `party_2` make one sell market order each with notional `1` USDT.  
-Verify that at the end of epoch `2` the metric `sum of fees paid` for `party_1` is:
+#### Funding reward accounts
+* `party_R` is funding multiple reward accounts for the same metric and same market to be paid in different assets (`$VEGA`, `USDC`)
+   * `party_R` makes a transfer of `90` `$VEGA` to `ETHUSD-MAR22 | Sum of fees paid | VEGA` in epoch `2`. (`ETHUSD-MAR22` is just brevity, this should be the market id not name).
```
```suggestion
   * `party_R` makes a transfer of `90` `$VEGA` to `ETHUSD-MAR22 | Sum of fees paid | VEGA` in epoch `2`. (`ETHUSD-MAR22` is just for brevity here, the transfer is specified by market id not its name).
```

protocol/0056-REWA-rewards_overview.md
```diff
@@ -58,66 +57,174 @@ That is if we have reward account balance `R`
...
+   * `party_R` makes a transfer of `120` `USDC` to `ETHUSD-MAR22 | Sum of fees paid | USDC` in epoch `2`. (`ETHUSD-MAR22` is just brevity, this should be the market id not name).
```
```suggestion
   * `party_R` makes a transfer of `120` `USDC` to `ETHUSD-MAR22 | Sum of fees paid | USDC` in epoch `2`. (`ETHUSD-MAR22` is just for brevity here, the transfer is specified by market id not its name).
```

davidsiska-vega (approved):

barnabee (commented):
Some questions / things we may need to revisit on market creation rewards, the rest is good though.

protocol/0056-REWA-rewards_overview.md
ze97286: it can be easily made this way but that's not how it was spec-ed as far as I remember. Also, especially now that it's done by market, one can transfer the funds to reward account when the market gets close to the threshold and worst case it will be paid in the following epoch. But also if we do that, we need to also store in the checkpoint information about whether the bonus has been paid.

protocol/0056-REWA-rewards_overview.md
ze97286: Well, the alternative is to save the balance of the value of the market in the checkpoint.

protocol/0056-REWA-rewards_overview.md
ze97286: although maybe it doesn't matter. If there's a balance in the reward account for the market in some asset, and the market has gone above the threshold - it gets paid. For the other comment we need to store in the checkpoint the market value and then after restore we continue the count and if/when the market is above the threshold AND there's funds in the market proposer reward account for the market in any asset - it gets paid. Does it make sense? @barnabee

protocol/0056-REWA-rewards_overview.md
barnabee: Yeah this sounds like the right approach, but per the above this is probably only worth doing if we can work how we will be able to realistically use this reward type…

protocol/0056-REWA-rewards_overview.md
barnabee: It makes sense, and your points are all correct but he more I think about it the more the market creation reward doesn't make sense at a market level and I'm not sure if we will be able to use it.

We can have a manual process to topup but it it may also be more of a regulatory concern in this case. We can't pay people directly and had designed the reward system originally to be an automated payment over quite a long time for that reason, but now we'd be directly paying a reward pot that we know will go exactly to one key, when it needs to go to that key. We could just transfer the money to the key, the only value is the transfer type metadata, but it might well be too close to paying directly for us to get comfort to do it.

So for now I am tempted to leave it, but we'll need to evaluate in our risks call if we can use it.

PR #1076, opened 2022-04-27
Made fee calc more readable

### LINKED ISSUE #1075: Update liquidity fee spec to be more clear around fee calculation
The current [spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0042-LIQF-setting_fees_and_rewarding_lps.md) has a very fact-only example of calculating the market liquidity fee. This ticket is to add some more details to make it more readable.

peterbarrow:
Updating the test around fee calculation to make it easier to follow.

Closes #1075


candida-d (approved):
Much clearer!

PR #1078, opened 2022-04-28
Update validator node specs to match implementation

### LINKED ISSUE #1077: Update validator node specs to remove info that does not match implementation
Not covered: 

1. There is very little information about the lowest tier of validating nodes: pending (or candidate?)

2. 0069 - This acceptance criteria doesn't seem to make sense as running a non-validator node is not a requirement, the requirement is to run a validator node. 

> 1. A running non-validator node can submit a transaction to become a validator. 

3. Should this whole section be removed from 0069? A validator with any amount of stake at all is not removed, because then stakers could not remove their stake. 

> ## Removing non-performing candidate validators
> 
> - Any party with fewer than `reward.staking.delegation.minimumValidatorStake` at the start of an epoch is removed. (the aim is to remove parties who don't delegate to self, however we achieve it is not too important)
> - Any party with performance score == `0` at the end of an epoch is removed.

candida-d:
Closes #1077 
These changes are based on feedback on the documentation of validators.


protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -53,14 +52,13 @@ for example:
         "avatar_url": "https://www.greenfield.one/avatar.png"
 }
 ```
-## Removing non-performing candidate validators
+## Removing non-performing pending validators
```
candida-d, line 55: This section is unclear, should it be removed? Any validator that has stake can't be 'removed' or people can't unstake them. What actually happens, if anything?

protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -144,7 +142,7 @@ See [limited network life spec](../non-protocol-specs/0005-NP-LIMN-limited_netwo
 
 ## Joining / leaving VEGA chain (<a name="0069-VCBS-001" href="#0069-VCBS-001">0069-VCBS-001</a>)
 1. A running non-validator node can submit a transaction to become a validator.
```
candida-d: This AC doesn't sound right, as a basic requirement is to run a validator node. Perhaps... "A running non-consensus validator node can submit a transaction to become a Tendermint validator"?

davidsiska-vega (approved):

PR #1085, opened 2022-05-04
Update oracles spec

### LINKED ISSUE #1086: Rework the Oracle spec

ValentinTrinque:
Close #1086

### DIFF
--- a/protocol/0045-DSRC-data_sourcing.md
+++ b/protocol/0045-DSRC-data_sourcing.md
@@ -28,7 +28,7 @@ Note that this approach means:
 
 **Oracle:** A provider of `Oracle Data`.
 
-**Internal Oracle:** An `Oracle` that lives inside the node, and sends `Oracle Data` from the inside.
+**Internal Oracle:** An `Oracle` that provides data internally available within Vega core nodes (e.g. time, prices on other markets).
 
 **External Oracle:** An `Oracle` that lives outside the node, and sends `Oracle Data` from the outside.
 
--- a/protocol/0045-DSRC-data_sourcing.md
+++ b/protocol/0045-DSRC-data_sourcing.md
@@ -1,35 +1,52 @@
-# Data sourcing (aka oracles)
+# Oracles (aka data sourcing)
 
 ## 1. Principles and summary
 
-The Vega network runs on data. Market settlement, risk models, and other features require a supplied price (or other data), which must come from somewhere, often completely external to Vega. This necessitates the use of both internal and external data sources for a variety of purposes.
+The Vega network runs on data. Market settlement, risk models, and other features require a supplied price (or other data), which must come from somewhere, often completely external to Vega. This necessitates the use of both internal and external data sources, called oracles, for a variety of purposes.
 
 
-a) The goals of Vega Protocol with regards to data sourcing are:
+a) The goals of Vega Protocol with regard to oracles are:
 
 1. To provide access to data internal to the Vega network in a standardised way, including data and triggers related to the "Vega Time" and market data (prices, etc.)
 1. To support a wide range of third party data sourcing solutions for external data rather than to implement a complete solution in-house.
 1. To be a source of definitive and final data to Products and Risk Models that can be trusted by market participants.
 1. To build simple, generic and anti-fragile data sourcing functionality, and not to introduce third party dependencies.
 
-b) Things that are explicitly NOT goals of the data sourcing framework at this time:
+b) Things that are explicitly NOT goals of the oracle framework at this time:
 
 1. Calculations or processing of data other than selecting the value of a specific field and filtering events in the data stream are out of scope.
-1. Processing arbitrary message formats is out-of-scope. Each required format should be specified explicitly. For instance we may specify that "Vega native protobuf message of key/value pairs" or "ABI encoded data in the OpenOracle format" must be valid data (and there may be more than one required format), but do not require general consumption of arbitrary data.
+1. Processing arbitrary message formats is out-of-scope. Each required format should be specified explicitly. For instance, we may specify that "Vega native protobuf message of key/value pairs" or "ABI encoded data in the OpenOracle format" must be valid data (and there may be more than one required format), but do not require general consumption of arbitrary data.
 1. Whilst we do need to build a framework that will be *extensible* with new sources and transformation/aggregation options, and *composable* by combining options, we are not aiming to build a large library of such features for unproven use cases initially. The MVP can be built and will be very useful with a small number of features.
 
 
 Note that this approach means:
 
 1. Vega will not integrate directly with specific oracle/data providers at the protocol level. Rather, we provide APIs and protocol capabilities to support a wide range of data sourcing styles and standards (so that oracles implementing these standards will hopefully be compatible with little or no work).
-1. External data sources must be able to provide a measure of finality that is either definitive or a configurable threshold on a probabilistic measure (‘upstream finality’).
+1. External oracles must be able to provide a measure of finality that is either definitive or a configurable threshold on a probabilistic measure (‘upstream finality’).
 1. Once upstream finality is achieved, Vega may in future provide optional mechanisms for querying, verification or dispute resolution that are independent of the source. These would be composable steps that could be added to any source.
 1. Vega will allow composition of data sources, including those with disparate sources, and may in future provide a variety of methods to aggregate and filter/validate data provided by each. 
 
+## 2. Terminology
+**Oracle:** A provider of `Oracle Data`.
 
-## 2. Data sourcing framework
+**Internal Oracle:** An `Oracle` that lives inside the node, and sends `Oracle Data` from the inside.
 
-Any part of Vega requiring a data source should be able to use any type of data source. **This means that there is a single common method for specifying a data source where one is required.**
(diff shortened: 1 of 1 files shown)


protocol/0045-DSRC-data_sourcing.md
```diff
@@ -1,35 +1,52 @@
...
+## 2. Terminology
+**Oracle:** A provider of `Oracle Data`.
 
-## 2. Data sourcing framework
+**Internal Oracle:** An `Oracle` that lives inside the node, and sends `Oracle Data` from the inside.
 
-Any part of Vega requiring a data source should be able to use any type of data source. **This means that there is a single common method for specifying a data source where one is required.**
+**External Oracle:** An `Oracle` that lives outside the node, and sends `Oracle Data` through the node API.
+
+**Oracle Data:** A signed message containing data, one or multiple signatures, and the public keys used to sign these data.
```
barnabee: This definition seems like it may be confusing?

What terminology would we use for data from an internal oracle with no signatures required? Or data from from the Ethereum bridge when we have Ethereum oracles? 

The text of this definition seems more like it is defining the message type that is used to submit oracle data for signed message oracles…

barnabee (commented):
One comment/question otherwise this seems good

protocol/0045-DSRC-data_sourcing.md
ValentinTrinque: Right, we might want to clarify it.

```suggestion
**Oracle Data:** All data emitted by an `Oracle`.
**Internal Oracle Data**: `Oracle Data` emitted by an `Internal Oracle`.
**External Oracle Data**: `Oracle Data` emitted by an `External Oracle`.
**Ethereum Oracle Data**: `Oracle Data` emitted by an `Ethereum Oracle`, through a bridge.
```

ValentinTrinque: Is it better ?

barnabee: Yes I think that’s clearer

protocol/0045-DSRC-data_sourcing.md
```diff
@@ -1,144 +1,269 @@
...
-1. Vega will allow composition of data sources, including those with disparate sources, and may in future provide a variety of methods to aggregate and filter/validate data provided by each. 
+1. Vega will allow composition of data sources, including those with disparate sources, and may in future provide a variety of methods to aggregate and filter/validate data provided by each.
 
+## 2. Terminology
 
-## 2. Data sourcing framework
+**Oracle:** A provider of `Oracle Data`.
 
-Any part of Vega requiring a data source should be able to use any type of data source. **This means that there is a single common method for specifying a data source where one is required.**
+**Internal Oracle:** An `Oracle` that lives inside the node, and sends `Oracle Data` from the inside.
```
davidsiska-vega:
```suggestion
**Internal Oracle:** An `Oracle` that provides data internally available within Vega core nodes (e.g. time, prices on other markets).
```

protocol/0045-DSRC-data_sourcing.md
```diff
@@ -1,144 +1,269 @@
-# Data sourcing (aka oracles)
+# Oracles (aka data sourcing)
 
 ## 1. Principles and summary
 
-The Vega network runs on data. Market settlement, risk models, and other features require a supplied price (or other data), which must come from somewhere, often completely external to Vega. This necessitates the use of both internal and external data sources for a variety of purposes.
+The Vega network runs on data. Market settlement, risk models, and other features require a supplied price (or other data), which must come from somewhere, often completely external to Vega. This necessitates the use of both internal and external data sources, called oracles, for a variety of purposes.
```
barnabee, line 5: I find it weird that we are changing data source to oracle everywhere. Are all data sources oracles? The definition I have in my head for an oracle is a protocol for bringing off-chain data on chain [hopefully] in a trust-minimising way, not just any data source.

So it seems to me like we are changing from having a framework for data sources, _some_ of which are oracle protocols to now just defining any data source as an oracle, and at the same time somewhat conflating the framework for dealing with them within Vega with the protocol or method for sourcing and verifying the data. 

I don't really understand the rationale for doing these things but would like to if we are going down this route.

protocol/0045-DSRC-data_sourcing.md
```diff
@@ -1,144 +1,269 @@
...
+  "oracle_spec_binding": {
+    "settlement_price": "price.BTC.value"
+  }
+}
+```
+
+```
+TO REVIEW
+
+I need a concrete example to fully understand the following specification
```
barnabee, line 182: So for 2, something like this, given the slightly toy example:

```
a = make_signed_message_oracle(pub_key_list)
b = a.filter(price_source == 'uniswap')
c = b.filter(ticket == 'ZECUSD')
d = c.filter(timestamp >= 2022-06-25T00:00)
e = d.select('closing_price')
```

a, b, c, d, e are all data sources within the framework and you could swap out any valid data source definition at any point (as long as the data types match), so I could replace a, b, with a value:

```
c = internal_data_soure_value({ timestamp: 2022-06-25T00:02, closing_price = 420.69 })
d = c.filter(timestamp >= 2022-06-25T00:00)
e = d.select('closing_price')
```

…and that would work, or I could in future replace bits with data source definition that references another market in Vega, or uses an Ethereum event observed by Vega validators as an oracle, or a governance vote, etc.

When I wrote the spec the assumption was that data sources might get created/have IDs and be reused by ID., so that if multiple different markets cared about Uniswap Zcash prices (this would be common if there are multiple expiries or products on the same underlying, which there will be), you might have something like:

```
a = make_signed_message_oracle(pub_key_list)
b = a.filter(price_source == 'uniswap')
c = b.filter(ticket == 'ZECUSD')
d = c.filter(timestamp >= 2022-06-25T00:00)
e = d.select('closing_price')

zcash_oracle_id = deploy_oracle_definition_and_return_id(c)
OR
zcash_oracle_id = # ID obtained by inspecting the oracle definition from the market using "e" above

THEN you can do, for the next expiry:

settlement_price_oracle = zcash_oracle_id.filter(timestamp >= 2022-12-25T00:00).select('closing_price')

etc.....
```

I don't really know if we are supporting this… perhaps you have to copy and paste the whole definition… hopefully if that happens at least 3 intermediate data sources that are the same only get evaluated once (or at least we can get there…) but it probably doesn't matter a lot at the moment (will do more so when we have things like options which could have many markets with fully or partially duplicated data source parts).

**So if it is not supported yet, and not useful to state now, I guess the thing to do is remove it…**

protocol/0045-DSRC-data_sourcing.md
```diff
@@ -1,144 +1,269 @@
...
+For Vega to support sufficient number types to enable processing of any reasonably expected message for each format,all values of an `Oracle Data` are defined as strings, and, then, converted into the requested type. This allows complex or big numbers to be express without hitting programming language limits.
+
+In future there will likely be other types.
+
+### Type checking
+
+```
+TO REVIEW
+
+Should we rethink the part in the light of the current wording and choosen components architecture ?
```
barnabee, line 223: Can you explain what needs rethinking?

barnabee (commented):
Left some comments/responses

protocol/0045-DSRC-data_sourcing.md
ValentinTrinque: That might have been my mistake, since we are using the term "Oracle" interchangeably with data source. That's probably an abuse of language, and we should probably rework our protobuf definitions to better aligned with the "data-source" lingo.

protocol/0045-DSRC-data_sourcing.md
ValentinTrinque: Alright, I think we were bothe out of sync in the way we approach the "data-source specification". 

I didn't get that it should have been apprehended as an independent entity. Now, the activated and deactivated part of them, as well as the re-usability between them, make sense.

I saw it as something solely tied to the market, living it's own life. If another market wanted to do the same thing, it would just declare the same data-source spec.

protocol/0045-DSRC-data_sourcing.md
ValentinTrinque: Data sources that emit data of an incorrect type to a defined data source should trigger an event or log of some sort

In our current implementation, we do not have this aspect of a data-source emitting a data to a "defined data source" (as I understand, it's a "defined data source" is the result of data source spec). Today, the data comes in, and is compare against all "defined data sources", without specific target, and if it matches one of them, it's forwarded to the subscribers, otherwise it's not.

protocol/0045-DSRC-data_sourcing.md
barnabee: > something solely tied to the market, living it's own life. If another market wanted to do the same thing, it would just declare the same data-source spec.

I don’t think this is a problem for now, we can think of it like that, operate that way, and update the spec to match it then revisit the design properly for “V2” of the protocol, ie when we introduce more programmable products and markets via WASM where it will be more important.

gordsport:
@ValentinTrinque 

Can this be closed as superseded by the other linked issues in:
- https://github.com/vegaprotocol/specs/issues/1086

PR #1093, opened 2022-05-23
Update 0072-SPPW-spam-protection-PoW.md

Vegaklaus:
Added parameter change spec.

### DIFF
--- a/protocol/0072-SPPW-spam-protection-PoW.md
+++ b/protocol/0072-SPPW-spam-protection-PoW.md
@@ -14,11 +14,11 @@ spam.pow.increaseDifficulty = 0 (range: 0/1)
 If there is a governance vote on parameter changes taking effect at blockheight h, then the parameter is valid for all PoWs that are tied to a block of height h or later; that is, old pre-computations remain valid, and only new ones need to respect the new parameters. If spam.pow.numberOfPastBlocks is changed at height h, the parameter is enforced starting at block h+spam.pow.numberOfPastBlocks (being the new value). 
 There is a (theoretical) possibility that a parameter is changed repeatedly so fast that the new parameter is not enforced before it is changed again. To avoid programming complexity, the intermediate parameter is ignored, i.e., we keep the current parameter until the new onw is valid.
 Example: Suppose, we have numberOfPastBlocks = 100, and difficulty =15
- If we change difficulty to 20 in blockheight 20.000, then all transactions tied to block 20.000 or later need to solve difficulty 20
- However, if we then change the difficulty to 25 at blockheight 20.005, and the current blockheight is 19990, then we keep difficulty
- 20 for all transactions tied to block 20004 or earlier, and require 25 for block 20.005 or later. 
+ If we change difficulty to 20 in blockheight 20 000, then all transactions tied to block 20 000 or later need to solve difficulty 20
+ However, if we then change the difficulty to 25 at blockheight 20 005, and the current blockheight is 19 990, then we keep difficulty
+ 20 for all transactions tied to block 20 004 or earlier, and require 25 for block 20 005 or later. 
  
- Note: The later is done to se don't need to do complext implementations for special cases that will never occur in reality; we need to store the current parameter, and a blockheight when it switches to a new one, and no intermediates. A way to avoid that would be to make parameter changes valid immediatelly; this however would cause special cases if users need to throw away transactions and recompute the PoW, which this approach avoids.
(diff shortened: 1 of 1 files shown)


protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -11,6 +11,15 @@ spam.pow.numberOfTxPerBlock = 2 (range: 1-1000)
 spam.pow.increaseDifficulty = 0 (range: 0/1)
 ```
 
+If there is a governance vote on parameter changes taking effect at blockheight h, then the parameter is valid for all PoWs that are tied to a block of height h or later; that is, old pre-computations remain valid, and only new ones need to respect the new parameters. If spam.pow.numberOfPastBlocks is changed at height h, the parameter is enforced starting at block h+spam.pow.numberOfPastBlocks (being the new value). 
+There is a (theoretical) possibility that a parameter is changed repeatedly so fast that the new parameter is not enforced before it is changed again. To avoid programming complexity, the intermediate parameter is ignored, i.e., we keep the current parameter until the new onw is valid.
+Example: Suppose, we have numberOfPastBlocks = 100, and difficulty =15
+ If we change difficulty to 20 in blockheight 20.000, then all transactions tied to block 20.000 or later need to solve difficulty 20
```
davidsiska-vega:
```suggestion
 If we change difficulty to 20 in blockheight 20 000, then all transactions tied to block 20 000 or later need to solve difficulty 20
```

davidsiska-vega:
```suggestion
 If we change difficulty to 20 in blockheight 20 000, then all transactions tied to block 20 000 or later need to solve difficulty 20
```

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -11,6 +11,15 @@ spam.pow.numberOfTxPerBlock = 2 (range: 1-1000)
...
+ However, if we then change the difficulty to 25 at blockheight 20.005, and the current blockheight is 19990, then we keep difficulty
```
davidsiska-vega:
```suggestion
 However, if we then change the difficulty to 25 at blockheight 20 005, and the current blockheight is 19 990, then we keep difficulty
```

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -11,6 +11,15 @@ spam.pow.numberOfTxPerBlock = 2 (range: 1-1000)
...
+ 20 for all transactions tied to block 20004 or earlier, and require 25 for block 20.005 or later.
```
davidsiska-vega:
```suggestion
 20 for all transactions tied to block 20 004 or earlier, and require 25 for block 20 005 or later. 
```

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -11,6 +11,15 @@ spam.pow.numberOfTxPerBlock = 2 (range: 1-1000)
...
+There is a (theoretical) possibility that a parameter is changed repeatedly so fast that the new parameter is not enforced before it is changed again. To avoid programming complexity, the intermediate parameter is ignored, i.e., we keep the current parameter until the new onw is valid.
+Example: Suppose, we have numberOfPastBlocks = 100, and difficulty =15
+ If we change difficulty to 20 in blockheight 20.000, then all transactions tied to block 20.000 or later need to solve difficulty 20
+ However, if we then change the difficulty to 25 at blockheight 20.005, and the current blockheight is 19990, then we keep difficulty
+ 20 for all transactions tied to block 20004 or earlier, and require 25 for block 20.005 or later. 
+ 
+ Note: The later is done to se don't need to do complext implementations for special cases that will never occur in reality; we need to store the current parameter, and a blockheight when it switches to a new one, and no intermediates. A way to avoid that would be to make parameter changes valid immediatelly; this however would cause special cases if users need to throw away transactions and recompute the PoW, which this approach avoids.
```
davidsiska-vega:
```suggestion
 Note: The later is done to se don't need to do complex implementations for special cases that will never occur in reality; we need to store the current parameter, and a blockheight when it switches to a new one, and no intermediates. A way to avoid that would be to make parameter changes valid immediatelly; this however would cause special cases if users need to throw away transactions and recompute the PoW, which this approach avoids.
```

davidsiska-vega (approved):

PR #1098, opened 2022-05-31
chore: make 0046-DSRM-016 for later milestone and reword

gordsport:
As per the spreadsheet being used I have updated the text from:

> Reject any SubmitData tx that is a duplicate (i.e. contains exactly the same data payload and is for the same data source), even if it is signed by a different signer (assuming the source has multiple configured signers) or was submitted by a different Vega key.

to the suggested:

> Reject any SubmitData tx that is a duplicate. For each active oracle definition/spec we store the messages / hash of the messages that have been processed and "used" sucesssfully and will not pass a message on for that specific target/definition if it has already received it (by checking that hash). This info can be discarded along with the rest of the oracle spec/definition once the oracle is no longer active (market is settled, etc.) and the check would be done on a per market basis..


vega-paul:
Currently, at no point do we reject any well-formatted transaction submitted to the market. All trxs are accepted via the wallet, processed, logged on the core and a hash is returned back through the wallet.

gordsport:
closing as now covered in:
- https://github.com/vegaprotocol/specs-internal/pull/1057

PR #1119, opened 2022-06-12
Explicitly allow wash trading to leave an auction

davidsiska-vega:
We are explicitly *not* allowing wash trades during continuous trading. 

Spec was not explicit about allowing wash trading during auctions.


witgaw (commented):
LGTM

witgaw (approved):
LGTM

PR #1120, opened 2022-06-13
chore: make `remove_asset` ACs for next milestone

gordsport:
As the `remove_asset` is not currently supported/used the ACs related to this have been given the COSMICELEVATOR tag.


witgaw (approved):
LGTM

PR #1131, opened 2022-06-28
chore: move LNL spec to protocol folder

### LINKED ISSUE #1124: Move LNL spec from non-protocol to protocol folder
Move [LNL spec](https://github.com/vegaprotocol/specs-internal/blob/master/non-protocol-specs/0005-NP-LIMN-limited_network_life.md) from non-protocol to protocol folder.

gordsport:
Move the LNL spec to the protocol folder and update all links point to that spec

Linked to AC code updates in:
- https://github.com/vegaprotocol/system-tests/pull/880


edd (approved):
Nice and tidy! In time, *all* non-protocol specs will move over like this, this is a great example to follow when we get around to it.

PR #1133, opened 2022-06-29
Update 0066-VALW-validator_tendermint_weights.md

Vegaklaus:
Added ability to turn of/scale down the effect of weights.


gordsport:
Hey @Vegaklaus 
Speaking to @ze97286 its thought this is not required.

If this is the case can you please confirm so we can close this PR?

davidsiska-vega:
I don't think it's needed: https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1656498079092239 closing. 

We may want some kind of scaling for multisig weights - when we get there - which should be reviewed with the rest of the bridge work.

PR #1137, opened 2022-07-05
chore: mark AC codes as not implemented

gordsport:
There are 3 AC’s on the spec which are intentional not-implement and are potential future work.

they are:
- 0069-VCBS-002
- 0069-VCBS-003
- 0069-VCBS-004


witgaw (approved):
LGTM

wwestgarth (approved):

PR #1139, opened 2022-07-08
Drafts of all User interface acceptance criteria. DO NOT MERGE

campbellssource:
This branch contains drafts for all acceptance criteria. We are prioritising the review an merge of a subset of these ACs, but this branch contains the all the latest merged ACs as well as drafts of the ones that may be merged later.

Easiest way to view them is to browser the branch here: https://github.com/vegaprotocol/specs/tree/Add-interface-acceptance-criteria/user-interface#user-interface-acceptance-criteria

SORD ONLY: https://github.com/vegaprotocol/specs/pull/1181
WITH only: https://github.com/vegaprotocol/specs/pull/1223


jtsang586:
Overall, I like the level of detail throughout. The ACs are perfectly readable and pretty clear in how we expect features to behave. 
From a QA perspective I'm not entirely sure I find the `must` and `shoulds` that necessary now that I'm seeing them in a list. QA would write tests for every scenario listed if possible and naturally if a `must` step fails then a bunch of tests for the `shoulds` would also fail.

e.g. If associating tokens fails then I won't be able to see a balance of associated tokens ready to use

I wouldn't mind having the musts and shoulds merged to `should` as that's what we normally use to assert in Cypress.
And have `can` scenarios as features that we'd like to have but are not developed yet which will eventually become `should` when they are ready for testing?

campbellssource:
Files have been significantly changed and the root that lists all the files is now on master so don't need to keep this open anymore

PR #1140, opened 2022-07-11
Update 0016-PFUT-product_builtin_future.md

vega-paul:


edd:
This change isn't correct.

https://github.com/vegaprotocol/specs-internal/commit/6d9365888847d9dcb9d97e07465c5b12816afe48 removed the Acceptance Criteria that this refers to. So the real fix should be to just remove this code. Sadly my commit message on that change is terrible, so we don't have a reason for why it was removed without searching Slack.

Initially I thought that the mistake was that the acceptance criteria didn't match the filename (e.g. `0001-EDD-001` should only live in a file called `0001-EDD-Edd_is_the_title_here.md`, so because `PFUT` didn't match `PROD` it wouldn't count as an acceptance criteria for this file. **However here that was intentional** - it was a reference to another file's acceptance criteria, _that should be covered as part of the coverage of that file_, not this one. But that was broken when the code this link pointed to was removed.

edd:
~So I’ve proposed a different change. However! If you don’t think my suggestion is correct, and there’s no other acceptance criteria that covers this, then this needs a new code, and that code has to be `0016-PFUT-00x`, which the intial change didn't do.~

@davidsiska-vega's suggestion is much better 👇

davidsiska-vega:
I think we could point it to `0051-PROD-003` which says:
```
The settlement asset / settlement assets cannot be changed on a product via governance  (<a name="0051-PROD-003" href="#0051-PROD-003">0051-PROD-003</a>)
```

vega-paul:
Ive updated the test to use  0051-PROD-003 - The settlement asset / settlement assets cannot be changed on a product via governance


So can we merge this PR now which removes the AC in this spec @edd

edd (approved):

PR #1157, opened 2022-07-16
chore: remove outdated note

fkondej:


edd (approved):
Thank you for catching that

PR #1165, opened 2022-07-26
Update 0042-LIQF-setting_fees_and_rewarding_lps.md

Jiajia-Cui:
update formula in spec 0042 line 86


davidsiska-vega:
I don't think this is correct, the spec is correct as it was. Start with `A(0) := T(0)`.  Say that `T(1)` now comes in (because the period ended) and you want to calculate `A(1)=T(1)` by the formula. Now `T(2)` comes in. This is the running average which is just 
```
T(1)/2 + T(2)/2 = A(1)/2 + T(2)/2 = A(n-1) x (n-1)/n + T(2)/n ,
```
since `n=2`. So you see it's the average. You could recursively check that it gives the average for any `n`.

Jiajia-Cui:
Its not a typo

PR #1166, opened 2022-07-28
fix: duplicate spec/AC code

gordsport:
Fix the duplicate AC code for liquidity

0044-LIQM becomes 0044-LIME

Should be merged at the same time as:
- https://github.com/vegaprotocol/vega/pull/5637

to ensure coverage metrics are not incorrect for long.


davidsiska-vega (approved):
LGTM

PR #1167, opened 2022-07-28
chore: remove AC for `updateFrequency` netparam

gordsport:
The following issue was merged in core to remove the `updateFrequency` network parameter. This PR is to remove the AC related to this parameter as its not present thus not testable:

- https://github.com/vegaprotocol/vega/pull/5625


davidsiska-vega (approved):
Yes this can go away, the functionality is handled by the state variable.

PR #1168, opened 2022-07-29
fix: clarify dynamic orders

davidsiska-vega:


edd (approved):
Nice. I nearly commented that your 'such as' meant this didn't clear it up, but it does.

PR #1175, opened 2022-08-04
fix: rename els to growth adjusted LP share

davidsiska-vega:


gordsport (approved):
LGTM

Checked all other files in the repo and 'equity like' is not present too.

gordsport:
@EVODelavega - when you did the core work were the network parameters updated in line with: protocol/0028-GOVE-governance.md

PR #1179, opened 2022-08-09
Batch Instruction spec

barnabee:
Adds a spec for batch instructions


barnabee:
I have made a first draft for batch instructions.


# Outstanding questions 

## Outstanding question 1: (partly-)mixed types

I have drafted this to allow all three of cancel, amend, and new order submission in a single batch. HOWEVER, I have drafted it in such a way thay ALL cancels are processed, then ALL amends, then ALL submisisons. 

This should mean that:

- An order that's created in the batch can't then be amended or cancelled (i.e. the batch can't interact with it's created orders, given that wash trading orders are also stopped)

- An order that's amended in the batch can't then be cancelled.

I have also explicitly stated that more than one amend for the same order ID is an error and only the first for a given ID will be processed.

I *think* this makes it safe to do the batch like this, which would have two advantages: 

1. Simpler implementation: there's a single transaction type and we always process all three lists in the same order and way. 

2. More functionality: particularly, it is potentially more useful for traders, for example some may prefer to cancel all orders and submit new ones rather than amend, as post-amend it is ambiguous whether the order exists (if it had traded away, it won't).


## Outstanding question 2: auction behaviour

I have specified that auction behaviour is as normal when processing each instruction, i.e. it all works as if the instrucitons in the batch were separate transactions. EXCEPT that for liquidity monitoring auctions I've stated that these should only be entered if the conditions to do so exist after processing the whole batch.

I did this to reduce the chance of "flash auctions" and markets flickering in and out of auctions, especially as market makers will use the batches a lot.

1. Confirm that we are ok with this and it is not bad/exploitable.

2. Confirm this would not unnecessarily complicate the implementation. 


# Remaining work to do 

1. Review and refine the content (research, then engineering)

2. Write some ACs!

protocol/0074-BTCH-batch-market-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+1. Any errors encountered in processing an instruction after it passes validation must cause it to be skipped, and the errors, as well as the instruction to which they relate, must be available in the result of the transaction.
+
+1. In addition to the usual validation and other errors that can occur in processing an instruction, the following also apply:
+
+   - Any second or subsequent Amend Order instruction for the same order ID within a single Batch Instruction transaction is an error
+
+
+## Auction behaviour
+
+1. If processing an instruction would cause the end of an Opening auction, the auction must be ended and the uncrossing must occur immediately as it would if the instruction was processed as a standalone transaction.
```
MM0819: This is interesting because it would potentially cause many later instructions to be invalid (e.g. if I submit 20 submissions of new orders, and the first one causes an auction to exit, then the remaining submissions might be rejected if they are GFA orders). 

I don't know if this matters, but it feels more intuitive to me to uncross the auction after the entire batch has been processed.

barnabee: Felt like there's a manipulation/exploit opportunity if you can place order all over the causing it to uncross.

Pretty sure you could manipulate the price that way.

Besides any GFA you place might be invalid by the time it gets there 🤷

protocol/0074-BTCH-batch-market-instructions.md
```diff
@@ -0,0 +1,72 @@
...
+
+1. If processing an instruction would cause entry into a Price Monitoring auction, the auction must be entered immediately as it would be if the instruction was processed as a standalone transaction.
+
+1. If processing an instruction would cause entry into a Liquidity Monitoring auction, the auction **must not be entered immediately**. The conditions for entering a Liquidity Monitoring auction must be evaluated only after processing the entire batch instruction. This is to avoid the situation where order cancellation and/or amends cause a Liquidity Monitoring auction part way through the batch that would no longer be required once further amendments and/or new order submissions in the batch have been processed. If this condition were not in place, the auction would would be held for at least the minimum period, even though there would be no reason to do so.
```
TomMcL: This seems reasonable in that I can't immediately think how it could lead to exploits. It does feel a little worrying leaving the market in a 'non-auction' state when if we weren't within a batch we would be in an auction, but feels like the 'worst' that hopefully would happen would be someone building a larger position than they otherwise would have. 

One thing that comes to mind is @witgaw mentioning about checks which don't allow an action if they would result in the market dropping into auction. Presumably we'd still disallow any actions which would be disallowed in a standard transaction (and possibly this was only for price auctions?)

barnabee: > It does feel a little worrying leaving the market in a 'non-auction' state when if we weren't within a batch we would be in an auction

Yeah, however this is all within one atomic transaction. It's definitely a change to what you can do in one txn (though so is batching more generally), but my general feeling is that we _never want to be in a liquidity auction if we don't need to be_ so if you can do stuff all within one transaction to prevent a liquidity auction we should be pleased with that.

The downside is you can take away _more_ of the liquidity that's there before the auction is entered, but only to the extent you could have done so anyway with a single larger order.

barnabee: > One thing that comes to mind is @witgaw mentioning about checks which don't allow an action if they would result in the market dropping into auction. Presumably we'd still disallow any actions which would be disallowed in a standard transaction (and possibly this was only for price auctions?)

This is indeed one to check. I believe that applies only to price monitoring, as it's important to apply the circuit breaker and ascertain the fair price level by auction _before_ moving the mark price there, but we should confirm all the cases where this applies before proceeding.

davidsiska-vega: I feel this is potentially the most disruptive change to include from implementation/ testing view - all the auction interaction is pretty baroque. 

I see the logic but this change on its own would make me to *not* introduce it before we launch trading.... due to the potential for disruption.

barnabee: This will change with the liquidity monitoring trigger check frequency PR

protocol/0074-BTCH-batch-market-instructions.md
davidsiska-vega: The auction behavior can be taken as given from https://github.com/vegaprotocol/specs/pull/1180/

protocol/0074-BTCH-batch-market-instructions.md
davidsiska-vega: This is taken care of via https://github.com/vegaprotocol/specs/pull/1180/

TomMcL:
Added ACs for perusal

gordsport:
Remaining:
- [ ] remove reference to auctions 
- [ ] add ACs / AC codes

barnabee:
I have updated the description of how this interacts with auctions given the changes in https://github.com/vegaprotocol/specs/pull/1180

~This PR should not be without #1180. If #1180 does not go ahead, this spec will need to change.~

#1180 is merged and must be delivered before or alongside this functionality, but merging this spec is no longer blocked.

davidsiska-vega:
This https://github.com/vegaprotocol/specs/pull/1180 just went in, so we should be good.

protocol/0074-BTCH-batch-market-instructions.md
```diff
@@ -0,0 +1,88 @@
...
+ - [ ] Any batch transaction containing more than one amend to the same order ID should attempt to execute the first as normal but all further amends should error without being executed. (<a name="0074-BTCH-002" href="#0074-BTCH-002">0074-BTCH-002</a>)
+ - [ ] An error in any instruction should be logged and returned to the caller but later instructions should still be attempted. (<a name="0074-BTCH-003" href="#0074-BTCH-003">0074-BTCH-003</a>)
+ - [ ] If an instruction causes the market to exit an Opening Auction the market should exit the auction immediately before continuing with later instructions. (<a name="0074-BTCH-004" href="#0074-BTCH-004">0074-BTCH-004</a>)
+ - [ ] If an instruction causes the market to enter a Price Monitoring Auction the market should enter the auction immediately before continuing with later instructions. (<a name="0074-BTCH-005" href="#0074-BTCH-005">0074-BTCH-005</a>)
+ - [ ] An instruction which is valid at the start of the batch execution but becomes invalid before it is executed should fail. (<a name="0074-BTCH-006" href="#0074-BTCH-006">0074-BTCH-006</a>)
```
davidsiska-vega:
```suggestion
 - [ ] An instruction which is valid at the start of the batch execution but becomes invalid before it is executed should fail. In particular 1. a batch consisting of two limit order placements C1 and C2 where the party has enough balance to place either of them individually but not both should place C1 but reject C2.  2. a batch transaction containing aggressive limit order C1 which moves the market into price monitoring auction and a C2 which is marked `GFN` (good for normal) should execute C1 but reject C2. (<a name="0074-BTCH-006" href="#0074-BTCH-006">0074-BTCH-006</a>)
```

MM0819:
I've reviewed this spec and added one more AC. Aside from that, I'm happy with the spec as it stands.

davidsiska-vega (approved):

PR #1180, opened 2022-08-09
Liquidity auction at end-of-block spec

TomMcL:
Adds spec for liquidity auction transitions only happening at the end of block rather than at any point within it.


protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,28 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
 ## What happens during the auction?
 
 The auction proceeds as usual. Please see the auction spec for details.
+## Frequency of checking for auction entry triggers
+
+ Through a series of actions which occur coincidently the market may be moved into a state in which a liquidity auction is expected and then back out of said state. Ideally, liquidity auctions should only be entered when the market truly requires one as once entered a minimum auction length must be observed. Even without a minimum auction length, a market flickering between two states is suboptimal.
```
davidsiska-vega:
```suggestion
 Through a sequence of actions which occur with the same timestamp the market may be moved into a state in which a liquidity auction is expected and then back out of said state. Ideally, liquidity auctions should only be entered when the market truly requires one as once entered a minimum auction length (controlled by `market.auction.minimumDuration`) must be observed. Even with a very short a minimum auction length, a market flickering between two states is suboptimal. 
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,28 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
 ## What happens during the auction?
 
 The auction proceeds as usual. Please see the auction spec for details.
+## Frequency of checking for auction entry triggers
+
+ Through a series of actions which occur coincidently the market may be moved into a state in which a liquidity auction is expected and then back out of said state. Ideally, liquidity auctions should only be entered when the market truly requires one as once entered a minimum auction length must be observed. Even without a minimum auction length, a market flickering between two states is suboptimal. 
+
+ To resolve this, the triggers should only be checked at the end of each batch of transactions occurring with an identical timestamp (in the current Tendermint implementation this is equivalent to once per block). At the end of each such period the auction conditions should be checked and the market moved into liquidity auction state if the triggers have been hit. Similarly, the criteria for exiting the auction should be checked on the same frequency, so that a market cannot leave a liquidity auction only to immediately re-enter it at the end of the block.
```
davidsiska-vega:
```suggestion
 To resolve this, the conditions for entering a liquidity auction should only be checked at the end of each batch of transactions occurring with an identical timestamp (in the current Tendermint implementation this is equivalent to once per block). At the end of each such period the auction conditions should be checked and the market moved into liquidity auction state if the conditions for entering a liquidity auction are satisfied. 
The criteria for exiting any auction (liquidity or price monitoring) should be checked only on timestamp change (ie block boundary with Tendermint). This means that a market cannot leave a liquidity auction only to immediately re-enter it at the end of the block.
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,28 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
...
+
+ A liquidity provider should still be unable to remove their stake if doing so would reduce the total stake below the target stake even though the market would remain open between the removal and the end of the block.
```
davidsiska-vega:
```suggestion
 A liquidity provider amending LP provision order can reduce their stake (as long as total stake >= target stake) even if doing so would mean that at the end of block the system enters liquidity auction (because e.g. max open interest increased or because another LP was removed due to not meeting margin commitments). 
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,28 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
...
 
 ## Acceptance Criteria
 
 1. The scenarios in the feature test [0026-AUCT-auction_interaction.feature](https://github.com/vegaprotocol/vega/tree/develop/integration/features/verified/0026-AUCT-auction_interaction.feature) are verified and pass. (<a name="0035-LIQM-001" href="#0035-LIQM-001">0035-LIQM-001</a>)
 
-2. An incoming order that would consume `best_bid` or `best_offer` gets executed (unless it will also trigger price monitoring auction at the same time), the trades are generated and only then the market goes into a liquidity auction (because there is a peg missing to deploy the liquidity provision volume). (<a name="0035-LIQM-002" href="#0035-LIQM-002">0035-LIQM-002</a>)
\ No newline at end of file
+2. An incoming order that would consume `best_bid` or `best_offer` gets executed (unless it will also trigger price monitoring auction at the same time), the trades are generated and only then the market goes into a liquidity auction (because there is a peg missing to deploy the liquidity provision volume). (<a name="0035-LIQM-002" href="#0035-LIQM-002">0035-LIQM-002</a>)
```
davidsiska-vega:
```suggestion
2. An incoming order that would consume `best_bid` or `best_offer` gets executed (unless it will also trigger price monitoring auction at the same time), the trades are generated. The market goes into a liquidity auction at the end of a block (because there is a peg missing to deploy the liquidity provision volume). (<a name="0035-LIQM-002" href="#0035-LIQM-002">0035-LIQM-002</a>)
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,28 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
...
+   
+4. A market which enters a state requiring liquidity auction at the end of a block through decreased total stake (e.g. through LP bankruptcy) remains in open trading between entering that state and the end of the block. (<a name="0035-LIQM-004" href="#0035-LIQM-004">0035-LIQM-004</a>)
+
+5. A market which enters a state requiring liquidity auction through increased open interest during a block but then leaves state again prior to block completion never enters liquidity auction. (<a name="0035-LIQM-005" href="#0035-LIQM-005">0035-LIQM-005</a>)
+
+6. A market which enters a state requiring liquidity auction through reduced current stake (e.g. through LP bankruptcy) during a block but then leaves state again prior to block completion never enters liquidity auction. (<a name="0035-LIQM-006" href="#0035-LIQM-006">0035-LIQM-006</a>)
+
+7. A liquidity provider should still be unable to remove their liquidity within the block if this would bring the current stake below the target stake as of that transaction. (<a name="0035-LIQM-007" href="#0035-LIQM-007">0035-LIQM-007</a>)
```
davidsiska-vega:
```suggestion
7. A liquidity provider should cannot remove their liquidity within the block if this would bring the current total stake below the target stake as of that transaction. (<a name="0035-LIQM-007" href="#0035-LIQM-007">0035-LIQM-007</a>)
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,29 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
...
+   
+4. A market which enters a state requiring liquidity auction at the end of a block through decreased total stake (e.g. through LP bankruptcy) remains in open trading between entering that state and the end of the block. (<a name="0035-LIQM-004" href="#0035-LIQM-004">0035-LIQM-004</a>)
+
+5. A market which enters a state requiring liquidity auction through increased open interest during a block but then leaves state again prior to block completion never enters liquidity auction. (<a name="0035-LIQM-005" href="#0035-LIQM-005">0035-LIQM-005</a>)
+
+6. A market which enters a state requiring liquidity auction through reduced current stake (e.g. through LP bankruptcy) during a block but then leaves state again prior to block completion never enters liquidity auction. (<a name="0035-LIQM-006" href="#0035-LIQM-006">0035-LIQM-006</a>)
+
+7. A liquidity provider should cannot remove their liquidity within the block if this would bring the current total stake below the target stake as of that transaction. (<a name="0035-LIQM-007" href="#0035-LIQM-007">0035-LIQM-007</a>)
```
TomMcL:
```suggestion
7. A liquidity provider cannot remove their liquidity within the block if this would bring the current total stake below the target stake as of that transaction. (<a name="0035-LIQM-007" href="#0035-LIQM-007">0035-LIQM-007</a>)
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,29 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
...
 
 ## Acceptance Criteria
 
 1. The scenarios in the feature test [0026-AUCT-auction_interaction.feature](https://github.com/vegaprotocol/vega/tree/develop/integration/features/verified/0026-AUCT-auction_interaction.feature) are verified and pass. (<a name="0035-LIQM-001" href="#0035-LIQM-001">0035-LIQM-001</a>)
 
-2. An incoming order that would consume `best_bid` or `best_offer` gets executed (unless it will also trigger price monitoring auction at the same time), the trades are generated and only then the market goes into a liquidity auction (because there is a peg missing to deploy the liquidity provision volume). (<a name="0035-LIQM-002" href="#0035-LIQM-002">0035-LIQM-002</a>)
\ No newline at end of file
+2. An incoming order that would consume `best_bid` or `best_offer` gets executed (unless it will also trigger price monitoring auction at the same time), the trades are generated. The market goes into a liquidity auction at the end of a block (because there is a peg missing to deploy the liquidity provision volume). (<a name="0035-LIQM-002" href="#0035-LIQM-002">0035-LIQM-002</a>)
```
EVODelavega: Seeing as we could be in a situation where these pegs are missing, I'm assuming any pegged order submissions during this time should be parked. Should we try and redeploy them as soon as we have the relevant pegs available again, or do we deploy all pegged orders at the end of the block/after liquidity auction?

I think it makes most sense to deploy them ASAP.

Another possible scenario to consider is how we handle LP submissions/amendments when these pegs are missing. Do we keep some LP orders on the book, and keep the LP submission as active, or remove all LP submissions that reference the missing peg instead, and try to redeploy it ASAP (like we would with pegged orders)? What about LP's that are pegged to the mid price? They could still stay deployed, in which case there's some implications WRT fees/equity like shares perhaps

davidsiska-vega: > Should we try and redeploy them as soon as we have the relevant pegs available again 

Yes. 

> remove all LP submissions that reference the missing peg instead, and try to redeploy it ASAP 

Ideally yes. 

> What about LP's that are pegged to the mid price? 

If they peg to mid price then they'll be removed. 

> in which case there's some implications WRT fees/equity like shares perhaps 

Hypothetically yes, but the incentive mechanism isn't exact science so I'd ignore. 

I'll update the spec if you think my answers are ok.

davidsiska-vega:
```suggestion
2. An incoming order that would consume `best_bid` or `best_offer` gets executed (unless it will also trigger price monitoring auction at the same time), the trades are generated. The volume implied by LP provision is removed (from both sides of the book and for all LPs). If `best_bid` is missing but `best_ask` is present then all the "normal pegged orders" (i.e. not the LP ones) which use `best_ask` as peg are still deployed. If `best_ask` is missing but `best_bid` is present then all the "normal pegged orders" (i.e. not the LP ones) which use `best_bid` as peg are still deployed. The market goes into a liquidity auction at the end of a block (because there is a peg missing and  the liquidity provision volume is not deployed). (<a name="0035-LIQM-002" href="#0035-LIQM-002">0035-LIQM-002</a>)
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,29 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
 ## What happens during the auction?
 
 The auction proceeds as usual. Please see the auction spec for details.
+## Frequency of checking for auction entry triggers
```
davidsiska-vega:
```suggestion
## Frequency of checking for liquidity auction entry conditions
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -62,9 +62,29 @@ During the liquidity monitoring auction new or existing LPs can commit more stak
...
+
+ To resolve this, the conditions for entering a liquidity auction should only be checked at the end of each batch of transactions occurring with an identical timestamp (in the current Tendermint implementation this is equivalent to once per block). At the end of each such period the auction conditions should be checked and the market moved into liquidity auction state if the conditions for entering a liquidity auction are satisfied. 
+The criteria for exiting any auction (liquidity or price monitoring) should be checked only on timestamp change (ie block boundary with Tendermint). This means that a market cannot leave a liquidity auction only to immediately re-enter it at the end of the block.
+
+ A liquidity provider amending LP provision order can reduce their stake (as long as total stake >= target stake) even if doing so would mean that at the end of block the system enters liquidity auction (because e.g. max open interest increased or because another LP was removed due to not meeting margin commitments).
```
davidsiska-vega, line 74:
```suggestion
An implication is that within the same time stamp an aggressive order may remove the `best_bid` or `best_ask`. At that point all LP provision volume is removed from the book (and by implication at least one side of the book is empty). If a subsequent limit order re-creates the peg (still with the same timestamp / from the same block) then the LP volume is re-deployed. If no subsequent limit order places a limit order restoring the peg and we reach end of the block we end up in a liquidity auction. 

As mentioned, as a consequence, intra-block, we may end with one side of the book empty which means that a party not meeting margin requirement *cannot* be closed out. We accept this consequence, their margin will be checked again when the mark price changes and either the book is restored (and they can get closed out) or the market is in liquidity auction.
```

EVODelavega (approved):
LGTM

PR #1192, opened 2022-08-17
Update market creation rewards

barnabee:
Clarify the calculation of the market creation reward

### DIFF
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -66,16 +66,17 @@ Metrics will be calculated using the [decimal precision of the settlement asset]
 
 ### For market creation metrics
 
-Every epoch the entire reward account for every [market in scope, payout asset] will be distributed pro-rata to the market parties that submitted the market creation governance proposal for any markets that have the market creation metric (described above) `>0`. Similarly to the other metrics market creation may be paid in more than one asset if someone funds a reward account with the corresponding metric type and market in scope with an arbitrary payout asset. 
+Every epoch the entire reward account for every [market in scope, payout asset] will be distributed to the parties that submitted the original market creation governance proposal for any markets that have the market creation metric (described above) `>0`. 
+The payout for each market having a non-zero market creation metric will be the same, that is, the eligible markets share the reward pool equally. 
+This means that the total market creation reward received by a party can vary, as their reward may reflect having created multiple eligible markets, each of which earns the same payout.
+Similarly to the other metrics market creation rewards may be paid in more than one asset if someone funds a reward account with the corresponding metric type and market in scope with an arbitrary payout asset. 
 
-That is if we have reward account balance `R`
-```
-[p_1,m_1]
-[p_2,m_2]
-...
-[p_n,m_n]
-```
-then calculate `M:= m_1+m_2+...+m_n` and transfer `R x m_i / M` to party `p_i` at the end of each epoch. 
+That is if we have:
+- reward account balance: `R` of some asset
+- number of markets with a non-zero market creation metric: `M`
+- parties `p_0 … p_M` being the creators of the markets with non-zero market creations metrics, where the creator is defined as the party that submitted the original market proposal that was enacted to create the market
+
+Then transfer `R / M` to each party `p_i` (`i` from 0 to `M`) identified above. 
 If `M = 0` (no markets newly met or exceeded the reward threshold since the last payout) then nothing is paid out of the reward account and the balance rolls into next epoch. 
 
 Market creation metrics are not reset at the end of the epoch, so the cumulative volume for each market continues to accrue across epochs and is always equal to the total trade value for fee purposes since the creation of the market.
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -402,7 +402,7 @@ Market has been trading but not yet eligible for proposer bonus.
 * Setup and fund multiple reward account for the market ETHUSDT:
     * Transfer 10000 $VEGA to `ETHUSDT | market creation | $VEGA` 
     * Transfer 20000 USDC to `ETHUSDT | market creation | USDC` 
-* start trading in the market such that trading volume in USDT is less than 10^6
+* start trading in the market such that traded value for fee purposes in USDT is less than 10^6
 
 #### Expectation
 At the end of the epoch no payout has been made for the market ETHUSDT and the reward account balances should remain unchanged.
@@ -418,7 +418,7 @@ Market has been trading but not yet eligible for proposer bonus.
 * Setup and fund multiple reward account for the market ETHUSDT:
     * Transfer 10000 $VEGA to `ETHUSDT | market creation | $VEGA` 
     * Transfer 20000 USDC to `ETHUSDT | market creation | USDC` 
-* start trading in the market such that trading volume in USDT is less than 10^6
+* start trading in the market such that traded value for fee purposes in USDT is less than 10^6
 * During the epoch 2 let the traded value be greater than 10^6
 
 #### Expectation
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -26,7 +26,10 @@ There will be the following fee metrics:
 1. Sum of LP fees received (in a given market).
 
 There will be the following market creation metrics:
-1. Market total [trade value for fee purposes](0029-FEES-fees.md) since market creation multiplied by either `1` if the market creation reward for this market has never been paid or by `0` if the reward has already been paid.
+1. Where `cumulative volume` is defined as market cumulative total [trade value for fee purposes](0029-FEES-fees.md) since market creation:
+   - **IF** `cumulative volume < rewards.marketCreationQuantumMultiple * quantum` (where [quantum is described here](0040-ASSF-asset_framework.md)) **THEN** `market creation metric := 0`
+   - **ELSE IF** any non-zero market creation reward has previously been paid out for this {reward account, market} combination **THEN** `market creation metric := 0`
+   - **ELSE** `market creation metric := cumulative volume`
 
 Reward metrics are not stored in [LNL checkpoints](./0073-LIMN-limited_network_life.md). 
 
@@ -63,11 +66,24 @@ Metrics will be calculated using the [decimal precision of the settlement asset]
 
 ### For market creation metrics
 
-In the *first* epoch during which the market total [trade value for fee purposes](0029-FEES-fees.md) since market creation exceeds `quantum x rewards.marketCreationQuantumMultiple` - if there is balance in the reward account for the corresponding reward type for the market, the proposer of the market gets paid in the reward payout asset. 
+Every epoch the entire reward account for every [market in scope, payout asset] will be distributed pro-rata to the market parties that submitted the market creation governance proposal for any markets that have the market creation metric (described above) `>0`. Similarly to the other metrics market creation may be paid in more than one asset if someone funds a reward account with the corresponding metric type and market in scope with an arbitrary payout asset. 
 
-Similarly to the other metrics market creation may be paid in more than one asset if someone funds a reward account with the corresponding metric type and market in scope with an arbitrary payout asset. 
+That is if we have reward account balance `R`
+```
+[p_1,m_1]
+[p_2,m_2]
+...
+[p_n,m_n]
+```
+then calculate `M:= m_1+m_2+...+m_n` and transfer `R x m_i / M` to party `p_i` at the end of each epoch. 
+If `M = 0` (no markets newly met or exceeded the reward threshold since the last payout) then nothing is paid out of the reward account and the balance rolls into next epoch. 
+
+Market creation metrics are not reset at the end of the epoch, so the cumulative volume for each market continues to accrue across epochs and is always equal to the total trade value for fee purposes since the creation of the market.
+
+Metrics will be calculated using the [decimal precision of the settlement asset](0070-MKTD-market-decimal-places.md).
 
 NB: if a market is being recreated from checkpoint, and trading in it is restarted - the tracking of trading value restarts from 0, regardless of how much trading was done before and if a proposer bonus has been given before the checkpoint was taken. Therefore after the checkpoint reload, the proposer may be eligible to the market proposer bonus again once the market trading value goes above the threshold for the first time.
+
 
 ## Acceptance criteria
 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -107,7 +107,7 @@ That is, if we have:
 - number of markets with a non-zero market creation metric: `n`
 - parties `p_1 … p_n` being the creators of the markets with non-zero market creation metrics, where the creator is defined as the party that submitted the original market proposal that was enacted to create the market
 
-Then transfer `R / M` to each party `p_i` (`i` from `1 to n`) identified above. 
+If `n > 0` then transfer `R / n` to each party `p_i` (`i` from `1 to n`) identified above. 
 If `n = 0` (i.e. no markets newly met or exceeded the reward threshold since the last payout) then nothing is paid out of the reward account and the balance rolls into next epoch. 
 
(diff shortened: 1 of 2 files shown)


davidsiska-vega (dismissed):

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
+The payout for each market having a non-zero market creation metric will be the same, that is, the eligible markets share the reward account balance equally. 
+This means that the total market creation reward received by a party can vary, as their reward may reflect having created multiple eligible markets, each of which earns the same payout.
+
+That is, if we have:
+- reward account balance: `R` of some asset
+- number of markets with a non-zero market creation metric: `n`
+- parties `p_1 … p_n` being the creators of the markets with non-zero market creation metrics, where the creator is defined as the party that submitted the original market proposal that was enacted to create the market
 
-Similarly to the other metrics market creation may be paid in more than one asset if someone funds a reward account with the corresponding metric type and market in scope with an arbitrary payout asset. 
+Then transfer `R / M` to each party `p_i` (`i` from `1 to n`) identified above.
```
davidsiska-vega, line 110: Is the `R / n` ?

```suggestion
If `n>0` then transfer `R / n` to each party `p_i` (`i` from `1 to n`) identified above. 
```

barnabee: Good spot, thanks

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
+
+## Reward process high level
+
+At a high level, rewards work as follows:
+
+- Reward metrics are calculated for each combination of [reward type, party, market].
+ The calculation used for the reward metric is specific to each reward type.
+-  A transfer is made to the reward account(s) for a specific reward type, for one or more markets. This is either a one-off transfer for a single market, or a recurring transfer for one, several, or all markets. (See [transfers](./0057-TRAN-transfers.md).)
+- At the end of the epoch, the entire balance of each reward account is distributed to the parties with a non-zero reward metric pro-rata by their reward metric.
+If the sum of all reward metrics is zero, the balance rolls over to the next epoch.
```
ze97286: not sure what you mean here. suppose we have a recurring transfer of 1000 USDC with a fee based metric - if there were no fees generated, the transfer doesn't happen in this epoch, and will just be re-evaluated in the next epoch, but it's not that the balance is carried over and next epoch 2000 will be transferred if there are fees generated.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
+Therefore, for example, to reward futures markets when they reach a lifetime traded notional over 1 mil USD, then this parameter should be set to around `1000000`. Any decimal value strictly greater than `0` is valid. 
+
+
+## Reward process high level
+
+At a high level, rewards work as follows:
+
+- Reward metrics are calculated for each combination of [reward type, party, market].
+ The calculation used for the reward metric is specific to each reward type.
+-  A transfer is made to the reward account(s) for a specific reward type, for one or more markets. This is either a one-off transfer for a single market, or a recurring transfer for one, several, or all markets. (See [transfers](./0057-TRAN-transfers.md).)
```
ze97286: metrics are only supported for recurring transfers

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
+
+1. Sum of maker fees paid by the party on the market this epoch
+1. Sum of maker fees received by the party on the market this epoch
+1. Sum of LP fees received by the party on the market this epoch
+
+Theese metrics apply only to the sum of fees for the epoch in question.
+That is, the metrics are reset to zero for all parties at the end of the epoch.
+If the reward account balance is 0 at the end of the epoch for a given market, any parties with non-zero metrics will not be rewarded for that epoch and their metric scores do not roll over (they are still zeroed).
+
+Market activity (fee based) reward metrics are not stored in [LNL checkpoints](./0073-LIMN-limited_network_life.md) and are reset after a checkpoint restart.
```
ze97286: that's not how it's implemented - as discussed [here](https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1660744259282999?thread_ts=1660727163.794979&cid=CAHA5EX0F) (although in the context of market bonus) they are actually stored in the checkpoint for all metrics and are continued after the checkpoint from that point. Not sure what the desired behaviour is.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
 
-There will be the following fee metrics:
-1. Sum of maker fees paid (in a given market).
-1. Sum of maker fees received (in a given market).
-1. Sum of LP fees received (in a given market).
+### Market creation reward metrics
 
-There will be the following market creation metrics:
-1. Market total [trade value for fee purposes](0029-FEES-fees.md) since market creation multiplied by either `1` if the market creation reward for this market has never been paid or by `0` if the reward has already been paid.
+There will be a single creation reward metricand reward tyype.
```
ze97286: metricand => metric and
tyype => type

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
+The reward metric for the single *market creator* party is as follows:
 
-It must be possible for any party to run a one off [transfer](0057-TRAN-transfers.md) or create a [recurring transfer](0057-TRAN-transfers.md) to any of these reward accounts. 
-Note that the market settlement asset has nothing to do in particular with the asset used to pay out a reward for the market for any of the relevant trading rewards. 
+- **IF** `cumulative volume < rewards.marketCreationQuantumMultiple * quantum` **THEN** `market creation metric := 0`
+- **ELSE** `market creation metric := 1` (NB: this is 1 as market creation rewards are paid equally to all qualifying creators for *reaching* the volume threshold, not pro-rata based on cumulative volume)
+
+When the `market creation metric` for a party is `>0` and the reward account balance for a specific reward asset is also `>0` (i.e. when a creator is rewarded):
+
+- A flag is added for each of the recorded `funders` of the reward account paired with the market and reward asset.
```
ze97286: I think it needs to be clarified here. My understanding from the discussion on slack is that it should be possible to do the following:

transfer 1: funder: zohar, payout_asset: VEGA, markets_in_scope: [A,B,C]
transfer 2: funder: zohar, payout_asset: VEGA, markets_in_scope: [A,B]
transfer 3: funder: zohar, payout_asset: VEGA, markets_in_scope: []

all of the above will pay out A,B
the first and the third will pay out A,B,C 

Maybe should be clarified that the key for the flag is [funder, payout_asset, market_in_scope]. If a payout has been made for a combo, it shall never happen again, and is recorded in checkpoint as such.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
+
+## Reward accounts
+
+Trading reward accounts are defined by the reward asset (the asset in which the reward is paid out), the market, and the reward type (metric). 
+That is, there can be multiple rewards with the same type paid in different assets for the same market.
+
+Note that the market settlement asset has nothing to do in particular with the asset used to pay out a reward for a market. 
+That is, a participant might recieve rewards in the settlement asset of the market, in VEGA governance tokens, and in any number of other unrelated tokens (perhaps governance of "loyalty"/reward tokens issued by LPs or market creators, or stablecoins like DAI).
+
+Reward accounts can be individually funded by normal single transfers. They can also be funded on an ongoing (recurring) basis in groups, pro-rata by their relative total reward metrics for the given reward type. See [transfers](./0057-TRAN-transfers.md).
```
ze97286: one off currently do not support metrics. the same can be achieved though by a recurring transfer with an end epoch being the following one

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
-then calculate `M:= m_1+m_2+...+m_n` and transfer `R x m_i / M` to party `p_i` at the end of each epoch. 
-If `M = 0` (no-one incurred / received fees for a given reward account)  then nothing is paid out of the reward account and the balance rolls into next epoch. 
-
-Metrics are reset at the end of the epoch, so in the case where there is no reward account or no reward balance at the end of the epoch for the given market in scope, the participants contributing to the relevant metric will not be compensated for their contribution. Their contribution to fees is not being carried over to the next epoch. 
 
-Metrics will be calculated using the [decimal precision of the settlement asset](0070-MKTD-market-decimal-places.md).
+Then calculate `M:= m_1+m_2+...+m_n` and transfer `R x m_i / M` to party `p_i` (for each `p_i`) at the end of the epoch.
 
-### For market creation metrics
+If `M = 0` (no-one incurred or received fees as specified by the metric type for a given *market scope*) then nothing is paid out of the reward account for that *market scope* and the balance rolls into next epoch.
```
ze97286: if M = 0 the reward account would not have been funded because the transfer would not have happened.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -1,73 +1,120 @@
...
 
 ## Reward distribution
 
 All rewards are paid out at the end of any epoch *after* [recurring transfers](0057-TRAN-Transfers.md) have been executed. 
-There are no fractional payouts and no delays.
+The entire reward account balance is paid out every epoch unless the total value of the metric over all parties is zero (there are no fractional payouts). 
+There are no payout delays, rewards are paid out instantly at epoch end.
 
-### For fee based metrics
+Rewards will be distributed pro-rata by the party's reward metric value to all parties that have metric values `>0`. That is, if we have reward account balance `R` and parties `p_1 – p_n` with non-zero metrics on the market in question:
```
ze97286: market proposer bonus is an exception as the total balance is divided equally between all the proposers of the markets that are eligible for it at the given point in time. Unless you're suggesting to do it pro-rata of the value generated? that's also possible.

protocol/0056-REWA-rewards_overview.md
barnabee: Is is possible to do a "normal" single trasnfer to a single market's reward account?

protocol/0056-REWA-rewards_overview.md
barnabee: My assumption was that a reward account could also be funded by a non-recurring transfer and that it was therefore possible to haqve a reward account balance even with no party having non-zero metrics. If not then I can remove this.

protocol/0056-REWA-rewards_overview.md
ze97286: by normal you mean say I want to transfer x VEGA to market1 account type LP fees reward? I think we removed this from the API.

protocol/0056-REWA-rewards_overview.md
barnabee: This is discussing the fee based metrics (e.g. cumulative maker fees paid this epoch) - my understanding was that these are not stored but that the market's cumulative trade value for fee purposes and the flags for market creator rewards paid would be in LNL checkpoints.

I think that understanding is in line with the linked slack convo, but apologise if I'm missing something - let me know if so.

protocol/0056-REWA-rewards_overview.md
ze97286: one off to market are currently not supported in the API

barnabee: Ok cool, that makes things simpler

protocol/0056-REWA-rewards_overview.md
barnabee: Cool, will update

protocol/0056-REWA-rewards_overview.md
ze97286: well tbf in the chat we only discussed the market creation aspect, but in the checkpoint we store the whole party contribution to fees in the market, so in case we take a checkpoint close to the end of the epoch, when we restore, they don't lose reward (if there's balance) at the end of the epoch.

barnabee: Ah ok, this came from the original spec which said no reward metrics were stored in checkpoints and I just split it out because we discussed that market creation rewards are. It sounds like the fee metrics are too so will update.

protocol/0056-REWA-rewards_overview.md
barnabee: I tweaked the definition of the reward metric calculation above such that it is always either `0` or `1` (and added an explanation as to why) which has the effect of making the pro-rata distribute it equally but without having to write up two different bits of logic.

As long as if the same party created 2 (or more) markets that become eligible in the same epoch that party will receive 2 (or more) times the payout of a party with just one eligible market then we should be good.

protocol/0056-REWA-rewards_overview.md
barnabee: updated

protocol/0056-REWA-rewards_overview.md
barnabee: So the flag will be [funder, payout_asset, market_scope_set]? 

Just want to clarify as your last sentence used "market_in_scope" rather than "markets_in_scope" but I read the rest of the comment as suggesting there would be a separate flag for every scope.

ze97286: market**s**, as the examples read :)

barnabee: cool thanks

protocol/0056-REWA-rewards_overview.md
barnabee: I've added a clarification here too.

gordsport:
Hey @barnabee 
This has been implemented in:
- https://github.com/vegaprotocol/vega/issues/5937

As per the spec now. It looks like all comments have been addressed so can this be merged or is there further changes to come?
Thanks

protocol/0057-TRAN-transfers.md
```diff
@@ -5,63 +5,101 @@ These transfers are not to be confused with the internal concept of transfers wh
...
 ## Recurring transfers
-A party can also setup recurring transfers which will happen at the end of every epoch. These transfers will happen at the end of the epoch, and before the next one can happen.
-A party is limited to a maximum of 1 running recurring transfer to between two given accounts. 
+
+A party can also setup recurring transfers which will happen at the end of every epoch, before the next epoch begins. 
+These transfers happen at the end of the epoch, but before processing any rewards. 
+Trading or staking rewards to be received for that epoch will not be available to be used by a recurring transfer. 
+Recurring transfers to reward accounts will happen before rewards are paid out. 
+
+//TODO: did we change this?
```
ze97286: they are processed in the order they were received now.

protocol/0057-TRAN-transfers.md
```diff
@@ -5,63 +5,101 @@ These transfers are not to be confused with the internal concept of transfers wh
...
+
+- When transferring to a reward account, it is possible to define the reward metric, the reward metric asset, and a subset of reward markets. If the reward markets are not defined - it is taken as all the markets that settle in the reward metric asset which must be specified when creating the recurring transfer. 
+
+- At the end of the epoch when the transfer is about to be distributed, it first calculates the contribution of each market (either out of all the markets that settle in the reward metric asset or only the ones in scope of the transfer) to the total reward metric and then distributes the transfer to the corresponding accounts of the markets pro-rata. 
+
+Where the reward metric type is "market creation rewards", it is important that no market creator will receive more than one market creation reward paid in the same asset from the same source account (reward funder). 
+Therefore: 
+
+- A list of [market, source account, reward asset] combinations that have already been rewarded is maintained.
```
ze97286: I think this should be a mapping from market -> [[markets_in_scope], source account, reward asset], otherwise I don't understand it. Where the market on the LHS is the market whose proposer has been paid, and the RHS is the combo for which they've been paid for. 

As written or as I understand it at least, in the following example, a, b will only be paid by the earlier of [t1, t2] which is not what I thought we wanted nor what is implemented. 
 
t1: funder = zohar, markets_in_scope = [a,b,c], payout_asset = vega
t2:funder = zohar, markets_in_scope = [a,b], payout_asset = vega

barnabee: Oops I missed that update, fixed

barnabee:
@gordsport once @ze97286 and @davidsiska-vega agree, it is ready to merge

ze97286 (approved):

PR #1198, opened 2022-08-23
Update 0069-VCBS-validators_chosen_by_stake.md

Vegaklaus:
Fixed AC that is untestable

### DIFF
--- a/protocol/0069-VCBS-validators_chosen_by_stake.md
+++ b/protocol/0069-VCBS-validators_chosen_by_stake.md
@@ -328,7 +328,7 @@ See [limited network life spec](./0073-LIMN-limited_network_life.md).
 9. Swap due to better score (<a name="0069-VCBS-042" href="#0069-VCBS-042">0069-VCBS-042</a>):
   * Setup a network with 5 Tendermint validators and self-delegate 1000 tokens to each of them. 
   * Announce a new node at the beginning of the epoch, self-delegate to them a total that is 10000 tokens. 
-  * At the beginning of the next epoch the new validator should have ranking score *equal* to all of the Tendermint validators so it doesn’t get promoted. 
+  * At the beginning of the next epoch the new validator should have ranking score *equal or lower* to all of the Tendermint validators so it doesn’t get promoted. 
   * In the middle of the epoch, shut node 1 down. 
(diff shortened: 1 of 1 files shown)


davidsiska-vega (approved):

protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -328,7 +328,7 @@ See [limited network life spec](./0073-LIMN-limited_network_life.md).
 9. Swap due to better score (<a name="0069-VCBS-042" href="#0069-VCBS-042">0069-VCBS-042</a>):
   * Setup a network with 5 Tendermint validators and self-delegate 1000 tokens to each of them. 
   * Announce a new node at the beginning of the epoch, self-delegate to them a total that is 10000 tokens. 
-  * At the beginning of the next epoch the new validator should have ranking score *equal* to all of the Tendermint validators so it doesn’t get promoted. 
+  * At the beginning of the next epoch the new validator should have ranking score *equal or lower* to all of the Tendermint validators so it doesn’t get promoted.
```
wwestgarth, line 331: This makes no sense. It's the tendermint validators that never quite have a perf score of 1, with the new node having an exact perf score of 0.1, so multiplied by their stake score, in this step the new node will almost always have a higher ranking score than one of the tendermint nodes and so will get promoted.

That's exactly why it's untestable and this change doesn't help that.

Vegaklaus:
Don't they get a incumbent bonus factor in the ranking score that should
compensate for that ?

wwestgarth:
> Don't they get a incumbent bonus factor in the ranking score that should compensate for that ?
> 

My take on the AC is that we are intentionally trying to flex "when all things are equal, no swap happens, node goes down swap then happens" which is why we delegate x10 to the new node, and check for equality when its perf-score hits 0.1. So I assumed the incumbent-factor would be 0 here. 

If thats not the case thats fine, but we would then we need to clarify that:
- `multipleOfTendermintValidators` being 0 (otherwise it'll become an ersatz and the incumbent factor will apply)
- the incumbent factor is the reason for the promotion not to happen in this step
- what value to set the incumbent factor
- the delegations on the new node to be such that if the incumbent factor _was not applied correctly_ that the new node will definitely be promoted (because that seems to now be what we are testing now)

Otherwise there is very little difference between this AC and AC 0069-VCBS-040

PR #1206, opened 2022-08-28
fix: opening auction volume shouldn't set the whole 0th period volume

davidsiska-vega:
Previously we were using the volume of the opening auction to set the volume used in the 0th period. 

But that makes no sense because then in a situation when `market.value.windowLength` is very short the growth will probably be negative while in situations where `market.value.windowLength` is very long the growth will likely be huge resulting in huge virtual stakes for LPs who have been present during the opening auction.


Jiajia-Cui (approved):
LGTM

PR #1207, opened 2022-08-28
[SPEC ONLY] Clarify unregistering data sources

barnabee:
Clarfiy that once a single use data source is used on a market it is "unregistered" immediately and we don't wait for the market to be removed entirely from core.

### DIFF
--- a/protocol/0016-PFUT-product_builtin_future.md
+++ b/protocol/0016-PFUT-product_builtin_future.md
@@ -36,6 +36,7 @@ cash_settled_future.value(quote) {
 
 ```javascript
 cash_settled_future.trading_termination_trigger(event) {
+	unregister_data_source_and_stop_listening(cash_settled_future.trading_termination_trigger)
 	setMarketStatus(TRADING_TERMINATED)
 }
 ```
@@ -45,6 +46,7 @@ cash_settled_future.trading_termination_trigger(event) {
 
 ```javascript
 cash_settled_future.settlement_data(event) {
+	unregister_data_source_and_stop_listening(cash_settled_future.settlement_data)
 
 	// If settlement data was received prior to trading termination use the last value received, otherwise use the first value received after trading is terminated 
 	while market.status != TRADING_TERMINATED {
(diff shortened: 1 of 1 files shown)


edd (dismissed):
👍 Good ACs

AnExsomnis (dismissed):

davidsiska-vega (changes requested):
It seems to me that 0016-PFUT-015 contradicts 0016-PFUT-008.
It seems tome that 0016-PFUT-015 and 0016-PFUT-016 seem to be duplicates of each other.

protocol/0016-PFUT-product_builtin_future.md
```diff
@@ -45,16 +79,15 @@ cash_settled_future.trading_termination_trigger(event) {
 
 ```javascript
 cash_settled_future.settlement_data(event) {
-
-	// If settlement data was received prior to trading termination use the last value received, otherwise use the first value received after trading is terminated 
-	while market.status != TRADING_TERMINATED {
-		waitForMarketStatus(TRADING_TERMINATED)
+	
+	// Store settlement data (replaces data from previous event so always use latest event up to the point we settle)
+	cash_settled_future.settlement_data = event.data
```
davidsiska-vega, line 84: I wonder whether we always want to do that. Imagine data that are of the form `[timestamp, price]`. Now imagine I receive first `[2022-09-01-23:59, price1]` and later I receive something with earlier timestamp `[2022-09-01-12:00, price2]`. In this case I doubt we'd want to replace the settlement data with the newer event.

protocol/0016-PFUT-product_builtin_future.md
```diff
@@ -74,3 +107,7 @@ cash_settled_future.settlement_data(event) {
...
+1. After trading termination has been triggered the trading terminated data source is no longer active (assuming it is not used anywhere else) and data from that source is no longer processed. (<a name="0016-PFUT-013" href="#0016-PFUT-013">0016-PFUT-013</a>)
+1. After settlement data has been recieved and the market has settled, the settlement data source is no longer active (assuming it is not used anywhere else) and data from that source is no longer processed. (<a name="0016-PFUT-014" href="#0016-PFUT-014">0016-PFUT-014</a>)
+1. After settlement data has been recieved and the market has not settled yet because trading termination has not been triggered, the settlement data source is remains active (assuming it is not used anywhere else) and data from that source is still processed (because each updated data event recieved is stored and eventually used for settlement up until trading terminated it triggered). (<a name="0016-PFUT-015" href="#0016-PFUT-015">0016-PFUT-015</a>)
```
davidsiska-vega, line 112:
```suggestion
1. After settlement data has been recieved and the market has not settled yet because trading termination has not been triggered, the settlement data source is remains active (regardless of whether it's used elsewhere) and data from that source is still processed (because each updated data event received is stored and eventually used for settlement up until trading terminated it triggered). (<a name="0016-PFUT-015" href="#0016-PFUT-015">0016-PFUT-015</a>)
```

davidsiska-vega (changes requested):
Left a comment and a suggestion.

protocol/0016-PFUT-product_builtin_future.md
barnabee: That was the thinking behind the original spec attempts — normally settlement, valuation, etc. happen with the first price printed after or exactly for the appropriate timestamp. I've never seen a system where if you want to expire something and you have two prices in a feed that are after the fixing time you'd use the later one. If you want more recent data, usually you would specify that (with a filter or by having a data source that doesn't send data that is too old at all). But here we are…

There have been three iterations of the spec for this:

**1st attempt:** take the first data that matches and is received _after_ you are in a state where you are ready to settle the market. 
**2nd attempt:** take the first data that matches and is received after the enactment of the market _whether or not_ it is ready to settle yet
**3rd attempt (now):** take the most recent data received after the enactment of the market (if trading is terminated before any data is received, this will also be the first data received)

Note that my preference and the original design was for (1). Though it was vulnerable to potentially dropping the real preferred data due to a late trading termination trigger, and perhaps even never settling at all:

- dropping the "correct" data but settling with later data I did see as a problem, but one could be avoided by using a no-later-than timestamp filter (so that it'd never settle in that case); but
- dropping the correct data and never settling I saw as a _feature_, not a bug, because if trading termination was delayed until after the price was known then it made sense to me that you might want to agree a settlement price via a governance update rather than settle to a price that some people had already seen with potentially no warning.

We then went via iteration (2) to always use the first data, which, like (1) avoids the problem you mention in your comment but with trade-off of less likely to need governance but more likely to suffer whatever issues out of order termination vs settlement might cause. I am not sure why this change was made but it may simply have been to try to align with what we thought had been built and because it seeemed "good enough".

Now we have (3), which I'm pretty sure boils down to discovering it doesn't work like (2) either and agreeing to change the spec to match implementation. I can't remember if anyone argued it was actually preferable to (1) or (2) but I think there might have been some discussion in that direction…

_**Anyhow:**_

1. I doubt it's a problem for trading enabled, though (3) is definitely my least favourite of the iterations.
2. We could probably change it fairly simply though. 
3. We could also add a product param to the cash settled future at some point to specify the behaviour (but see next point - that is probably not be the best way to deal with it)
4. With the time triggered data source modifier/combinastor (which was not yet built), it would be possible to trivially emulate iteration (2) given (1), and with additional data source features (or WASM or similar), you could also create (3) entirely within the data source definition. This is almost certainly preferable to building all this logic into each product, and also speaks to why iteration (1), which is the simplest (either you consumer the data source now and settle or you throw it away and store no state), is also probably the better design.

So I don't know if we want to change anything now. But I do agree what we have isn't particularly good, doesn't compose well with other features, adds complexity, etc. (though it clearly _is_ workable).

davidsiska-vega: I think we want to go back to (1). I wouldn't require it for "enable trading" but we may want to do it anyway... (because writing one set of tests now and rewriting is a pain. 

My reasoning: it's better to miss a settlement data transaction (this can be fixed be governance) than a situation where someone can try to submit an "old coinbase price" on a market without a filter just before trading terminated.

protocol/0016-PFUT-product_builtin_future.md
barnabee: For this test I think it needs to no be used elsewhere rather than left up to the implementer whether it is or not. Or perhaps there should even be 2 ACs?

davidsiska-vega: Maybe two ACs.

PR #1209, opened 2022-08-31
fix: correct i.e. remove per asset delay

davidsiska-vega:


edd (approved):
Lazy reviewing on my part for the previous PR. 👍

PR #1214, opened 2022-09-01
fix: clean up auction ACs

davidsiska-vega:


protocol/0026-AUCT-auctions.md
```diff
@@ -166,9 +166,8 @@ message Market {
 }
 ```
 # Acceptance Criteria
-- As a user, I can configure a market through the market configuration to use auction mode (<a name="0026-AUCT-001" href="#0026-AUCT-001">0026-AUCT-001</a>)
-  - I can define an opening auction for a continuous trading market, and the duration of the call period. (<a name="0026-AUCT-002" href="#0026-AUCT-002">0026-AUCT-002</a>)
-  - The duration of the auction period at market creation cannot be below the minimum auction period defined within the network (<a name="0026-AUCT-003" href="#0026-AUCT-003">0026-AUCT-003</a>)
+- A market with default trading mode "continuous trading" will start with an opening auction. The opening auction will run from the close of voting on the market proposal (assumed to pass sucessfully) until the enactment time assuming there are orders crossing on the book, [liquidity is supplied](./0038-OLIQ-liquidity_provision_order_type.md) and after the auction uncrossing we will have best bid and best ask so that [liquidity can be deployed](./0038-OLIQ-liquidity_provision_order_type.md). (<a name="0026-AUCT-002" href="#0026-AUCT-002">0026-AUCT-002</a>)
```
edd: I think this could be split up in to multiple criteria, but otherwise it's good.

edd (dismissed):
Approved, but with one suggested change

protocol/0026-AUCT-auctions.md
```diff
@@ -177,4 +176,4 @@ message Market {
 - As an API user, I can identify: (<a name="0026-AUCT-008" href="#0026-AUCT-008">0026-AUCT-008</a>)
   - If a market is temporarily in an auction period
   - Why it is in that period (e.g. Auction at open, liquidity sourcing)
-  - When the auction mode ends
+  - When the auction will next attempt to uncross and end
```
edd: Note for pedants: this can be *right this second* in the case of an extending liquidity auction, in which case it's probably in the past, which looks odd.

edd (approved):

PR #1219, opened 2022-09-02
docs: Adding/updating ACs for rewards spec

### LINKED ISSUE #1210: Market creation rewards need more ACs
See https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1660727163794979 (interal Vega Slack, sorry)

TomMcL:
Adding/updating ACs for the newly changed market creations rewards

Closes #1210

### DIFF
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -500,10 +500,10 @@ A market should be able to be rewarded multiple times if several reward pools ar
     * Transfer 10000 $VEGA to `ETHUSDT | market creation | $VEGA` 
 * start trading in the market such that traded value for fee purposes in USDT is less than 10^6
 * During the epoch 2 let the traded value be greater than 10^6
-* During epoch 3, transfer 20000 USDC to `ETHUSDT | market creation | USDC` 
+* During epoch 3, transfer 20000 USDC to `all | market creation | USDC` 
 
 #### Expectation
-At the end of epoch 2 the VEGA rewards should be distributed. Then, at the end of epoch 3, the funded USDC rewards should be distributed
+At the end of epoch 2 the VEGA rewards should be distributed. Then, at the end of epoch 3, the funded USDC rewards should be distributed again to the market creator
 
 ### Distributing market creation rewards - Same asset multiple party rewards (<a name="0056-REWA-044" href="#0056-REWA-044">0056-REWA-044</a>)
 #### Rationale 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -505,7 +505,7 @@ Then, at the end of epoch 3, the 20000 USDC rewards should be distributed again 
 The reward pool balance should be 0.
 
 
-### Distributing market creation rewards - multiple asset rewards simultaneous payout (<a name="0056-REWA-043" href="#0056-REWA-043">0056-REWA-043</a>)
+### Distributing market creation rewards - multiple asset rewards simultaneous payout (<a name="0056-REWA-045" href="#0056-REWA-045">0056-REWA-045</a>)
 #### Rationale 
 A market should be able to be rewarded multiple times if several reward pools are created with different payout assets.
 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -34,12 +34,12 @@ At a high level, rewards work as follows:
 
 ## Reward metrics
 
-Reward metrics are scoped by [reward type, market, party] (this triplet can be thought of as a primary key for reward metrics).
-Therefore a party may be in scope for the same reward type multiple times but no more than once per market.
+Fee-based reward metrics are scoped by [`reward type`, `market`, `party`] (this triplet can be thought of as a primary key for fee-based reward metrics).
+Therefore a party may be in scope for the same reward type multiple times but no more than once per market per epoch.
 Metrics will be calculated at the end of every epoch, for every eligible party, in each market for each reward type.
 Metrics only need to be calculated where the [market, reward type] reward account has a non-zero balance of at least one asset. 
 
-Reward metrics will be calculated once for each party/market combination in the reward metric asseet which is the [settlement asset](0070-MKTD-market-decimal-places.md) of the market. 
+Reward metrics will be calculated once for each party/market combination in the reward metric asset which is the [settlement asset](0070-MKTD-market-decimal-places.md) of the market. 
 This is the original precision for the metric source data. 
 
 
@@ -125,7 +125,7 @@ Then calculate `M := m_1 + m_2 + … + m_n` and transfer `R ✖️ m_i / M` to p
 If `M=0` (no-one incurred or received fees as specified by the metric type for the given market) then no transfer will have been made to the reward account and therefore there are no rewards to pay out.
 The transfer will be retried the next epoch if it is still active. 
 
-Reward payouts will be calculated using the decimal precision of the reward payout asset. If this allows less precision than the rewward metric asset (the market's settlement asset) then the ratios between reward payouts may not match exactly the ratio between the reward maetrics for any two parties. All funds will always be paid out.
+Reward payouts will be calculated using the decimal precision of the reward payout asset. If this allows less precision than the reward metric asset (the market's settlement asset) then the ratios between reward payouts may not match exactly the ratio between the reward maetrics for any two parties. All funds will always be paid out.
 
 
 ## Acceptance criteria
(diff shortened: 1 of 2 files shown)


davidsiska-vega (commented):
On line 454 (ac 041) the Rationale is copied from ac 040 and is incorrect / confusing.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -486,10 +486,58 @@ Market goes above the threshold in trading value in an epoch before the reward a
...
+* start trading in the market such that traded value for fee purposes in USDT is less than 10^6
+* During the epoch 2 let the traded value be greater than 10^6
+* During epoch 3, transfer 20000 USDC to `all | market creation | USDC` 
+
+#### Expectation
+At the end of epoch 2 the VEGA rewards should be distributed. Then, at the end of epoch 3, the funded USDC rewards should be distributed again to the market creator
+
+### Distributing market creation rewards - Same asset multiple party rewards (<a name="0056-REWA-044" href="#0056-REWA-044">0056-REWA-044</a>)
+#### Rationale 
+A market reward pool funded with the same asset by different parties should pay out to eligible markets twice
```
```suggestion
A market reward pool funded with the same asset by different parties should pay out to eligible markets as many times as there are parties, assuming threshold is reached.
```

davidsiska-vega (dismissed):
LGTM, let's check with Zohar too.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -466,15 +466,15 @@ Market has been trading but not yet eligible for proposer bonus.
 #### Expectation
 At the end of the epoch 2 the proposer of the market ETHUSDT is paid 10000 `$VEGA` and 20000 `USDC`
 
-Then during epoch 3 make the following transfers:
+Then during epoch 3 set up the following transfers:
```
ze97286: the transfer setup should be part of the setup. here the expectation should say - that in epoch 3 the balance of the reward account for $VEGA/market_creation/ETHUSDT and USDC/market_creation/ETHUSDT is 0.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -486,10 +485,58 @@ Market goes above the threshold in trading value in an epoch before the reward a
...
+At the end of epoch 2 the VEGA rewards should be distributed. Then, at the end of epoch 3, the VEGA rewards should again be distributed
+
+
+### Distributing market creation rewards - Multiple markets eligible, one already paid (<a name="0056-REWA-046" href="#0056-REWA-046">0056-REWA-046</a>)
+#### Rationale 
+A market reward pool funded with the same asset by the same party with different market scopes should pay to all markets even if already paid
+#### Setup
+* Setup a market ETHUSDT settling in USDT.
+* Setup a market BTCDAI settling in DAI with a different party.
+* The value of `marketCreationQuantumMultiple` is `10^6` and `quantum` for `USDT` is `1`.
```
ze97286: and quantum for DAI?

ze97286: also, would be nice to have a similar test but where we also qualify the metric asset on the transfer. In this case it has to be left empty because you have two different settlement assets but I think we also want one where it's defined to be lets say USDT and then the second transfer pays only ETHUSDT proposer rather than both.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -486,10 +485,58 @@ Market goes above the threshold in trading value in an epoch before the reward a
...
+* The value of `marketCreationQuantumMultiple` is `10^6` and `quantum` for `USDT` is `1`. 
+* Setup and fund multiple reward account for the market ETHUSDT:
+    * Transfer 10000 $VEGA to `ETHUSDT | market creation | $VEGA` 
+* start trading in the market such that traded value for fee purposes in USDT is less than 10^6
+* During the epoch 2 let the traded value be greater than 10^6
+* During epoch 3, setup and fund multiple reward account for the market ETHUSDT with a different party:
+    * Transfer 10000 $VEGA to `ETHUSDT | market creation | $VEGA` 
+
+#### Expectation
+At the end of epoch 2 the VEGA rewards should be distributed. Then, at the end of epoch 3, the VEGA rewards should again be distributed
```
ze97286: I think it's easier here to specify it in terms of balances:
end of epoch 2:
reward account balance = 0
proposer of ETHUSDT general balance in $VEGA = 10000k + extra from the infra fees of the transfers, not sure what the % is so can't tell an exact number here. 
end of epoch 3:
reward account balance = 0
proposer of ETHUSDT general balance in $VEGA = 20000k + 2 *extra from the infra fees of the transfers

end of epoch 4: (to show no more transfers take place)
reward account balance = 0
proposer of ETHUSDT general balance in $VEGA = 20000k + 2 *extra from the infra fees of the transfers

protocol/0056-REWA-rewards_overview.md
```diff
@@ -486,10 +485,58 @@ Market goes above the threshold in trading value in an epoch before the reward a
...
+A market should be able to be rewarded multiple times if several reward pools are created with different payout assets.
 
+#### Setup
+* Setup a market ETHUSDT settling in USDT.
+* The value of `marketCreationQuantumMultiple` is `10^6` and `quantum` for `USDT` is `1`. 
+* Setup and fund multiple reward account for the market ETHUSDT:
+    * Transfer 10000 $VEGA to `ETHUSDT | market creation | $VEGA` 
+* start trading in the market such that traded value for fee purposes in USDT is less than 10^6
+* During the epoch 2 let the traded value be greater than 10^6
+* During epoch 3, transfer 20000 USDC to `all | market creation | USDC`
```
ze97286, line 500: I think we need two cases here, one as you have it where the payout happens sequentially (in separate epochs), and one where the payouts takes place at the same time (i.e. setup both transfers up front, then when threshold is reached payout is made in both assets at the same time).

protocol/0056-REWA-rewards_overview.md
```diff
@@ -462,19 +462,18 @@ Market has been trading but not yet eligible for proposer bonus.
     * Transfer 20000 USDC to `ETHUSDT | market creation | USDC` 
 * start trading in the market such that traded value for fee purposes in USDT is less than 10^6
 * During the epoch 2 let the traded value be greater than 10^6
+* Then during epoch 3 set up the following transfers with the same party as the first set:
+    * Transfer 10000 $VEGA to `ETHUSDT | market creation | $VEGA` 
+    * Transfer 20000 USDC to `ETHUSDT | market creation | USDC`
```
ze97286: all of these transfers must be recurring so 1) this is not needed and 2) this is not possible because you already have an active transfer from line 462. 

It is probably worth clarifying that all of the transfers in these ACs must be recurring, using the metric market_proposer and with metric asset being either empty or specified to qualify the scope of markets to only a single asset or to all.

protocol/0056-REWA-rewards_overview.md
TomMcL: I'm not sure I'm entirely clear on what the second test would be. Would we not want the two markets having different settlement assets if the metric asset on the transfer is going to make a difference?

TomMcL: Or actually perhaps metric asset isn't part of the 'key' and doesn't have to be specified so blank and `USDT` shouldn't be classed as a 'different' payment?

TomMcL:
@ze97286 Addressed comments. Added a new test but not sure whether the interpretation there is correct so grateful if you could check that one 👍

protocol/0056-REWA-rewards_overview.md
```diff
@@ -466,30 +466,143 @@ Market has been trading but not yet eligible for proposer bonus.
...
+
+#### Expectation
+At the end of epoch 2 the full 10000 VEGA rewards should be distributed to only the ETHUSDT creator. At the end of epoch 3 the full 10000 VEGA rewards should be distributed to the BTCUSDT creator.
 
+At the end of epoch 2, 10000 VEGA rewards should be distributed to only the ETHUSDT creator.
+    The general account balance of the ETHUSDT creator should be 10000.
+    The general account balance of the BTCUSDT creator should be 0.
+    The reward pool balance should be 0.
 
+At the end of epoch 3, 10000 VEGA should be distributed split between the BTCUSDT creator and the ETHUSDT creator.
```
ze97286, line 620: I think the balances below are incorrect. I'd expect the balance of the proposer of ETHUSDT to be 15k and of the proposer of BTCUSDT to be 5k. also, when specifying the markets it's worth mentioning that you want them to be proposed by different parties or the same. (EDIT: you are saying that, it's just the wording: I would replace "with a different party." with "with a different proposing party" or "with a different proposer")

protocol/0056-REWA-rewards_overview.md
```diff
@@ -466,30 +466,143 @@ Market has been trading but not yet eligible for proposer bonus.
...
+    * Transfer 10000 $VEGA to `all | market creation | $VEGA` 
+
+#### Expectation
+At the end of epoch 2, 10000 VEGA rewards should be distributed to only the ETHUSDT creator.
+    The general account balance of the ETHUSDT creator should be 10000.
+    The general account balance of the BTCDAI creator should be 0.
+    The reward pool balance should be 0.
+
+At the end of epoch 3, 10000 VEGA should be distributed split between the BTCDAI creator and the ETHUSDT creator.
+    The general account balance of the ETHUSDT creator should be 15000.
```
ze97286, line 574: I don't think this is correct. As you're setting the metric asset in this example, it means that it is restricted to markets settling in USDT, which means again only ETHUSDT gets the whole lot, meaning they would have a balance of 20k and poor BTCDAI would have a balance of 0.

protocol/0056-REWA-rewards_overview.md
TomMcL: Thanks for catching, I even said as much in the line above so not sure where the 10000 each came from! Updating

protocol/0056-REWA-rewards_overview.md
TomMcL: Fixed up + added a new test

ze97286 (approved):

PR #1248, opened 2022-09-15
clarification on virtual stake used in "entry valuation" in spec 0042

Jiajia-Cui:


protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -120,6 +120,8 @@ So `Delta S > 0` (and so `S+Delta S > 0`) in what follows.
 ```
 (entry valuation) = sum over j from 1 to N of (LP j virtual stake)
 ```
+Note, the `virtual stake` used in the calcuation of `entry valuation` is after the change of the LP commitmnet is applied.
+
```
davidsiska-vega:
```suggestion
This in particular means that if this is the first LP commitment on the market then the `(entry valuation) = Delta S`.
```

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -120,21 +120,29 @@ So `Delta S > 0` (and so `S+Delta S > 0`) in what follows.
...
+Example 2: 
+A new LP1 has `0` stake and they wish to add `Delta S = 8000` and a new LP2 has `0` stake and they wish to add `Delta S = 2000`. Currently the sum of all virtual stakes is `10000` after the LP commmitments added. The average entry valuations are:
+```
+(average entry valuation LP1) <- 0 + 8000 x 8000 / (0 + 8000) = 8000
+(average entry valuation LP2) <- 0 + (8000 + 2000) x 2000 / (0 + 2000) = 10000
+```
+Example 3:
 An existing LP has `average entry valuation 1000` and `S=100`. Currently the sum of all virtual stakes is `2000`. They wish to add `10` to their stake. 
 ```
 (average entry valuation) <- 1000 x 100 / (100 + 10) + 2000 x 10 / (100 + 10) = 1090.9....
```
EVODelavega, line 143: So essentially 3 LP's could commit the same amount (say, during opening auction), depending on which one increases their stake by 10, the AEV would be different:

```
LP1 -> commits 900 (AEV == 900)
LP2 -> commits 100 (AEV == 1000)
LP3 -> commits 100 (AEV == 1100)
LP4 -> commits 100 (AEV == 1200)
LP5 -> commits 800 (AEV == 2000)
```

Then LP's 2, 3, and 4 add 10:

```
LP2 -> AEV = 1000 * 100/ 110 + 2000 * 10/110 == 1090.909...
LP3 -> AEV = 1100 * 100/110 + 2000 * 10/110 = 1181.8181...
LP4 -> AEV = 1200 * 100/110 + 2000 * 10/110 = 1272.7272...
```

And if 2 or more change their stake, depending on the order in which they do so, the AEV would be different still:

```
LP2 -> AEV = 1000 * 100/ 110 + 2000 * 10/110 == 1090.909...
LP3 -> AEV = 1100 * 100/110 + 2010 * 10/110 = 1182.7272...
LP4 -> AEV = 1200 * 100/110 + 2020 * 10/110 = 1274.5454...
```

But then if we swap LP3 and LP4 (so LP4 increases their stake first)
```
LP2 -> AEV = 1000 * 100/ 110 + 2000 * 10/110 == 1090.909...
LP4 -> AEV = 1200 * 100/110 + 2010 * 10/110 = 1273.6363...
LP4 -> AEV = 1100 * 100/110 + 2020 * 10/110 = 1183.6363...
```

That's what would happen, based on example 2

EVODelavega: @davidsiska-vega @Jiajia-Cui

Jiajia-Cui: yes.

Do you mean, if LP3 and LP4 made the same commitment (changes) then they should end up with the same AEV (even when they made the change in different order)?

EVODelavega (approved):

PR #1251, opened 2022-09-16
fix: add AC making it clear that quantum must be greater than .

davidsiska-vega:


edd (dismissed):
This doesn't clarify that it applies to both built in and ERC20 (etc) assets. But I don't think it needs to, as long as it's covered in tests for both.

edd (approved):

PR #1262, opened 2022-09-23
User interface: check market status when submitting order `SORD`

campbellssource:
Adds a few more ACs for dealing with markets in non-tradable modes.
removes codes for non-musts
renames codes


user-interface/6001-SORD-submit_orders.md
```diff
@@ -1,11 +1,43 @@
...
+## Before seeing a "deal ticket"
+
+When looking at a market, I...
+
+- **must** see/select the [Market](./7001-DATA-data_display.md#market) I am submitting the order for (<a name="6001-SORD-001" href="#6001-SORD-001">6001-SORD-001</a>)
+- **must** see the current `status` of the market (<a name="6001-SORD-061" href="#6001-SORD-061">6001-SORD-061</a>)
+  
+  if the market is in a state of `rejected`, `canceled` or `closed`:
+
+  - **should** not see a deal ticket, and see that the market is not accepting orders and never will be (<a name="6001-SORD-062" href="#6001-SORD-062">6001-SORD-062</a>)
```
mattrussell36: not sure about completely 'not seeing' a deal ticket (for this AC and the rest), seems a little over bearing, I could see valid designs where the deal ticket has an overlay to prevent interaction, or where order prevention happens on submission validation

campbellssource: Fair enough
```suggestion
  - **should** see that the market is not accepting orders and never will be (<a name="6001-SORD-062" href="#6001-SORD-062">6001-SORD-062</a>)
```

radoslaw-szpiech (approved):

PR #1285, opened 2022-10-03
chore: remove old datanode spec and update ACs

gordsport:
Remove the old 0004-NP-APIN specs

Move across the ACs not already obviously covered.

@davidsiska-vega @vega-paul @ettec  - some of the ported over ACs look like they may be covered already but i'm not sure if we should keep in case there are nuanced differences.


protocol/0076-DANO-data-node.md
```diff
@@ -168,3 +167,13 @@ It must be possible to add to the data node APIs that return the result of calcu
 
 ### Schema compatibility:
 1. It is possible to identify if schema versions are NOT backwards compatible. Pull existing network snapshots start network, run a protocol upgrade to at later version and ensure both the core state and data-node data is correct (<a name="0076-COSMICELEVATOR-013" href="#0076-COSMICELEVATOR-013">0076-COSMICELEVATOR-013</a>)
+
+### General Acceptance
+* The DataNode must be able to handle brief network outages and disconnects (<a name="0076-DANO-015" href="#0076-DANO-015">0076-DANO-015</a>) 
+* All API functionality currently in the Core node must be available from the DataNode (<a name="0076-DANO-016" href="#0076-DANO-016">0076-DANO-016</a>)
```
davidsiska-vega: I think this can go - that was for the migration when data node was being created.

protocol/0076-DANO-data-node.md
```diff
@@ -168,3 +167,13 @@ It must be possible to add to the data node APIs that return the result of calcu
 
 ### Schema compatibility:
 1. It is possible to identify if schema versions are NOT backwards compatible. Pull existing network snapshots start network, run a protocol upgrade to at later version and ensure both the core state and data-node data is correct (<a name="0076-COSMICELEVATOR-013" href="#0076-COSMICELEVATOR-013">0076-COSMICELEVATOR-013</a>)
+
+### General Acceptance
+* The DataNode must be able to handle brief network outages and disconnects (<a name="0076-DANO-015" href="#0076-DANO-015">0076-DANO-015</a>) 
+* All API functionality currently in the Core node must be available from the DataNode (<a name="0076-DANO-016" href="#0076-DANO-016">0076-DANO-016</a>) 
+* The validator node will only accept requests for event bus subscriptions. All other API requests will be invalid. (<a name="0076-DANO-017" href="#0076-DANO-017">0076-DANO-017</a>)
```
davidsiska-vega: Thinking about this: @jgsbennett - to what extent is this a well testable AC? If it's not let's get rid of it.

protocol/0076-DANO-data-node.md
```diff
@@ -168,3 +167,13 @@ It must be possible to add to the data node APIs that return the result of calcu
 
 ### Schema compatibility:
 1. It is possible to identify if schema versions are NOT backwards compatible. Pull existing network snapshots start network, run a protocol upgrade to at later version and ensure both the core state and data-node data is correct (<a name="0076-COSMICELEVATOR-013" href="#0076-COSMICELEVATOR-013">0076-COSMICELEVATOR-013</a>)
+
+### General Acceptance
+* The DataNode must be able to handle brief network outages and disconnects (<a name="0076-DANO-015" href="#0076-DANO-015">0076-DANO-015</a>) 
+* All API functionality currently in the Core node must be available from the DataNode (<a name="0076-DANO-016" href="#0076-DANO-016">0076-DANO-016</a>) 
+* The validator node will only accept requests for event bus subscriptions. All other API requests will be invalid. (<a name="0076-DANO-017" href="#0076-DANO-017">0076-DANO-017</a>)  
+* The event bus stream is only available from the stream node and not the validator or DataNode (<a name="0076-DANO-018" href="#0076-DANO-018">0076-DANO-018</a>)
```
davidsiska-vega: I think this is incorrect, validator node can also produce the event stream.

gordsport: @ze97286 - can you also confirm. Looks to me that this needs re-wording/removing

protocol/0076-DANO-data-node.md
gordsport: removed

protocol/0076-DANO-data-node.md
jgsbennett: I guess we could try sending data node requests to core... there's a question as to how valuable this test is. Are all the endpoints defined in a yaml somewhere? Feels like it's possibly better answered by review than writing a system test. This effectively says "Check that we don't duplicate data node APIs on core..." which isn't the kind of thing I'd expect we could do by accident.

jgsbennett: From memory though, core _does_ support some non-event-bus-subscription APIs, which are different to the data node ones. Is that valid, or is this spec saying that that should not be the case?

davidsiska-vega: Let's just get rid of the AC.

gordsport:
@peterbarrow - please can can you review these copied ACs and specifically the term 'streaming node'

peterbarrow:
I've updated the spec as best as I can, but we are waiting for some more information from the data-node team around how it works with network disconnects. @pscott31 is working on this when he gets time.

PR #1309, opened 2022-10-10
fix: Amend acceptance criteria for reusing proof of work details

MuthuVega:
Amend AC's for reusing proof of work details.
Transactions are now rejected rather than the parties being banned.


witgaw (approved):
Approving based on the linked Slack discussion.

PR #1319, opened 2022-10-12
Added more ACs for reasons

### LINKED ISSUE #1190: 0024-OSTA-order_status Orders to have field for the `Reason` an order has its status
[Order Status Spec](https://github.com/vegaprotocol/specs/blob/master/protocol/0024-OSTA-order_status.md) does not currently mention anything that gives detail on why an order has its status.
But the implementation does include a `rejectedReason` (e.g. Not enough collateral)

This ticket was recently raise to include a reason orders were `Stopped`: https://github.com/vegaprotocol/vega/issues/5915 (e.g. "insufficient balance to maintain open orders", or "FOK order but insufficient liquidity on book to fill")

This ticket is to include the reason as a requirement in the API and write ACs so that the different reasons can be tested.

peterbarrow:
Added ACs for every possible reason value that can be returned from inside the core.
This does not cover the validation codes coming back from the wallet.

Closes #1190


gordsport (approved):
Approved on the protocol design meeting.

PR #1353, opened 2022-10-23
Modify market observable priority in margin calculations during auctions

witgaw:
Prioritise use of mark price, if unavailable fall back on indicative uncrossing price, if that's unavailable too (for orders only) use vwap of orders.

### DIFF
--- a/protocol/0019-MCAL-margin_calculator.md
+++ b/protocol/0019-MCAL-margin_calculator.md
@@ -139,15 +139,16 @@ where meanings of terms in Step 1 apply except for:
 ## Margin calculation for auctions
 
 We are assuming that:
+
 - `indicative_uncrossing_price` is *not* the mark price, so no mark-to-market transfers happen (update mark-to-market spec)
 - mark price never changes during an auction, so it's the last mark price from before auction,
 - during an auction we never release money from the margin account, however we top-it-up as required,
 - no closeouts during auctions
 
-Use the same calculation as above with the following re-defined: 
+Use the same calculation as above with the following re-defined:
+
 - in `slippage_per_unit` we use `indicative_uncrossing_price` instead of `exit_price`. If there is no `indicative_uncrossing_price` then use `slippage_per_unit = 0`.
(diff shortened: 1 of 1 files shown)


protocol/0019-MCAL-margin_calculator.md
```diff
@@ -139,15 +139,16 @@ where meanings of terms in Step 1 apply except for:
...
 - `indicative_uncrossing_price` is *not* the mark price, so no mark-to-market transfers happen (update mark-to-market spec)
 - mark price never changes during an auction, so it's the last mark price from before auction,
 - during an auction we never release money from the margin account, however we top-it-up as required,
 - no closeouts during auctions
 
-Use the same calculation as above with the following re-defined: 
+Use the same calculation as above with the following re-defined:
+
 - in `slippage_per_unit` we use `indicative_uncrossing_price` instead of `exit_price`. If there is no `indicative_uncrossing_price` then use `slippage_per_unit = 0`.
-- For the open position part of the margin: `market_observable = indicative_uncrossing_price`. If there is no current `indicative_uncrossing_price`, then use the previous value for `market_observable` whatever it was (i.e. the last `indicative_uncrossing_price` or `mark_price`).
```
davidsiska-vega, line 149: We still need to say what we do for the position part, don't we?

witgaw: Oh I thought since we always have mark price if have open positions (we're out of opening auction) it's not longer a special case and hence don't need to mention it, but maybe better to spell it out.

witgaw: Added now.

davidsiska-vega: > we always have mark price if have open positions

True! I was just reading the diff and not the whole file. But maybe better to have it spelled out as you did.

davidsiska-vega (approved):

davidsiska-vega:
~~@gordsport if you're watching this: this is a spec change aligning spec with what we've implemented and tested so no core work should arise.~~

Apparently there will be a ticket for one sub-case.

gordsport:
Thanks @davidsiska-vega please let me know the ticket for the sub-case and it's priority (pre/post Oregon Trail timings).

PR #1356, opened 2022-10-24
fix: update mark price only once per block

### LINKED ISSUE #1354: Mark-to-market only at timestamp change...
... as a amplified spam mitigation, see https://github.com/vegaprotocol/specs/issues/1326.

davidsiska-vega:
Closes https://github.com/vegaprotocol/specs/issues/1354

### DIFF
--- a/protocol/0009-MRKP-mark_price.md
+++ b/protocol/0009-MRKP-mark_price.md
@@ -9,7 +9,7 @@ A *Mark Price* is a concept derived from traditional markets.  It is a calculate
 
 Algorithm 1:
 
-- [ ] Out of a sequence of transactions with the same time-stamp the last transaction transaction that results in one or more trades causes the mark price to change to the value of the last trade and only the last trade. (<a name="0009-MRKP-007" href="#0009-MRKP-007">0009-MRKP-007</a>)
+- [ ] Out of a sequence of transactions with the same time-stamp the last transaction that results in one or more trades causes the mark price to change to the value of the last trade and only the last trade. (<a name="0009-MRKP-007" href="#0009-MRKP-007">0009-MRKP-007</a>)
 - [ ] A transaction that doesn't result in a trade does not cause the mark price to change.  (<a name="0009-MRKP-004" href="#0009-MRKP-004">0009-MRKP-004</a>)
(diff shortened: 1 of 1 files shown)


witgaw:
Could we not have it update once `n` blocks where `n` is either hardcoded or a network parameter?

Slightly more work with the counter etc., but more flexibility for the future.

Could also allow `n == 0` which would preserve current behaviour, again, more work, but perhaps a saving of effort overall as we wouldn't have to update the existing tests.

davidsiska-vega:
> Could we not have it update once `n` blocks where `n` is either hardcoded or a network parameter?
> 
> Slightly more work with the counter etc., but more flexibility for the future.
> 
> Could also allow `n == 0` which would preserve current behaviour, again, more work, but perhaps a saving of effort overall as we wouldn't have to update the existing tests.

I broadly agree; though do we want to tie it to blocks or would we rather have passage of time (e.g. at least 5s elapsed)? I agree to include `0s` which would preserve current behaviour.

witgaw:
> > Could we not have it update once `n` blocks where `n` is either hardcoded or a network parameter?
> > Slightly more work with the counter etc., but more flexibility for the future.
> > Could also allow `n == 0` which would preserve current behaviour, again, more work, but perhaps a saving of effort overall as we wouldn't have to update the existing tests.
> 
> I broadly agree; though do we want to tie it to blocks or would we rather have passage of time (e.g. at least 5s elapsed)? I agree to include `0s` which would preserve current behaviour.

Yeah, that sounds even better.

protocol/0009-MRKP-mark_price.md
```diff
@@ -2,16 +2,20 @@ Feature name: Mark Price
 # Summary
 A *Mark Price* is a concept derived from traditional markets.  It is a calculated value for the 'current market price' on a market.
 
+Introduce a network parameter `network.markPriceUpdateMaximumFrequency` with minimum allowable value of `0s` maximum allowable value of `1h` and a default of `5s`.
```
witgaw, line 5: Isn't it just `network.markPriceUpdateFrequency`? Or are you trying to convey that exact timings are not guaranteed?

witgaw (approved):
LGTM, just a small comment/question:

PR #1361, opened 2022-10-25
Update categories.json

JonRay15:
Needed to remove all the commas to avoid issues in the CSV output. 

Have opted for a painful number of &s ....


edd (approved):
Of course, we could have used semicolons and avo... never mind

PR #1362, opened 2022-10-25
Network wide limits

### LINKED ISSUE #1355: Network wide limits on orders / positions etc.
Write new spec for:

- limit the number of public keys on a market e.g. 10^6 users, so that's across orders *and* positions (netparam)
- limit the number of orders on a book e.g. 10^7 orders (netparams) (check if order will trade immediately - in which case allow)
- limit the number of pegged orders e.g. 10^4 orders (netparam) @peterbarrow to test
- LP shapes have limit on the number offsets
- LP have to commit a minimum amount but maybe we set like at most 10^3 (netparam)
- limit the number of markets on Vega (netparam)
- limit the number of assets on Vega (netparam)

witgaw:
First stab at the network wide limits spec.

Next steps
- [x] add default values for each netparam
- [x] add API entries showing when a market has reached any of the limits
- [x] ACs

Closes https://github.com/vegaprotocol/specs/issues/1355 

We want this implemented for Oregon Train (research hour 31st Oct decision).


protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Number of assets
+
+`limits.assets.max`
+
+TODO: Do we really care? Each asset is added via a governance vote, so it doesn't seem likely we'll get a ton of unnecessary ones (plus we'd then have to implement a process of removing an asset with a vote if the limit was already reached, but there was some new asset that community perceived as very important). Also, does asset itself really increase the load that much?
+
+## Number of markets
+
+Introduce a new network parameter `limits.markets.max` controlling the maximum number of markets allowed within a network.
```
witgaw: In the future we may want to make these market parameters, so some markets that are deemed to be important and constraint by the limits are allowed to grow without increasing the overall limit, but making them network parameters is a sensible first implementation.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+If it gets decreased below the current total party count in any given market then all parties in that market remain unaffected (their positions and orders remain open), but no new parties are allowed to participate in the market until the total party count drops below the new limit.
+
+## Limit orders on a book
+
+Introduce a new network parameter `limits.markets.maxLimitOrders` controlling the maximum number of limit orders that can rest on a book in any market. If the limit is reached limit orders can still be submitted, but:
+
+* if the order is aggressive (results in a trade) the trade proceeds as normal,
+* if the order is passive (it doesn't match on entry) and would get added to the order book it gets rejected.
+
+Pegged orders and [liquidity provision orders](./0038-OLIQ-liquidity_provision_order_type.md) do not get counted towards the limit.
```
witgaw, line 43: The consequence of that is that if the maximum number of orders gets reached in a market then LPs can only exercise their strategy via the automated order shapes. This may not be ideal, but allowing exclusions for LPs would complicate the design and also open up the door for malicious party entering the market as an LP to circumvent the limits, so I'd leave it as is specified here for the first implementation and change it only if we can clear requests from LPs that it impedes their ability to exercise their strategy.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+
+## LP order shapes
+
+Each [LP order shape](./0038-OLIQ-liquidity_provision_order_type.md#how-they-are-submitted) has a hardcoded limit of at most 20 entries (offsets).
+
+## LPs in a market
+
+Introduce a new network parameter `limits.markets.maxLPs` controlling the maximum number of liquidity providers that can be active in any market. Each liquidity provider that successfully submits a liquidity provision transaction for a given market gets counted towards the **active LP count** for that market. When an LP in a given market reduces their commitment amount to 0 or gets closed out the LP count for that market gets decremented by 1.
+
+Once the active LP count in a market reaches the limit no further LP commitments are accepted. Once the count drops below the limit the LP commitments are accepted again.
```
witgaw: I think we will need to complicate this design a bit more to not allow a possibility of a market becoming deadlocked in a liquidity auction. 

So if market is in liquidity auction and the max LP limit is reached we should probably allow new LPs to kick out (their commitment gets cancelled) existing ones if their commitment is larger (in the commitment amount & seniority priority - we traverse the list by lowest commitment and within any commitment amount by time in market).

witgaw (commented):
Ready for initial comments. 

Please note that the strategy for dealing with limits getting decreased below the current state of the market is attrition based now - we just wait till things get back to normal, but don't do anything more drastic. This allows to avoid a lot of implementation complication and allows some form of "net neutrality" in the markets, but may not be enough to "rescue" a malfunctioning network if it turns out ex post that limits were too lax.

Please see my other comments below, curious to hear your thoughts:

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Number of assets
+
+`limits.assets.max`
+
+TODO: Do we really care? Each asset is added via a governance vote, so it doesn't seem likely we'll get a ton of unnecessary ones (plus we'd then have to implement a process of removing an asset with a vote if the limit was already reached, but there was some new asset that community perceived as very important). Also, does an additional asset itself really increase the load that much?
```
davidsiska-vega: I agree - we shouldn't have a limit on assets unless we have a process for removing them (which is super unclear TBH).

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Number of markets
+
+Introduce a new network parameter `limits.markets.max` controlling the maximum number of markets allowed within a network.
+
+Each market in a `Proposed` [state](./0043-MKTL-market_lifecycle.md) should increment the **total market count** by 1. If a market reaches a `Rejected`, `Cancelled`, `Closed` or `Settled` state the total market count should be decremented by 1.
```
davidsiska-vega: I would only increment once in opening auction. Otherwise someone with enough tokens to submit a makret proposal can trivially exhaust the max.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+As soon as the total market count reaches the value specified by `limits.markets.max` parmeter no other further market proposals should be accepted. If the total market count drops below `limits.markets.max` the network should start accepting market proposals again.
+
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
```
davidsiska-vega, line 21:
```suggestion
If the network parameter gets increased via a governance vote no further actions are needed.
```

protocol/0078-NWLI-network_wide_limits.md
davidsiska-vega: If the market is approved by voting *but* it would exceed the max then it's the same outcome as if it was rejected by voting.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+If it gets decreased below the current total market count then no further actions are needed. Specifically: all existing markets continue functioning in their respective states, but no further proposals are accepted until the count drops below the new limit.
+
+## Parties in a market
+
+Introduce a new network parameter `limits.markets.maxParties` controlling the maximum number of parties allowed within any given market.
+
+A party gets counted towards the limit if it has either open orders or open positions in the market. Once party has no open orders and no open positions it gets removed from a **total party count** within the market. Once the limit gets reached the market accepts no further orders.
```
davidsiska-vega:
```suggestion
A party gets counted towards the limit if it has either open orders or open positions in the market. Once party has no open orders and no open positions it gets removed from a **total party count** within the market. Once the limit gets reached the market accepts no further orders from parties that are not already in the market.
```

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+Once the total party count drops below the limit market accepts orders again (provided its [state](./0043-MKTL-market_lifecycle.md) allows that).
```
davidsiska-vega:
```suggestion
Once the total party count drops below the limit market accepts orders from any party again (provided its [state](./0043-MKTL-market_lifecycle.md) allows that).
```

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+
+The limit does not apply to liquidity providers.
+
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
```
davidsiska-vega, line 30:
```suggestion
If the network parameter gets increased via a governance vote no further actions are needed.
```

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+If it gets decreased below the current total party count in any given market then no further actions are needed. Specifically: all parties in that market remain unaffected (their positions and orders remain open), but no new parties are allowed to participate in the market until the total party count drops below the new limit.
+
+## Limit orders on a book
+
+Introduce a new network parameter `limits.markets.maxLimitOrders` controlling the maximum number of limit orders that can rest on a book in any given market. If the limit is reached limit orders can still be submitted, but:
+
+* if the order is aggressive (results in a trade) the trade proceeds as normal,
+* if the order is passive (it doesn't match on entry) and would get added to the order boo (it's a persistent order) it gets rejected.
```
davidsiska-vega:
```suggestion
* if the order is passive (it doesn't match on entry) and would get added to the order book (it's a persistent order) it gets rejected.
```

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+The limit does not apply to liquidity providers.
+
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
+If it gets decreased below the current total party count in any given market then no further actions are needed. Specifically: all parties in that market remain unaffected (their positions and orders remain open), but no new parties are allowed to participate in the market until the total party count drops below the new limit.
+
+## Limit orders on a book
+
+Introduce a new network parameter `limits.markets.maxLimitOrders` controlling the maximum number of limit orders that can rest on a book in any given market. If the limit is reached limit orders can still be submitted, but:
```
davidsiska-vega: A limit order of arbitrary volume contributes +1 to the count.

protocol/0078-NWLI-network_wide_limits.md
davidsiska-vega: That sounds essential, yes. In fact why just liq. auction? I'd say that at any point in time we are sitting at the max, if a newly arriving LP would commit enough stake so that they have more ELS than the LP with the least ELS then the smallest LP is booted out.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,80 @@
...
+Once the active LP count in a market reaches the limit no further LP commitments are accepted. Once the count drops below the limit the LP commitments are accepted again.
+
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
+If it gets decreased below the current total number of LPs in the market then no further actions are needed. Specifically: all LPs active in the market remain unaffected, however no new liquidity provisions are accepted until the total LP count drops below the new limit.
+
+## Acceptance Criteria
+
+* (<a name="0078-NWLI-001" href="#0078-NWLI-001">0078-NWLI-001</a>)
```
davidsiska-vega: ;-)

protocol/0078-NWLI-network_wide_limits.md
witgaw, line 16: This is just a line drop for readability.

protocol/0078-NWLI-network_wide_limits.md
witgaw: Deleted now.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -40,7 +40,7 @@ If it gets decreased below the current total party count in any given market the
 Introduce a new network parameter `limits.markets.maxLimitOrders` controlling the maximum number of limit orders that can rest on a book in any given market. If the limit is reached limit orders can still be submitted, but:
 
 * if the order is aggressive (results in a trade) the trade proceeds as normal,
-* if the order is passive (it doesn't match on entry) and would get added to the order boo (it's a persistent order) it gets rejected.
+* if the order is passive (it doesn't match on entry) and would get added to the order book (it's a persistent order) it gets rejected.
```
jeremyletang: Well, it would still be easy to submit very small order, which would match on entry and stay on the book just by using 2 different vega wallets right?

protocol/0078-NWLI-network_wide_limits.md
witgaw: Added now.

protocol/0078-NWLI-network_wide_limits.md
witgaw: Good point, modified now.

protocol/0078-NWLI-network_wide_limits.md
witgaw: Added now.

protocol/0078-NWLI-network_wide_limits.md
witgaw: I don't think so, I've clarified that part now, but not sure if that's what you were after.

witgaw (commented):
Ready for review

witgaw:
@peterbarrow could you please take a look at the proposed default values and let me know if they should either be relaxed or made more stringent based on your testing?

@vegaprotocol/qa could you please take a look at the ACs?

protocol/0078-NWLI-network_wide_limits.md
davidsiska-vega: 👍

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
+If it gets decreased below the current total number of orders on the book in any given market then no further actions are needed. Specifically: all open orders in that market remain unaffected, but no new orders are allowed until the **total open order count** drops below the new limit.
+
+## Pegged orders on a market
+
+Introduce a new network parameter `limits.markets.maxPeggedOrders` controlling the maximum number of pegged orders that can be active in any given market.\
+Default value: `10,000`.
```
davidsiska-vega:
```suggestion
Default value: `1,500`.
```

davidsiska-vega: https://vegaprotocol.slack.com/archives/C03GJPCERFY/p1666969465173389

davidsiska-vega: Frome @peterbarrow 's testing. 

image

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+Each [LP order shape](./0038-OLIQ-liquidity_provision_order_type.md#how-they-are-submitted) has a hardcoded limit of at most 20 entries (offsets).
+
+## LPs in a market
+
+Introduce a new network parameter `limits.markets.maxLPs` controlling the maximum number of liquidity providers that can be active in any market.\
+Default value: `1000`.
+
+Each liquidity provider that successfully submits a liquidity provision transaction for a given market gets counted towards the **active LP count** for that market. When an LP in a given market reduces their commitment amount to 0 or gets closed out the LP count for that market gets decremented by 1.
+
+Once the active LP count in a given market reaches the limit only the LP transactions with commitment amount larger then the lowest commitment already active in that market are accepted. If such a situation occurs the existing LP commitment with the lowest value which has been in the market the shortest should get forcibly cancelled.\
```
davidsiska-vega: I am not sure we want that. Imagine I was the first LP on the market; small commitment but *huge* ELS. This would be a way for someone to force me out.

davidsiska-vega:
```suggestion
Once the active LP count in a given market reaches the limit only the LP transactions with commitment amount implying [equity-like-share](0042-LIQF-setting_fees_and_rewarding_lps.md) larger then the lowest equity-like-share already active in that market are accepted. If such a situation occurs the existing LP commitment with equity-like-share which is getting replaced have their commitment cancelled (bond returned to margin account, volume no longer placed on book but positions and orders kept in place).\
```

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+
+### LP order shapes
+
+* Submitting a [liquidity provision order](./0038-OLIQ-liquidity_provision_order_type.md) with 20 buy shapes and 20 sell shapes proceeds without any errors and liquidity provision becomes active. (<a name="0078-NWLI-014" href="#0078-NWLI-014">0078-NWLI-014</a>)
+* Submitting a liquidity provision order with 21 buy shapes and 20 sell shapes results in a failure. Error message attributes the rejection to the limit. (<a name="0078-NWLI-015" href="#0078-NWLI-015">0078-NWLI-015</a>)
+* Submitting a liquidity provision order with 20 buy shapes and 21 sell shapes results in a failure. Error message attributes the rejection to the limit. (<a name="0078-NWLI-016" href="#0078-NWLI-016">0078-NWLI-016</a>)
+
+### LPs
+
+* Attempt to place an additional liquidity provision with valid commitment amount below the lowest one in the market when the number of liquidity provisions in the market is equal to `limits.markets.maxLPs` results in a rejection. Error message attributes the rejection to the limit. (<a name="0078-NWLI-017" href="#0078-NWLI-017">0078-NWLI-017</a>)
```
davidsiska-vega: This will potentially need updates, depending on how the discussion above pans out.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+[Liquidity provision orders](./0038-OLIQ-liquidity_provision_order_type.md) do not get counted towards the limit.
+
+## LP order shapes
+
+Each [LP order shape](./0038-OLIQ-liquidity_provision_order_type.md#how-they-are-submitted) has a hardcoded limit of at most 20 entries (offsets).
+
+## LPs in a market
+
+Introduce a new network parameter `limits.markets.maxLPs` controlling the maximum number of liquidity providers that can be active in any market.\
+Default value: `1000`.
```
davidsiska-vega: I suspect this is too high because it is somehow potentially equivalent to `1 000 x 20 x 2 = 40k` pegged orders but let's wait for @peterbarrow 's comment. Even if we don't resolve this my suggestion would be to pick a default of `50` so that the PR can be merged.

davidsiska-vega (changes requested):

protocol/0078-NWLI-network_wide_limits.md
witgaw: Good point, committed your suggestion.

protocol/0078-NWLI-network_wide_limits.md
witgaw: Ok, changed to 50 now.

protocol/0078-NWLI-network_wide_limits.md
witgaw: Committed your ELS suggestion and modified the ACs.

davidsiska-vega (approved):

davidsiska-vega:
We effectively have "minimum order margin cost limit" in terms of position decimal places. We now allow negative PDPs.
For example a market that’s trading at 0.001 USD (some kind of shitcoin) and has RF short of 0.1 so it will allow 10x leverage you would get the following minimum order costs on Vega:
PDPs = 0 ==> 0.001 x 0.1 = 1e-4 USD to place an order locked in margin.
PDPs = -5 ==> 0.001 x 0.1 x 1e5 = 10 USD to place an order locked in margin.

We could also of course consider a hard "order margin cost limit” in the sense that regardless of how the market is proposed there will be “reject orders which will lock less that x USD in margin”. 

This would have the small advantage that a market for an underlying which starts trading at 100 USD when the market is proposed and then the price tanks to 0.1 USD  two months later doesn’t allow stupidly tiny orders.

* Rejected * (at least for now) on research hour call. Not for oregon trail.

protocol/0021-MDAT-market_data_spec.md
```diff
@@ -55,6 +55,7 @@ List of market data fields to be available via the API. All these values can be
   - **Supplied stake** market's supplied stake.
   - **Price monitoring bounds** one or more price monitoring bounds for the current timestamp.
   - **Liquidity provider fee share** share of the accrued fees each liquidity provider is eligible to.
+  - **Active limits** list of [network wide limits](./0078-NWLI-network_wide_limits.md) and a boolean indicating if they are active for the queried market
```
jeremyletang, line 58: What do we mean by active here? that the limit have been breached and the market won't accept anymore e.g: orders?

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+If it gets decreased below the current total market count then no further actions are needed. Specifically: all existing markets continue functioning in their respective states, but no further proposals are accepted until the count drops below the new limit.
+
+## Parties in a market
+
+Introduce a new network parameter `limits.markets.maxParties` controlling the maximum number of parties allowed within any given market. Default value: `100,000`.
+
+A party gets counted towards the limit if it has either open orders or open positions in the market. Once party has no open orders and no open positions it gets removed from a **total party count** within the market. Once the limit gets reached the market accepts no further orders from parties that are not already in the market.
+Once the total party count drops below the limit market accepts orders from any party again (provided its [state](./0043-MKTL-market_lifecycle.md#market-lifecycle-statuses) allows that).
+
+The limit does not apply to liquidity providers.
```
jeremyletang, line 26: I'm not sure I understand? Does that mean that a party which is a liquidity order doesn't take part of this limit? Or that if this limit is reached, and an LP want to join AND the max LP limit is not reached, then the new lp submission should be accepted?

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
+If it gets decreased below the current total party count in any given market then no further actions are needed. Specifically: all parties in that market remain unaffected (their positions and orders remain open), but no new parties are allowed to participate in the market until the total party count drops below the new limit.
```
jeremyletang, line 31: Would it make sense to say -> all parties with open orders and no open positions are removed? I'm not sure, and it'd complicate a little bit things though

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Limit orders on a book
+
+Introduce a new network parameter `limits.markets.maxLimitOrders` controlling the maximum number of limit orders that can rest on a book in any given market.\
```
jeremyletang, line 35: Would it make sense to have that per side instead of per book? Just to make it less likely to have a book too massively filled on a specific side? that might also be an acceptable behaviour/.

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+Pegged orders and [liquidity provision orders](./0038-OLIQ-liquidity_provision_order_type.md) do not get counted towards the limit.
+
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
+If it gets decreased below the current total number of orders on the book in any given market then no further actions are needed. Specifically: all open orders in that market remain unaffected, but no new orders are allowed until the **total open order count** drops below the new limit.
+
+## Pegged orders on a market
+
+Introduce a new network parameter `limits.markets.maxPeggedOrders` controlling the maximum number of pegged orders that can be active in any given market.\
```
jeremyletang, line 52: Shouldn't that be a limit per party? it is so low that it seems that a few parties could easily reach that point?

protocol/0078-NWLI-network_wide_limits.md
```diff
@@ -0,0 +1,128 @@
...
+### Change of network parameter
+
+If the network parameter gets increased via a governance vote no further actions are needed.\
+If it gets decreased below the current total number of orders on the book in any given market then no further actions are needed. Specifically: all pegged orders already present in that market remain unaffected, but no new pegged orders are allowed until the **total pegged order count** drops below the new limit.
+
+[Liquidity provision orders](./0038-OLIQ-liquidity_provision_order_type.md) do not get counted towards the limit.
+
+## LP order shapes
+
+Each [LP order shape](./0038-OLIQ-liquidity_provision_order_type.md#how-they-are-submitted) has a hardcoded limit of at most 20 entries (offsets).
```
jeremyletang, line 64: This seems to go against the current default of 100 and maximum per side of 1000. Althought I'd rather reduce it a lot as suggested

PR #1363, opened 2022-10-26
chore: add description for snapshot.interval.length

### LINKED ISSUE #1290: Snapshot.interval.length not mentioned in 0009-NP-SNAP spec
The following network parameter is not mentioned in the [0009-NP-SNAP spec](https://github.com/vegaprotocol/specs/blob/master/non-protocol-specs/0009-NP-SNAP-snapshots.md):

`Snapshot.interval.length`

Can the spec be updated to include this param and describe its function / use

davidsiska-vega:
closes #1290


gordsport (approved):
LGTM

PR #1365, opened 2022-10-26
chore: tidy up net params spec

### LINKED ISSUE #1241: Update 0054-NETP-network_parameters.md to include new NetParams
### Update spec
Starting with the network parameters deemed to be a priority we need to add AC codes to the table to provide a granular record of coverage:
- [ ] Provide a link to the code where we have all the defults/ranges (https://github.com/vegaprotocol/vega/blob/develop/core/netparams/defaults.go) as a replacement for the table
- [ ] reference `updateDisallowed` in https://github.com/vegaprotocol/vega/blob/develop/core/netparams/netparams.go as a list of those not able to be changed via an on-chain governance vote.
- [ ] Check all specs reference the correct parameters that match what’s in code (update if needed)
- [ ] list out any params that do not exist in the specs 
- [ ] list any that are not referenced in ACs

gordsport:
Tidy's up the spec by removing the table and pointing to the code where the params and their ranges are set.

Also updates the threshold info and example according to the implementation 

close: vegaprotocol/specs#1241 .


gordsport:
@davidsiska-vega 
please can you review and also add link for the naming: 
https://github.com/vegaprotocol/vega/blob/develop/core/netparams/keys.go

davidsiska-vega (approved):

PR #1371, opened 2022-10-27
Added GFN and GFA behaviour ACs

### LINKED ISSUE #1370: Add ACs to auctions spec to cover coming in and out of auction
We identified some missing coverage with orders being cancelled when going in or out of auction.

GFN should be cancelled when going into auction, GFA should be cancelled when coming out of auction.

peterbarrow:
Added ACs to test for removal of orders when moving in and out of auctions.

Closes #1370


witgaw (approved):
LGTM

davidsiska-vega (approved):

PR #1388, opened 2022-11-08
Update 0069-VCBS-validators_chosen_by_stake.md

### LINKED ISSUE #1384: Antiwhale ersatz
As per https://github.com/vegaprotocol/vega/issues/6711 and https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1667414246465009

Vegaklaus:
Added Ersatzvalidator anti-whaling and AC.

Closes #1384 .


protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -383,6 +387,11 @@ Verify that 2 validators are replaced, one in each epoch
 
 Verify that one validator is replaced the following epoch, one in the epoch after
 
+13. Ersatzvalidator reward (<a name="0069-VCBS-051" href="#0069-VCBS-051">0069-VCBS-051</a>)
```
ze97286: in this case the antiwhaling has no meaning as if the demoted ersatz validator is the only ersatzvalidator, they take the whole lot allocated to ersatzvalidators. Depending on the value of network.validators.ersatz.rewardFactor this can definitely give them a higher reward than tendermint validators. It actually highlights another aspect of it, suppose you have 2 ersatzvalidators, one as in your example where it has 50% of the total delegation of tendermint validators - and another earsatz validator with half of the delegation of any other tendermint validator (so lets say all tendermint validators have 10k, and the downgraded ersatz validator has 40k, and the other ersatz validator has 5k) - assuming perfect performance - the split between the ersatzvalidators after antiwhaling will be 2:1 but the split of the reward pool will be 50:(45 * scale_down), so depending on how much we scale the ersatz delegation down, there's still a chance that the ersatz validators (even the small one with 5k delegation) will get paid more than the tendermint validators. 

Let's say scale_down=0.5 this gives the ersatz validators 22.5/72.5 to be split 2:1 between them, which gives the bigger validator (2/3)*22.5/72.5 = 0.2068965517 of the total reward which is still bigger share than any of the tendermint validators which would get 0.2 * 50/72.5 = 0.1379310345

Vegaklaus: I fixed the AC; the other issue is something that'd require some redesign (or implementing parameter-dependencies), so I'd propose that's post-launch.

protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -115,7 +115,11 @@ Their performance is also defined by the number of heartbeats they sent out of t
 
 ### Rewards for Ersatz validators
 In terms of rewards, Ersatz validators are treated in line with Tendermint validators, see details in [validator rewards spec](./0064-VALP-validator_performance_based_rewards.md) and [performance measurement](./0064-VALP-validator_performance_based_rewards.md).
-However `network.validators.ersatz.rewardFactor` in `[0,1]` is taken into account to scale their rewards.
+However `network.validators.ersatz.rewardFactor` in `[0,1]` is taken into account to scale their rewards. Also, the same scoring
+function is aopplied as for the normal validators, so anti-whaling rules apply for Ersatzvalidators as well.
```
ze97286: aopplied => applied

protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -115,7 +115,11 @@ Their performance is also defined by the number of heartbeats they sent out of t
...
+An Ersatzvalidator being affected by the whaling rule (i.e., getting sufficient stake to havbe their reward lowered by the anti-whaling rule)
```
ze97286: havbe => have

protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -115,7 +115,11 @@ Their performance is also defined by the number of heartbeats they sent out of t
 
 ### Rewards for Ersatz validators
 In terms of rewards, Ersatz validators are treated in line with Tendermint validators, see details in [validator rewards spec](./0064-VALP-validator_performance_based_rewards.md) and [performance measurement](./0064-VALP-validator_performance_based_rewards.md).
-However `network.validators.ersatz.rewardFactor` in `[0,1]` is taken into account to scale their rewards.
+However `network.validators.ersatz.rewardFactor` in `[0,1]` is taken into account to scale their rewards. Also, the same scoring
+function is aopplied as for the normal validators, so anti-whaling rules apply for Ersatzvalidators as well.
+An Ersatzvalidator being affected by the whaling rule (i.e., getting sufficient stake to havbe their reward lowered by the anti-whaling rule) 
+is a out-of-the ordinary event and should be logged.
```
ze97286, line 121: I'm not sure what's the point of this logging. who's gonna scan the log for these? if you already suspect someone is overpaid, might as well do the math manually rather than scan the log.

protocol/0069-VCBS-validators_chosen_by_stake.md
ze97286: (we already have implemented param dependency, but happy to wait till after launch)

protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -383,6 +387,11 @@ Verify that 2 validators are replaced, one in each epoch
 
 Verify that one validator is replaced the following epoch, one in the epoch after
 
+13. Ersatzvalidator reward (<a name="0069-VCBS-051" href="#0069-VCBS-051">0069-VCBS-051</a>)
+    Set up a network with 5 validators one of them with sufficient stake to be affected by the anti-whaling rule (e.g., 50% stake), as well as 5 Ersatzvalidators with equal (low) stake.  Demote
```
ze97286: would be good to give the full details here so it's easier to implement:

1. setup a network with 5 validators with the following distribution of delegation:
10%, 10%, 10%, 10%. 60% of the total delegation of tendermint validators

2. setup 5 ersatz validators each with the minimum delegation 
3. at the end of the epoch verify that the stake score of the validator with 60% of the delegation (under reward) is anti-whaled
4. shutdown the validator with 60% of the delegation
5. run for an epoch with it down
6. at the end of the epoch expect the validator with 60% of the stake to be swapped as a tendermint validator for one of the ersatz validators. 
7. restart the validator, run until the end of the epoch
8. verify that this validator is paid reward as ersatz validator and that their stake score under reward is anti-whaled

protocol/0069-VCBS-validators_chosen_by_stake.md
davidsiska-vega:
```suggestion
function is applied as for the normal validators, so anti-whaling rules apply for Ersatzvalidators as well.
```

protocol/0069-VCBS-validators_chosen_by_stake.md
davidsiska-vega:
```suggestion
An Ersatzvalidator being affected by the whaling rule (i.e., getting sufficient stake to have their reward lowered by the anti-whaling rule) 
```

protocol/0069-VCBS-validators_chosen_by_stake.md
davidsiska-vega: Hypothetically monitoring could be set up for these logged events and it may be worth noticing. A logged message won't harm.

protocol/0069-VCBS-validators_chosen_by_stake.md
davidsiska-vega: We leave it for now.

davidsiska-vega (approved):

PR #1389, opened 2022-11-08
Specify closeout behaviour upon exiting auction + AC

### LINKED ISSUE #1322: AC needs a home - at the point of leaving auction to continuous what happens if parties become distressed. Define expected behaviour
Find home spec for this..

AC - at the point of leaving auction to continuous what happens if parties become distressed. Define expected behaviour.

witgaw:
As per title☝️ 

Closes https://github.com/vegaprotocol/specs/issues/1322


Jiajia-Cui (approved):
LGTM

PR #1390, opened 2022-11-08
Reflect latest stance on the network wide limits in the specs

### LINKED ISSUE #1379: Re-write 0078-NWLI network limits spec
After a review we have decided that we will rework this spec to only include:

- Limit the number of pegged orders Vega wide.
- Add limits of the number/type of transactions that can be added to a block (based on Tendermint gas weighting events). @jeremyletang to add a proposal

We also need to review and update existing network parameters:
- Set the minimum an LP can submit to $10,000 (value TBC)
- Update the LP shapes param to be 5 each side (currently set to 100

The information in this spec will remain in git history should we wish to revisit in the future

witgaw:
My bit of https://github.com/vegaprotocol/specs/issues/1379

### DIFF
--- a/protocol/0020-APIS-core_api.md
+++ b/protocol/0020-APIS-core_api.md
@@ -36,6 +36,11 @@ To *observe the operation, and validate the state of the protocol, we must be ab
 
 - List all known [parties](./0016-PART-party.md).
 
+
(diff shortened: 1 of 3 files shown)


Jiajia-Cui (approved):
LGTM

PR #1397, opened 2022-11-10
Vega transaction gas costs and priorities

davidsiska-vega:


protocol/0079-TGAS-transaction_gas_costs.md
```diff
@@ -0,0 +1,109 @@
...
+                                        + level factor x levels
+gasOrder = min(maxGas-1,gasOrder)
+```
+
+### Cancellation of any order
+
+```
+gasCancel = network.transaction.defaultgas + peg cost factor x pegs 
+                                        + LP shape cost factor x shapes 
+                                        + position factor x levels
```
TomMcL: `position factor x levels`?

Does removing an order from the book entail any notably less processing/fewer logs than creating an order? If not then presumably this should be the same as creating an order (which would also reduce the number of different calculations)

TomMcL:
Put together a very simple [calculator](https://docs.google.com/spreadsheets/d/1w8W98n42iuSQXFLUH_tq0SCpqZdj3Hko3SnK8YWwFFc/edit?usp=sharing) and it seems like with current suggested parameters we very quickly end up with a tiny number of orders allowed per block (and are always quite low even to start with). So if the weight formulae are reasonably representative of growth we probably want to bump up the minimum/default maxgas numbers?

protocol/0079-TGAS-transaction_gas_costs.md
davidsiska-vega: ~~My reasoning here was that an incoming order might result in a trade (and associated costs due to impacting positions) whereas removing an order won't. ~~

~~ But I'd also be happy to treat them all the same. ~~

Sorry, I can't read. Fixed the typo now.

davidsiska-vega:
> calculator

The calculator is very handy indeed; we the cost factors and default max gas will be up to @peterbarrow / engineering to decide.

Vegaklaus (commented):
The 'include transactions in sequence until the gas limit is reached' can mean two things:
- go through transactions in sequence until the next one doesn't fit anymore
- as above, but then keep going through the next transactions to see if there's one later that's cheap enough to fit

davidsiska-vega:
@Vegaklaus currently Tendermint does: keep adding stuff into block as long as gas fits, if the next transaction is too big, that's it, nothing more goes in. 

Other than that, spoke with @jeremyletang - we're not adding transactions that call for signing: `WithdrawSubmission` `IssueSignatures`, `EthereumKeyRotateSubmission`, `KeyRotateSubmission`.  

So it's ready for review. And also already implemented 🤣 .

TomMcL (dismissed):

protocol/0079-TGAS-transaction_gas_costs.md
```diff
@@ -0,0 +1,112 @@
...
+```
+gasCancel = network.transaction.defaultgas + peg cost factor x pegs 
+                                        + LP shape cost factor x shapes 
+                                        + level factor x levels
+gasCancel = min(maxGas-1,gasCancel)
+```
+
+### Batch orders 
+
+Define `batchFactor` (a hard coded parameter) set to something between `0.1 and 0.9`.
```
barnabee: This should be between 0 and 1. It is quite reasonable to imagine wanting to charge the same for every item in the batch or to charge nothing more than the cost of one of each type of transaction, especially as we optimise other bits.

It's also possible that we may in future want a different batchFactor for each instructon type but I probably wouldn't bother adding that now?

protocol/0079-TGAS-transaction_gas_costs.md
```diff
@@ -0,0 +1,112 @@
...
+- `batchFactor = 0.5` decimal between `0.1 and 0.9`.
+
+### Any type of limit or market order
+
+```
+gasOrder = network.transaction.defaultgas + peg cost factor x pegs 
+                                        + LP shape cost factor x shapes 
+                                        + position factor x positions 
+                                        + level factor x levels
+gasOrder = min(maxGas-1,gasOrder)
```
barnabee: Seems like the implication of this formula is that it is relatively easy to have an order, or indeed a cancellation that **has to be the first order in the block**.

I worry that this could make it really hard for someone to bring a market out of auction by adding one or more priced limit orders or adding/updating liquidity commitments. If we have 100 orders per block you have a 1% chance of getting in.

Perhaps something lke: 

```
gasOrder = min((maxGas/minBlockCapacity)-1,gasOrder)
```

might solve this, where minBlockCapacity is some constant that we might set to something like 16 or 32.

protocol/0079-TGAS-transaction_gas_costs.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+### Cancellation of any order
+
+```
+gasCancel = network.transaction.defaultgas + peg cost factor x pegs
```
barnabee: We need a cost for the two special mass cancel tx types:

- cancel all orders on a market for a party = same cost as cancelling ONE order on that market (we only do any other processing like reprice once ALL are removed)
- cancel all orders on all markets for a party = same cost as cancelling ONE order on each market the party has orders on

protocol/0079-TGAS-transaction_gas_costs.md
davidsiska-vega: ✅

protocol/0079-TGAS-transaction_gas_costs.md
davidsiska-vega: ✅

protocol/0079-TGAS-transaction_gas_costs.md
davidsiska-vega: ✅

barnabee (dismissed):

barnabee (approved):

PR #1398, opened 2022-11-11
Use existing netparam for LP shape

witgaw:
We already have a network parameter `market.liquidityProvision.shapes.maxSize` so let's use it.


Jiajia-Cui (approved):
LGTM

PR #1400, opened 2022-11-11
refactor: modify AC to reflect current implementation

witgaw:


davidsiska-vega (dismissed):
While you're doing this could you also say what increasing it will do? I am assuming it will increase the current period, but would be worth checking; spelling out.

gordsport (approved):
Approved on PD weekly call

PR #1412, opened 2022-11-15
remove parked pegged orders from

RomanLeca:
Acting on this comment
https://github.com/vegaprotocol/vega/issues/6823#issuecomment-1315248756


davidsiska-vega:
We should check that the parked orders don't incur margin (since they're not on the book). But the ACs can stay. They can be tested via feature tests only.

PR #1418, opened 2022-11-18
chore: correct AC code number 0027-ASSP-020

gordsport:
@vega-paul 

The AC code `0027-ASSP-020` is reference in `./system-tests/tests/assets/erc20_asset_test.py`

However in this spec its marked as COSMICELEVATOR.

- Should the reference in the test be removed to avoid possible false coverage once this AC is made current?
- Should this AC be made current as it is indeed testable and has been covered?

cc @davidsiska-vega for info


davidsiska-vega (dismissed):

vega-paul:
This AC be made current as it is indeed testable and has been covered

vega-paul:
Our system test for this AC is not covering the defined scenario. I can adjust easily to match and can do that now on a new PR.

vega-paul (approved):

vega-paul:
Spec reflects new AC and existing AC covered so we are good.

PR #1429, opened 2022-11-21
Update 0049-TVAL-validate_transaction_preconsensus.md

vega-paul:


protocol/0049-TVAL-validate_transaction_preconsensus.md
```diff
@@ -35,4 +35,5 @@ Note that separate pre-consensus validation is carried out as part of PoW anti-s
...
 1. [ ] Transaction is rejected (never included in a block) if a party has strictly less than the [quantum](./0041-TSTK-target_stake.md)  balance in the settlement asset of the market and it is submitting any kind of orders (limit, market, LP provision)  (<a name="0049-COSMICELEVATOR-004" href="#0049-COSMICELEVATOR-004">0049-COSMICELEVATOR-004</a>)
 1. [ ] Transaction interacting in a market is included in a block if signed with a key from a non-validator party with a balance >= [quantum](./0041-TSTK-target_stake.md) of the settlement asset for the market, where an identical (apart from PoW proof and block data details) transaction was rejected from a previous block when the party had relevant balance < less than relevant [quantum](./0041-TSTK-target_stake.md) (<a name="0049-COSMICELEVATOR-005" href="#0049-COSMICELEVATOR-005">0049-COSMICELEVATOR-005</a>)
+1. [ ] A party with no balance in any account must have their transactions rejected. (<a name="0049-TVAL-006" href="#0049-TVAL-006">0049-TVAL-006</a>)
```
wwestgarth, line 38: What do you mean by rejected? `0049-TVAL-001` suggests that someone with no balance in any accounts that submits a non-governance transaction will **_not_** be blocked by pre-consensus spam (for OT at least). 

The discussion about it is here: https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1658312125720139?thread_ts=1657884254.730619&cid=CAHA5EX0F

vega-paul: We cover this by just ensuring orders are rejected with no asset balance

gordsport:
@vega-paul - please can you catch up with @davidsiska-vega as this AC should not apply to ALL transactions

vega-paul:
@davidsiska-vega we need to think of a way to describe this behaviour and test it. Currently we only check balances and order success but we would refactor to match.

gordsport:
Closing as this AC is not correct - further discussions to be had.

Can we create a table for each TX type what can cause it to be rejected pre-consensus?

PR #1435, opened 2022-11-23
chore: Remove no longer valid OPEG AC

TomMcL:
Removing acceptance criterion for 0037-OPEG.

Criterion is no longer needed as for now we only have one type of market, so pegged orders existing at all fulfils it.


Jiajia-Cui (approved):
LGTM

PR #1436, opened 2022-11-24
chore: update DSRI AC to match the intention

ze97286:


gordsport (approved):
Approved in the protocol design meeting

PR #1449, opened 2022-12-01
docs: Adding clarity on PoW specs

### LINKED ISSUE #1448: Improve ACs around increasing PoW difficulty with number of transactions
From @barnabee 

We need to ensure this is covered with good ACs fo reach case, too. I also think the spec. could be made clearer and more detailed around exactly how this works, it's written quite vaguely, for instance:
- not clear that it is max txns per block for each party
- not explicit that this means the party needs to count their txns for each key and increase difficulty themselves and they won't get any help from the core/data node APIs for that part
not clear how much difficulty needs to increase to be compliant
- etc.

TomMcL:


Vegaklaus (commented):
If we add the increased difficulty, it should be spam.pow.difficulty+x, if the validator has seen at least x*spam.pow.numberrOfTransactionsPerBlock.
instead of
, or `spam.pow.difficulty` + 1 if the validator has already seen `spam.pow.numberOfTxPerBlock` transactions from this party within the same block.

TomMcL:
@Vegaklaus it looks like, after having dug into the current code, the way we currently have it is once we pass `numberOfTxPerBlock` the difficulty starts increasing by 1 for each transaction it receives (e.g. 10 for the first 100 tx, then 11 for tx 101, then 12 for tx 102 etc). What do you think on this setup? Will be a much 'harder' cutoff as I imagine you'll reach a computationally slow number much faster than otherwise.

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -12,46 +12,49 @@ spam.pow.increaseDifficulty = 0 (range: 0/1)
...
 - The same identifier has not been used for another transaction from a previously committed block. If the same identifier is used for 
   several transactions in the same block, these transactions need to be removed during post-processing, and the initiator blocked as a 
   spammer.
-- The same block has not been used for more than `spam.pow.numberOfTxPerBlock` transactions per spam difficulty level 
+- The same block has not been used for more than `spam.pow.numberOfTxPerBlock` transactions by the same party per spam difficulty level 
   (i.e., if `spam.pow.increaseDifficulty` is `> 1`, the same block can be used for more transactions if the PoW accordingly increases in difficulty).
  
  Violations of the latter rules cannot lead to a transaction being removed, as different validators have a different view on 
- this; however, they can be verified post-agreement, and the offending vega-key can be banished for 4 epochs.
+ this; however, they can be verified post-agreement, and the offending vega-key can be banished for the maximum of 1/48 of an Epoch (30 minutes if an epoch is 1 day) or 30 seconds.
```
ze97286: maybe better as:
can be banished for the duration of 1/48 of an Epoch with a minimum duration of 30 seconds. E.g. if the epoch duration is 1 day, then the ban period is 30 minutes. If however the epoch is 10 seconds, then the ban period is 30 seconds.

protocol/0072-SPPW-spam-protection-PoW.md
```diff
@@ -12,46 +12,49 @@ spam.pow.increaseDifficulty = 0 (range: 0/1)
...
-3. The user then attaches the PoW to a transaction, and sends it of together with `x` and `H(b)`.
+2. The user downloads the hash `H(b)` of the latest block it has seen (or uses any other block hash within `spam.pow.numberOfPastBlocks` of the block to-be-produced), and brute forces values of `x` such that `hash("Vega_SPAM_PoW", H(b), tid, x)` as bytes starts with `spam.pow.difficulty` zeros.
+   1. The user must monitor how many transactions they have sent with PoW tied to a given block.
+      1. If this number is less than or equal to `spam.pow.numberOfTxPerBlock` the hash must start with `spam.pow.difficulty` zeros.
+      2. If this number is greater than `spam.pow.numberOfTxPerBlock` the hash must start with `spam.pow.difficulty` zeros plus one additional for each `spam.pow.numberOfTxPerBlock` sized batch of transactions beyond the limit (e.g. if `spam.pow.numberOfTxPerBlock` is 10 and `spam.pow.difficulty` is 2 the 11th - 20th transactions will require 2 zeros, the 21st-30th 3 zeros, the 31st-40th 4 zeros and so on).
```
ze97286: I'm confused by the example, why the first batch of 10 is free? 
if `spam.pow.numberOfTxPerBlock` is 10 and `spam.pow.difficulty` is 2 then:
1-10 - 2 preceding 0s
11-20 - 3 preceding 0s
21-30 - 4 preceding 0s
31-40 - 5 preceding 0s

protocol/0072-SPPW-spam-protection-PoW.md
TomMcL: Sounds good

protocol/0072-SPPW-spam-protection-PoW.md
TomMcL: Yeah seem to be off by 1 with my example zeroes there, fixed

ze97286 (approved):

PR #1415, opened 2022-11-17
Update 0072-SPPW-spam-protection-PoW.md

Vegaklaus:
Acs added for parameter changes

close: #1407 .

PR #1480, opened 2022-12-19
fix: Update 0062-SPAM-008 ban length to one 48th of an epoch

gordsport:


gordsport:
Closing as this is now accounted for in: https://github.com/vegaprotocol/specs/pull/1415

PR #1485, opened 2022-12-21
feat: add datanode modes to spec

### LINKED ISSUE #1432: Update datanode spec to define the 3 default modes
In the data nodes it is possible to set data retention policies. During a recent session on the length of the data retention it was discussed that we are likely to have 3 base default types of data node retention policies.

The spec should mention the minimum and the archival settings

ettec:
closes #1432


davidsiska-vega (commented):
Looks great, would you be able to formulate some ACs for this? 

We can then speak with QA how best test those (again could be a meta test - something that gets checked against the whole run).

davidsiska-vega (approved):

PR #1487, opened 2022-12-23
Remove never raised error AC

davidsiska-vega:


peterbarrow (approved):
The code which generates this reason is currently not able to run as the conditions required are handled above it and the code returns early. Removing this AC feels like the right thing to do in case the order of things change in the future.

PR #1496, opened 2023-01-06
feat: builtin product Spot

### LINKED ISSUE #871: New Product: Spot

Create a new spec with for new product spot markets.

 - [ ] The spec for spot has been written
 - [ ] The specification contains acceptance criteria with IDs that reflect the scope of the work required

cdummett:


TomMcL:
Thought that is perhaps already agreed/in the spec but I missed. What does 'staking' liquidity in a spot market even entail? Given that there are no 'margin' positions to maintain, is there any way to be slashed? Perhaps the bond could be used to post orders without slashing as the default behaviour, but then what is an LP providing to be worthy of setting the fee vs anyone else posting orders? (Especially given there is no lock so they can submit and remove as frequently as they want).

Along those lines, I think I prefer option 1 as someone posting with a silly fee when it isn't needed shouldn't be able to affect the fee even if they stake a huge bond. 

One could have timed bond stakes which can only be removed after a certain period of time but that might be adding a lot of complexity! Additionally should an LP here stake tokens on both sides of the trade? Again though, implementation complexity..

cdummett:
@TomMcL 

> Perhaps the bond could be used to post orders without slashing as the default behaviour

Bond slashing is an interesting point which I hadn't considered so thanks! I think this suggestion would work for both Option 1 and Option 2.

> but then what is an LP providing to be worthy of setting the fee vs anyone else posting orders?

I think the main thing would be LPs are forced to commit an equal amount of liquidity on either side of the book. A trader positing limit orders isn't forced to do this. You are right though, this "disadvantage" for LPs is really quite negated by the fact they can just remove a commitment.

Liquidity providers still have an incentive to not  frequently submit and cancel liquidity provisions though because they would lose any virtual stake gained from market growth. If that isn't enough to dissuade LPs from doing so, maybe this is where we add bond-slashing...? (i.e. If an LP want's to remove a commitment some of their bond is slashed and added to the insurance pool?)

> Along those lines, I think I prefer option 1 as someone posting with a silly fee when it isn't needed shouldn't be able to affect the fee even if they stake a huge bond.

I did wonder this, in the end my reasoning for not thinking this was an issue was "Why would someone do this? What do they have to gain?". The network parameter `Market.Liquidity.MaximumLiquidityFeeFactorLevel` should also help here.

protocol/0080-SPOT-product_builtin_spot.md
```diff
@@ -0,0 +1,167 @@
...
+
+A single LP will never be able to reduce their commitment to `0`. They can either keep reducing to a sufficiently small amount they're willing to ignore, or they can submit a governance vote to cancel the market, see the [governance spec](./0028-GOVE-governance.md).
+
+For market cancellation proposal a sole LP in the market holds all the voting power (unless governance token holders override them).
+
+### Liquidity Shortfalls
+
+If at any point in time, a liquidity provider has insufficient capital in their general accounts to cover a transfer arising from a filled liquidity order, the network will utilise the liquidity commitment, held in the relevant bond account to cover the shortfall, applying the bond penalty factor (slashing the bond).
+
+As there is no market insurance pool, funds from bond slashing in the result of shortfall will be transferred to the global insurance pool for that asset.
```
ze97286: there is also no such thing as global insurance pool for the asset. Maybe change to network treasury?

cdummett: Thanks, have updated to network treasury

protocol/0080-SPOT-product_builtin_spot.md
```diff
@@ -0,0 +1,167 @@
...
+        reference: "example_liquidity_provision_submission"
+    }
+}
+```
+
+As the LP now has a different commitment amount on each side of the book, the following considerations must be made:
+
+- Physical Stake:
+  - An LPs `physical_stake` should be treated separately for each side of the book - call these the `buy_physical_stake` and the `sell_physical_stake`.
+  - The current `physical_stake` for market stake calculations is the smaller of the two values, where the `sell_physical_stake` is converted into the `quote_asset` at the current `mark_price`.
```
ze97286, line 70: "at the current mark_price" - does that mean the last traded price?

cdummett: Was thinking the `mark_price` should be calculated the same way as in a Futures market, so the `mark_price` wouldn't be the last traded price if `network.markPriceUpdateMaximumFrequency` is greater than `0s`.

davidsiska-vega (approved):

PR #1499, opened 2023-01-09
Added Acs to. 0062-SPAM

### LINKED ISSUE #1482: additional tests for 0062 spec around 0062-SPAM-008
1. After epoch ends - so the ban ends as well - check that we can propose again;
2. Check that a party with less than 50% if it is post-rejected then it is not banned;
3. Get a party to be banned again once the initial ban expires;
4. once the ban is lifted vote again on an issue that already had full votes to check that the vote is rejected but party is not banned
5. check that during the ban all governance transactions are rejected
6. check what happens during the ban (the ban is applied when doing market proposals) if you raise orders - during and after the ban

https://github.com/vegaprotocol/specs/blob/master/protocol/0062-SPAM-spam_protection.md#0062-SPAM-008

the text from the spec is here:
`Any tokenholder that had more than 50% if its post-rejected is banned for max (30 seconds, 1/48 of an epoch) or until the next epoch starts, and all its governance related transactions ( but no no trading related transactions) are immediately rejected. E.g. if the epoch duration is 1 day, then the ban period is 30 minutes. If however the epoch is 10 seconds, then the ban period is 30 seconds (or until the start of the next epoch). The test for 50% of the governance transactions is repeated once the next governance related transaction is post-rejected, so a it is possible for a violating party to get banned quite quickly again; the test is only done in case of a new post-rejection, so the account does not get banned twice just because the 50% quata is still excveeded when the ban ends. The voting counters are unaffected by the ban, so voting again on an issue that already had the full number of votes in the epoch will lead to a rejection of the new vote; this is now likely to not trigger a new ban, as this rejection will happen pre-consensus, and thus not affect the 50% rule.`

Vegaklaus:


protocol/0062-SPAM-spam_protection.md
```diff
@@ -108,6 +108,12 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
...
  - Any rejection due to spam protection is reported to the user upon transaction submission detailing which criteria the key exceeded / not met  (<a name="0062-SPAM-013" href="#0062-SPAM-013">0062-SPAM-013</a>)  
  - If a party is banned for too many voting-rejections, it still can send trading related transactions which are not banned. (<a name="0062-SPAM-014" href="#0062-SPAM-014">0062-SPAM-013</a>)  
+ - If the ban of a party ends because the banning time is up, transactions from that party are no longer rejected (<a name="0062-SPAM-015" href="#0062-SPAM-015">0062-SPAM-015</a>)  
+ - If the ban of a party ends because the epoch ends, transactions from that party are no longer rejected (<a name="0062-SPAM-016" href="#0062-SPAM-016">0062-SPAM-063</a>)  
+ - If a party gets banned, the ban ends due to the epoch ending, and it gets banned again at the beginning of the new epoch, the bann still laasts the entire time (or until the next epoch end), i.e., the ban-expiration timer is reset. (<a name="0062-SPAM-017" href="#0062-SPAM-017">0062-SPAM-017</a>)
```
```suggestion
 - If a party gets banned, the ban ends due to the epoch ending, and it gets banned again at the beginning of the new epoch, the ban still lasts the entire time (or until the next epoch end), i.e., the ban-expiration timer is reset. (<a name="0062-SPAM-017" href="#0062-SPAM-017">0062-SPAM-017</a>)  
```

protocol/0062-SPAM-spam_protection.md
```diff
@@ -108,6 +108,12 @@ is then not increased for another 10 blocks. At the beginning of every epoch, th
...
  - If a party is banned for too many voting-rejections, it still can send trading related transactions which are not banned. (<a name="0062-SPAM-014" href="#0062-SPAM-014">0062-SPAM-013</a>)  
+ - If the ban of a party ends because the banning time is up, transactions from that party are no longer rejected (<a name="0062-SPAM-015" href="#0062-SPAM-015">0062-SPAM-015</a>)  
+ - If the ban of a party ends because the epoch ends, transactions from that party are no longer rejected (<a name="0062-SPAM-016" href="#0062-SPAM-016">0062-SPAM-063</a>)  
+ - If a party gets banned, the ban ends due to the epoch ending, and it gets banned again at the beginning of the new epoch, the bann still laasts the entire time (or until the next epoch end), i.e., the ban-expiration timer is reset. (<a name="0062-SPAM-017" href="#0062-SPAM-017">0062-SPAM-017</a>)  
+ - If a party gets banned several times duting an epoch, all banns last for the defined time or until the epich ends (try with at least three banns) (<a name="0062-SPAM-018" href="#0062-SPAM-018">0062-SPAM-018</a>)
```
```suggestion
 - If a party gets banned several times during an epoch, all banns last for the defined time or until the epoch ends (try with at least three banns) (<a name="0062-SPAM-018" href="#0062-SPAM-018">0062-SPAM-018</a>)  
```

gordsport (dismissed):
changes requested related to spelling not content 

@claudiumilea or @vega-paul to review content

vega-paul:
Can I suggest we add a few more ACS to cover undelegation trx and to cover ensuring this trx does not impact core performance unduly.

Could we add....

More than 360 undelegation changes in one epoch (or, respectively, the value of spam.protection.max.undelegation)

Undelegating while having less than one vega (10^18 of our smallest unit) (spam.protection.undelegation.min.tokens)

gordsport:
@Vegaklaus 
from paul:
> Can I suggest we add a few more ACS to cover undelegation trx and to cover ensuring this trx does not impact core performance unduly.

Could we add....

More than 360 undelegation changes in one epoch (or, respectively, the value of spam.protection.max.undelegation)

Undelegating while having less than one vega (10^18 of our smallest unit) (spam.protection.undelegation.min.tokens)

claudiumilea (approved):

PR #1500, opened 2023-01-09
feat: successor market

### LINKED ISSUE #1440: Successor markets
Propose a market and specify that it is a continuation of an existing market (starts with opening auction, keep the equity like shares)

davidsiska-vega:
Closes #1440 

Probably start reading the changes to the governance spec first, followed by liquidity.


protocol/0043-MKTL-market_lifecycle.md
```diff
@@ -85,9 +85,9 @@ When a Market Proposal is not successful, see [governance proposal](./0028-GOVE-
 
 ### Pending
 
-When a Market Proposal is successful at the end of the voting period, the Market state becomes "Pending". Currently a Pending Market is always in an [auction call period](./0026-AUCT-auctions.md) that ends at the enactment date as specified in the Market Proposal. 
+When a Market Proposal is successful at the end of the voting period, the Market state becomes "Pending". Currently a Pending Market is always in an [auction call period](./0026-AUCT-auctions.md) that ends at the enactment date as specified in the Market Proposal if the other conditions for exiting auction period are met (liquidity committed, best static bid / ask present). If, initially,  at the specified enacment date it was not possible to leave the auction period, the auction will conclude as soon as possible once the conditions for auction exit are met.
```
TomMcL: Probably want to specify that a successor market will remain in pending until it's parent is settled, then (assuming other conditions are met) the opening auction will immediately resolve (in the same block I would expect?)

protocol/0028-GOVE-governance.md
```diff
@@ -213,8 +213,19 @@ Note the following key points from the market lifecycle spec:
 * A market usually enters Active status at the proposal's enactment date/time, but some conditions may delay this or cause the market to be Cancelled instead
 
 A proposal to create a market contains
-1. a complete market specification as per the Market Framework (see spec) that describes the market to be created.
+1. a complete market specification as per the [Market Framework](./0001-MKTF-market_framework.md) that describes the market to be created.
 1. an enactment time that is at least the *minimum auction duration* after the vote closing time (see [auction spec](./0026-AUCT-auctions.md))
+1. if the market is meant to be a *sucessor* of a given market then it contains the marketId of the market it's suceeding (parent market) and certain entries in the market proposal must be identical to those of the market it's succeeding.
```
TomMcL: Possibly those entries (whatever they end up being) shouldn't even be set in a successor market proposal? Although perhaps there will be different unchangeable fields depending on the market type in future.

protocol/0028-GOVE-governance.md
```diff
@@ -213,8 +213,19 @@ Note the following key points from the market lifecycle spec:
...
+The parent market must be in one of `active` or `suspended` or `trading terminated` states. 
+If the parent market is `settled` or `proposed` or `pending` or `cancelled` then the proposal should be rejected at validation stage with an error "parent market cannot be in * state" with * being one of the dis-allowed states above. 
+The point of setting up a market to be successor of existing market is to 
+a) allow LPs who wish to continue claim their equity-like-share (ELS) by commiting liquidity to the successor market during the pending period and
+b) allow the successor market to inherit the insurance pool of the parent market, once the parent market moves from "trading terminated" to "settled" state. 
+
+Note that each market can have exactly one market as a *sucessor* market. 
+- if there already is a market (possibly pending i.e. in opening auction, see [lifecycle spec](./0043-MKTL-market_lifecycle.md)) naming a parent market which is referenced in the proposal then the proposal is rejected with error parent market no longer available.
```
TomMcL: Somewhere we probably want the ability to amend a market to remove a child successor (which would also immediately terminate said child I expect)

protocol/0043-MKTL-market_lifecycle.md
davidsiska-vega: I thought I'd want to allow to have the successor and parent running at the same time. For example CDS index markets do overlap... 

What you suggest would also make sense - would imply no overlap... so it's a decision. @barnabee thoughts?

protocol/0028-GOVE-governance.md
davidsiska-vega: Yes, that's a good point, if the successor settles or gets cancelled maybe we want to free the parent to have another successor.

protocol/0043-MKTL-market_lifecycle.md
TomMcL: Running at the same time would also make sense, but for things like the insurance pool it would have to be shared for a while etc which might be a more complex setup than just shifting.

And CDS Indices intertwine terrifyingly indeed 😉.

Though that does bring up a thought. Assuming we do want market parent/successors overlapping in existence, do we want an ability to set said inheritance of liquidity/insurance pool to happen prior to settlement/termination? Settlement works fine for a rolling weekly etc but having that move occur at a more customisable point would allow futures to roll from on-the-run to off-the-run with liquidity etc moving to the new one whilst the old one still exists and settles at some later date.

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -247,3 +250,11 @@ When the time defined by `market.liquidity.providers.fee.distributionTimeStep` e
 ### Distribution 
 
 - [ ] If `market.liquidity.providers.fee.distributionTimeStep > 0` and an LP submits a new liquidity commitment halfway through the distribution step then they receive roughly 1/2 the fee income compared with the next epoch when they maintain their commitment and that sees the same trade value. (<a name="0042-LIQF-018" href="#0042-LIQF-018">0042-LIQF-018</a>)  
+
+
+### Sucessor market
+- [ ] If an LP has ELS of `0.5` and stake of `10000` on a parent marketId=`m1` and a new market is proposed and enacted as `m2` with `m1` as its parent market and the LP submits a commitment of `10000` to `m2` during the "Pending" period, see [lifecycle](./0043-MKTL-market_lifecycle.md) then for the duration of the first `market.value.windowLength` after the opening auction ends the LP has ELS of `0.5` and stake of `10000` on `m2`. (<a name="0042-LIQF-019" href="#0042-LIQF-019">0042-LIQF-019</a>)
```
witgaw: It's not really `ELS`, but `virtual stake` that's meant to be the same on `m2`, right? I wonder if we should expose it too on the API.

protocol/0028-GOVE-governance.md
davidsiska-vega: Agree. I'll update.

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
davidsiska-vega: I agree we should be exposing virtual stake on the APIs, probably for a different spec update though?

witgaw: Definitely, but probably worth rephrasing it to "(...) after the opening auction ends the LP has **virtual stake** of **`xxx`** and stake of `10000` on `m2`." in this one.

davidsiska-vega: TODO: david

protocol/0043-MKTL-market_lifecycle.md
davidsiska-vega: We allow overlaps.

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
davidsiska-vega: Done.

protocol/0028-GOVE-governance.md
davidsiska-vega: Done.

wordlist.txt
```diff
@@ -149,6 +149,8 @@ mainnet
 malus
 margined
 margining
+marketId
+marketID
```
TomMcL: Do we want to add both of these or take a stance on which one to use? 😄

protocol/0001-MKTF-market_framework.md
```diff
@@ -216,9 +216,24 @@ Market {
...
 ```
 
+## Successor market
+
+If a market proposal, see [governance](./0028-GOVE-governance.md), designates an existing market as a *parent market* then it must have the same *product*, *settlement asset(s)* and *margin asset(s)*.
+It may propose new risk model and parameters, price monitoring parameters, position and market decimal places and oracles, both for trading terminated and for settlement data.
+Each market can have exactly one market as a *successor* market.
+
+1. if there already is a market (possibly pending i.e. in opening auction, see [lifecycle spec](./0043-MKTL-market_lifecycle.md)) naming a parent market which is referenced in the proposal then the proposal is rejected.
+1. if there are two proposals naming the same parent market then whichever one gets into the pending state first (i.e. passes governance vote) becomes the successor of the named parent; the other proposal is cancelled with reason "parent market not available".
```
candida-d, line 226:
```suggestion
It may propose a new risk model and parameters, price monitoring parameters, position and market decimal places and oracles, both for trading terminated and for settlement data.
Each market can have exactly one market as a *successor* market.

1. if there already is a market (possibly pending, i.e. in opening auction, see [lifecycle spec](./0043-MKTL-market_lifecycle.md)), naming a parent market, then a subsequent proposal referencing that market is rejected.
1. if there are two proposals naming the same parent market then whichever one gets into the pending state first (i.e. passes governance vote) becomes the successor of the named parent; the other proposal is cancelled with reason "parent market not available".
```

candida-d: "may propose a new risk model and parameters" - Might this be better phrased "may propose new risk model parameters"?

protocol/0001-MKTF-market_framework.md
```diff
@@ -216,9 +216,24 @@ Market {
...
+
+Some markets can be labelled as *perpetual* at the time of proposal; in this case they cannot serve as a parent market.
```
candida-d: And would a market proposal that chooses a perpetual market as its parent then be rejected?

protocol/0028-GOVE-governance.md
```diff
@@ -215,10 +215,23 @@ Note the following key points from the market lifecycle spec:
...
 - A market usually enters Active status at the proposal's enactment date/time, but some conditions may delay this or cause the market to be Cancelled instead
 
-A proposal to create a market contains:
+A proposal to create a market contains
 
-1. a complete market specification as per the Market Framework (see spec) that describes the market to be created.
+1. a complete market specification as per the [Market Framework](./0001-MKTF-market_framework.md) that describes the market to be created.
 1. an enactment time that is at least the _minimum auction duration_ after the vote closing time (see [auction spec](./0026-AUCT-auctions.md))
+1. if the market is meant to be a _successor_ of a given market then it contains the marketId of the market it's succeeding (parent market), a parameter called `insurancePoolFraction` which is a decimal in `[0,1]` (i.e. it can be `0` or `1` or anything in between) and certain entries in the market proposal must be identical to those of the market it's succeeding OR the proposal should simply not contain the fields that cannot be changed.
+In particular instrument, settlement asset, margin asset, and `market.value.windowLength` must match.
```
candida-d:
```suggestion
In a particular instrument, the settlement asset, margin asset, and `market.value.windowLength` must match.
```

protocol/0028-GOVE-governance.md
```diff
@@ -215,10 +215,23 @@ Note the following key points from the market lifecycle spec:
...
+The parent market must be in one of `active` or `suspended` or `trading terminated` states.
+If the parent market is `settled` or `proposed` or `pending` or `cancelled` then the proposal should be rejected at validation stage with an error "parent market cannot be in `*` state" with `*` being one of the dis-allowed states above.
+The point of setting up a market to be successor of existing market is to
+a) allow LPs who wish to continue claim their equity-like-share (ELS) by committing liquidity to the successor market during the pending period and
+b) allow the successor market to inherit the insurance pool of the parent market. When the successor market leaves the opening auction (moves from pending to active) the amount equal to `insurancePoolFraction x parent market insurance pool balance` is transferred to the successor market insurance pool. Once the parent market moves from "trading terminated" to "settled" state the entire remaining insurance pool of the successor market is transferred to the successor market insurance pool.
```
candida-d:
```suggestion
The parent market must be in one of `active` or `suspended` or `trading terminated` states.
If the parent market is `settled` or `proposed` or `pending` or `cancelled` then the proposal should be rejected at the validation stage with an error "parent market cannot be in `*` state" with `*` being one of the dis-allowed states above.
The point of setting up a market to be successor of an existing market is to
a) allow LPs continue claim their equity-like-share (ELS) by committing liquidity to the successor market during the pending period if they wish to, and
b) allow the successor market to inherit the insurance pool of the parent market. When the successor market leaves the opening auction (moves from pending to active) the amount equal to `insurancePoolFraction x parent market insurance pool balance` is transferred to the successor market insurance pool. Once the parent market moves from "trading terminated" to "settled" state, the entire remaining insurance pool of the successor market is transferred to the successor market insurance pool.
```

protocol/0028-GOVE-governance.md
```diff
@@ -215,10 +215,23 @@ Note the following key points from the market lifecycle spec:
...
+
+Note that each market can have exactly one market as a _successor_ market.
+
+- if there already is a market (possibly pending i.e. in opening auction, see [lifecycle spec](./0043-MKTL-market_lifecycle.md)) naming a parent market which is referenced in the proposal then the proposal is rejected with error parent market no longer available.
+- if there are two proposals naming the same parent market then whichever one gets into the pending state first (i.e. passes governance vote) becomes the successor of the named parent; the other proposal is cancelled with reason "parent market no longer available".
+- if there is a successor market naming a parent market that terminates and settles or is cancelled by governance before the parent market (for whatever reason) then the parent market successor is set to empty and a different successor can be proposed.
```
candida-d:
```suggestion
- if there already is a market (possibly pending, i.e. in opening auction, see [lifecycle spec](./0043-MKTL-market_lifecycle.md)), naming a parent market, then a subsequent proposal referencing that market is rejected.
- if there are two proposals naming the same parent market then whichever one gets into the pending state first (i.e. passes governance vote) becomes the successor of the named parent; the other proposal is cancelled with reason "parent market no longer available".
- if there is a successor market naming a parent market that terminates and settles or is cancelled by governance before the parent market (for whatever reason) then the parent market successor is set to empty and a different successor can be proposed.
```
can the updated successor market be proposed using an "update market" proposal? Or is it some other action? Does that make that field conditionally immutable?

protocol/0043-MKTL-market_lifecycle.md
```diff
@@ -81,9 +81,9 @@ When a Market Proposal is not successful, see [governance proposal](./0028-GOVE-
 
 ### Pending
 
-When a Market Proposal is successful at the end of the voting period, the Market state becomes "Pending". Currently a Pending Market is always in an [auction call period](./0026-AUCT-auctions.md) that ends at the enactment date as specified in the Market Proposal.
+When a Market Proposal is successful at the end of the voting period, the Market state becomes "Pending". Currently a Pending Market is always in an [auction call period](./0026-AUCT-auctions.md) that ends at the enactment date as specified in the Market Proposal if the other conditions for exiting auction period are met (liquidity committed, best static bid / ask present). If, initially,  at the specified enactment date it was not possible to leave the auction period, the auction will conclude as soon as possible once the conditions for auction exit are met.
```
candida-d:
```suggestion
When a Market Proposal is successful at the end of the voting period, the Market state becomes "Pending". Currently a Pending Market is always in an [auction call period](./0026-AUCT-auctions.md) that ends at the enactment date as specified in the Market Proposal if the other conditions for exiting auction period are met (liquidity committed, best static bid / ask present). If, initially, at the specified enactment date it was not possible to leave the auction period, the auction will conclude as soon as possible once the conditions for auction exit are met.
```

protocol/0001-MKTF-market_framework.md
davidsiska-vega: I am not sure; the meaning I am trying to convey is that the parameters may be new but also the risk model can be different (and if that happens new parameters are needed because different risk models have different parameter sets).

protocol/0001-MKTF-market_framework.md
davidsiska-vega: Yes.

witgaw:
```suggestion
Some markets can be labelled as *perpetual* (e.g. perpetual futures or spot markets) at the time of proposal; in this case they cannot serve as a parent market.
```

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -70,25 +70,32 @@ The guiding principle of this section is that by committing stake a liquidity pr
 The parameter which determines the period over which market value and hence growth is `market.value.windowLength` which could be e.g. a week.
 From the end of the opening auction, which we will refer to as `t0` until `t0+market.value.windowLength` is the `0th` or "bootstrap period". Then from `t0+market.value.windowLength` until `t0 + 2 x market.value.windowLength` is the `1st` period and so on.
 For each LP we track the stake they have and also their virtual stake.
-Before and during the 0th (bootstrap) any stake commitment or removal is mirrored in the virtual stake.
+For markets that have no "parent" market, see [governance](./0028-GOVE-governance.md)  we postulate that before and during the 0th (bootstrap) any stake commitment or removal is mirrored in the virtual stake.
 
-For any period `n >= 1` LP can add stake or remove stake but virtual stake is treated differently:
+For any period `n >= 1` LP can add stake or remove stake but virtual stake is treated differently or for markets with "parent market":
```
davidsiska-vega:
```suggestion
For any period `n >= 1` LP can add stake or remove stake but virtual stake is treated differently for markets with "parent market":
```

wordlist.txt
davidsiska-vega: ID

protocol/0001-MKTF-market_framework.md
```diff
@@ -216,9 +216,24 @@ Market {
...
+## Successor market
+
+If a market proposal, see [governance](./0028-GOVE-governance.md), designates an existing market as a *parent market* then it must have the same *product*, *settlement asset(s)* and *margin asset(s)*.
+It may propose new risk model and parameters, price monitoring parameters, position and market decimal places and oracles, both for trading terminated and for settlement data.
+Each market can have exactly one market as a *successor* market.
+
+1. if there already is a market (possibly pending i.e. in opening auction, see [lifecycle spec](./0043-MKTL-market_lifecycle.md)) naming a parent market which is referenced in the proposal then the proposal is rejected.
+1. if there are two proposals naming the same parent market then whichever one gets into the pending state first (i.e. passes governance vote) becomes the successor of the named parent; the other proposal is cancelled with reason "parent market not available".
+
+Some markets can be labelled as *perpetual* (e.g. perpetual futures or spot markets) at the time of proposal; in this case they cannot serve as a parent market.
```
gordsport: Revisit.

protocol/0028-GOVE-governance.md
```diff
@@ -215,10 +215,23 @@ Note the following key points from the market lifecycle spec:
...
 
-A proposal to create a market contains:
+A proposal to create a market contains
 
-1. a complete market specification as per the Market Framework (see spec) that describes the market to be created.
+1. a complete market specification as per the [Market Framework](./0001-MKTF-market_framework.md) that describes the market to be created.
 1. an enactment time that is at least the _minimum auction duration_ after the vote closing time (see [auction spec](./0026-AUCT-auctions.md))
+1. if the market is meant to be a _successor_ of a given market then it contains the marketID of the market it's succeeding (parent market), a parameter called `insurancePoolFraction` which is a decimal in `[0,1]` (i.e. it can be `0` or `1` or anything in between) and certain entries in the market proposal must be identical to those of the market it's succeeding OR the proposal should simply not contain the fields that cannot be changed.
+In a particular instrument, the settlement asset, margin asset, and `market.value.windowLength` must match.
+The parent market must be in one of `active` or `suspended` or `trading terminated` states.
```
gordsport: Allow settled with `successor launch window`.

EVODelavega: Something else that needs clarifying is when a successor proposal is enacted, and the parent market is still in `active` (or `suspended` due to auction) state: we want to carry over the ELS, but I don't see any reason why it wouldn't be possible for a successor market to be in opening auction (or even leaving the opening auction) while its parent market is in active state. At which point, then, should we transfer the insurance pool fraction, and what should happen with the ELS?

Arguably, LP's would probably prefer it if their ELS transfers over to the successor market when they provide liquidity on that new market for a couple of reasons:

1. Being the first LP to provide stake on the successor market with ELS carry-over is advantageous if other LP's are still on the parent market, being the only LP with ELS from a previous market is a force multiplier.
2. Depending on the window length, this could allow you to keep an eye on `r` for the parent market and with a single transaction, bail if the parent market experiences negative growth, preserving ELS by moving it to a successor market in opening auction (ELS being "frozen" is preferable to ELS shrinking)
3. What if an LP provides liquidity on the successor market, removes stake from the parent, and then (either accidentally or on purpose) creates an LP on the parent market again? Is the ELS shared between parent and successor markets? Does an LP risk losing the ELS they built up over time by accidentally providing stake on the wrong market like this?
4. Based on what @davidsiska-vega said when we discussed this, the ELS is only preserved if the LP provides stake on the new market and _before_ removing stake from the parent market (or waits for their bond to be released once the parent market is settled). That's fine, but we have to be very careful and clear on what happens when stake is no longer provided on either the parent/successor market when both are up and running (e.g. ELS is carried over to the successor, but by providing stake on the parent market after committing to the successor, though a bizarre thing to do just feels incredibly risky - for the LP if we transfer ELS between markets, or very weird and risky WRT fairness if we duplicate/share ELS between parent and successor.

davidsiska-vega: >  I don't see any reason why it wouldn't be possible for a successor market to be in opening auction (or even leaving the opening auction) while its parent market is in active state. At which point, then, should we transfer the insurance pool fraction, and what should happen with the ELS?

1. the party keeps the virtual stake / ELS on the parent market and gets the same virtual stake (if they provided full stake / or virtual stake using the usual delta way).
2. the insurance pool fraction is transferred at that point

davidsiska-vega: @EVODelavega in fact the spec says that the virtual stake is *copied* not *transferred*. This is important.

> If the market has a "parent market" then each LP which commits liquidity (to this market) gets the virtual stake copied from the parent market as the 1st step of the process and the stake they are committing here minus the stake on parent market is treated as the `delta` here.

barnabee: As David says, virtual stake (ELS) is not transferred it is copied.

> very weird and risky WRT fairness if we duplicate/share ELS between parent and successor.
I am not sure what is odd here. Once successor market leaves opening auction, ELS should be _copied_ to the new market, once. It is unchanged on the parent market. 

Where is the fairness issue you see @EVODelavega ?

protocol/0001-MKTF-market_framework.md
davidsiska-vega: We allow perps and spot to also serve as parent as this will allow us to "reconfigure" those.

davidsiska-vega:
```suggestion
```

protocol/0028-GOVE-governance.md
davidsiska-vega:
```suggestion
The parent market must be in one of `active` or `suspended` or `trading terminated` states or `settled` but with time since settlement less than or equal `network.liquidity.successorLaunchWindowLength`.
```

protocol/0028-GOVE-governance.md
```diff
@@ -215,10 +215,23 @@ Note the following key points from the market lifecycle spec:
...
+A proposal to create a market contains
 
-1. a complete market specification as per the Market Framework (see spec) that describes the market to be created.
+1. a complete market specification as per the [Market Framework](./0001-MKTF-market_framework.md) that describes the market to be created.
 1. an enactment time that is at least the _minimum auction duration_ after the vote closing time (see [auction spec](./0026-AUCT-auctions.md))
+1. if the market is meant to be a _successor_ of a given market then it contains the marketID of the market it's succeeding (parent market), a parameter called `insurancePoolFraction` which is a decimal in `[0,1]` (i.e. it can be `0` or `1` or anything in between) and certain entries in the market proposal must be identical to those of the market it's succeeding OR the proposal should simply not contain the fields that cannot be changed.
+In a particular instrument, the settlement asset, margin asset, and `market.value.windowLength` must match.
+The parent market must be in one of `active` or `suspended` or `trading terminated` states or `settled` but with time since settlement less than or equal `network.liquidity.successorLaunchWindowLength`.
```
davidsiska-vega:
```suggestion
The parent market must be in one of `active` or `suspended` or `trading terminated` states or `settled` but with time since settlement less than or equal `network.liquidity.successorLaunchWindowLength`. In particular the market cannot be `pending`.
```

protocol/0028-GOVE-governance.md
```diff
@@ -215,10 +215,23 @@ Note the following key points from the market lifecycle spec:
...
+1. a complete market specification as per the [Market Framework](./0001-MKTF-market_framework.md) that describes the market to be created.
 1. an enactment time that is at least the _minimum auction duration_ after the vote closing time (see [auction spec](./0026-AUCT-auctions.md))
+1. if the market is meant to be a _successor_ of a given market then it contains the marketID of the market it's succeeding (parent market), a parameter called `insurancePoolFraction` which is a decimal in `[0,1]` (i.e. it can be `0` or `1` or anything in between) and certain entries in the market proposal must be identical to those of the market it's succeeding OR the proposal should simply not contain the fields that cannot be changed.
+In a particular instrument, the settlement asset, margin asset, and `market.value.windowLength` must match.
+The parent market must be in one of `active` or `suspended` or `trading terminated` states or `settled` but with time since settlement less than or equal `network.liquidity.successorLaunchWindowLength`. In particular the market cannot be `pending`.
```
davidsiska-vega:
```suggestion
The parent market must be in one of `active` or `suspended` or `trading terminated` states or `settled` but with time since settlement less than or equal `network.liquidity.successorLaunchWindowLength` or `closed by governance` but with the closing time less than or equal `network.liquidity.successorLaunchWindowLength`. 
In particular the market cannot be `pending`.
```

gordsport (dismissed):
approved on the call.

New PR with more ACs to come

gordsport (approved):
Approving as per the discussion in the protocol design meeting 17th May

PR #1504, opened 2023-01-09
Perps spec update

### LINKED ISSUE #868: Review/Update spec for new product 'Perpetuals'

Is this covered in: https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0053-PERP-product_builtin_perpetual_future.md

Details of the product type Perpetuals to enable the expected work related to the prioritisation meeting held on the [Notion Board](https://www.notion.so/vegaprotocol/New-product-Perpetuals-097a5735bc1147f98fad0d891ff2030e

 - [ ] The spec has been updated to indicate what are the requirements for new product 'Perpetuals'
 - [ ] The specification contains acceptance criteria with IDs that reflect the scope of the work required for Q2 2022

witgaw:
Closes https://github.com/vegaprotocol/specs/issues/868


protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -4,12 +4,16 @@ This built-in product provides perpetual futures that are cash-settled, i.e. the
...
 
 ## 1. Product parameters
 
-1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows. The receipt of this data triggers this calculation and the transfers between parties to "true up" to the external reference price.
 1. `settlement_asset (Settlement Asset)`: this is used to specify the single asset that an instrument using this product settles in.
+1. `settlement_cue (Data Source)`: this data is used to indicate that next periodic settlement should happen imminently.
+1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows. The receipt of this data triggers this calculation and the transfers between parties to "true up" to the external reference price.
+1. `settlement_cue_auction_duration`: value expressed in seconds which specifies the duration of an auction started once settlement cue is received. The auction ends when the specified time elapses or when the settlement data is received. A value of `0` indicates no auction.
```
davidsiska-vega: I would insist on seconds, simply a time interval e.g. "60s" or "10min".

witgaw: Good point, thanks, changed now.

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -4,12 +4,16 @@ This built-in product provides perpetual futures that are cash-settled, i.e. the
...
+1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows. The receipt of this data triggers this calculation and the transfers between parties to "true up" to the external reference price.
+1. `settlement_cue_auction_duration`: a time interval which specifies the duration of an auction started once settlement cue is received. The auction ends when the specified time elapses or when the settlement data is received. A value of `0s` indicates no auction.
+1. `max_settlement_gap`: a time interval which specifies the amount of time without periodic settlement after which the market will go into protective auction and remain in that mode until settlement data is received.
+1. `settlement_price_monitoring`: a boolean flag indicating if periodic settlement price should go through the [price monitoring](0032-PRIM-price_monitoring.md) logic. If set to `true` any valid `settlement_data` ingested by the market will go through the price monitoring engine and contribute to its price history as well as trigger a price monitoring auction if it falls outside the current valid price bounds.
```
davidsiska-vega: So if the settlement value comes outside bounds then what happens to mark-to-market? Let's say bounds are [80,120] I get settlement data with 125 and the `settlement_price_monitoring = true`. Now we go into price monitoring auction if:
1) price monitoring resolves at say 124 or 126 - this value is used for mark to market. 
2) price monitoring resolves at say 100 - this value is used for mark to market (I guess). 
So then the settlement data ends up effectively not contributing to settlement, so if the time we've not settled goes past `max_settlement_gap` we go into an auction, correct?

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -4,12 +4,16 @@ This built-in product provides perpetual futures that are cash-settled, i.e. the
...
 1. `settlement_asset (Settlement Asset)`: this is used to specify the single asset that an instrument using this product settles in.
+1. `settlement_cue (Data Source)`: this data is used to indicate that next periodic settlement should happen imminently.
+1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows. The receipt of this data triggers this calculation and the transfers between parties to "true up" to the external reference price.
+1. `settlement_cue_auction_duration`: a time interval which specifies the duration of an auction started once settlement cue is received. The auction ends when the specified time elapses or when the settlement data is received. A value of `0s` indicates no auction.
+1. `max_settlement_gap`: a time interval which specifies the amount of time without periodic settlement after which the market will go into protective auction and remain in that mode until settlement data is received.
```
TomMcL: Do we also want/need a `min_settlement_gap`? (Might be somewhat confusing if you're expecting settlement once a day and someone sends a settlement half way through. But perhaps that's not an issue?)

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: That depends on how we resolve the TODO at line 55. 

I think what we'd like to end up with is:
1. if settlement data is very different from previous settlement prices or market prices we want to discard it (I think it makes sense to not use at all if we don't use it immediately (otherwise we're deliberately making the data stale).
1. if it's persistently different (a few consecutive ones are similar) we want to eventually accept it.
1. if it goes all over the place then some form of manual intervention wouldn't be terrible.

I think price monitoring bounds + max_settlement_gap that's not larger than say 2-3 settlement anniversaries gets us mostly there. 

An open question is if we want to use the discarded data in history, but it'd just ignore it not to create a special case in the engine and assume that if that's indeed a sensible price point then auction will resolve not far from it).

If oracle supplies consecutive datapoints that are all over the place, then it probably makes sense to either keep market in auction for longer or change the oracle (including a change to the internal oracle that just supplies a value that's been agreed on via consensus).

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: I was thinking the `settlement_data` oracle will handle that (we still need to spec it of course), but it could have a mechanism that fires up the data on same dates as `settlement_cue` subject to data meeting some criteria. 

`settlement_cue` will probably just be an internal oracle that fires with a specified frequency in most cases so in a sense it will already contain the min gap.

protocol/0053-PERP-product_builtin_perpetual_future.md
TomMcL: It still feels somewhat wrong for the market to have the ability to largely ignore a settlement price given that those settlements are the only things giving the trading value 😄. But I guess not the end of the world as long as it's settling frequently

protocol/0053-PERP-product_builtin_perpetual_future.md
TomMcL: Ah yes makes sense :+1:

protocol/0048-DSRI-data_source_internal.md
```diff
@@ -51,6 +53,32 @@ on: {
...
+	{
+		timestamp: '202112311T12:00:00'
+		frequency: '24h'
+	},
+	{
+		timestamp: '202112311T20:00:00'
+		frequency: '24h'
+	},
+}
+
```
witgaw, line 59: TODO: 
- Add ACs.
- Be explicit about "frequencies" we allow.
- Allow `0s` frequency?
- Can be less explicit with pseudcode and let core decide.
- Limit number of entires (netparam `(...).spam.(...)`)

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -1,15 +1,20 @@
...
 
 ## 1. Product parameters
 
-1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows. The receipt of this data triggers this calculation and the transfers between parties to "true up" to the external reference price.
 1. `settlement_asset (Settlement Asset)`: this is used to specify the single asset that an instrument using this product settles in.
+1. `settlement_cue (Data Source)`: this data is used to indicate that next periodic settlement should happen imminently.
+1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows. The receipt of this data triggers this calculation and the transfers between parties to "true up" to the external reference price.
```
witgaw: TODO: Specify how we want to filter the data so that we only consider a subset of it (based on filter).

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -1,15 +1,20 @@
...
+1. `data_ingestion_period`: specifies the length of time window since `settlement_cue` event during which data from `settlement_data` data source will be accepted by the market. Once the first value is received no further data is accepted.
+1. `max_settlement_gap`: a time interval which specifies the amount of time without periodic settlement after which the market will go into protective auction and remain in that mode until settlement data is received.
+1. `settlement_data_monitoring`: if set to `true` any valid `settlement_data` ingested by the market will be checked against the market's [price monitoring](0032-PRIM-price_monitoring.md) engine. Specifically, the incoming settlement data will be checked against market's active [price monitoring bounds](0021-MDAT-market_data_spec.md#market-data-fields), if it falls within the market the periodic settlement proceeds, otherwise the system behaves as if the data was never received.
```
witgaw: TODO: keep waiting for value within PM range for the duration of the original `data_ingestion_period`.

protocol/0048-DSRI-data_source_internal.md
witgaw: Made the repeating internal time triggered oracle bit more general for now, we can make it more prescriptive after discussing it with engineering.

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: Done: https://github.com/vegaprotocol/specs/pull/1504/files#diff-9c47979319e99ba93f2f121a920be94964b074266e01196f107414ce2121187dR17

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: Added pseudocode: 
https://github.com/vegaprotocol/specs/pull/1504/files#diff-9c47979319e99ba93f2f121a920be94964b074266e01196f107414ce2121187dR25-R46

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -1,18 +1,48 @@
...
+            internal_time_oracle:
+                repeating:
+                    - every 24h from 20230201T09:30:00
+                    - every 168h from 20230203T12:00:00
+       settlement_data:
+            data_source: SignedMessage{ pubkey=0xA45e...d6 }
+            field: 'price'
+            filters: 
+                - received: >= 'settlement_cue.time'
+                - received: <= 'settlement_cue.time' + "35min"
```
witgaw: This may need to go into `0047-DSRF-data_source_filter.md` as an extra key allowed for any oracle, but let's maybe first agree here that we want it at all.

witgaw (commented):
Ready for review.

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -1,18 +1,51 @@
...
 
 ## 1. Product parameters
 
-1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows. The receipt of this data triggers this calculation and the transfers between parties to "true up" to the external reference price.
 1. `settlement_asset (Settlement Asset)`: this is used to specify the single asset that an instrument using this product settles in.
+1. `settlement_schedule (Data Source: datetime)`: this data is used to indicate when the next periodic settlement should be carried out.
+1. `settlement_data_cue (Data Source: datetime)`: this data is used to indicate the earliest time at which the next `settlement_data` should be expected.
```
witgaw: Do we want to make the settlement cue optional and just ingest all the oracle data if it's not specified

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -28,23 +61,84 @@ cash_settled_perpetual_future.value(quote) {
...
+    d := data_points[i].x - data_points[i].y
+    sd += d * t
+    st += t
 }
+funding_rate = sd / st
 ```
 
+All the funding rate data points except for the last one (it should get carried over as the first data point for the next period) can then be deleted.
+
+Last step is to calculate each party's cash flows as $-\text{open volume} * \text{funding rate}$ where cashflows are first collected from parties that are making the payment (negative value of the cashflow, i.e. longs when the funding rate is positive) and distributed to those receiving it. Any shortfall should be made-up from the insurance pool and if that's not possible loss socialisation should be applied (exactly as per mark-to-market settlement methodology).
```
witgaw: Do we want to use the insurance pool for funding rate cashflow shortfalls? Entire amount or a fraction to discourage oracle manipulation?

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -28,23 +61,84 @@ cash_settled_perpetual_future.value(quote) {
...
+
+### 4.4.1. Periodic settlement during [auction](0026-AUCT-auctions.md)
+
+Periodic settlement is not allowed during the opening auction and it's extensions.
+If periodic settlement data happens whilst market is in auction of any other type then periodic settlement should be carried out as per above methodology and the market should remain in auction until it's allowed to move back into market's default trading mode.
```
witgaw, line 89: Is that ok or do we want to allow uncrossing (without leaving the auction) at the current (at point of periodic settlement) price before exchanging the funding rate payments?

witgaw (commented):
A few points to discuss

protocol/0053-PERP-product_builtin_perpetual_future.md
davidsiska-vega: Optional would work I think. As long as we can record both the value and the time for the incoming price stream so that it can be used to calculate the funding payments.

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -28,23 +61,84 @@ cash_settled_perpetual_future.value(quote) {
...
+When the `settlement_schedule` event is received the latest funding rate data point gets repeated with the timestamp set to the current value of `vegaprotocol.builtin.timestamp`.
+
+The next step is to calculate the periodic settlement funding rate. If there are no periodic settlement data points then the periodic settlement is skipped. Otherwise, consider all the periodic settlement data points and calculate the time-weighted average price difference as:
+
+```go
+sd := 0
+st := 0
+for i := 0; i < len(data_points) - 1; i++ {
+    t := data_points[i+1].t-data_points[i].t
+    d := data_points[i].x - data_points[i].y
```
davidsiska-vega:
```suggestion
    d := data_points[i].x - data_points[i].y     // recall that x stands for mark price and y for the external price source input
```

protocol/0053-PERP-product_builtin_perpetual_future.md
davidsiska-vega: I'd say we just make the entire insurance pool available. The data sourcing (in the future) should deal with making the data valid; on the level of the product we shouldn't care.

protocol/0053-PERP-product_builtin_perpetual_future.md
davidsiska-vega: I think that's ok.

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -28,23 +61,84 @@ cash_settled_perpetual_future.value(quote) {
...
+All the funding rate data points except for the last one (it should get carried over as the first data point for the next period) can then be deleted.
+
+Last step is to calculate each party's cash flows as $-\text{open volume} * \text{funding rate}$ where cashflows are first collected from parties that are making the payment (negative value of the cashflow, i.e. longs when the funding rate is positive) and distributed to those receiving it. Any shortfall should be made-up from the insurance pool and if that's not possible loss socialisation should be applied (exactly as per mark-to-market settlement methodology).
+
+### 4.4.1. Periodic settlement during [auction](0026-AUCT-auctions.md)
+
+Periodic settlement is not allowed during the opening auction and it's extensions.
+If periodic settlement data happens whilst market is in auction of any other type then periodic settlement should be carried out as per above methodology and the market should remain in auction until it's allowed to move back into market's default trading mode.
+
+### 4.5. Protective auctions
```
davidsiska-vega: Discussed with @barnabee and we're thinking that if funding inputs aren't coming we just don't care; 
- the funding rate payments don't happen; you can exit your position if you don't like it
- if this persist we should be able to suspend by market governance
- price monitoring should protect from completely crazy price moves.

davidsiska-vega:
We must remember the interest rate parameter that's needed - presumably a market parameter.

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: Made it optional, but perhaps it should just be removed altogether now?

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: Removed that section.

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: Removed now.

wwestgarth:
A few things that have popped to mind while thinking about perps:
- Successor markets: I'm assuming they make no sense for perps? Or is succeeding a fixed-expiry futures markets with a perp something reasonable?
- Update markets: theres an AC about updating the data-sources, should we also be allows to update `margin_funding_factor`?
- Any special handing during checkpoint restores wrt to data-points from pre-checkpoint restore?

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -28,23 +53,71 @@ cash_settled_perpetual_future.value(quote) {
...
+
+### 4.3.1. Periodic settlement during [auction](0026-AUCT-auctions.md)
+
+Periodic settlement is not allowed during the opening auction and it's extensions.
+If periodic settlement data happens whilst market is in auction of any other type then periodic settlement should be carried out as per above methodology and the market should remain in auction until it's allowed to move back into market's default trading mode.
+
+### 5. Margin considerations
+
+To assure adequate solvency we need to include the estimate of the upcoming funding payment in maintenance margin estimate for the party. Let $t_{k-1}$ be the time of the last funding payment. Let $t$ be current time ($t < t_k$).
+Calculate $G_t$ as the [funding payment](#44-periodic-settlement) between $t_k$ and $t$.
```
wwestgarth: Is this suppose to be `between $t_{k-1}$ and $t$`?  If `t_k` is in the future I don't know how we can know its value?

davidsiska-vega (dismissed):

davidsiska-vega:
@wwestgarth 
> Successor markets: I'm assuming they make no sense for perps? Or is succeeding a fixed-expiry futures markets with a perp something reasonable?

I think we want the whole successor market stuff to work with perps (the way we want it to work with spot). The classical workflow would be: 1) governance proposal to cancel a perps market. 2) A new (successor) perps market which changes market decimals / position decimals (and possible other things that a mere market change proposal doesn't allow). 

> Update markets: theres an AC about updating the data-sources, should we also be allows to update `margin_funding_factor`?

Yes, we want to allow updates to this in market updates. 

> Any special handing during checkpoint restores wrt to data-points from pre-checkpoint restore?

Is this LNL checkpoint? I don't think any special handling is needed; no one will have any positions / orders. @witgaw ?

protocol/0053-PERP-product_builtin_perpetual_future.md
davidsiska-vega:
```suggestion
Calculate $G_t$ as the [funding payment](#44-periodic-settlement) between $t_{k-1}$ and $t$.
```

davidsiska-vega:
> Or is succeeding a fixed-expiry futures markets with a perp something reasonable?

In successor market spec we say "The product, settlement asset, and margin asset must match" which I think is the saner setup. So the answer to this is a no. I think allowing successor markets to change the product is a bit of a can of worms so I'd rather not go there.

witgaw:
> @wwestgarth
> 
> > Successor markets: I'm assuming they make no sense for perps? Or is succeeding a fixed-expiry futures markets with a perp something reasonable?
> 
> I think we want the whole successor market stuff to work with perps (the way we want it to work with spot). The classical workflow would be: 1) governance proposal to cancel a perps market. 2) A new (successor) perps market which changes market decimals / position decimals (and possible other things that a mere market change proposal doesn't allow).
> 
> > Update markets: theres an AC about updating the data-sources, should we also be allows to update `margin_funding_factor`?
> 
> Yes, we want to allow updates to this in market updates.
> 
> > Any special handing during checkpoint restores wrt to data-points from pre-checkpoint restore?
> 
> Is this LNL checkpoint? I don't think any special handling is needed; no one will have any positions / orders. @witgaw ?

Yes, if positions disappear then we don't need to worry about storing data required for funding rate payment calculation.

protocol/0081-SUCM-successor_markets.md
```diff
@@ -67,6 +67,8 @@ Market proposal may specify parent market ID. If it does then:
 
 It is possibly to cancel a [spot market](./0080-SPOT-product_builtin_spot.md) via governance and propose a new spot market as a successor with different `market_decimal_places` and `position_decimal_places` (aka `size_decimal_places` for spot); the LPs virtual stakes are carried over (<a name="0081-SUCM-004" href="#0081-SUCM-004">0081-SUCM-004</a>).
 
+It is possibly to cancel a [perpetual futures](./0053-PERP-product_builtin_perpetual_future.md) market via governance and propose a new perpetual futures market as a successor with different `market_decimal_places` and `position_decimal_places` (aka `size_decimal_places` for spot); the LPs virtual stakes are carried over (<a name="0081-SUCM-015" href="#0081-SUCM-015">0081-SUCM-015</a>).
```
davidsiska-vega:
```suggestion
It is possibly to cancel a [perpetual futures](./0053-PERP-product_builtin_perpetual_future.md) market via governance and propose a new perpetual futures market as a successor of the aforementioned cancelled / to be cancelled with different `market_decimal_places` and `position_decimal_places`; the LPs virtual stakes are carried over (<a name="0081-SUCM-015" href="#0081-SUCM-015">0081-SUCM-015</a>).
```

davidsiska-vega:
```suggestion
It is possibly to cancel a [perpetual futures](./0053-PERP-product_builtin_perpetual_future.md) market via governance and propose a new perpetual futures market as a successor of the aforementioned cancelled / to be cancelled market with different `market_decimal_places` and `position_decimal_places`; the LPs virtual stakes are carried over (<a name="0081-SUCM-015" href="#0081-SUCM-015">0081-SUCM-015</a>).
```

davidsiska-vega (approved):

PR #1505, opened 2023-01-09
Account abstraction (WIP)

### LINKED ISSUE #866: [SPEC ONLY]: Gather requirements for fund accounts
In order to be ready for Comsic Elevator development we need to agree upon the scope for the engineering team to deliver in this release.

[Notion Board](https://www.notion.so/vegaprotocol/Vega-mainnet-has-fund-accounts-Fund-accounts-provides-semi-passive-LP-following-liquidity-provisio-c01a87eea6e94cba8c54395d692771f0)

Market making and trading in general require a high level of expertise which means that only a small subset of potential users can engage with these activities. Finance solves for this with the existence of funds (of which there are many types) that allow someone with capital (usually after paying a fee/price) deposit that capital with someone else who will trade on their behalf. This has the benefit of allow more and different types of participants in the marketplace and also of increasing the amount of capital available, for example, to provide liquidity. This increased capital and liquidity benefits all participants through lower spread, reduced slippage, and more viable products/markets.

In addition to people/entities doing this "off-chain" with traditional structures and risks, the blockchain enables funds to be created that are non-custodial, not give their operators direct access to funds, and that enforce certain rules and transparency around trading activities. Vega intends to support such a feature to enable on-chain fund creation.

Any MVP of this feature must enable deplyoment of funds for liquidity provision on Vega, including the fund account making and keeping to an LP commitment. This will enable participants in the network to deploy liquidity to markets without trading expertise. 

- [ ] A fund account spec has been written
- [ ] The specification contains acceptance criteria with IDs that reflect the scope of work required for Cosmic Elevator
- [ ] The QA team has reviewed and understands the specification and approves or has added to the set of ACs

barnabee:
Summary of account abstraction feature for discussion and design and spec work.

tl;dr:

1. Add multisig support with:
    - ability to update signer account IDs
    - ability to change signing threshold
    - ability to separate controller account ID (could be the same or a diff. multisig or single user) and signer accounts
    - multisig account ID can be used anywhere (?) a normal pubkey can be
2. Add fund accounts with:
   - separation of "investor", "controller", and "fund manager/trader" roles (account ID, each can be a multisig, at least investors can be other fund accounts, though with a general abstraction all 3 could be)
   - tracking of users' share of fund NAV
   - optional rules for how long it takes to remove funds
   - ability to apply a "2 and 20" style fee (fraction of NAV per period X + fraction of gains net of losses from prior periods per period X) to be sent to any account ID of fund creator's choosing (can be multisig or another fund)
   - ability to restrict "investors" to a limited set of account IDs or leave open to all
   - ability to cap total "invested" or close to new money coming in
   - ability to restrict the fund to a subset of assets/markets/funds
   - ability to restrict the fund to a subset of trader/LP/fund investor roles
   - fund account ID can be used anywhere (?) a normal pubkey can be as long as not restricted by fund rules
   - delay for changing fund rules >= delay for redemption? 
3. Ability to create a "meta" account that allows other accounts do a a rules based subset of activities
    - rules control other accounts ability to act "as" the meta account for any normal tx type
    - can therefore subdivide one "meta" account to diffrent accounts for each type of rule set / responsibility
    - can restrict to specific tx types, markets, assets, etc.

### DIFF
--- a/0000-account-abstraction.md
+++ b/0000-account-abstraction.md
@@ -1,0 +1,246 @@
+# Account abstraction
+
+Note: this is a WIP document outlining the features and requirements, not a spec, at this point.
+
+
+## Background
+
+Vega has the concept of a "party" which is an entity within the system (unrelated to the concept of legal entities) that can to change the state of the system.
+
+Prior to implementation of this feature, parties map 1:1 with owners of public/private Vega keypairs.
+This means that there is a single private key which can sign messages to act as any given party.
+It also means that all parties can be identified by one (and only one) public key.
+The keypair representing a given party cannot be changed.
+**As a result, public keys in Vega are currently reused directly as party IDs.**
+
+
+## Change summary
+
+This PR will add specs for four major changes:
+
+1. Decoupling of party "accounts" from the type of account (e.g. keypair) and implementation detail (e.g. ed25519 public key).
+
+2. Addition of an account type enabling flexible "multisig" functionality.
+
+3. Addition of an account type that allows restriction, permissioning, and delegation of the capability to perform various types of action to to other accounts.
+
+4. Addition of an account type that operates as a non-custodial fund, separating deposits and withdrawals from fund deployment decisions, and only allowing a depositor to withdraw their share of the fund's value.
+
+
+## Account abstraction requirements
+
+Account abstraction decouples party acccounts as a concept from public/private keypairs. It lays the groundwork for new account types like multisigs and "fund" accounts.
+
+- After this change there will be multiple types of account.
+
+- This means that account IDs will need to change.
+  They can no longer be public keys but instead might, for instance, be deterministically generated at account creation time (similar to order IDs).
+
+- It is CRITICAL that ANY account type can be used ANYWHERE that an account is expcted.
+
+- The first type of account will be a standard public/private keypair.
+
+- The other three are described below.
+
+- There will need to be appropriate core and data node APIs for querying accounts and connecting between accounts (for example getting the account ID for a public key).
+
+- Will need a way to wrap to inform Vega of the account to be used.
+  For example, whether I am acting theough my pubkey account, a multisig, fund account, etc.
+	To prevent excessive complexity, layered wrapping of account types could be done when setting an account up but it would probably make sense to generate a single account ID for each potential destination at that point. Note to review and get comfortable with the approach here.
+
+- There should be a fee and appropriate spam protection to create or update account contructs.
+
+- A portion of the fee could be stored on creation and released if an account is destroyed to encourage not having accounts live forever.
+
+- Destroying an account is only possible for an account with no open positions and no liquidity commitments.
+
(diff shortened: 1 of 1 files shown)


barnabee:
Thinking about it, we can probably live without item 3 of the tl;dr in the first instance, though I'd need to discuss with some MMs and other community members to understand the specific needs of their use cases better.

0000-account-abstraction.md
```diff
@@ -0,0 +1,246 @@
...
+
+- The first type of account will be a standard public/private keypair.
+
+- The other three are described below.
+
+- There will need to be appropriate core and data node APIs for querying accounts and connecting between accounts (for example getting the account ID for a public key).
+
+- Will need a way to wrap to inform Vega of the account to be used.
+  For example, whether I am acting theough my pubkey account, a multisig, fund account, etc.
+	To prevent excessive complexity, layered wrapping of account types could be done when setting an account up but it would probably make sense to generate a single account ID for each potential destination at that point. Note to review and get comfortable with the approach here.
```
witgaw, line 49: Could you clarify the bit about "generate a single account ID for each potential destination". 
Does it mean account B acting on behalf of account A gets labeled as account C everywhere or something else?

0000-account-abstraction.md
```diff
@@ -0,0 +1,246 @@
...
+  Their share is reduced by the fraction of the asset that is removed.
+
+- A fund account has a manager account ID.
+
+- The manager account can deploy funds through trading, liquidity provision, transfers, or by placing funds in other fund accounts.
+  Actions of the manager account may be delegated or restricted by making it permissioned account.
+	The actions may be performed by a DAO or committee by making it a multisig.
+  The manager account could be a multisig with permissions applied, or some combinations of permissions and several multisigs, etc. to create more complex structures.
+
+- The
```
witgaw, line 212: How would fee distribution work? Get the total value of general + margin account balances for the fund account and transfer a given percentage from general if available? I suppose it'd be up to the fund account manager to assure enough is available there (so no auto-liquidation or anything like that).

0000-account-abstraction.md
barnabee: Yes I think it means something like that. I am a bit unsure about this bit and happy for you to rip any of it up and start again if you can comeup with something better.

0000-account-abstraction.md
barnabee: Good question. That sounds like the simplest option.

0000-account-abstraction.md
```diff
@@ -0,0 +1,246 @@
...
+- The controller can update the weights of each account.
+
+- The controller can change the approval number threshold.
+
+- The controller can change the approcal weight-fraction threshold.
+
+- The controller can change the account-wide timeout.
+
+
+## Account permissions and restrictions
```
barnabee, line 107: See my comment on the PR @witgaw, I think it'd be good to descope this from this PR and include basic restrictions of scope of use of funds in the fund account spec directly.

PR #1511, opened 2023-01-11
feat: update dano spec connection behavoiur

ettec:
Based on this discussion-> https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1673275647392279

As long as the client can determine the staleness or otherwise of the datanode data, the node should not automatically shutdown if it stops receiving blocks.


edd:
I agree with this change, and I suspect it's current behaviour anyway - but I'll let someone close to the issue hit the approve button

protocol/0076-DANO-data-node.md
```diff
@@ -186,4 +186,4 @@ It must be possible to add to the data node APIs that return the result of calcu
 1. All events that are emitted on the full unfiltered event stream are processed by the DataNode (no data is lost) (<a name="0076-DANO-018" href="#0076-DANO-018">0076-DANO-018</a>)  
 1. If a DataNode loses connection to a Vega node if will attempt to reconnect and if the cached data received from the Vega node is enough to continue working it can resume being a DataNode. (<a name="0076-DANO-019" href="#0076-DANO-019">0076-DANO-019</a>)  
 1. If the DataNode loses connection to a Vega node and it is unable to reconnect in time to see all the missing data, it will shutdown. (<a name="0076-DANO-020" href="#0076-DANO-020">0076-DANO-020</a>)
```
MuthuVega: Should we have a config value for time to reconnect and shutdown after that time has elapsed?
OR
Remove this AC ?

ettec: the only way for it to not reconnect 'in time' is if the cores event buffer fills up, at this point the core node will shutdown , to be consistent with the change to point 7 I suggest we just rely on the client checking the response headers for staleness and not do any autoshutdown in datanode.

MuthuVega: Thanks @ettec , in that case , can we please remove AC `0076-DANO-020`

ettec: done

MuthuVega (approved):

PR #1512, opened 2023-01-11
`DATA`. Quantum decimals and accounts

### LINKED ISSUE #3514: Acceptance criteria for common data formatting
link may change but at time of writing: https://github.com/vegaprotocol/specs/blob/user-interface-placeholders-for-all/user-interface/DATA-data_display.md

This will probably be added to as people write other ACs. 
it is intended to house common rules about how common data should be displayed. e.g. decimal places, public keys. 
This ticket to review its content once all (or most) ACs have been done.

campbellssource:
closes: https://github.com/vegaprotocol/frontend-monorepo/issues/3514

This PR takes some information from Slack and github tickets about common ways we format data and commits it to the user-interface specs so that it can be remembered and tested against.

https://vegaprotocol.slack.com/archives/C037STCKMGC/p1669812189426259?thread_ts=1669806984.217019&cid=C037STCKMGC
https://vegaprotocol.slack.com/archives/C033K0B1N2J/p1673437654522559?thread_ts=1673434908.700639&cid=C033K0B1N2J
https://github.com/vegaprotocol/vega.xyz/issues/707

The DATA file is referenced by other ACs as to avoid having to write the same information in every instance a particular data point is displayed.

### DIFF
--- a/user-interface/9001-DATA-data_display.md
+++ b/user-interface/9001-DATA-data_display.md
@@ -82,7 +82,7 @@ In the case of the Governance token the party may also have staked balances
 
 When listing markets in a breakdown it may good to see a total exposure to a market, that is then broken down by Margin and Bond
 
-All part accounts should be displayed as an [asset balance](#asset-balances) with the appropriate number of decimal places.
+All party accounts should be displayed as an [asset balance](#asset-balances) with the appropriate number of decimal places.
 
 ## Transaction hash
 
--- a/user-interface/9001-DATA-data_display.md
+++ b/user-interface/9001-DATA-data_display.md
@@ -63,6 +63,17 @@ Markets can have several statuses and it may be sensible when listing markets to
 When looking at a public key it is important that the user can get the full public key but it is often appropriate just to show an abbreviated form. The first 6 and last 6 characters are preferable, with an indication that it is truncated e.g. `56d1e6739deac3c5c1ddc6fee876b3217e504a161b5b00fda96b40ed3e8f89b8` as `56d1e6...8f89b8` or just `8f89b8` if enough of a convention has been established. In cases where the key being shows comes from your connected wallet it should also show the name (aka alias) of the key.
 Vega public keys are hexadecimal, but the convention is to display them without the preceding `0x` as this is what the Vega API returns.
 
+## Party accounts
+
+> aka account balances
+
+A can have a number of different accounts for any given asset. 
+- General account - For assets that have not been deployed to one of the following
+- Margin accounts - one for each market the part has orders/position
+- Bond accounts - (aka liquidity commitment) one for each market the user has committed to providing liquidity on
+- Locked for withdraw - one for each withdraw that is waiting to be finalised (e.g. where withdrawal limits/time locks are in place)
+  
+In the case of the Governance token the party may also have staked balances
 ## Transaction hash
 
 > aka Transaction ID, txn, tx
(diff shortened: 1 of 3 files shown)


user-interface/9001-DATA-data_display.md
```diff
@@ -33,6 +33,15 @@ Vega does not return a `total balance` that is a sum of all accounts in a curren
...
 `Asset.decimals` tells the UIs where to put the decimal place. Ethereum assets often have 18 decimal places, but can have less. Forms where you deposit, withdraw or transfer assets must show all decimal places. It may be appropriate to truncate at a certain number of DP in many cases (e.g. `0.01` instead of `0.012345678912345678` event though `0.001 wBTC` may be worth as much as than `0.01 ETH`). At the moment Vega does not have a source of information that allows conversion of currencies or way of knowing that the significant value of an asset is.
 
+## Quantum
+
+The quantum is a value that is used to define "The minimum economically meaningful amount of this specific asset
+". It can be used to show only the number of decimal places that are meaningful in summary information. A quantum is defined on each asset.
+
+Formula below will make sure that USD is shown to 1 cent .... and for every order of magnitude above / below that we show 1 less / more DP, ie. ETH is big so will show more DP.
+
+`Number DP = MAX(0, LOG10(100/Quantum))`
```
dexturr: This makes sense to me however there is an edge case here.

1x10^-18 vega is not a meaningful amount of money, so if I had this associated for some reason this formula would round the amount down to zero even if the user had a non-zero balance. Are we ok with this trade off?

dexturr: Perhaps a more meaninful example is, rewards are paid out based on share. Many people will have very small shares, 0.000002% or something. On the rewards page it is likely if my share is not big I will appear to have earnt no rewards in some assets with this formula if the amount distributed to me is below quantum.

campbellssource: I stole that from a ticket @JonRay15 wrote. I don't know the answer your question, maybe he does?

dexturr: The more generic point I guess I am trying to make is that the amount of DP to show can be dependent on multiple things:

* The amount the user has, if I am a millionaire I don't care about pence. If I'm not so wealthy I really do, possibly even less than pence.
* The amount the asset is worth, quantum
* The amount of decimal places a UI can support. It would be great to show 18 d.p. for everything but if there is a max supply of billions 27 digits is probably a bit much

Only considering one of these means that we may end up with overflow errors and misrepresenting small user balances. That's fine, but as long as we have said we are ok with that.

user-interface/9001-DATA-data_display.md
```diff
@@ -46,6 +55,28 @@ Markets can have several statuses and it may be sensible when listing markets to
...
+
+When listing asset balances for a given key it is useful to show a total of all accounts for a given asset (except locked for withdrawal this should be left out of account).
+
+In some cases the user may only be interested in the total, and the detailed breakdown can be hidden behind a link.
+
+When showing a summary of asset balance but not a full breakdown: it is useful to show the amount that is in general distinct from the amounts that are locked for other purposes. e.g. 90 Available, 10 locked
+
+When listing markets in a breakdown it may good to see a total exposure to a market, that is then broken down by Margin and Bond
+
+All part accounts should be displayed as an [asset balance](#asset-balances) with the appropriate number of decimal places.
```
edd:
```suggestion
All party accounts should be displayed as an [asset balance](#asset-balances) with the appropriate number of decimal places.
```

PR #1514, opened 2023-01-11
refactor: amend ACs relating to insurance pool balance transfer 

witgaw:


davidsiska-vega (commented):
Should we specify that each bucket gets equal amount or is that too much gold plating?

witgaw:
> Should we specify that each bucket gets equal amount or is that too much gold plating?

`protocol/0015-INSR-market_insurance_pool_collateral.md` which most of the ACs link to now says

>  When a market is finalised / closed remaining funds are distributed equally among other same-currency insurance pools (including the on-chain treasury for the asset, if no such treasury exists it gets created at this point).

so I'd think that's enough.

davidsiska-vega (approved):

PR #1529, opened 2023-01-17
refactor: let max OI drop during auction

witgaw:


protocol/0041-TSTK-target_stake.md
```diff
@@ -21,6 +21,8 @@ The parameter c_1 is a market parameter (with network parameter `market.liquidit
 First, `max_oi` is defined  maximum (open interest) measured over a time window,
 `t_window = [max(t-market.stake.target.timeWindow,t0),t]`. Here `t` is current time with `t0` being the end of market opening auction. Note that `max_oi` should be calculated recorded per transaction, so if there are multiple OI changes withing the same block (which implies the same timestamp), we should pick the max one, NOT the last one that was processed.
 
+If the market is in auction mode the `max_oi` can only increase while `auction duration` <= `market.stake.target.timeWindow`. Once the market's been in the auction for more than `market.stake.target.timeWindow` the `max_oi` is whatever the current positions and `indicative_uncrossing_volume` imply - specifically, this allows the `target_stake` to drop as a result of trades generated in the auction so that `target_stake` > `supplied_stake` (even in absence of changes to `supplied_stake`) and the market can go back to its default trading mode.
```
TomMcL, line 24: I don't know whether it simplifies or complicates things, but rather than the auction having to be longer than `market.stake.target.timeWindow` could the auction period not just be included in the window (allowing it to roll off in less than `market.stake.target.timeWindow`)? (The whole auction 'period' could be counted as having OI equal to whatever currently implied OI at resolution would be, as the spec says)

witgaw: That's effectively the same thing, but I thought describing it in the above way stresses it's implications on the ability to decrease target stake etc.

TomMcL (approved):

davidsiska-vega (approved):

PR #1534, opened 2023-01-19
fix: bring market.liquidity.targetstake.triggering.ratio in line with…

davidsiska-vega:
… implementation. 

The only change to ACs is the removal of 0035-LIQM-009 but I took the opportunity to use the proper markdown syntax for numbering - which makes it look like they all changed.


gordsport:
@vega-paul 
FYI as you are working on:
- https://github.com/vegaprotocol/system-tests/issues/1643

witgaw (approved):

PR #1556, opened 2023-01-27
Add quality checks to the makefile

gordsport:
In order to be able to run the newly added quality checks locally this PR adds in spelling and markdown linting checks to the makefile.

Running `make` will by default run **all** in the order specified below, or you can run each individually

- spellcheck 
- markdownlint 
- names 
- codes 
- links 
- references

This PR also fixes a number of formatting and spelling errors that have crept in


gordsport:
@fkondej - Added you as a reviewer. Ive testing this locally and all seems to work, but would appreciate some devops eyes over this to make sure ive not done something stupid

davidsiska-vega (commented):
For me it runs but seems to think lot of the specs have broken links - is that expected?

gordsport:
> For me it runs but seems to think lot of the specs have broken links - is that expected?

Im not entirely sure, i've not change the links or the approbation checks so these should be as they were before this PR.

I will investigate these separately

gordsport:
I will also in the near future update the bash scripts to check to see if the linting/spelling has been installed already and skip that step to speed it up. However, for now it works and is quicker than waiting for the GH CI to run etc..

fkondej (dismissed):
Please remove the tmp files and you can add *.tmp to gitignore

davidsiska-vega (approved):

PR #1586, opened 2023-02-09
chore: Remove untestable acceptance criteria - 0076-DANO-018

MuthuVega:
Removing AC 0076-DANO-018 as per the comments below:

`TBH that sounds like a criteria that is easy to write but pretty intense to implement without basically reimplementing the datanode in python in the system tests :slightly_smiling_face: I guess you might have have to have some sort of big switch statement that looks at each event that was emitted, and then does some API call to check that the right stuff got put in the DB`

**Slack thread on this discussion for reference:**
https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1675873898560439


gordsport (approved):
Approving based on the agreement in the linked slack convo

PR #1587, opened 2023-02-10
feat: add rate limit spec

### LINKED ISSUE #1578: Add ACs around the data node rate limiting
The following core issue is to implement rate limiting on the data node API requests

- https://github.com/vegaprotocol/vega/issues/7445

We need to update the [0076-DANO-data-node.md](https://github.com/vegaprotocol/specs/blob/master/protocol/0076-DANO-data-node.md) spec with any details we think should be in the specification (maybe about the configurability of the rate limiting?) and add  new ACs to ensure that we do not ban people in error or visa versa.

pscott31:
close #1578 .


gordsport:
@MuthuVega - please can you review these changes and add AC codes as you se fit?

protocol/0076-DANO-data-node.md
```diff
@@ -183,6 +183,21 @@ It must be possible to add to the data node APIs that return the result of calcu
 2. Standard nodes should retain data in accordance with the configured data retention policy (<a name="0076-DANO-027" href="#0076-DANO-027">0076-DANO-027</a>)
 3. Archival nodes should retain all data from the height at which they joined the network (<a name="0076-DANO-028" href="#0076-DANO-028">0076-DANO-028</a>)
 
+### API Request Rate Limiting
+
+1. Datanode should provide an optional mechanism for limiting the average number of requests per second over on its API
+2. That rate should be specified in the datanode configuration file
+3. A client may, over a short period of time, make requests at a greater frequency than the limit as long as the average rate over a longer period of time is not exceeded. (<a name="0076-DANO-029" href="#0076-DANO-029">0076-DANO-029</a>)
+4. The extent to which clients may of 'burst' requests should also be capped and specified in the datanode configuration file (<a name="0076-DANO-030" href="#0076-DANO-030">0076-DANO-030</a>)
+5. Limits should be enforced on a per-client basis. Source IP address is a sufficient discriminator (<a name="0076-DANO-031" href="#0076-DANO-031">0076-DANO-031</a>)
```
MuthuVega, line 192: Source IP is a good discriminator, however we may have issues testing it from system tests as vegacapsule will use the same IP. This may require some changes to vegacapsule for this to be tested. Or some other solution might be needed to overcome this limitation.

protocol/0076-DANO-data-node.md
```diff
@@ -183,6 +183,21 @@ It must be possible to add to the data node APIs that return the result of calcu
...
+6. Headers or metadata should be included in each API response indicating to the client what the limits are, and how close they currently are to exceeding them (<a name="0076-DANO-032" href="#0076-DANO-032">0076-DANO-032</a>)
+7. If limits are exceeded an API appropriate error response should be returned, containing similar headers or metadata (<a name="0076-DANO-033" href="#0076-DANO-033">0076-DANO-033</a>)
+8. If the client continues to make requests that generate error responses due to rate limiting, eventually the client should banned (<a name="0076-DANO-034" href="#0076-DANO-034">0076-DANO-034</a>)
```
MuthuVega: `eventually the client should banned` - the trigger point for the banning should be defined for the AC to be testable.

protocol/0076-DANO-data-node.md
```diff
@@ -183,6 +183,21 @@ It must be possible to add to the data node APIs that return the result of calcu
...
+9. The ban denies all access to the API for a configurable length of time (<a name="0076-DANO-035" href="#0076-DANO-035">0076-DANO-035</a>)
+10. For that time, any requests will receive an API appropriate error response indicating that they are banned (<a name="0076-DANO-036" href="#0076-DANO-036">0076-DANO-036</a>)
+11. The rate limit for the GraphQL API should be configurable separately from the gRPC API and it's REST wrapper since a single GraphQL request can trigger many internal gRPC requests (<a name="0076-DANO-037" href="#0076-DANO-037">0076-DANO-037</a>)
+12. Where one API makes use of another (e.g. GraphQL making use of gRPC), rate limits should be enforced only once, on the side that faces the client (<a name="0076-DANO-038" href="#0076-DANO-038">0076-DANO-038</a>)
```
MuthuVega, line 199: Are there any other examples other than `GraphQL making use of gRPC` ?

protocol/0076-DANO-data-node.md
pscott31: So for the all the existing system tests, rate limiting is simply disabled, so it doesn't interfere with those.. obviously that's no good for when we actually _want_ to test rate limiting.

It's quite straightforward on Linux to assign arbitrary IP addresses to an interface:

```
➜  ~ sudo ifconfig eth0:0 127.0.0.2 up
➜  ~ ping 127.0.0.2
PING 127.0.0.2 (127.0.0.2) 56(84) bytes of data.
64 bytes from 127.0.0.2: icmp_seq=1 ttl=64 time=0.026 ms
64 bytes from 127.0.0.2: icmp_seq=2 ttl=64 time=0.021 ms

➜  ~ sudo ifconfig eth0:0 127.0.0.3 up
➜  ~ ping 127.0.0.3
PING 127.0.0.3 (127.0.0.3) 56(84) bytes of data.
64 bytes from 127.0.0.3: icmp_seq=1 ttl=64 time=0.017 ms
64 bytes from 127.0.0.3: icmp_seq=2 ttl=64 time=0.022 ms
```

I think you should be able to use those as source addresses when connecting to the API.

I'm sure we can figure something out, but in any case I think this isn't part of the spec, which is how the datanode should behave, not how we test it.

protocol/0076-DANO-data-node.md
pscott31: OK. There is a defined mechanism for this (a second bucket of 'naughty' tokens of the same size/refill rate, from which a token is removed on every unsuccessful-due-to-rate-limiting request). I'll was slightly reluctant to put that in the spec as I wasn't sure that was going to be the final solution but you're right - to test it you need to know so I'll update it.

protocol/0076-DANO-data-node.md
pscott31: Not exactly. The only other example is REST making use of gRPC, but because there is a 1:1 correspondence between REST calls and gRPC calls I didn't add rate limiting on the REST API directly, rather it relies on the rate limiting provided by gRPC.

One GraphQL API call can generate many gRPC calls (which is a separate, complicated problem) and a graphQL client has no way to know beforehand how many it will generate - so it seemed unfair to limit them based on our internal architecture, which is why we rate limit `client -> gql gateway` calls but not `gql gateway -> grpc` calls.

protocol/0076-DANO-data-node.md
pscott31: I've changed the definition of point 8 to be a bit more concrete. I wanted to keep the specification kind of short and snappy so didn't go into massive detail, if you want me to lay out how it actually works I can.

MuthuVega (approved):

PR #1599, opened 2023-02-23
fix: Add extra cases for various things that might be affected by val…

jgsbennett:
…idators joining and leaving.


protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -487,6 +487,9 @@ See [limited network life spec](./0073-LIMN-limited_network_life.md).
     - Verify that at the end of the Epoch, the Ersatz validators get 40% of the reward that the validators get (in total)
     - Set the `rewardFactor` to 0.32832979375934745648654893643856748734895749785943759843759437549837534987593483498
     - Verify that all validators round the value of reward for the Ersatz validators to the same value.
+1. Ensure multisig threshold is set to '999' (require all signatures). Demote one of the original validators and replace with a new validator. Update the multisig. Attempt a withdrawal. (<a name="0069-VCBS-069" href="#0069-VCBS-069">0069-VCBS-069</a>)
```
davidsiska-vega: I'm not expecting this to work. 

If I need all signatures and demote a validator - then I'll be 1 signature short when requesting the update.

jgsbennett: @wwestgarth this was your suggest. Any comments please?

wwestgarth: Thats a fair point. The intention of setting the threshold to 999 was to check that during the creation of signatures for the withdrawal that the new node's signature was necessary and valid (at 666, 4/5 would be enough so if the new node wasn't producing signatures or producing wrong signatures the bundle would still be valid). So I think we can just reword the AC so that the threshold is changed to 999 _after_ the "Update the multisig." part.

protocol/0069-VCBS-validators_chosen_by_stake.md
```diff
@@ -487,6 +487,9 @@ See [limited network life spec](./0073-LIMN-limited_network_life.md).
...
+1. On a network with n original validators, gradually replace (via demotion of existing node and promotion of a new node) and stop all of the original validators. (Original nodes not even participating as ersatz or pending). Ensure that consensus continues, and that asset withdrawals are possible. (<a name="0069-VCBS-070" href="#0069-VCBS-070">0069-VCBS-070</a>)
+1. Ensure multisig threshold is set to '666'. Request an asset withdrawal (but do not yet exercise this in the smart contract). Demote one of the original validators and replace with a new validator. Update the multisig. Attempt to enact the withdrawal on the same contract. (<a name="0069-VCBS-072" href="#0069-VCBS-072">0069-VCBS-072</a>)
```
wwestgarth: Minor and petty: `Attempt to enact the withdrawal` doesn't say anything about the expected outcome.

protocol/0069-VCBS-validators_chosen_by_stake.md
jgsbennett: Done. Thanks both,

protocol/0069-VCBS-validators_chosen_by_stake.md
jgsbennett: Thanks. Alternate version in new commit.

wwestgarth (dismissed):

davidsiska-vega (dismissed):

davidsiska-vega (approved):

PR #1601, opened 2023-02-24
fix: More validator join/leave test cases involving checkpoints.

jgsbennett:


davidsiska-vega (dismissed):

wwestgarth (dismissed):

gordsport (dismissed):
re-approving as last change was just a spelling fix

gordsport (approved):
reapproving as only an AC code number update.

PR #1603, opened 2023-03-01
fix: Relabel existing PLUP mainnet case to reflect the fact that it i…

### LINKED ISSUE #1576: Add AC to 0075-PLUP-protocol_upgrades.md for upgrade with a DB schema change
In the [0075-PLUP-protocol_upgrades.md](https://github.com/vegaprotocol/specs/blob/master/protocol/0075-PLUP-protocol_upgrades.md) speac we do not have an acceptance criteria covering the scenario whereby we have a database schema change in the version when using the protocol upgrade feature.

- [ ] review this scenario 
- [ ] add an acceptance criteria

jgsbennett:
…s only valid once PLUP-compatible versions are deployed. Add a new case to cover what is worth testing now, namely, that our best guess is that mainnet data is upgradable in future.

close #1576


gordsport (approved):
Reviewed in the protocol upgrade meeting - approved.

PR #1607, opened 2023-03-06
fix: set numbers on ds part

AnExsomnis:
This change fixes the numbering in the data sources.
It needs to be fixed in  the rest of the specs as well, because it is only 1's everywhere.
There is no issue for that, happy to move forward with that work and create an issue if that is accepted


AnExsomnis:
Closing, because numbering is handled in another way

PR #1619, opened 2023-03-10
Create multis user update

### LINKED ISSUE #1503: Multisig improvements for scaling

Vegaklaus:
close #1503 .


Vegaklaus:
One other option: Don't have an explicit evaluate function, but evaluate every time someone votes (shouldn't make any significant difference)

PR #1633, opened 2023-03-21
Amend margin spec

witgaw:
A while back I created a PR which both:
1. clarified the margin calculation method that we settled on
1. made a change to include margins in slippage part of the margin

This got merged to `master`, then reverted and merged to `cosmicelevator`.

This PR brings 1. to `master` so that diff with `cosmicelevator` will only be 2.


gordsport (approved):
Looking at the cosmic / master changes this LGTM. Approving based on the fact the changes were previously approved.

PR #1637, opened 2023-03-23
feat: Add a test for checking performance scores when checkpoint rest…

jgsbennett:
…art happens in a new epoch.


gordsport (dismissed):
Approved as per the agreement in the protocol design meeting 29th March

gordsport (approved):
Approved as per the agreement in the protocol design meeting 29th March

PR #1640, opened 2023-03-28
User interface: add Post only and Reduce only

campbellssource:
Updates the User interface ACs now that Post and Reduce only exist


user-interface/7002-SORD-submit_orders.md
```diff
@@ -114,6 +114,22 @@ When populating a deal ticket I...
 
 ... so I can control if and how my order stays on the order book
 
+## Execution flags
+
+... could select a some execution flags, When selecting execution flags (aka order conditions)...
```
candida-d:
```suggestion
... could select some execution flags. When selecting execution flags (aka order conditions)...
```

mattrussell36 (approved):

PR #1643, opened 2023-03-29
Clarify rewards if multisig contains a key it should not

### LINKED ISSUE #1593: Clarify validator rewards in case `0` score
In the specs https://github.com/vegaprotocol/specs/blob/master/protocol/0069-VCBS-validators_chosen_by_stake.md we have
```
In the case where a node is removed due reduced delegation, or due to not meeting self-delegation criteria, or due to lack of performance, or due to a reduction in the value of network.validators.tendermint.number, the onus is on all of the remaining validators to remove the demoted member from the Multisig contract. They are incentivised to do so by all receiving a validator_score of 0 in the reward calculation until the excess member is removed.
but it’s not clear that in that case the consequence is that there are no validator rewards and the reward asset is kept in the pool until the next epoch end.
```

But in transfers spec we have:

```
A user's recurring transfer to a reward account does not occur if there are no parties eligible for a reward in the current epoch (0057-TRAN-057) 
```

Which seems to be inconsistent with the above.

There is a reason: validator rewards come from trading fees - those cannot be “not transferred”. 

We need a spec update to clarify this. 

Btw. from @ze97286 "A transfer without a metric, as is a regular recurring transfer to a reward account is no different from a recurring transfer to a general account, it’s not aware of what’s on the other side or whether there’s anyone eligible or not."

davidsiska-vega:
Closes #1593

### DIFF
--- a/protocol/0069-VCBS-validators_chosen_by_stake.md
+++ b/protocol/0069-VCBS-validators_chosen_by_stake.md
@@ -103,7 +103,7 @@ Thus a validator who is not there but should be has incentive to pay gas to upda
 
 As a consequence, if a potential validator joined the Vega chain validators but has *not* updated the Multisig members (and/or weights) then at the end of the epoch their score will be `0`. They will not get any rewards.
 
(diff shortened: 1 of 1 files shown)


gordsport (dismissed):
LGTM

gordsport (approved):
LGTM

PR #1645, opened 2023-03-29
spec: draft for time-based mempool pruning

### LINKED ISSUE #1639: Timebased pruning of the mempool
In order to best manage the mempool (before the below mentioned further optimisations) we need to investigate and implement time based pruning of the mempool

- [ ] Investigate how we can prune the mempool with Tendermint 0.34
- [ ] Update the specification(s) where applicable
- [ ] Add any ACs where required
- [ ] Implement the changes (https://github.com/vegaprotocol/vega/issues/7969)

Further optimisations to the mempool are to be designed along with the upcoming tendermint releases and ABCI++
- https://github.com/vegaprotocol/specs/issues/1439

EVODelavega:
Closes #1639


protocol/0080-TBPM-time_based_pruning_mempool.md
```diff
@@ -0,0 +1,24 @@
+# Time based pruning of the mempool
+
+Transactions currently can stay in the mempool indefinitely. For non-validator commands, we want to be able to limit the time they are considered valid, once a certain amount of time has passed, the transaction should be rejected and removed from the mempool.
+
+## Network parameters
+
+- `network.transactions.maxttl` is a network parameter with type unsigned integer with the minimum value of `10` and a maximum value of `50`. This value determines the max number of blocks a given transaction is valid for. A transaction that specifies the `TTL`, this value *must* be less than, or equal to this parameter. If no `TTL` is specified, the transaction `TTL` will be set to this value.
```
davidsiska-vega, line 7: I understand that I have a maximum number of blocks that a transaction can be current for. 

How / where is the current block / block hash added to the transaction? Or is this not needed? Maybe the "current block and block hash" is already part of the transaction submitted and I am just not aware but it may be worth spelling out.

protocol/0080-TBPM-time_based_pruning_mempool.md
```diff
@@ -0,0 +1,24 @@
...
+
+This `TTL` is applied to all transactions, except for validator commands. Users have the ability to set the `TTL` manually, using transaction version 4, previous versions use the network parameter mentioned above as the `TTL`.
+
+## Validation
+
+For commands using the V4 transaction, the wallet will ensure the `TTL` does not exceed the maximum as specified by the network parameter. Vega/core will additionally validate the `TTL` on `CheckTx`. Should the `TTL` be greater than the network paramter, the network paramter will override the `TTL` specified in the transaction, both in the wallet and the core.
+
+## Acceptance criteria
+
+### Basic network paramter test
```
davidsiska-vega:
```suggestion
### Basic network paramter test (<a name="0080-TBPM-001" href="#0080-TBPM-001">0080-TBPM-001</a>)
```

protocol/0080-TBPM-time_based_pruning_mempool.md
EVODelavega: Will clarify, but essentially, the transaction contains the block height at which the transaction was made. The TTL therefore will be `tx.InputData.BlockHeight + (tx.InputData.GoodForBlocks || netparams.MaxTTL)` where the TTL is either the specified number of blocks (for TxV4) or the max number of blocks defined by the network parameter

davidsiska-vega: > the transaction contains the block height 

Does it also contain the block hash? Otherwise the submitting client may lie about it if they want to mess with the chain.

EVODelavega: Yes, if the network is not specified when signing the transaction, there is _some_ leniency, but the tx hash + height both have to be provided, and should be fairly close to the current height/hash:

```go
	cmd.Flags().Uint64Var(&f.TxBlockHeight,
		"tx-height",
		0,
		"It should be close to the current block height when the transaction is applied, with a threshold of ~ - 150 blocks, not required if --network is set",
	)
	// ...
	cmd.Flags().StringVar(&f.TxBlockHash,
		"tx-block-hash",
		"",
		"The block-hash corresponding to tx-height which will be used to generate proof-of-work (hex encoded)",
	)
```

That does mean that, in combination with the TTL, specifying a block height that is outdated, so to speak, essentially means signing a transaction with a specified height/hash that is too far in the past is pretty much a waste of time. Not sure what we should do about this... Only validator commands don't expire, any other transaction should specify a height that is < current - TTL, so the 150 block thing is a bit pointless

davidsiska-vega: Okay, I'll wait for you to clarify and then read again. Thx!

protocol/0080-TBPM-time_based_pruning_mempool.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Network parameters
+
+- `network.transactions.maxttl` is a network parameter with type unsigned integer with the minimum value of `10` and a maximum value of `50`. This value determines the max number of blocks a given transaction is valid for. A transaction that specifies the `TTL`, this value *must* be less than, or equal to this parameter. If no `TTL` is specified, the transaction `TTL` will be set to this value.
+
+Part of the transaction data is the block height at which the transaction was created. Transactions will remain current until this block height + the `TTL`.
+
+## Including transactions
+
+This `TTL` is applied to all transactions, except for validator commands. Users have the ability to set the `TTL` manually, using transaction version 4, previous versions use the network parameter mentioned above as the `TTL`.
```
wwestgarth: Do we even need a new transaction version? If we default to the value of `network.transactions.maxttl` if its not set in the transaction then we're fine.

protocol/0080-TBPM-time_based_pruning_mempool.md
EVODelavega: Should be better now

protocol/0080-TBPM-time_based_pruning_mempool.md
EVODelavega: Indeed, TxV4 was removed from the spec and implementation - adding the field is non-breaking.

davidsiska-vega (approved):

protocol/0080-TBPM-time_based_pruning_mempool.md
Vegaklaus, line 1: It is still possible to spam the mempool/make your transaction live forever by adding an insanely high block number as the base block (if the chain is at block 1000, and I say 'here's the hash from block 10.000, and my TTL is 20, an individual validator cannot prove me wrong). The utility of this can be mitigated by making the condition "No transaction can get into a block smaller than the base block it gives", which would turn those transactions into zombies. Alternatively, a validator who gets a transaction far ahead of where they are can outright reject it, under the assumption that it's either spam or the validator is so far behind that it won't contribute anyhow, though the might break things.
Relatedly, are all validator commands signed by the validator issuing them ?

Vegaklaus: There probably is a way to identify zombies (and then remove them) - if I'm a leader and my block got accepted, I'm clearly not 10000 blocks behind. Would need some work to make sure that's synchronised though

gordsport:
Merging as approved and the development PR has been merged too:
- https://github.com/vegaprotocol/vega/pull/8028

PR #1674, opened 2023-04-20
Update 0045-DSRC-data_sourcing.md

vega-paul:


protocol/0045-DSRC-data_sourcing.md
```diff
@@ -246,3 +246,6 @@ Vega should reject any data source tx that is not explicitly required, so this w
...
+1. It should be possible to update the market termination spec from external to internal and prove that the original termination key and filter no longer supports termination and that the updated market termination key and filter supports termination. This should apply to filter key types Timestamp/Boolean and Integer. (<a name="0045-DSRC-022" href="#0045-DSRC-022">0045-DSRC-022</a>)
+1. It should be possible to update the market termination spec from internal to external and prove that the original termination key and filter no longer supports termination and that the updated market termination key and filter supports termination. This should apply to filter key types Timestamp/Boolean and Integer. (<a name="0045-DSRC-023" href="#0045-DSRC-023">0045-DSRC-023</a>)
+1. It should not be possible to update the market settlement spec from external to internal type (<a name="0045-DSRC-024" href="#0045-DSRC-024">0045-DSRC-024</a>)
```
davidsiska-vega: What does this mean exactly?

AnExsomnis: @davidsiska-vega , these conditions mean that 
1. We should be able to update a market termination from external to internal
2. We should be able to update a market termination from internal to external
3. Market settlement is only external
After updating the markets, the old conditions should not work.

In my opinion the changes are relevant, approving now.

AnExsomnis (approved):

protocol/0045-DSRC-data_sourcing.md
davidsiska-vega: I agree with "0045-DSRC-022" and "0045-DSRC-023". 

I don't uderstand "0045-DSRC-024". Why should I not be allowed to change market settlement spec from external to internal?

AnExsomnis: Because market settlement can be only external.

AnExsomnis: @davidsiska-vega , currently the internal data source is implemented only for time terminated type, please refer to this:
https://github.com/vegaprotocol/specs/blob/master/protocol/0048-DSRI-data_source_internal.md#13-vega-time-changed

In the current core implementation this type does not contain anything except a timestamp and the builtin key name for it.

barnabee: We should be able to update an external trading termination trigger to be an internal time based trigger. And in future there will be more internal options. This AC looks like it's invalid, even now.

Obviously if I am switching type of data source it still needs to be one that provides the expected data type for the consumer, though I guess going internal to external this can't necessarily be verified at the time the data source is set/changed, and invalid data needs to be handled gracefully and emit appropriate events / set appropriate state.

barnabee: This doesn't mean we need to support [internal value sources](https://github.com/vegaprotocol/specs/blob/master/protocol/0048-DSRI-data_source_internal.md#11-value), etc. at this time, to switch to, just that if there is a valid internal data source for a given consumer, we should be able to switch to/from it.

AnExsomnis:
```suggestion
1. It should not be possible to update the market settlement spec from external to internal type (there is no requirement to prevent making settlement data an internal source, it's just currently true that no internal sources emit data that could be used for settlement so it should in practice not be possible at the moment). (<a name="0045-DSRC-024" href="#0045-DSRC-024">0045-DSRC-024</a>)
```

AnExsomnis (changes requested):

AnExsomnis (approved):

davidsiska-vega (approved):

PR #1681, opened 2023-04-27
feat: add stop orders to order types spec

### LINKED ISSUE #1666: stop orders
### Issue

We need to introduce stop order types, this issue is to specify update [0014-ORDT-order_types.md](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0014-ORDT-order_types.md) and other specs? to add the new order type

 The spec has been updated to indicate the requirements
 The specification contains acceptance criteria with IDs that reflect the scope of the work required
 The engineering team has reviewed the specifications and understand the scope
 The spec is merged into the specs repo before engineering start development/test work

barnabee:
close #1666  . Add stop orders.


protocol/0014-ORDT-order_types.md
```diff
@@ -47,8 +47,61 @@ Notes on scope of current version of this spec:
...
+These are generally used to exit positions under pre-defined conditions, either as a "stop loss" order that controls the maximum losses a position may take, a "take profit" order that closes a position once a defined level of profit has been made, or both.
+To prevent traders from "hiding" order book depth behind conditional orders, stop orders can only be used to close some or all of a trader's position, and therefore must be "reduce only" orders.
+
+A stop order submission can be made (stop loss or take profit are probably both just called a stop order internally).
+
+* Stop order submissions must include either a trigger price OR trailing stop distance as a % move from the reference price in addition to a normal order submission.
+
+* Stop order submissions must include a trigger direction.
+Direction may be _rises above_ or _falls below_.
+_Rises above_ stops trigger if the last traded price is higher than the trigger level, and _falls below_ stops trigger if the last traded price is lower than the trigger level.
```
witgaw: Would it take a lot away from the feature if we made this conditional on mark price instead to save some computation? Can of course start with last traded price and modify it if does become an issue.

protocol/0014-ORDT-order_types.md
```diff
@@ -47,8 +47,61 @@ Notes on scope of current version of this spec:
...
+
+* A stop order can be entered during an auction, and can then be triggered by the auction uncrossing price if the auction results in a trade, as well as any trades (including auction uncrossing trades) after that.
+
+* GFA is not a valid TIF for a stop order submission.
+
+* Spam prevention:
+
+    * Stop orders will only be accepted from keys with either a non-zero open position or at least one active order.
+
+    * A network parameter will control the maximum number of stop orders per party (suggested initial value: between 4 and 10).
```
witgaw: Could we not just accept stop orders with volume up to the current riskiest long / riskiest short?

protocol/0014-ORDT-order_types.md
```diff
@@ -47,8 +47,61 @@ Notes on scope of current version of this spec:
...
+
+    * (TODO: confirm this makes sense) If the trader's position size gets to zero or changes direction all their stop orders are cancelled.
```
witgaw: Probably not, one can always have some other passive limit orders they wish to stop-loss-manage in case they fill.

protocol/0014-ORDT-order_types.md
barnabee: I think it isn't great to do that, would be quite surprising for users I think. Let's start with last traded price, stops shouldn't be a huge amount of computation to evaluate as you can maintain them in a sorted data structure.

protocol/0014-ORDT-order_types.md
barnabee: You could still have a lot of them, then. The thing this is trying to control is the number of stop orders rather than the total volume of your stops.

witgaw: Ah sure, position decimal places...

protocol/0014-ORDT-order_types.md
barnabee: The problem is that if your stuff changes direction then the stops that are currently reduce only (which is a condition of stops in this spec) wont apply because they no longer reduce.

I guess we could allow them to sit there and only evaluate whether they are reduce only at the time they are triggered…

The other option will be to add "OTO" (one triggers the other) stops so that you can connect a stop to an active order and it won't become an "active" stop order unless/until that trades. But idk if it is worth it…

Maybe not having this condition and just checking the reduce only at trigger time saves it, what do you think @witgaw?

witgaw: That'd be fine, as you say the reduce only bit will guard against stop order filling on the wrong side of the position. We should "garbage collect" all of them if at any point party has 0 open volume and no limit orders posted, but that's implementation detail.

barnabee: I think we should explicitly state they are always removed if you have zero position and orders not leave it to implementation but will allow change of direction to guard against wrong way stops. 

I'll update the spec.

ValentinTrinque:
At the very top, there is mention of:

https://github.com/vegaprotocol/specs/blob/eb06c5e98460f96c733be155af0460870c29c2ef/protocol/0014-ORDT-order_types.md?plain=1#L10

Now the spec is being written, this probably should be removed.

protocol/0014-ORDT-order_types.md
```diff
@@ -47,8 +47,61 @@ Notes on scope of current version of this spec:
...
+
+- GFA is not a valid TIF for a stop order submission.
+
+- Spam prevention:
+
+  - Stop orders will only be accepted from keys with either a non-zero open position or at least one active order.
+
+  - A network parameter will control the maximum number of stop orders per party (suggested initial value: between 4 and 10).
+
+  - TODO: confirm this makes sense) If the trader's position size gets to zero or changes direction all their stop orders are cancelled.
```
TomMcL: Unsure on this one. The reduce-only-ness of the order should be validated at trade time rather than stop creation time, so it's possible the position touches 0 one way or another between creation and execution. That said, it doesn't seem to *logically* make much sense to have a stop order, close your position/change sides then want it still to trigger. 

Thinking about it though I could see a use for it so I'd probably lean towards not closing them in this case (for example, one might put in both a large buy and sell for if BTC drops below `x` price knowing that only one will execute but you will close out your position whatever happens)

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +142,45 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+- A trailing stop order for a 5% rise placed when the price is `50`, followed by a drop to `40` will:
+  - Be triggered by a rise to `42`. (<a name="0014-ORDT-028" href="#0014-ORDT-028">0014-ORDT-028</a>)
+  - Not be triggered by a rise to `41`. (<a name="0014-ORDT-029" href="#0014-ORDT-029">0014-ORDT-029</a>)
+- A trailing stop order for a 25% drop placed when the price is `50`, followed by a price rise to `60`, then to `50`, then another rise to `57` will:
+  - Be triggered by a fall to `45`. (<a name="0014-ORDT-030" href="#0014-ORDT-030">0014-ORDT-030</a>)
+  - Not be triggered by a fall to `46`. (<a name="0014-ORDT-031" href="#0014-ORDT-031">0014-ORDT-031</a>)
+
+- A stop order placed either prior to or during an auction will not execute during an auction, nor will it participate in the uncrossing. (<a name="0014-ORDT-032" href="#0014-ORDT-032">0014-ORDT-032</a>)
+- A stop order placed either prior to or during an auction, where the uncrossing price is within the triggering range, will immediately execute following uncrossing. (<a name="0014-ORDT-033" href="#0014-ORDT-033">0014-ORDT-033</a>)
+
```
MuthuVega, line 183: The AC's for stop loss orders look good to me.

MuthuVega:
The AC's for stop loss orders look good to me.

TomMcL:
Closing as GitHub caused chaos changing PR branch base, new PR is https://github.com/vegaprotocol/specs/pull/1735

PR #1685, opened 2023-04-30
SLA liquidity

### LINKED ISSUE #1665: RFC: Liquidity mechanism overhaul — replace forced volume with SLA
This is a request for comments, at an early stage, for consideration.

It would make quite a large change to the LP mechanics, but it would significantly simplify some things, namely:

- No more LP shapes, LP orders, or LP pegs
- LP mechanics become just a commit/uncommit transaction, plus liqudity monitoring and calculation of fees and rewards

### The problem:

* LPs aren't used to being forced to quote at all times and don't want to.
* This may lead either to fewer LPs and smalller commitments, or in fact not making commitments at all.
* This threatens the proper functioning of the liquidity incentivisation mechanism, with knock-on potential impacts to credit risk management (margin calcs).
* If the network is forced to operate with target stake multiplier set to zero, then this also opens the liquidity incentives up to a vampire/leech(?) attack by potentially any trader, diverting rewards that ought to be received by a real LP who may spend say >95% of the time providing liqudiity towards far more fickle traders.

### Historical context

The liquidation system, and therefore a significant part of credit risk management (particularly given that in a pseudonymous, decentralised system, the assumed recovery rate if a party owes more in cashflows during MTM or final settlement is zero), relies on there being sufficient liquidity to close out distressed or delinquent positions against other traders. The financial scale of the risk is also dependent on the expected slippage during closeout.

As a result of this, early designs aimed to ensure that a market with non-distressed committed (via on-chain transaction and bond) LPs would always remain operational with a minimum level of liquidity available.

This proved impractical for a number of reasons and was always incompatible with price monitoring ("circuit breakers") so it was never a hard requirement of the overall system design. However, the liquidity mechanism retained features related to this idea, in the form of the forced deployment of LP pegged orders.

On reflection this was a poor design decision, given that markets can be suspended (in auction) for a number of reasons in any case and we are comfortable with this.

This ticket aims to set out a path forward.

### Proposed solution overview

At a high level, the solution to this would be to replace the system that forces LPs to be on the book all the time with one that:
1. Requires LPs to meet an SLA (i.e. % of time spent providing liquidity within some range) to be rewarded, ensuring LP remains a defined and useful role.
2. Rewards LPs more for better performance against the SLA (already done for price, this would add similar for time spent providing liquidity), ensuring there is an incentive to do more than the bare minimum if market conditions allow.
3. Penalises LPs that commit and do not meet the SLA, to reduce the attractiveness of opportunism/leeching style attacks.

A side effect of this change will be to remove the LP "shape" from LP commitments, as LPs will now only place normal orders, pegged or otherwise. In order to allow LPs to remain on the book after trading with greater certainty while limiting the amount that might trade in any one transaction, the forced refresh of the orders implied by the LP "shape" will be replaced by an optional iceberg feature available on all persistent order types, including pegged persistent orders.

Another side effect is that LPs can effectively suspend a market by withdrawing their orders. They can sometimes do this anyway if there are no other priced orders, so I don't see that as a big deal.

### Proposed solution detail

This is a high level summary of a possible approach, primarily intended to be complete enough to generate useful discussion about a quite considerably 

#### SLA

- [x] New market parameter: SLA (decimal) = minimum fraction of time LPs must spend "on the book" providing their committed liquidity.
- [x] New market parameter: SLA evaluation period length = number of past epochs used to identify worst SLA performance. The reason for this is so that payouts can be made regularly and a bad performance concentrated on a single day can be penalised, but so that performance penalties are not so short lived as to be shrugged off. This may be expected to be 7 or even 30 or more epochs.
- [x] New market parameter: SLA  competition factor, determines how much of their accrued rewards an LP only just achieveing the SLA receives. If this is set to 1 there is no penalty for achieving the SLA but being on the book less than other LPs in practive, if this is set to 0 the LP will receive nothing if they only just achieve the SLA.
- [x] The SLA (and therefore payment of LP rewards) are calculated and applied per epoch. Performance for an LP can be counted during an epoch by a counter of nanoseconds spent on the book meeting the SLA for each LP (we'd need to record the Vega time when they start meeting the SLA [again] or nothing if they aren't meeting it, so when they stop meeting it, the counter can be incremented).
- [x] LPs take a haircut on LP rewards/revenue for each epoch equal to the worst of their downtime penalty for the current epoch and their moving average downtime downtime penalty over the last N epochs where N is the "SLA evaluation period length" parameter. 
- [x] The downtime penalty is 100% if they are below the SLA level, or between zero and `1 - market.liqudity.slaCompetitionFactor` if they achieve 100% performance or the minimum required by the SLA, respectively.
- Example superceded by changes, refer to spec. for more detail. ~So if the SLA is 0.8 (80%, 20% downtime is allowed), and the period length is 7 (NB: this SLA is probably much more generous than would be used in reality, where something like 0.98 or higher may be more reasonable):~
    - ~For an LP that quoted 85% of the time (15% downtime) 6 epochs ago, and 98% of the time in all other epochs, their rewards are scaled by `1-(0.15/0.2) = 0.25` (i.e. a quarter).~
    - ~In 2 epochs time, once the 85% falls outside the period length (7 in this case), if the LP has consistently quoted >=98% of the time in the prior 7 epochs, then their rewards would start to be scaled by `1-(0.02/0.2)=0.9`.~
- [x] Once an LP hits or goes below the worst allowable performance under the SLA, they get a 100% penalty on accrued rewards, i.e. no rewards for that epoch and 100% penalty applied to the moving average for that epoch while it remains within the evaluation period.

#### Committing liquidity

- [x] LPs would commit liquidity as they do now, but without providing any LP "shape" / pegged orders.
- [x] LPs can now always cancel their commitment entirely but will lose their entire bond if they do so when then the target stake is not met. If they reduce their commitment when target stake is not met, they lose a pro-rata amount of their bond based on the amount by which they reduced their commitment.
- [x] If an LP's bond goes below the level implied by their committment (this can only happen in a way thay implies the LP has paid a penalty) then the LP's commitment is reduced pro-rata (and their ELS is reduced, as normal).

#### Providing liquiditing

- [x] Remove LP "shape" and automated, forced deployment of LP pegged orders. LPs are no longer forced to automatically quote all the time. 
- [x] LPs now always and only provide liquidity using normal order types available to all traders. They have full access to features like "post only" and can be "off" the book  _at any time_ simply by cancelling all their orders.
- [x] Add an optional `display quantity` to persistent orders (must be <= order quantity/size). If this is smaller than the full order size then the order is an iceberg order and is effectively split into a series of orders of quantity = display quantity, with only . This will be specced fully if the proposal moves forward.
- [x] Rewards are calculated the same way as before, except for the additional application of the SLA derived penalty to the LPs' relative reward shares.

#### Liquidity monitoring

- [x] LM is still triggered when commitment drops below the trigger ratio and exited when target stake is reached or exceeded.
- [x] Auctions are no longer triggered if there are no priced orders on the book.
- [x] Auctions are not triggered if LPs are off the book so there's less liquidity than committed or less liquidity than required by target stake or the triggering ratio.
- [x] Low supplied liqudiity situations (i.e. 2, 3 above) are assumed not to need to directly trigger an auction because if the lack of liquidity is an issue it should trigger a price monitoring auction on a well configured market.

### Open questions:

- [x] 1. Can we reuse proof-of-stake epochs for this or is there any reason for a separately defined LP epoch? (I can't think of one, I think the same epochs for both is probably fine.) **Yes we can.**
- [x] 2. Should both the SLA and SLA period be market parameters? **Yes.**
- [x] 3. Can we really do without ad-hoc liquidity auctions and rely on price monitoring? **Yes.**
- [x] 5. Is a linear haircut based on time_off_book as a fraction of max allowed time off book under the SLA the right penaltly. This could be non-linear or calculated some other way. **Linear is ok. We have seperated penalty forfeit of rewards and penalty confiscaiton of bond as two distinct concepts in the spec.**
- [x] 7. Is it right to have a max penalty per epoch of the contents of your bond account. **Yes.**
- [x] 8. Is it right to allow LPs to always reduce their commitment? I think maybe this is ok because you lose accured LP share that you've built up by doing so anyway. **Yes this is ok. Key goal is to make sure LPs always have an out with a known max. cost (their bond).**
- [x] 9. Do we still want ghost pegs or similar, or would we just expect LPs to use iceberg non-pegged limits? **We don't need ghost pegs now, yay!**
- [ ] 10. Is there a way to unify some or all of this model with liquidity mechanics for spot. **Yes, this can [mostly?] be the liquidity mechanism for spot.** TODO: make sure it applies fully.
- [ ] 11. Is this even a good way to solve the problem? What alternatives exist? 
- [x] 12. [added 20/04] In response to a good question by @Jiajia-Cui: should we treat an LP failing to meet their SLA either somewhat or even exactly like exiting their commitment for that time? So they could lose all or part of any accrued increase in ELS over the baseline if they had made the same commitment now. **We reduce the LP's commitment if their bond is depleted and not refilled now, and will apply bond penalties in some scenarios, but we wont reduce their commitment just for SLA failure.**

Comments and discussion are welcome, including from the wider community, especially market makers and LPs who have integrated with the protocol to date (we will also happilty accept feedback via private channels if preferred).

davidsiska-vega:
Closes #1665


protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+* `initial peak size` - this specifies the amount displayed and available on the order book for a new or newly refreshed glassberg order.
+
+* `minimum peak size` - this determines when a glassberg order is eligible for refresh.
+The glassberg is refreshed any time the order's displayed quantity less than the minimum peak size.
+
+* `refresh policy` - this specifies when a glassberg order that is eligible for refresh and has remaining volume >0 is refreshed:
+
+    * `IMMEDIATE` - the refresh occurs _after_ processing the transaction that depleted the display quantity to zero.
+
+    * `BLOCK_END` - the rerresh occurs after processing the _entire block_ containing the transaction that depleted the display quantity to zero.
```
TomMcL: These should both be “decreased to less than minimum peak size” right?

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+Glassberg orders are created by populating three additional fields on any valid persistent limit order:
+
+* `initial peak size` - this specifies the amount displayed and available on the order book for a new or newly refreshed glassberg order.
+
+* `minimum peak size` - this determines when a glassberg order is eligible for refresh.
+The glassberg is refreshed any time the order's displayed quantity less than the minimum peak size.
+
+* `refresh policy` - this specifies when a glassberg order that is eligible for refresh and has remaining volume >0 is refreshed:
+
+    * `IMMEDIATE` - the refresh occurs _after_ processing the transaction that depleted the display quantity to zero.
```
TomMcL: Thought, how does this interact with batch transactions? Our current rule is that multiple orders in a batch should execute as if each was an individual transaction. However with glassbergs this would mean that, for immediate refresh ones, you might well get a better price by splitting your order up into as many pieces as the batch allows. Do we care or just assume that’s still not a huge number of orders so probably ok?

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -9,12 +9,11 @@ Similarly to [price monitoring](./0032-PRIM-price_monitoring.md), we need to be
 Note that as long as all pegs that LP batch orders can peg to exists on the book there is one-to-one correspondence between the total stake committed by liquidity providers (LPs), see [LP mechanics](./0044-LIME-lp_mechanics.md) spec, and the total supplied liquidity.
 Indeed
 
-`lp_liquidity_obligation_in_ccy_siskas = stake_to_ccy_siskas ⨉ stake`.
+`lp_liquidity_obligation_in_ccy_volume = market.liquidity.stakeToCcyVolume ⨉ stake`.
 
-Thus it is sufficient to compare `target_stake` with `total_stake` while also ensuring that `best_bid` and `best_offer` are present on the book (*).
+Thus it is sufficient to compare `target_stake`.
```
TomMcL:
```suggestion
Thus it is sufficient to compare `target_stake` with `total_stake`. 
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -40,21 +39,17 @@ If an incoming order would match orders on the book resulting in trades increasi
 
 ### Decreasing supplied stake
 
-If the [liquidity provision transaction would decrease](./0044-LIME-lp_mechanics.md#liquidity-provider-proposes-to-amend-commitment-amount) `supplied_stake` so that liquidity auction gets triggered then the liquidity provision amendment should be rejected and market should stay in it's current trading mode.
+If the [liquidity provision transaction would decrease](./0044-LIME-lp_mechanics.md#liquidity-provider-proposes-to-amend-commitment-amount) `supplied_stake` so that liquidity auction gets triggered then a liquidity auction is triggered the next time the next time conditions for liquidity auctions are evaluated.
```
TomMcL: No more lock-in at all then? I would worry the combination of no lock in and still having liquidity auctions might mean in a crashing market the LPs all pull out and leave people with positions in the market trapped in a liquidity auction?

TomMcL: Might it be possible for liquidity auctions to periodically allow uncrossing of some trades which reduce market OI without actually exiting the auction (I.e it reduces but just not enough to drop below target)

TomMcL: That might breach the simplicity rules though..

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -1,16 +1,56 @@
...
+- `market.liquidity.maximumLiquidityFeeFactorLevel` - used in validating fee amounts that are submitted as part of [lp order type](./0038-OLIQ-liquidity_provision_order_type.md). Note that a value of `0.05 = 5%`. Valid values are: any decimal number `>0` and `<=1`. Default value `1`.
+- `market.liquidity.stakeToCcyVolume` - used to translate a commitment to an obligation. Any decimal number `>0` with default value `1.0`.
+- `validators.epoch.length` - LP rewards from liquidity fees are paid out once per epoch according to whether they met the "SLA" (implied by `market.liquidity.committmentMinTimeFraction`) and their previous performance (for the last n epochs defined by `market.liqudity.performanceHysteresisEpochs`), see [epoch spec](./0050-EPOC-epochs.md).
+- `market.liquidity.earlyExitPenalty` (decimal between 0 and 1 inclusive), sets how much LP forfeits of their bond in case the market is below target stake and they wish to reduce their commitment. If set to `0` there is no penalty for early exit, if set to `1` their entire bond is forfeited.
```
TomMcL: Ah I guess this plays into my concern above!

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -115,38 +141,58 @@ The [liquidity_fee](./0029-FEES-fees.md) of a market on Vega takes as an input,
...
 
-`lp_liquidity_obligation_in_ccy_volume = market.liquidity.stakeToCcyVolume ⨉ stake`.
 
-Note here "ccy" stands for "currency". Liquidity measure units are 'currency x volume'. This is because the calculation is basically `volume ⨉ price of the volume` and the price of the volume is in the said currency.
+Note: we don't evaluate whether LPs meet the SLA during opening auctions so there will always be a mark price. 
 
-### How liquidity is supplied
 
-When a liquidity provider commits to a market, the LP Commitment transaction includes a _buy shape_ and _sell shape_ which allow the LP to spread their liquidity provision over a number of pegged orders at varying distances from the best prices on each side of the book. These 'shapes' are used to generate pegged orders (see the [liquidity provision order type spec](./0038-OLIQ-liquidity_provision_order_type.md)).
+### Penalty for not meeting the SLA
```
TomMcL: Is there anything stopping an LP near the end of the epoch who hasn’t met the SLA just withdrawing their stake (assuming over target stake) and avoiding any penalty?

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
 1. **Reduce-Only (True/False):** Only valid for Non-Persistent orders. Cannot be True at the same time as Post-Only. If set, order will only be executed if the outcome of the trade moves the trader's position closer to 0. In addition, a Reduce-Only order will not move a position to the opposite side to the trader's current position (e.g. if short, a Reduce-Only order cannot make the trader long as a result). If submitted as IOC, where the full volume would switch sides, only the amount required to move the position to 0 will be executed.
 
+
+### 'Glassberg' transparent iceberg orders
+
+On centralised exchanges, icebergs are a type of order that enables a trader to make an order with a relatively small visible "display quantiy" and a larger hidden total size.
```
davidsiska-vega, line 55:
```suggestion
On centralised exchanges, icebergs are a type of order that enables a trader to make an order with a relatively small visible "display quantity" and a larger hidden total size.
```

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+Like an iceberg, most of the order is not visible to other traders.
```
davidsiska-vega, line 56:
```suggestion
Like an iceberg, most of the order is not active on the order book.
```

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+After the full size of the visible portion of the order has traded away, the order is "refreshed" and reappears with its maximum display quantity.
```
davidsiska-vega, line 57:
```suggestion
After the full active portion (specified amount, see below) of the order has traded away, the order is "refreshed" and re-activates with its maximum display quantity.
```

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+This is repeated until the order is cancelled or expires, or its full volume trades away.
+Some platforms also allow the display quantity to be randomised on each refresh to further frustrate traders trying to identify the existence of the iceberg ordeer).
```
davidsiska-vega, line 59:
```suggestion
Some platforms also allow the display quantity to be randomised on each refresh to further frustrate traders trying to identify the existence of the iceberg ordeer).
```

protocol/0014-ORDT-order_types.md
davidsiska-vega:
```suggestion
On centralised exchanges, icebergs are a type of order that enables a trader to make an order with a relatively small visible "display quantity" and a larger hidden total size.
```

protocol/0014-ORDT-order_types.md
davidsiska-vega: I think that's ok; as the submitter of the glassberg has the option to set `BLOCK_END` if they don't want this.

protocol/0014-ORDT-order_types.md
davidsiska-vega:
```suggestion
    * `BLOCK_END` - the rerresh occurs after processing the _entire block_ containing the transaction that depleted the display quantity to less than minimum peak size.
```

protocol/0014-ORDT-order_types.md
davidsiska-vega:
```suggestion
    * `IMMEDIATE` - the refresh occurs _after_ processing the transaction that depleted the display quantity to less than minimum peak size.
```

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+* The order's non-glassberg-related fields must be set so as to make a valid order.
+
+* Any persistent TIF (GTC, GTT, GFA, GFN) can be a glassberg order.
+
+* A glassberg order may have either an ordinary or pegged limit price. 
+Market glassberg orders are not supported, even if with a persistent TIF.
+
+* Glassbergs may be post only.
+
+* `initial peak size` must be greater than TODO
```
davidsiska-vega:
```suggestion
* `initial peak size` must be greater than or equal to minimum position size (i.e. minimum order size).
```

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+* Glassberg orders never trade more than their `displayed quantity` at the start of the transaction, as the result of any one transaction.
+
+* When `displayed quantity < minimum peak size` and `remaining > displayed quantity` the order will be refreshed:
+
+    * The refresh either happens at the end of the transaction when the order became eligible for refresh, if the `refresh policy == IMMEDIATE`; or at the end of the block containing that transaction if the `refresh policy == BLOCK_END`
+
+    * On refresh `display quantity` is set to `min(remaining, initial peak size)`.
+
+* Once the remaning quantity is equal to the displayed quantity, no further refresh is possible.
+The order now behaves like a normal limit order and will leave the book if it trades away completely.
```
davidsiska-vega, line 136: I suppose on refresh all the other limit orders at that level have higher priority for execution (if we're doing the refresh at block end).

protocol/0014-ORDT-order_types.md
```diff
@@ -49,6 +49,117 @@ Notes on scope of current version of this spec:
...
+
+
+#### APIs
+
+* The fields `displayed quantity`, `remaining`, `quantity`, `initial peak size`, `minimum peak size`, `refresh policy` must be exposed by data node APIs in addition to all normal fields for an order.
+
+* Glassberg refresh must generate an event bus event.
+
+
+
```
davidsiska-vega, line 172: I would add a comment that an LP that meets their obligation via glassbergs that refresh at end of block will always meet the SLA commitment (though perhaps that's obvious as we're counting time for SLA and it never moves inside a block...).

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -40,21 +39,17 @@ If an incoming order would match orders on the book resulting in trades increasi
...
 
 ## Trigger for exiting the auction
 
 We exit if
 
-`total_stake >= target_stake AND there is best_bid AND there is best_offer`.
+`total_stake >= target_stake`.
 
-During the liquidity monitoring auction new or existing LPs can commit more stake (and hence liquidity) through the special market making order type and enable this by posting enough margin - see the [liquidity provision mechanics](./0044-LIME-lp_mechanics.md) spec for details. These need to be monitored to see if auction mode can be exit.
+During the liquidity monitoring auction new or existing LPs can commit more stake (and hence liquidity) through the special market making order type and enable this by posting enough margin - see the [liquidity provision mechanics](./0044-LIME-lp_mechanics.md) spec for details. These need to be monitored to see if auction mode can be exitted.
```
witgaw:
```suggestion
During the liquidity monitoring auction new or existing LPs can commit more stake (and hence liquidity) through the special market making transaction and enable this by posting enough margin - see the [liquidity provision mechanics](./0044-LIME-lp_mechanics.md) spec for details. These need to be monitored to see if auction mode can be exited.
```
Wondering if it still makes sense to refer to these as "order types", though it would probably require a broader renaming across other specs.

protocol/0014-ORDT-order_types.md
barnabee: I think we have two issues I might like to prevent that this has highlighted to me:

1. If an `IMMEDTIATELY` refreshing glassberg (assuming not post only) is crossed and trades on entry it will refresh and trade again. For a large quantity/small peak against significant volume this could create a large number of loops. Better to treat a glassberg as its full remaining size on entry during aggressive execution (this does not conflict with "disaplayed" quantity as aggressive size that trades on entry is never displayed on the book).

2. Although this is different to the usual policy for batch orders, perhaps we should make clear that glassberg refreshes happen at transaction end (_strictly_ as specified). I think it is a fairness violation / abuse of batch orders to be able to atomically take more than the `initial peak size` in a single transaction.

Thoughts?

protocol/0035-LIQM-liquidity_monitoring.md
barnabee: > No more lock-in at all then? I would worry the combination of no lock in and still having liquidity auctions might mean in a crashing market the LPs all pull out and leave people with positions in the market trapped in a liquidity auction?

No lock in but you do lose some or all of your bond for quitting if below target stake.

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -193,53 +201,104 @@ Then on every Vega time change, after `fractional instantaneous liquidity score`
...
-LP 2 els = 0.17
-LP 3 els = 0.07
-LP 4 els = 0.33
+        * Add the difference in nanoseconds between the current Vega time and the time the LP began meeting their commitment (stored in the step above) to `s_i`
+        
+        * Store `nothing` as the time the LP began meeting their commitment, to signify the LP not meeting their commitment
+
+* At the end of the epoch, calculate the actual observed epoch length `observed_epoch_length` = the difference in nanoseconds between the Vega time at the start of the epoch and the Vega time at the end of the epoch.
+
+Note that we only need to evaluate each LP's status at the end of each block, as no _Vega time_ passes between transactions in a block.
```
witgaw: Are we safeguarding against an LP trying to place the transactions so that they cancel their orders around the first transaction in the block an reinstate them towards the end of the block (so as to appear as having been actively providing liquidity throughout the block)? 
This may be covered, but I can't convince myself, so might be worth spelling it out.

protocol/0035-LIQM-liquidity_monitoring.md
barnabee: > Might it be possible for liquidity auctions to periodically allow uncrossing of some trades which reduce market OI without actually exiting the auction (I.e it reduces but just not enough to drop below target)

I like the idea of an uncrossing where the market remains in auction after uncrossing, that can only occur if OI is reduced. @TomMcL would you be able to take a look at how this would fit with the relevant liquidity monitoring and auction specs?

protocol/0044-LIME-lp_mechanics.md
barnabee: Good point!

There are three options I can think of:

1. Unforced (i.e. not by bond confiscation) commitment reductions only take effect at epoch end. New commitments take effect at epoch start, too, I guess to avoid unfair penalties.

2. Your SLA is evaluated at epoch end for the time period for which you still had a commitment. This is a bit nasty because what if you have multiple different commitment levels during an epoch.

3. Yolo! Leave it and LPs can deal with it.

I'm leaning towards 1. Thoughts @TomMcL @davidsiska-vega @witgaw ?

protocol/0014-ORDT-order_types.md
barnabee: Yes - latest commit this AM clarifies this and also that as increasing size amendments add to remaining but not display quantity they *do not lose time priority* unlike for normal orders.

protocol/0014-ORDT-order_types.md
barnabee: Added the following explicitly to [0042](https://github.com/vegaprotocol/specs/pull/1685/files#diff-71887ef730ec45323f5184892e551217f5e434a46bbabc0574a33a2a2d2d837aR230-R231):

> * Note that this happens _after_ any glassberg order refreshes, including those with `refresh policy == BLOCK_END`. This means that while a glassberg has sufficient `remaining` quantity, it will **never** be considered to be contributing less than its `minimum peak size` of liquidity.

protocol/0035-LIQM-liquidity_monitoring.md
barnabee: Agree this should be a transaction (I guess "liquidity commitment transaction") now.

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
barnabee: I don't think we are and I can't quite work out how big a problem this is, but I think probably big enough to do something about…

I think the LP's success rate in timing this is key to how effective an attack it might be. Each block where you "miss" re-upping your liquidity at the end will take you off book until the next block, so you can have `1-SLA%` chance of missing before you miss the SLA anyway, which may not be realistic. Though you would potentially manage to reduce your penalty even in this case.

That's assuming you're dropping fully to zero. There's also a chronic underprovision attack which I think is probably simpler and made easier by `refresh policy = BLOCK_END` glassbergs, as these help you to always be meeting the provision at the end of the block without doing so during it.

I prefer David's soluton below
~We could instead evaluate LPs' SLA achievement after each transaction (1=SLA met, 0=not met) and use the average of these for the block multiplied by the block length in nanos to decide how much additional time the LP spent meeting their commitment "during" the block.~

~This would seem strange in a way given how time evolves vs. transactions, but might be fairer. The strange thing is that the liquidity that's "visible" (i.e. on the book between blocks) exists for all the "Vega time" (and is what people see) but a tiny fraction of the considered provision. It would also make `refresh policy = BLOCK_END` pretty useless, so if we do this I'd propose removing that (in fact that may be an _enabler_ of under-provision in the way you describe).~

~This "solution" would create an incentive to supply liqudity then make many transactions that don't trade with it or only trade a little, though. That could perhaps be solved by excluding transactions that don't trade on the market, and weighting the average described above by the volume traded by each transaction.~

~Either solution increases the amount of work done per LP per blockto measure SLA performance from from O(1) to O(n) (n of [trading] transactions), and increases the complexity of the calculation a bit too.~

How else might we solve this?

witgaw: Coule we use the minimum liquidity within block? We'd then need to monitor all cancels and amends within the block and exclude glassberg fills from updating the minimum.

protocol/0044-LIME-lp_mechanics.md
witgaw: 1 sounds nice and clean to me

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
witgaw: So really just calculate it at the beginning of the block and monitor cancels, amends and fills (of non-glassberg orders) within the block to update the minimum.

witgaw: Or we just don't care about the fills at all, but not sure if it doesn't open a possibility of wash trading back and forth between two LPs.

barnabee: @davidsiska-vega came up with a much better and simpler solution:

We randomise the point in the block at which we determine whether an LP met the SLA for that block.

So for a block containing _N_ transactions (not _N_ orders, a batch is 1 txn), we pick `random_integer_between(0, N)` and asses the LP's liquidity after that transaction. We could seed the RNG with something like the hash of all transaction hashes in the block or the previous state/block hash.

Glassbergs with `BLOCK_END` would still refresh at block end, so to be certain they are providing some amount of liquidity every time you would need to use that as `minimum peak size` rather than `initial peak size`. I think this is ok.

protocol/0014-ORDT-order_types.md
barnabee: The above note will be amended if we do [this](https://github.com/vegaprotocol/specs/pull/1685#discussion_r1182395475)

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
barnabee: I have added the above to the PR.

protocol/0014-ORDT-order_types.md
barnabee: Amended, the note now reads as below:

>   * Note that this happens _before_ any glassberg order refreshes with `refresh policy == BLOCK_END`, even if `k == N`, and after glassbergs refreshed with `refresh policy == IMMEDIATELY` that need refreshing as a result of transaction `k`. This means that while a glassberg has sufficient `remaining` quantity, it will **never** be considered to be contributing less than its `initial peak size` of liquidity if it's policy is `IMMEDIATELY`. A `BLOCK_END` glassberg will **never** be refreshed within before the SLA achievement is evaluated for that block, and can therefore only be guaranteed to be contributing at least `minimum peak size` of liquidity, even when it has plenty of remaining volume.

barnabee:
Glassberg open question:

How to handle a passive glassberg against an aggressive persistent limit it crosses with, where the limit is > peak size and they would therefore be crossed after refresh.

protocol/0014-ORDT-order_types.md
barnabee: I updated to include the above btw, please shout if in disagreement with this

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
barnabee: > Coule we use the minimum liquidity within block? We'd then need to monitor all cancels and amends within the block and exclude glassberg fills from updating the minimum.

We _could_ use their minimum. I think that is somewhat harsh compared to what MMs will be used to. IIRC CEXs sample the data (less often than per block, even) cc @MM0819 is that correct?

MM0819: They are roughly doing this:

Randomly sample your orders on the book within 0.3% (for example) twice per minute. Then take the average of all of these samples over the day, and require the average to be > $100k (again, for example). You retain the additional MM benefits if you satisfy this requirement on more than 90% of the days within the month. In the case of CEXs, this will be negative maker fees.

This is nice because it allows you to quote e.g. $1M of depth during times of low volatility (weekends, when US markets are closed, etc), and quote $0 during times of predictably high volatility (e.g. non-farm payrolls, CPI, FOMC etc).

MM0819: I should add that the penalty is quite harsh for not meeting these requirements. If you do not satisfy these requirements in two consecutive months you are kicked out of the market making program completely... :smile:

MM0819: So, if we propose to have less strict penalties, we can surely have stricter requirements.

barnabee: Thanks, that’s useful info!

The actual strictness (amounts, depth, distance from best price) will be configurable by market and in any case the rewards will be competitive.

So the SLA to not get a penalty against you will be pretty permissive, but if others are consistently more competitive than you you’ll start to get less rewards pretty quickly (like a more advanced version of your “if they commit $10 I’ll commit $1000 and take all their fees” but measuring their actual performance on the SLA KPIs).

TomMcL:
> Glassberg open question:
> 
> How to handle a passive glassberg against an aggressive persistent limit it crosses with, where the limit is > peak size and they would therefore be crossed after refresh.

Assuming that is still an open question, I would say it uses up the visible chunk completely but waits until next refresh to use up the next chunk. We could hopefully upfront calculate the total traded volume if the glassberg is immediate refreshing to avoid too many loops in that case. (Perhaps they should still display as separate trades though to show that the glassberg contract is being honoured?)

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -19,87 +59,81 @@ Any Vega participant can apply to become a liquidity provider (LP) on a market b
...
-### liquidity provider proposes to amend commitment amount
+### Liquidity provider proposes to amend commitment amount
 
 The commitment transaction is also used to amend any aspect of their liquidity provision obligations.
 A participant may apply to amend their commitment amount by submitting a transaction for the market with a revised commitment amount.
 
 `proposed-commitment-variation = new-proposed-commitment-amount - old-commitment-amount`
 
-#### INCREASING COMMITMENT
+The amendment is actioned at the beginning of the next epoch (after the rewards / penalties for present LPs - including the party that's amending - have been evaluated). It's the amending LP's responsibility to have sufficient balance in their general account at the epoch boundary; if not the LP amendment is not actioned.
```
TomMcL: Thought on this, only allowing amending liquidity provisions at the end of the epoch is going to mean, for all intents and purposes and without the OI reducing stages in an auction, that dropping into a liquidity auction often means we stay there until the end of the epoch before we even have a chance to exit right? And if we miss that we wait until the next to try again?

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -1,16 +1,56 @@
...
+
+## Network and market parameters
+
+### Network parameters
+
+- `market.liquidity.bondPenaltyParameter` - used to calculate the penalty to liquidity providers when they fail to support their open position through sufficient `general+margin` balance.
+Valid values: any decimal number `>= 0` with a default value of `0.1`.
+- `market.liquidity.sla.nonPerformanceBondPenaltySlope` - used to calculate how much is the LP bond slashed if they fail to reach the minimum SLA. Valid values: any decimal number `>= 0` with a default value of `2.0`.
+- `market.liquidity.sla.nonPerformanceBondPenaltyMax` - used to calculate how much is the LP bond slashed if they fail to reach the minimum SLA. Valid values: any decimal number `>= 0` and `<=1.0` with a default value of `0.5`.
+- `market.liquidity.maximumLiquidityFeeFactorLevel` - used in validating fee amounts that are submitted as part of the LP commitment transaction. Note that a value of `0.05 = 5%`. Valid values are: any decimal number `>0` and `<=1`. Default value `1`.
```
```suggestion
- `market.liquidity.maximumLiquidityFeeFactorLevel` - used in validating fee amounts that are submitted as part of the LP commitment transaction. Note that a value of `0.05 = 5%`. Valid values are: any decimal number `>=0` and `<=1`. Default value `1`.
```

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -1,16 +1,56 @@
...
+- `market.liquidity.minimum.probabilityOfTrading.lpOrders` sets a lower bound on the result of the probability of trading calculation.
+
+### Market parameters
+
+- `market.liquidity.priceRange` (decimal) - this is a percentage price move (e.g. `0.05 = 5%`) from `mid_price` during continuous trading or indicative uncrossing price during auctions. This is set / can be modified as part of [market proposal](0028-GOVE-governance.md) / market change proposal.
+
+- `market.liquidity.committmentMinTimeFraction` (decimal) —  minimum fraction of time LPs must spend "on the book" providing their committed liquidity. This is a decimal number in the interval $[0,1]$ i.e. both limits included. When set to $0$ the SLA mechanics are switched off for the market entirely.
+
+- `market.liquidity.providers.fee.calculationTimeStep` (time period e.g. `12h` or `7d`) controls how often LP the quality of liquidity supplied by the LPs is evaluated and fees arising from that period are earmarked for specific parties.
```
```suggestion
- `market.liquidity.providers.fee.calculationTimeStep` (time period e.g. `1m`) controls how often LP the quality of liquidity supplied by the LPs is evaluated and fees arising from that period are earmarked for specific parties. 
```

protocol/0044-LIME-lp_mechanics.md
davidsiska-vega: Good point, we need to fix this @barnabee.

```suggestion
The amendment is actioned in two steps. 
1) the amount is immediately transferred from the party's general account to a temporary "pending" bond account. This amount counts towards the stake committed to the market and so in particular can get the market out of liquidity auction. 
2) at the beginning of the next epoch (after the rewards / penalties for present LPs - including the party that's amending - have been evaluated) the amount is transferred from the "pending" bond to the true bond account. ```

Comment (TomMcL) [reply protocol/0044-LIME-lp_mechanics.md]: Do they earn fees on this? I would assume they must otherwise why would someone do it, but then they could presumably increase them decrease again just before the epoch ends?

Comment (TomMcL) [reply protocol/0044-LIME-lp_mechanics.md]: This also could be a pain to track if someone added multiple times in an epoch. But I can't think of much else to allow intra-epoch adding without some sort of secondary account, hmm

Comment (TomMcL) [reply protocol/0044-LIME-lp_mechanics.md]: Thinking more, it might just work allowing increases whenever and decreases only at epoch end.

- Liquidity fees themselves are calculated and assigned at time of each trade, so you don't get a claim over historic fees in the epoch by submitting later
- SLA is based on your time on the book. We should ensure that if someone places their first commitment mid way through the block it counts as them having met their SLA perfectly until that point.
- Bonus payments are weighted by total fees received, so increasing later in the epoch should make that proportionally smaller. However it means your performance score is much more likely to be perfect, so I'm not quite sure how that ends up playing off against each other.

File: protocol/0044-LIME-lp_mechanics.md [hunk 19]
```diff
@@ -19,87 +59,81 @@ Any Vega participant can apply to become a liquidity provider (LP) on a market b
...
 
-If they do not have sufficient collateral the transaction is rejected in entirety. This means that any data from the fees or orders are not applied. This means that the  `old-commitment-amount` is retained.
+If they do not have sufficient collateral the transaction is rejected in entirety. This means that any change in fee bid is not applied and that the `old-commitment-amount` is retained.
 
-#### DECREASING COMMITMENT
+#### Decreasing commitment
 
 _Case:_ `proposed-commitment-variation < 0`
-We to calculate whether the liquidity provider may lower their commitment amount and if so, by how much. To do this we first evaluate the maximum amount that the market can reduce by given the current liquidity demand in the market.
+We to calculate how much the LP can reduce commitment without incurring a penalty.
```
TomMcL: Writing the ACs for this a further thought has come up. If we now only process these reductions in commitment at the end of an epoch there's a question of sequencing and thus penalties which might not be immediately obvious when first placing your reduction transaction (e.g. if someone else reduces their commitment such that the stake is much closer to target than it was when you first placed your reduction). 

The 'easiest' solution would just be to say we action these in the order they are placed, and LPs should just take this into account when they place their reductions, but we should probably then at least be explicit somewhere that reductions are executed in order of receipt.

davidsiska-vega: @TomMcL that's a good point, lets add a line saying that reductions are processed in the sequence they were placed; for each LP it's the most recent that counts.

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -193,52 +201,118 @@ Then on every Vega time change, after `fractional instantaneous liquidity score`
...
-If `market.liquidity.providers.fee.distributionTimeStep` is set to `0` then the balance is distributed either immediately upon collection or at then end of a block.
+The liquidity fees are transferred from the market LP fee account into the LP-per-market fee account, pro-rata depending on the `LP i equity-like share` multiplied by `LP i liquidity score` scaled back to `1` across all LPs at a given time.
 
-The liquidity fees are distributed pro-rata depending on the `LP i equity-like share` multiplied by `LP i liquidity score` scaled back to `1` across all LPs at a given time.
+The LP parties don't control the LP-per-market fee account; the fees from there are then tranferred to the LPs' general account at the end epoch as described below.
```
candida-d:
```suggestion
The LP parties don't control the LP-per-market fee account; the fees from there are then transferred to the LPs' general account at the end epoch as described below.
```

protocol/0014-ORDT-order_types.md
```diff
@@ -46,9 +46,127 @@ Notes on scope of current version of this spec:
...
+* Icebergs can be entered or carried into auctions if the underlying TIF is supported. 
+
+* Icebergs can trade in the auction uncrossing up to their current `displayed quantity` as for any other transaction that would cause a trade with an iceberg order.
+
+* Icebergs are refreshed after an auction uncrossing if they traded to below their `minimum peak size`, according to the same rules as for normal execution.
+
+
+#### APIs
+
+* The fields `displayed quantity`, `remaining`, `quantity`, `initial peak size`, `minimum peak size`, `refresh policy` must be exposed by data node APIs in addition to all normal fields for an order.
```
MuthuVega, line 168: What is `refresh policy` ?
Reading the comments below, its either refresh policy == BLOCK_END or refresh policy == IMMEDIATELY

protocol/0044-LIME-lp_mechanics.md
davidsiska-vega: Done.

protocol/0014-ORDT-order_types.md
```diff
@@ -46,9 +46,129 @@ Notes on scope of current version of this spec:
...
+
+* Once the remaining quantity is equal to the displayed quantity, no further refresh is possible.
+The order now behaves like a normal limit order and will leave the book if it trades away completely.
+
+* For an incoming order with size larger than the total displayed quantity at a given price level, the following procedure should be followed:
+  
+  * The incoming order trades with all visible volume at the price level, whether an iceberg order or a vanilla limit order
+
+  * If there is still remaining volume in the order once all visible volume at the price level is used up, the remaining quantity is split between the non-visible components of all iceberg orders at this level according to their remaining volumes. For example if there are three iceberg orders with remaining volume 200 lots, 100 lots and 100 lots, an order for 300 lots would be split 150 to the first order and 75 to the two 100 lot orders.
```
wwestgarth: Any preference on where we put the crumbs in the case where the distribution doesn't divide exactly? For example there is 100 left in the aggressive order and 3 icebergs with all equal remaining hidden volume so it'll be 33 each. Just stick the extra 1 onto the iceberg which is first in terms of time priority?

davidsiska-vega: Yes, let's. I'll update the spec.

protocol/0014-ORDT-order_types.md
```diff
@@ -76,20 +196,81 @@ Notes on scope of current version of this spec:
...
+
+#### Iceberg Order Cancellation
+
+1. Cancelling an iceberg order will cancel the order, remove it from the order book , release margin and update order book to reflect the change (<a name="0014-ORDT-026" href="#0014-ORDT-026">0014-ORDT-026</a>)
+
+#### Iceberg Order Execution
+
+1. An aggressive iceberg order that crosses with an order where volume > iceberg volume, the iceberg order gets fully filled on entry, the iceberg order status is filled, the remaining quantity = 0. Atomic trades are generated if matched against multiple orders (<a name="0014-ORDT-027" href="#0014-ORDT-027">0014-ORDT-027</a>)
+2. An aggressive iceberg order that crosses with an order where volume < iceberg volume. The initial display quantity is filled and the remaining volume is unfilled. Status of iceberg order is partially filled , the volume remaining = (quantity - initial volume) and the remaining volume sits on the book. When additional orders thar are submitted which consume the remaining volume on the iceberg order , the volume of the iceberg order is refreshed as and when the volume dips below the minimum peak size (<a name="0014-ORDT-028" href="#0014-ORDT-028">0014-ORDT-028</a>)
```
wwestgarth: Normal orders which trade when aggressive and then sit on the book are ACTIVE so I don't think icebergs should be different (partially filled is only set for IOC orders)
```suggestion
2. An aggressive iceberg order that crosses with an order where volume < iceberg volume. The initial display quantity is filled and the remaining volume is unfilled. Status of iceberg order is active , the volume remaining = (quantity - initial volume) and the remaining volume sits on the book. When additional orders are submitted which consume the remaining volume on the iceberg order , the volume of the iceberg order is refreshed as and when the volume dips below the minimum peak size (<a name="0014-ORDT-028" href="#0014-ORDT-028">0014-ORDT-028</a>)
```

protocol/0014-ORDT-order_types.md
```diff
@@ -46,9 +46,129 @@ Notes on scope of current version of this spec:
...
+* Amending the size/quantity of an iceberg order does not cause it to lose time priority.
+This is because the increase applies to the `remaining` quantity and not to the `displayed quantity`.
+This is allowed because the order will lose time priority on refresh, i.e. before the increased quantity is available to trade.
+
+
+#### Auctions
+
+* Icebergs can be entered or carried into auctions if the underlying TIF is supported.
+
+* Icebergs can trade in the auction uncrossing up to their current `displayed quantity` as for any other transaction that would cause a trade with an iceberg order.
```
wwestgarth: (hopefully the last comment): if we have an iceberg order on each side of the book and uncross only with their visible volume when coming out of auction, they'll refresh and remain crossed. When in continuous trading we don't have this problem because an aggressive iceberg counts as its full volume.

Thoughts on what we can do? Do we have the notion of an "aggressor" when auction uncrossing? Does it make sense to "un-iceberg" one side while uncrossing?

davidsiska-vega: We don't have a notion of aggressor when uncrossing (no maker fee, remaining fees are split equally between both sides). 

I guess it makes more sense to trade the whole volume (not just visible) at uncrossing. If the party placing the iceberg doesn't want this they can mark it GFN.

barnabee: Agree whole volume. 

It's all on the price level after all. It should be split into the peaks at their position on the book and the rest at lower time priority than everything else when executing. If the remainders of multiple icebergs sitt at the same prixe and are not fully used up the traded volume should be split between them pro-rata by their total total size.

That means the algo for doing execution in an auction should be able to reuse the one for executing against passive icebergs as it's the same.

wwestgarth (dismissed):

karlem (dismissed):

wwestgarth (dismissed):

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -19,87 +59,88 @@ Any Vega participant can apply to become a liquidity provider (LP) on a market b
...
 
 The commitment transaction is also used to amend any aspect of their liquidity provision obligations.
 A participant may apply to amend their commitment amount by submitting a transaction for the market with a revised commitment amount.
 
 `proposed-commitment-variation = new-proposed-commitment-amount - old-commitment-amount`
 
-#### INCREASING COMMITMENT
+The amendment is actioned in two steps.
+
+1) the amount is immediately transferred from the party's general account to a temporary "pending" bond account. This amount counts towards the stake committed to the market and so in particular can get the market out of liquidity auction.
```
witgaw, line 105: Do we really need another type of account? Could it not just go directly into the bond account with the understanding that the commitment only becomes binding at the beginning of the next epoch?

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -19,87 +59,88 @@ Any Vega participant can apply to become a liquidity provider (LP) on a market b
...
-If they do not have sufficient collateral the transaction is rejected in entirety. This means that any data from the fees or orders are not applied. This means that the  `old-commitment-amount` is retained.
+If they do not have sufficient collateral the transaction is rejected in entirety. This means that any change in fee bid is not applied and that the `old-commitment-amount` is retained.
 
-#### DECREASING COMMITMENT
+#### Decreasing commitment
 
 _Case:_ `proposed-commitment-variation < 0`
-We to calculate whether the liquidity provider may lower their commitment amount and if so, by how much. To do this we first evaluate the maximum amount that the market can reduce by given the current liquidity demand in the market.
+We to calculate how much the LP can reduce commitment without incurring a penalty.
+To do this we first evaluate the maximum amount that the market can reduce without penalty by given the current liquidity demand in the market.
```
witgaw:
```suggestion
We to calculate how much the LP can reduce commitment without incurring a penalty.
To do this we first evaluate the maximum amount that the market can reduce without penalty given by the current liquidity demand in the market.
```

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -115,38 +156,71 @@ The [liquidity_fee](./0029-FEES-fees.md) of a market on Vega takes as an input,
...
+## Liquidity provision and penalties
 
-A liquidity provider can amend their orders by providing a new set of liquidity provision orders in the liquidity provider network transaction. If the amended orders are invalid the transaction is rejected, and the previous set of orders will be retained.
+### Calculating liquidity from commitment
 
-### Checking margins for orders
+Committed Liquidity Providers are required to provide a multiple of their stake (supplied in the settlement currency of the market) in notional volume of orders within the range defined below.
```
witgaw: So it's no longer per book side? That is an LP which posts only buy orders within the LP range now fully meets their commitment?

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -115,38 +156,71 @@ The [liquidity_fee](./0029-FEES-fees.md) of a market on Vega takes as an input,
...
-### Calculating liquidity from commitment
+### Meeting the committed volume of notional
+
+#### During continuous trading
+
+If there is no mid price then no LP is meeting their committed volume of notional.
+If there is mid price then we calculate the volume of notional that is in the range
+
+```text
+(1.0-market.liquidity.priceRange) x mid =< price levels that count <= (1+market.liquidity.priceRange)x mid.
```
witgaw:
```suggestion
(1.0-market.liquidity.priceRange) x mid <= price levels <= (1+market.liquidity.priceRange)x mid.
```

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -115,38 +156,71 @@ The [liquidity_fee](./0029-FEES-fees.md) of a market on Vega takes as an input,
...
 
-`lp_liquidity_obligation_in_ccy_volume = market.liquidity.stakeToCcyVolume ⨉ stake`.
+#### During auctions
 
-Note here "ccy" stands for "currency". Liquidity measure units are 'currency x volume'. This is because the calculation is basically `volume ⨉ price of the volume` and the price of the volume is in the said currency.
+We calculate the volume of notional that is in the range
 
-### How liquidity is supplied
+```text
+(1.0-market.liquidity.priceRange) x min(last trade price, indicative uncrossing price) =<  price levels that count <= (1.0+market.liquidity.priceRange) x max(last trade price, indicative uncrossing price).
```
witgaw:
```suggestion
(1.0-market.liquidity.priceRange) x min(last trade price, indicative uncrossing price) <=  price levels <= (1.0+market.liquidity.priceRange) x max(last trade price, indicative uncrossing price).
```

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -173,30 +247,26 @@ The network will:
 
 Note:
 
-- As mentioned above, closeout should happen as per regular trader account (with the addition of cancelling the liquidity provision and the associated LP rewards & fees consequences). So, if after cancelling all open orders (both manually maintained and the ones created automatically as part of liquidity provision commitment) the party can afford to keep the open positions sufficiently collateralised they should be left open, otherwise the positions should get liquidated.
-- Bond account balance should never get directly confiscated. It should only be used to cover margin shortfalls with appropriate penalty applied each time it's done. Once the funds are in margin account they should be treated as per normal rules involving that account.
+- As mentioned above, closeout should happen as per regular trader account (with the addition of cancelling the liquidity provision and the associated LP rewards & fees consequences). So, if after cancelling all open orders the party can afford to keep the open positions sufficiently collateralised they should be left open, otherwise the positions should get liquidated.
```
witgaw, line 264: Is the early exit penalty applies in that case shouldn't we still charge it from the margin account?

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -173,30 +247,26 @@ The network will:
...
 
-- `market.liquidity.bondPenaltyParameter` - used to calculate the penalty to liquidity providers when they fail to meet their obligations.
-Valid values: any decimal number `>= 0` with a default value of `0.1`.
-- `market.liquidity.maximumLiquidityFeeFactorLevel` - used in validating fee amounts that are submitted as part of [lp order type](./0038-OLIQ-liquidity_provision_order_type.md). Note that a value of `0.05 = 5%`. Valid values are: any decimal number `>0` and `<=1`. Default value `1`.
-- `market.liquidity.stakeToCcySiskas` - used to translate a commitment to an obligation (in siskas). Any decimal number `>0` with default value `1`.
 
 ## What data do we keep relating to liquidity provision?
 
 1. List of all liquidity providers and their commitment sizes and their “equity-like share” for each market [see 0042-setting-fees-and-rewarding-lps](./0042-LIQF-setting_fees_and_rewarding_lps.md)
```
witgaw:
```suggestion
1. List of all liquidity providers and their commitment sizes, their “equity-like share” and "liquidity score" for each market [see 0042-setting-fees-and-rewarding-lps](./0042-LIQF-setting_fees_and_rewarding_lps.md)
```

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -207,9 +277,30 @@ Valid values: any decimal number `>= 0` with a default value of `0.1`.
...
+- If a liquidity provider with an active liquidity provision at the start of an epoch reduces their liquidity provision staked commitment during the epoch the initial committed level at the start of the epoch will remain in effect until the end of the epoch, at which point the protocol will attempt to reduce the bond to the new level. If the reduced level has been changed several times during an epoch, only the latest value will take effect (<a name="0044-LIME-019" href="#0044-LIME-019">0044-LIME-019</a>)
+- If a liquidity provider with an active liquidity provision at the start of an epoch reduces their liquidity provision staked commitment during the epoch the initial committed level at the start of the epoch will remain in effect until the end of the epoch, at which point the protocol will attempt to reduce the bond to the new level. If the bond stake has been slashed to a level lower than the amendment, this slashed level will be retained (i.e. the protocol will not attempt to now increase the commitment) (<a name="0044-LIME-020" href="#0044-LIME-020">0044-LIME-020</a>)
```
witgaw, line 297: Doesn't this contradict paragraph at line 253? It's not sufficiently clear to be when do we change their ELS and when (if at all) we change their commitment size.

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -207,9 +277,30 @@ Valid values: any decimal number `>= 0` with a default value of `0.1`.
...
+- For a market with `market.liquidity.earlyExitPenalty = 0.25` and `total stake = target stake + 40` already, a liquidity provider who reduces their commitment by `100` will receive a total of `85` back into their general account with `15` transferred into the market's insurance account (`40` received without penalty, then the remaining `60` receiving a `25%` penalty). (<a name="0044-LIME-024" href="#0044-LIME-024">0044-LIME-024</a>)
+
+- For a market with `market.liquidity.earlyExitPenalty = 0.25` and `total stake = target stake + 140` already, if one liquidity provider places a transaction to reduce their stake by `100` followed by a second liquidity provider who reduces their commitment by `100`, the first liquidity provider will receive a full `100` stake back whilst the second will receive a total of `85` back into their general account with `15` transferred into the market's insurance account (`40` received without penalty, then the remaining `60` receiving a `25%` penalty). (<a name="0044-LIME-025" href="#0044-LIME-025">0044-LIME-025</a>)
```
witgaw, line 304: Wouldn't it make sense to only assess this at the end of epoch and apply pro rata the penalty free amounts among those wishing to leave?

I know it's an additional complication, but wondering if without it we wouldn't have LPs rushing towards the exit at certain times to avoid the penalty or even spamming the network with commitment reduction transaction (to reserve a penalty-free reduction spot for themselves) which they'll then cancel shortly before the end of an epoch if they decide to stay on.

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -1,16 +1,56 @@
...
+- `market.liquidity.minimum.probabilityOfTrading.lpOrders` sets a lower bound on the result of the probability of trading calculation.
+
+### Market parameters
+
+- `market.liquidity.priceRange` (decimal) - this is a percentage price move (e.g. `0.05 = 5%`) from `mid_price` during continuous trading or indicative uncrossing price during auctions. This is set / can be modified as part of [market proposal](0028-GOVE-governance.md) / market change proposal.
+
+- `market.liquidity.committmentMinTimeFraction` (decimal) —  minimum fraction of time LPs must spend "on the book" providing their committed liquidity. This is a decimal number in the interval $[0,1]$ i.e. both limits included. When set to $0$ the SLA mechanics are switched off for the market entirely.
+
+- `market.liquidity.providers.fee.calculationTimeStep` (time period e.g. `1m`) controls how often LP the quality of liquidity supplied by the LPs is evaluated and fees arising from that period are earmarked for specific parties.
```
witgaw:
```suggestion
- `market.liquidity.providers.fee.calculationTimeStep` (time period e.g. `1m`) controls how often the quality of liquidity supplied by the LPs is evaluated and fees arising from that period are earmarked for specific parties.
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -40,21 +39,17 @@ If an incoming order would match orders on the book resulting in trades increasi
 
 ### Decreasing supplied stake
 
-If the [liquidity provision transaction would decrease](./0044-LIME-lp_mechanics.md#liquidity-provider-proposes-to-amend-commitment-amount) `supplied_stake` so that liquidity auction gets triggered then the liquidity provision amendment should be rejected and market should stay in it's current trading mode.
+If the [liquidity provision transaction would decrease](./0044-LIME-lp_mechanics.md#liquidity-provider-proposes-to-amend-commitment-amount) `supplied_stake` so that liquidity auction gets triggered then a liquidity auction is triggered the next time the next time conditions for liquidity auctions are evaluated.
```
witgaw:
```suggestion
If the [liquidity provision transaction would decrease](./0044-LIME-lp_mechanics.md#liquidity-provider-proposes-to-amend-commitment-amount) `supplied_stake` so that liquidity auction gets triggered then a liquidity auction is triggered the next time conditions for liquidity auctions are evaluated.
```

protocol/0035-LIQM-liquidity_monitoring.md
```diff
@@ -40,21 +39,17 @@ If an incoming order would match orders on the book resulting in trades increasi
...
 
-If the `supplied_stake` decreases as a result of a closeout of an insolvent liquidity provider, then closeout should proceed and market should go into liquidity auction.
-
-### Removing `best_bid` or `best_offer`
-
-If an incoming order would get matched so that entire side of the order book gets consumed and `best_bid` or `best_offer` no longer exists, then the order should be allowed to go through and market should go into liquidity auction after it gets matched and the resulting trades get generated.
+If the `supplied_stake` decreases as a result of a closeout of an insolvent liquidity provider, then closeout should proceed and market should go into liquidity auction the next time the next time conditions for liquidity auctions are evaluated.
```
witgaw:
```suggestion
If the `supplied_stake` decreases as a result of a closeout of an insolvent liquidity provider, then closeout should proceed and market should go into liquidity auction the next time conditions for liquidity auctions are evaluated.
```

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -200,59 +201,124 @@ Then on every Vega time change, after `fractional instantaneous liquidity score`
...
+For each LP where `fraction_of_time_on_book ≥ market.liquidity.commitmentMinTimeFraction`, the SLA penalty fraction `p_i` is calculated as follows:
+
+
+Let $t$ be `fraction_of_time_on_book`
+
+Let $s$ be `market.liquidity.commitmentMinTimeFraction`.
+
+Let $c$ be `market.liquidity.slaCompetitionFactor`.
 
 ```text
```
witgaw:
```suggestion
```

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -200,59 +201,124 @@ Then on every Vega time change, after `fractional instantaneous liquidity score`
...
-0.65 x 103.5 = 67.275 ETH to LP 1's LP account
-0.25 x 103.5 = 25.875 ETH to LP 2's LP account
-0.10 x 103.5 = 10.350 ETH to LP 3's LP account
+$$
+p_i = (1 - \frac{t - s}{1 - s}) \cdot c.
+$$
 ```
```
witgaw:
```suggestion
```

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -200,59 +201,124 @@ Then on every Vega time change, after `fractional instantaneous liquidity score`
...
 
-Then LP 4 made a delayed LP commitment, and updated share as below:
+#### Calculating the SLA performance penalty for over hysteresis period
+
+Now, for each LP $i$ take the $p_i$ values calculated over the last `market.liqudity.performanceHysteresisEpochs - 1`, call these $p_i^1, p_i^2, ..., p_i^{n-1}$ (if all the historical ones are not yet available, take as many as there are - i.e. expanding window till you get to the full length).
+
+Now calculate $p_i^n$ to be the arithmetic average of $p_i^k$ for $k = 1,2,...,n-1$.
+Finally set
 
 ```text
```
witgaw:
```suggestion
```

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -200,59 +201,124 @@ Then on every Vega time change, after `fractional instantaneous liquidity score`
...
-LP 1 els = 0.43
-LP 2 els = 0.17
-LP 3 els = 0.07
-LP 4 els = 0.33
+$$
+p_i^n \leftarrow \max(p_i,p_i^n).
+$$
 ```
```
witgaw:
```suggestion

```

protocol/0014-ORDT-order_types.md
```diff
@@ -46,9 +46,130 @@ Notes on scope of current version of this spec:
...
+Iceberg orders (or in other words, transparent iceberg orders) are still helpful, especially for market makers and in combination with pegged orders, as they allow a trader to remain competitively present on the order book through the refresh facility without needing to supply excessive volume into a large aggressive order.
+
+
+#### Definitions
+
+These terms are used to refer to fields on an order:
+
+* `quantity` - the full initial size of the order on entry.
+
+* `displayed quantity` - the current displayed quantity, i.e. the amount of the remaining quantity that is active on the book and can be hit.
```
witgaw, line 76: Is "quantity" not just "size" in the current API? Perhaps no point introducing another term?

protocol/0014-ORDT-order_types.md
```diff
@@ -46,9 +46,130 @@ Notes on scope of current version of this spec:
...
+Note that for a non-iceberg order, `displayed quantity == remaining`.
+
+* `remaining` - the total quantity of the order remaining that could trade.
+
+
+#### Creating iceberg orders
+
+Iceberg orders are created by populating three additional fields on any valid persistent limit order:
+
+* `initial peak size` - this specifies the amount displayed and available on the order book for a new or newly refreshed iceberg order.
```
witgaw, line 86: Maybe just "peak size" since it isn't only used on submission but also on refresh?

protocol/0014-ORDT-order_types.md
```diff
@@ -46,9 +46,130 @@ Notes on scope of current version of this spec:
...
+If the remainders of multiple icebergs sit at the same price and are not fully used up, the traded volume should be split between them pro-rata by their total total size. Any integer remainder should be allocated to the iceberg with the highest time priority.
+
+* Icebergs are refreshed after an auction uncrossing if they traded to below their `minimum peak size`, according to the same rules as for normal execution.
+
+
+#### APIs
+
+* The fields `displayed quantity`, `remaining`, `quantity`, `initial peak size`, `minimum peak size`, `refresh policy` must be exposed by data node APIs in addition to all normal fields for an order.
+
+* An iceberg order refresh must generate an event bus event.
```
witgaw:
```suggestion
* An iceberg order refresh must generate an event of the event bus.
```

witgaw (commented):
~Are we planning to switch futures to the SLA liquidity before releasing this along with spot? Because if not then we'll temporarily loose specs relating to futures.~

Please see my other comments below:

protocol/0044-LIME-lp_mechanics.md
davidsiska-vega: Accounts are good, double entry account enforced, transfers used to move.

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -19,87 +59,88 @@ Any Vega participant can apply to become a liquidity provider (LP) on a market b
...
 
-- the difference between their actual staked amount and new commitment is transferred back to their general account, i.e.
-`transferred-to-general-account-amount =  actual-stake-amount - new-actual-commitment-amount`
-- the revised fee amount and set of orders are processed.
+```text
+penalty-incuring-reduction-amount = -1 * proposed-commitment-variation - maximum-penalty-free-reduction-amount
+```
+
+Transfer `maximum-penalty-free-reduction-amount` to their general account.
+Now transfer `min((1-market.liquidity.earlyExitPenalty) x penalty-incuring-reduction-amount, bond account balance remaining)` to their general account and transfer `market.liquidity.earlyExitPenalty x penalty-incuring-reduction-amount` to the market insurance pool.
```
witgaw:
```suggestion
Now transfer `min((1-market.liquidity.earlyExitPenalty) x penalty-incuring-reduction-amount, bond account balance remaining)` to their general account and transfer `market.liquidity.earlyExitPenalty x min((1-market.liquidity.earlyExitPenalty) x penalty-incuring-reduction-amount, bond account balance remaining)` to the market insurance pool.
```

protocol/0044-LIME-lp_mechanics.md
witgaw:
```suggestion
Committed Liquidity Providers are required to provide a multiple of their stake (supplied in the settlement currency of the market) in notional volume of orders within the range defined below on both sides of the order book.
```

protocol/0044-LIME-lp_mechanics.md
witgaw: TODO: bond account balance == commitment amount ALWAYS

- spell it out and delete paragraph about bond search stuff,
- whenever we search bond account, look into temp bond account first,

protocol/0044-LIME-lp_mechanics.md
witgaw: TODO: Pro-rata (add to spec)

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -296,21 +364,49 @@ When the time defined by `market.liquidity.providers.fee.distributionTimeStep` e
...
 - An LP leaves a market that is above target stake when their fee bid is currently being used: fee changes to fee bid by the LP who takes their place in the bidding order (<a name="0042-LIQF-025" href="#0042-LIQF-025">0042-LIQF-025</a>)
 - An LP leaves a market that is above target stake when their fee bid is lower than the one currently being used and their commitment size changes the LP that meets the target stake: fee changes to fee bid by the LP that is now at the place in the bid order to provide the target stake (<a name="0042-LIQF-026" href="#0042-LIQF-026">0042-LIQF-026</a>)
-- An LP leaves a market that is above target stake when their fee bid is lower than the one currently being used and their commitment size doesn't change the LP that meets the target stake: fee doesn't change (<a name="0042-LIQF-027" href="#0042-LIQF-027">0042-LIQF-027</a>)
+- An LP leaves a market that is above target stake when their fee bid is lower than the one currently being used and their commitment size doesn't the LP that meets the target stake: fee doesn't change (<a name="0042-LIQF-027" href="#0042-LIQF-027">0042-LIQF-027</a>)
```
witgaw: @barnabee

barnabee:
```suggestion
- An LP leaves a market that is above target stake when their fee bid is lower than the one currently being used. The loss of their commitment doesn't change which LP meets the target stake: fee doesn't change (<a name="0042-LIQF-027" href="#0042-LIQF-027">0042-LIQF-027</a>)
```

protocol/0014-ORDT-order_types.md
```diff
@@ -76,20 +197,81 @@ Notes on scope of current version of this spec:
...
+8. Wash trading is not permitted for iceberg orders. The same party has one iceberg order that sits at the back of the queue , another normal order in opposite direction , the iceberg at the back comes in front and matches either fully OR partially and gets rejected( <a name="0014-ORDT-034" href="#0014-ORDT-034">0014-ORDT-034</a>)
+9. For a price level with multiple iceberg orders, if an aggressive order hits this price level, any volume greater than the displayed volume at a level is split proportionally between the hidden components of iceberg orders at that price level
+   1. If there are three iceberg orders with remaining volume 200 lots, 100 lots and 100 lots, an order for 300 lots would be split 150 to the first order and 75 to the two 100 lot orders. (<a name="0014-ORDT-037" href="#0014-ORDT-037">0014-ORDT-037</a>)
+   1. If there are three iceberg orders with remaining volume 200 lots, 100 lots and 100 lots, an order for 600 lots would be split 200 to the first order and 100 to the two 100 lot orders, with 200 lots then taking farther price levels. (<a name="0014-ORDT-038" href="#0014-ORDT-038">0014-ORDT-038</a>)
+
```
witgaw, line 266: TODO: Price level of a pegged iceberg order never changes during an execution of an incoming aggressive order (even as price levels get eaten up).

witgaw (approved):

witgaw (approved):

PR #1694, opened 2023-05-10
feat: align Spot product with SLA changes

cdummett:
PR makes changes to the Spot spec to align with the SLA [PR](https://github.com/vegaprotocol/specs/pull/1685).

As SLA updates remove automatic liquidity orders, there is no longer a requirement for two bond accounts (previously needed to ensure an LP has enough funds to support their commitment). Now if an LP can't support their commitment they are simply not rewarded.

As a result the liquidity mechanisms for the Spot spec can be simplified and aligned with the liquidity mechanisms used in CSF markets.

### DIFF
--- a/protocol/0080-SPOT-product_builtin_spot.md
+++ b/protocol/0080-SPOT-product_builtin_spot.md
@@ -45,20 +45,6 @@ submission = {
         marketId: "abcdefghiklkmnopqrstuvwxyz",
         fee: "0.01",
         commitmentAmount: 15000 (DAI)
-        buys: [
-            {
-                offset: "1"
-                proportion: "1"
-                reference: "PEGGED_REFERENCE_BEST_BID"
-            }
-        ]
-        sells: [
-            {
-                offset: "1"
-                proportion: "1"
-                reference: "PEGGED_REFERENCE_BEST_ASK"
-            }
-        ]
         reference: "example_liquidity_provision_submission"
     }
 }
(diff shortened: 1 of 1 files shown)


protocol/0080-SPOT-product_builtin_spot.md
```diff
@@ -63,61 +64,23 @@ submission = {
...
+As the target stake (and therefore `maximum_reduction_amount`) is some factor of the total stake (see [Target Stake](./0080-SPOT-product_builtin_spot.md#market-target-stake)), a single LP will never be able to reduce their commitment to `0` if they are the only LP in a market. They can either keep reducing to a sufficiently small amount they're willing to ignore, or they can submit a governance vote to cancel the market, see the [governance spec](./0028-GOVE-governance.md).
 
 For market cancellation proposal a sole LP in the market holds all the voting power (unless governance token holders override them).
 
-### Liquidity Shortfalls
-
-If at any point in time, a liquidity provider has insufficient capital in their general accounts to cover a transfer arising from a filled liquidity order, the network will utilise the liquidity commitment, held in the relevant bond account to cover the shortfall, applying the bond penalty factor (slashing the bond).
-
-As there is no market insurance pool, funds from bond slashing in the result of shortfall will be transferred to the global network treasury for that asset.
-
```
cdummett, line 115: This is the biggest implication of going with the single commitment amount route as a result of the SLA changes @davidsiska-vega @barnabee 

An LPs bond account can no longer be used to cover any shortfalls an LP may have when submitting orders (they need to transfer the full value of the order to a holding account).

Basically means an LP needs at least 3x their commitment amount to support their commitment.

davidsiska-vega: And that's fine, it just means they're not meeting their SLA. They will be penalised as per the spec.

davidsiska-vega (approved):

PR #1704, opened 2023-05-17
Fix/succesor markets

davidsiska-vega:
Few small updates on successor markets.


protocol/0028-GOVE-governance.md
```diff
@@ -146,7 +146,9 @@ For market change proposals the network will additionally calculate
 1. `LP participation rate = SUM (equity-like share of all LPs who cast a vote)` (no need to divide by anything as equity-like share sums up to `1`).
 1. `LP for rate = SUM (all who voted for) / LP participation rate`.
 
-A market parameter change is passed only when:
+If the market that the proposal is changing is pending at the vote resolution time then only token holder votes are used.
```
EVODelavega: So specifically: if a market has been accepted, enacted proposal still in opening auction?

protocol/0028-GOVE-governance.md
```diff
@@ -221,18 +223,22 @@ A proposal to create a market contains
...
+
+
 Note that each market can have exactly one market as a _successor_ market.
 
 - if there already is a market (possibly pending, i.e. in opening auction, see [lifecycle spec](./0043-MKTL-market_lifecycle.md)), naming a parent market, then a subsequent proposal referencing that market is rejected.
 - if there are two proposals naming the same parent market then whichever one gets into the pending state first (i.e. passes governance vote) becomes the successor of the named parent; the other proposal is cancelled with reason "parent market no longer available".
-- if there is a successor market naming a parent market that terminates and settles or is cancelled by governance before the parent market (for whatever reason) then the parent market successor is set to empty and a different successor can be proposed.
+- if there is a successor market naming a parent market that terminates and settles or is cancelled by governance before the parent market (for whatever reason) then the parent market successor field is set to empty and a different successor can be proposed by a future market proposal.
```
EVODelavega: parent markets don't have a successor field. Successors have a parent market field. Is there a reason why this should be a doubly linked list?

protocol/0028-GOVE-governance.md
davidsiska-vega: Yes. Clarified, will push.

protocol/0028-GOVE-governance.md
davidsiska-vega: I assumed it will be, yes. I want to have a look at a market and see immediately whether I can use it as my parent or not.

EVODelavega: From data-node, or does core have to provide this?

davidsiska-vega: When a new proposal comes saying marketId `xyz` is the proposed parent core needs to do validation. 

With a single linked lists I will need to look through all the markets to see if any of them use `xyz` as parent. With a doubly linked list I just go to `xyz` and check whether its available. 

To me that sounds saner (though of course there is maybe more work to ensure consistency). 

I'll leave it up to your implementation @EVODelavega and you can come back and edit this part to match implementation. 

Data node will have to provide the double linked info because hypothetically with some obscure data retention rules the successor may be gone already from the date node so the single linked version won't do, I think.

EVODelavega: I agree that datanode has to offer the capability to traverse the _"family tree"_/lineage of markets, and thus requires something like a doubly linked list, and datanode will therefore be able to provide you the information required (as in: market xyz has proposal abc as a successor).

As far as the core is concerned though, markets are pretty isolated things, they aren't really aware, nor do they have access to other markets that may exist. The execution engine, which instantiates the markets is aware of which market is the parent/successor, and once a successor market was enacted, all other proposals will get rejected as per spec (several competing successor proposals can exist after all). Once a successor proposal is enacted, any future proposal to succeed its parent will just get rejected outright, and by that time, the datanode can already provide the relevant information easily.

To ensure market state doesn't depend on the state of other markets once they are enacted, I think it best that core only concerns itself with setting the parent market id field on the market. Once that market is enacted, that field is either blank or not, and becomes static.

EVODelavega (dismissed):

EVODelavega (dismissed):

gordsport (approved):
reapproving as the only change after the approval from @EVODelavega was a typo (no content change)

PR #1716, opened 2023-05-25
chore: Add acceptance criteria for iceberg orders

MuthuVega:
First set of acceptance criteria for iceberg orders for review.

### DIFF
--- a/protocol/0014-ORDT-order_types.md
+++ b/protocol/0014-ORDT-order_types.md
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
   - A GTT order submitted at a time >= its expiry time is rejected. (<a name="0014-ORDT-005" href="#0014-ORDT-005">0014-ORDT-005</a>)
 - No party can submit a [network order type](#network-orders)  (<a name="0014-ORDT-006" href="#0014-ORDT-006">0014-ORDT-006</a>)
 
+# Iceberg Orders AC's:
+
+### Iceberg Order Submission:
+
+1. A persistent (GTC, GTT, GFA, GFN) iceberg order that is not crossed with the order book is included in the order book with order book volume == initial peak size. No trades are generated.
+2. An iceberg order with either an ordinary or pegged limit price can be submitted.
+3. An iceberg post only order can be submitted.
+4. An iceberg refuce only order is rejected.
+4. Margin requirement for iceberg order - is it based on initial peak size i.e. displayed quantity OR actual quantity ? Assume its displayed quantity - needs confirmation.
+5. For a persistent (GTC, GTT, GFA, GFN) iceberg order thats submitted margin is calculated and deducted correctly both during submission as well as during subsequent refreshes.
+6. For an iceberg order, the orders are refreshed immediately after producing a trade. Every time volume is taken from the displayed quantity , ensure the order is refreshed if if display quantity < minimum peak size.
+   If the order is successfully refreshed , then the order loses its time priority and is pushed to the back of the queue.
+7. For an iceberg order thats submitted when the market is in auction, only the dispalyed quantity is filled when coming out of auction. What other attributes need to be checked here ?
+
+### Iceberg Order Batch Submission:
+
+1. For multiple icebers orders submitted as a batch of orders with a mix of ordinary limit orders and market orders, the icebers orders are processed automically and the order book volume and price, margin calculations , order status are all correct - What else needs checking ?
+2. What other scenarios need testing ?
+
+### Iceberg Order Submission - Negative tests:
+
+1. An iceberg order with a non persistent TIF (IOC, FOK) is rejected with a valid error message.
+2. An iceberg market order with any TIF is rejected with a valid error message.
+3. A reduce-only iceberg order with any TIF is rejected with a valid error message.
+4. An iceberg order with initial peak size less than the minimum order size rejected with a valid error message.
+5. An iceberg order with initial peak size greater than the total order size is rejected with a valid error message.
+6. An iceberg order with minimum peak size less than 0 is rejected with a valid error message.
+7. An iceberg order with minimum peak size greater than initial peak size is rejected with a valid error message.
+
+### Iceberg Order Amendment:
+
+1. Amending an iceberg order to increase size will increase the total and remaining quantities of the order and time priority of the order is not lost.
+2. Amending an iceberg order to decrease size will decrease the total and remaining quantities and time priority of the order is not lost.
+3. Amend an iceberg order to decrease size so that the displayed quantity is decreased. Total and remaining quantites are decreaed, margin is recalculated and released and time priority is not lost.
+4. What other scenarios do we need to consider here ?
+
+### Iceberg Order Cancellation:
+
+1. Cancelling an iceberg order will cancel the order, remove it from the order book , release margin and update order book to reflect the change.
+
+### Iceberg Order Execution:
+
+1. An aggressive iceberg order that crosses with an order where volume > iceberg volume, the iceberg order gets fully filled on entry, the iceberg order status is filled, the remaining quantity = 0. Are atomic trades generated OR one single trade for each display quantty volume ???
+2. An aggressive iceberg order that crosses with an order where volume < iceberg volume. The initial display quantity is filled and the remaining volume is unfilled. Status of iceberg order is partially filled , the volume remaining = (quantity - initial volume) and the remaining volume sits on the book. When additional orders thar are submitted which consume the remaining volume on the iceberg order , the volume of the iceberg order is refreshed as and when the volume dips below the minimum peak size.
+3. A passive iceberg order (the only order at a particular price level) when crossed with another order that comes in which consumes the full volume of the iceberg order is fully filled. Status of iceberg order is filled and the remaining = 0. Atomic trades are produced.
+4. A passive iceberg order with a couple of order that sit behind the iceberg order at the same price that crosses with an order where volume > display quantity of iceberg order. After the first trade is produced , the iceberg order is pushed to the back of the queue and gets filled only when the other orders in front get fully filled.
+5. Submit an aggressive iceberg order for say size 100. There are multiple matching orders of size 30,40,50. Ensure the orders are matched and filled in time priority of the orders and any remaining volume on the orders is correctly left behind.
+6. Submit an aggressive iceberg order for say size 100. There are multiple matching orders of size 20,30. Ensure the orders are matched and filled in time priority of the orders. Ensure remaining volume on the iceberg order is (100 - (20+30))
+7. When a non iceberg order sitting on the book is amended such that it trades with with an iceberg order, then the iceberg order is refreshed.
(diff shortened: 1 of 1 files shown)


protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
 
+### Iceberg Orders AC's
+
+### Iceberg Order Submission
+
+1. A persistent (GTC, GTT, GFA, GFN) iceberg order that is not crossed with the order book is included in the order book with order book volume == initial peak size. No trades are generated.
+2. An iceberg order with either an ordinary or pegged limit price can be submitted.
+3. An iceberg post only order can be submitted.
+4. An iceberg reduce only order is rejected.
+5. Margin requirement for iceberg order - is it based on initial peak size i.e. displayed quantity OR actual quantity ? Assume its displayed quantity - needs confirmation.
```
TomMcL: I believe margin requirement should be full order quantity

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+6. For a persistent (GTC, GTT, GFA, GFN) iceberg order that's submitted margin is calculated and deducted correctly both during submission as well as during subsequent refreshes.
```
TomMcL: Let's make this one more explicit. i.e. "For an iceberg order that is submitted with total size `x` and visible size `y` the margin taken should be identical to a regular order of size `x` rather than one of size `y`"

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+7. For an iceberg order, the orders are refreshed immediately after producing a trade. Every time volume is taken from the displayed quantity , ensure the order is refreshed if if display quantity < minimum peak size.
+   If the order is successfully refreshed , then the order loses its time priority and is pushed to the back of the queue.
+8. For an iceberg order that's submitted when the market is in auction, only the displayed quantity is filled when coming out of auction. What other attributes need to be checked here ?
```
TomMcL: I think that would be the main one I'd think of

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+
+### Iceberg Order Batch Submission
+
+1. For multiple iceberg orders submitted as a batch of orders with a mix of ordinary limit orders and market orders, the iceberg orders are processed atomically and the order book volume and price, margin calculations , order status are all correct - What else needs checking ?
```
TomMcL: Shouldn't be too much weird which needs testing which I can think of, most should be covered by batch submission itself.

One thing which does need an AC is the behaviour of multiple orders when trading against an iceberg in a batch (I.e. ensure that the iceberg doesn't refresh until the end of the batch after it is depleted)

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+5. An iceberg order with initial peak size greater than the total order size is rejected with a valid error message.
+6. An iceberg order with minimum peak size less than 0 is rejected with a valid error message.
+7. An iceberg order with minimum peak size greater than initial peak size is rejected with a valid error message.
+
+### Iceberg Order Amendment
+
+1. Amending an iceberg order to increase size will increase the total and remaining quantities of the order and time priority of the order is not lost.
+2. Amending an iceberg order to decrease size will decrease the total and remaining quantities and time priority of the order is not lost.
+3. Amend an iceberg order to decrease size so that the displayed quantity is decreased. Total and remaining quantity is decreased, margin is recalculated and released and time priority is not lost.
+4. What other scenarios do we need to consider here ?
```
TomMcL: Amending total quantity to less than displayed quantity should also reduce displayed amount

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+
+### Iceberg Order Cancellation
+
+1. Cancelling an iceberg order will cancel the order, remove it from the order book , release margin and update order book to reflect the change.
+
+### Iceberg Order Execution
+
+1. An aggressive iceberg order that crosses with an order where volume > iceberg volume, the iceberg order gets fully filled on entry, the iceberg order status is filled, the remaining quantity = 0. Are atomic trades generated OR one single trade for each display quantity volume ???
```
TomMcL: It should trade as if it is a full order for that size, so the same number of trades as for a standard limit order at the same price

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+2. An aggressive iceberg order that crosses with an order where volume < iceberg volume. The initial display quantity is filled and the remaining volume is unfilled. Status of iceberg order is partially filled , the volume remaining = (quantity - initial volume) and the remaining volume sits on the book. When additional orders thar are submitted which consume the remaining volume on the iceberg order , the volume of the iceberg order is refreshed as and when the volume dips below the minimum peak size.
+3. A passive iceberg order (the only order at a particular price level) when crossed with another order that comes in which consumes the full volume of the iceberg order is fully filled. Status of iceberg order is filled and the remaining = 0. Atomic trades are produced.
```
TomMcL: @davidsiska-vega / @barnabee  I'm actually not sure how the orchestration here should work. Trading the full amount of the iceberg order probably makes sense technically but would conflict with the assertion that only the visible component can ever trade once the order is on the book. However if we only trade the visible component and there is still an order of the other side sitting on the book there will be further trades each time there is a refresh, and who would be the agressor in those trades?

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+4. A passive iceberg order with a couple of order that sit behind the iceberg order at the same price that crosses with an order where volume > display quantity of iceberg order. After the first trade is produced , the iceberg order is pushed to the back of the queue and gets filled only when the other orders in front get fully filled.
+5. Submit an aggressive iceberg order for say size 100. There are multiple matching orders of size 30,40,50. Ensure the orders are matched and filled in time priority of the orders and any remaining volume on the orders is correctly left behind.
+6. Submit an aggressive iceberg order for say size 100. There are multiple matching orders of size 20,30. Ensure the orders are matched and filled in time priority of the orders. Ensure remaining volume on the iceberg order is (100 - (20+30))
```
TomMcL: These two may not be worth worrying about, the general matching engine should probably be handling time priority fine already

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+7. When a non iceberg order sitting on the book is amended such that it trades with with an iceberg order, then the iceberg order is refreshed.
+8. Try for scenarios for wash trading of iceberg orders - is that even possible ? Or some complex scenarios - same party , one iceberg order that sits at the back of the queue , another normal order in opposite direction , the iceberg at the back slowly comes in front and matches either fully OR partially - is this possible ?
```
TomMcL: Wash trades should be handled as normal, but worth having one to ensure it works

protocol/0014-ORDT-order_types.md
```diff
@@ -90,6 +90,76 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
...
+1. API end points should be available to query initial peak size, minimum peak size, quantity, displayed quantity and remaining.
+2. The additional fields relating to iceberg orders should be available in the streaming api end points
+
+### Protocol Upgrade
+
+1. ???
+
+### Regression
+
+1. All other order types should behave as they were
```
TomMcL: Not necessary, other Acs continuing to pass should be the criteria for this already.

protocol/0014-ORDT-order_types.md
MuthuVega: Thanks for the confirmation.

protocol/0014-ORDT-order_types.md
MuthuVega: AC updated with the above suggestion.

protocol/0014-ORDT-order_types.md
MuthuVega: Thanks.

protocol/0014-ORDT-order_types.md
MuthuVega: New AC added as suggested.

protocol/0014-ORDT-order_types.md
MuthuVega: AC updated.

protocol/0014-ORDT-order_types.md
MuthuVega: AC updated.

protocol/0014-ORDT-order_types.md
MuthuVega: Thanks , have marked them as low priority.

protocol/0014-ORDT-order_types.md
MuthuVega: Updated the AC to state wash trading is not permitted between iceberg orders.

protocol/0014-ORDT-order_types.md
MuthuVega: Removed this section for the AC's.

protocol/0014-ORDT-order_types.md
davidsiska-vega: @TomMcL my take would be either:
a) make icebergs always post-only; in this case we would reject the order. 
b) make icebergs with post-only as on option; in this case I would trade potentially the full volume (both visible and hidden) if there is enough volume already on the book and the iceberg is aggressive.

TomMcL: @davidsiska-vega I'm not sure that fully resolves the scenario above thought, or I may be missing something. What I'm picturing is:

1. Party 1 places an iceberg order for price `x`, size `200`, visible size `10` is placed on the book at an empty price level
2. A limit order from party 2, happy to be a taker, for price `x`, size `100` is placed on the book and trades `10` as a taker
3. Party 2's `90` remaining is now sitting on the book at the end of the tx
4. Do we immediately refresh and match again, with party 2 still counted as the taker? That seems hard to keep track of. Though the alternative is presumably that the full size of an iceberg order can immediately trade with incoming limit orders which feels incorrect

barnabee (changes requested):
This PR looks like it is to the wrong branch. The ORDT spec will not contain the iceberg spec when it is merged, only the ACs. It needs to be rebased to the correct branch (@gordsport can confirm) before it can be merged.

PR #1736, opened 2023-06-05
Add AUTH v2 spec

### LINKED ISSUE #1575: [SPEC UPDATE] Update 0022-AUTH-auth.md to match latest implementation of the wallet
Update [0022-AUTH-auth.md](https://github.com/vegaprotocol/specs/blob/master/protocol/0022-AUTH-auth.md) to match latest implementation of the wallet.

- [ ] update the spec and ACs to match implementation
- [ ] Consider using new AC codes if the AC differs from the currently written test
- [ ] Align with qa team to make sure the tests are covered

ValentinTrinque:
This spec is supposed to explain the way we authenticate parties on the network, using public keys system. That part has been clarified. Details about the cryptographic algorithms used to generate such keys has been added. The high-level description of a "component" managing such keys, renamed _wallet management application_ (prior _wallet service_) has been updated.

However, the previous version also explain how a user would interact with the _wallet service_, from an implementation perspective. This is doesn't fit anymore, as the interaction between the user and the wallet doesn't relate to the "authentication" of a user and a party on the network.  As a result, the wallet API ACs has been moved out of that spec, to a non-protocol spec.

Close #1575

### DIFF
--- a/non-protocol-specs/0011-NP-CLIE-client-interaction-tests.md
+++ b/non-protocol-specs/0011-NP-CLIE-client-interaction-tests.md
@@ -6,28 +6,93 @@ This specification contains a set of tests/acceptance criteria that clients (wal
 
 ## Acceptance criteria
 
-1. The parameter `spam.pow.numberofTxPerBlock` is decreased.  Verify that:
-    - The new parameter is communicated to and adapted/used by the wallet, i.e., if a user has too many transactions according to the new parameter, the wallet does not submit the excess transactions and returns an appropriate error message OR reprocesses the transactions. (<a name="0011-NP-CLIE-001" href="#0011-NP-CLIE-001">0011-NP-CLIE-001</a>)
-    - The new parameter is communicated to and adapted/used by the wallet, i.e., if a user has too many transactions according to the new parameter, AND if the parameter `spam.pow.increaseDifficulty` is set to `1`, the wallet submits the excess transactions by submitting a PoW of higher difficulty. (<a name="0011-NP-CLIE-002" href="#0011-NP-CLIE-002">0011-NP-CLIE-002</a>)
+1. The parameter `spam.pow.numberofTxPerBlock` is decreased. Verify that:
+
+   - The new parameter is communicated to and adapted/used by the wallet, i.e., if a user has too many transactions according to the new parameter, the wallet does not submit the excess transactions and returns an appropriate error message OR reprocesses the transactions. (<a name="0011-NP-CLIE-001" href="#0011-NP-CLIE-001">0011-NP-CLIE-001</a>)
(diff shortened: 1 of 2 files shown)


gordsport:
@vega-paul This PR moves some ACs from the AUTH spec to the non-protocol CLIE spec. I will need to renumber them as CLIE ACs and will make the changes to the text comments in the respective system test files.

It would also be good for you to review the other CLIE ACs and add/update as you see fit. Lower priority for now in terms of implementation, however, it'd be good to get these in as the spec was very out-of-date

vega-paul (dismissed):

non-protocol-specs/0011-NP-CLIE-client-interaction-tests.md
```diff
@@ -6,28 +6,93 @@ This specification contains a set of tests/acceptance criteria that clients (wal
...
 1. Create 20 keys with sufficient tokens to be allowed proposals and delegations (i.e., more than `SpamProtectionMinTokensForDelegation` and `SpamProtectionMinTokensForProposal`. For all keys (within their respective wallets), submit random (but valid) votes, delegation changes, transactions and proposals in sets of at least 13 each at the same time, at a rate of one set per second. Let that run for 4 epochs, with an epoch length of at least 100 blocks. Verify that the wallet returns an error for the excess transactions OR reprocesses the transactions. Verify that no account ever gets banned, and that the clients never crash. (<a name="0011-NP-CLIE-019" href="#0011-NP-CLIE-019">0011-NP-CLIE-019</a>)
+
+### Wallet service
+
+This acts as a guideline to implement network-based (remote) wallet management applications with third-party applications.
+
+#### API v1
+
+DEPRECATED: The wallet API v1 is no longer officially supported.
+
```
Vegaklaus, line 54: Can we clarify what we mean with that ? It's still available, but not recommended to use ? The Vega client will reject all v1 requests ?

protocol/0022-AUTH-auth.md
```diff
@@ -2,149 +2,126 @@
...
-  - delete a party and its associated key pair
-- Sign blob of data
-  - accept a request from a authenticated user containing a blob of data and reference to a party (and therefore a key pair) to be used to sign the blob
+The process of sending the first transaction for a user is as follows:
 
-#### Wallet service: A user
+1. Bob creates a wallet and generates a party in that wallet using his wallet management application.
+2. Bob requests to his wallet management application a signature of a transaction by submitting the transaction and the public key of the party to use.
+3. The transaction is checked, signed, and bundled with the signature and the public key by the wallet management application.
+4. The signed transaction is then submitted to the Vega network, which verify the authenticity of the transaction and its content to ensure the transaction was signed on behalf of the correct party.
```
Vegaklaus, line 45: The network also verifies for a replay of an old transaction.

PR #1738, opened 2023-06-06
SLA liquidity (extras)

witgaw:
A PR to merge additional items into the original `feat/sla-liquidity` branch, including:

* extend description of the liquidity provider temporary bond account and specify that any search attempting to access the true bond account should first go via this account,
* the LPs commitment level is always equal to the balance of their "true" bond account,
* we should never try to automatically top up any of the bond accounts from general,
* LP amendments should all be processed at the beginning of the epoch and each LP's penalty free amount should be calculated pro rata based on the size of the reduction.

### DIFF
--- a/protocol/0014-ORDT-order_types.md
+++ b/protocol/0014-ORDT-order_types.md
@@ -210,6 +210,7 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
   * Any GTT limit order that [still] resides on the order book at its expiry time is cancelled and removed from the book before any events are processed that rely on its being present on the book, including any calculation that incorporates its volume and/or price level. (<a name="0014-ORDT-004" href="#0014-ORDT-004">0014-ORDT-004</a>)
   * A GTT order submitted at a time >= its expiry time is rejected. (<a name="0014-ORDT-005" href="#0014-ORDT-005">0014-ORDT-005</a>)
 * No party can submit a [network order type](#network-orders)  (<a name="0014-ORDT-006" href="#0014-ORDT-006">0014-ORDT-006</a>)
+* A pegged order (including iceberg pegged orders) never has it's price updated during the execution of an incoming aggressive order (even as price levels get consummed so that it's reference price changes after the execution). (<a name="0014-ORDT-039" href="#0014-ORDT-039">0014-ORDT-039</a>)
 
 ### Iceberg Orders AC's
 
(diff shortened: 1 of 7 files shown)


witgaw (commented):
Ready for review

protocol/0014-ORDT-order_types.md
```diff
@@ -210,6 +210,7 @@ Network orders are used during [position resolution](./0012-POSR-position_resolu
   * Any GTT limit order that [still] resides on the order book at its expiry time is cancelled and removed from the book before any events are processed that rely on its being present on the book, including any calculation that incorporates its volume and/or price level. (<a name="0014-ORDT-004" href="#0014-ORDT-004">0014-ORDT-004</a>)
   * A GTT order submitted at a time >= its expiry time is rejected. (<a name="0014-ORDT-005" href="#0014-ORDT-005">0014-ORDT-005</a>)
 * No party can submit a [network order type](#network-orders)  (<a name="0014-ORDT-006" href="#0014-ORDT-006">0014-ORDT-006</a>)
+* A pegged order (including iceberg pegged orders) never has it's price updated during the execution of an incoming aggressive order (even as price levels get consummed so that it's reference price changes after the execution). (<a name="0014-ORDT-039" href="#0014-ORDT-039">0014-ORDT-039</a>)
```
davidsiska-vega:
```suggestion
* A pegged order (including iceberg pegged orders) never has its price updated during the execution of an incoming aggressive order (even as price levels get consummed so that its reference price changes after the execution). (<a name="0014-ORDT-039" href="#0014-ORDT-039">0014-ORDT-039</a>)
```

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -105,42 +105,58 @@ The amendment is actioned in two steps.
 1) the amount is immediately transferred from the party's general account to a temporary "pending" bond account. This amount counts towards the stake committed to the market and so in particular can get the market out of liquidity auction.
 2) at the beginning of the next epoch (after the rewards / penalties for present LPs - including the party that's amending - have been evaluated) the amount is transferred from the "pending" bond to the true bond account.
 
-Commitment amendments should be processed in the time order of their latest updates within the epoch. For example if an LP places an amendment at time `t1`, followed by LP2 placing an amendment at `t2` and the first LP amending again at `t1`, LP2's amendment will be executed first.
+For each party only the most recent amendement should be considered. All the amendements get processed simultaneously, hence the relative arrival of amendments made by different LPs within the previous epoch is irrelevant.
```
witgaw:
```suggestion
For each party only the most recent amendement should be considered. All the amendements get processed simultaneously, hence the relative arrival of amendments made by different LPs within the previous epoch is irrelevant (as far as commitment reduction is concerned, it still has implications for other aspects of the mechanism).
```

davidsiska-vega (approved):

PR #1741, opened 2023-06-06
Update 0028-GOVE-governance.md 0028-GOVE-076

vega-paul:
Remove AC: 0028-GOVE-076


gordsport (approved):
Approved on the PD meeting: Agreed that AC 99 and 100 cover this.

PR #1745, opened 2023-06-07
Fix: another AC for changing oracle after trading termination

davidsiska-vega:
Fix: another AC for changing oracle after trading termination


daunatv (dismissed):
As far as what I have seen no update to the oracle is accepted after market termination, so this would cover what we are looking for.

daunatv (approved):
This will test what we were not able to complete in the current version where the oracle change to the marketupdate was accepted and voted through, but did not work for settling the market.

PR #1753, opened 2023-06-10
Update 0028-GOVE-governance.md

vega-paul:


protocol/0028-GOVE-governance.md
```diff
@@ -582,7 +585,7 @@ Below `*` stands for any of `asset, market, updateMarket, updateNetParam, freeFo
 
 - Enacted and active recurring governance transfers must be included in LNL banking checkpoint and resume after the checkpoint restore.(<a name="0028-GOVE-105" href="#0028-GOVE-105">0028-GOVE-105</a>)
 
-- When a transfer gets enacted it emits transfer event similar to regular transfer events from regular transfers, however with governance-recurring types. At the time of enactment no amount is attached to the transfer and it will show 0.(<a name="0028-GOVE-106" href="#0028-GOVE-106">0028-GOVE-106</a>)
+- When a transfer gets enacted it emits transfer event similar to regular transfer events from regular transfers, however with governance-recurring types. (<a name="0028-GOVE-107" href="#0028-GOVE-107">0028-GOVE-107</a>) At the time of enactment no amount is attached to the transfer and it will show 0.(<a name="0028-GOVE-106" href="#0028-GOVE-106">0028-GOVE-106</a>)
```
gordsport: @vega-paul - should this be formatted to two list items?

Also there is a duplication of `0028-GOVE-107` looks like this new AC should be given the number `-108` or checking that the `-107` ACs are not actually duplicated (I think the former in this case)

gordsport: Also `0028-GOVE-106` looks to be duplicated too. Please check the approbation CI run and adjust the ACs and/or numbers accordingly

PR #1758, opened 2023-06-13
Update 0080-SPOT-product_builtin_spot.md

VanithaVega:
Changed the Network parameter to SpotMarketTradingEnabled


cdummett (dismissed):

protocol/0080-SPOT-product_builtin_spot.md
```diff
@@ -11,7 +11,7 @@ When trading Spot products, parties can only use assets they own - there is no l
 
 ## 2. Network Parameter
 
-1. `spot_trading_enabled`: parameter defines whether markets using Spot products are enabled on the network.
+1. `SpotMarketTradingEnabled`: parameter defines whether markets using Spot products are enabled on the network.
```
gordsport:
```suggestion
1. `limits.markets.proposeSpotEnabled`: parameter defines whether markets using Spot products are enabled on the network.
```

@ze97286 - we have reviewed this in the protocol design meeting and suggest the net param should be named as suggested here

ze97286: proposeSpotEnabledFrom, is this a date?

gordsport: @davidsiska-vega - I think this was a boolean, but the suggested name implies a date. Please can you confirm if this should have been `limits.markets.proposeSpotEnabled` instead? 

Current markets (futures?) has: `limits.markets.proposeEnabledFrom`

davidsiska-vega: Good point! I think it should be `limits.markets.proposeSpotEnabled` and it should just be one boolean. 

I don't think we want to bother with the "enabled from" date.

gordsport: suggestion updated - can be reviewed and committed in todays meeting

ze97286 (approved):

Jiajia-Cui (approved):
LGTM

PR #1764, opened 2023-06-15
refactor: clarify iceberg wash trading AC

cdummett:
Current AC defines a case which is not possible as the normal order will be the order that is stopped as the iceberg order will be passive and the normal order will the aggressor.

PR updates the AC to say the normal order should be stopped.


witgaw (approved):
LGTM

PR #1768, opened 2023-06-15
Modify market change proposal requirement

witgaw:
We used to require both the governance token AND a minimum ELS for a party to be able to submit a market change proposal. This PR proposes to use OR instead. This will bring it in line with rules for the voting (a party with sufficient ELS, but no governance tokens is still able to successfully cast a vote).


davidsiska-vega (changes requested):
Would you like to drop in an AC to that effect too?

davidsiska-vega (approved):

PR #1775, opened 2023-06-19
chore(trading): 6004-CHAR-chart ACs created

bwallacee:
New ACs for the Chart on console.

Note: Currently missing ACs for "price monitoring bounds" overlay as there is a potential bug being invesitigated.

### DIFF
--- a/user-interface/6004-CHAR-chart.md
+++ b/user-interface/6004-CHAR-chart.md
@@ -1,0 +1,106 @@
+# Chart
+
+## Display options
+- **Must** be able to change time interval from a list of intervals ([1003-TRAN-001](#1003-TRAN-001)) :    
+    - 1m  
+    - 5m  
+    - 15m     
+    - 1H  
+    - 6H  
+    - 1D
+    
+- **Must** be able to change display type from a list of display types ([1003-TRAN-001](#1003-TRAN-001)):   
+    - Moutain     
+    - Candlestick     
+    - Line    
+    - OHLC    
+
+- **Must** icon will change depending on the display type selected([1003-TRAN-001](#1003-TRAN-001)):   
+    - Moutain     
+    - Candlestick     
+    - Line    
+    - OHLC   
+
+- **Must** be able to add overlays by selecting from a list of overlays ([1003-TRAN-001](#1003-TRAN-001)):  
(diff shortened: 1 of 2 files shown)


user-interface/6004-CHAR-chart.md
```diff
@@ -0,0 +1,106 @@
+# Chart
+
+## Display options
+- **Must** be able to change time interval from a list of intervals ([1003-TRAN-001](#1003-TRAN-001)) :
```
mattrussell36, line 4: should these not be 1000-CHAR-001?

user-interface/6004-CHAR-chart.md
```diff
@@ -0,0 +1,106 @@
...
+
+- **Must** be able to close any studies selected ([1003-TRAN-001](#1003-TRAN-001))
+
+- **Must** be able to add multiple overlays at the same time ([1003-TRAN-001](#1003-TRAN-001))
+
+- **Must** be able to close any overlays selected ([1003-TRAN-001](#1003-TRAN-001))
+
+## Price and Time
+
+- **Must** see details of price from where my mouse cursor is on the chart([1003-TRAN-001](#1003-TRAN-001)):
```
mattrussell36, line 49: We might have to leave any ACs that cover the actual chart rendering untested for now as we'll need some additional tooling due to it being rendered with canvas

mattrussell36 (changes requested):

user-interface/6004-CHAR-chart.md
bwallacee: yeah, Will update all the approbation tagging once the actual ACs are all finialised just so that I have to change the numbers once etc.

user-interface/6004-CHAR-chart.md
bwallacee: Agree with the tooling bit, But I think we still want ACs on the behaviours in the spec? So that we know in our coverage what coverage we actually have.

user-interface/6004-CHAR-chart.md
bwallacee: Fixed.

mattrussell36 (approved):

PR #1781, opened 2023-06-22
re-arrange ac to save mainetenance

dalebennett1992:


pennyandrews (approved):

dexturr (changes requested):
I wonder if we can put this in a `browser wallet` folder or something. With this I find it difficult to grok the list to find the ones that are browser and which ones are for other UIs

dexturr (approved):
Going to be addressed separately

PR #1794, opened 2023-06-27
update spec for parent in proposed state

EVODelavega:
Small change to the successor market spec (AC's 0081-SUCM-008 and 0081-SUCM-009)


protocol/0081-SUCM-successor_markets.md
```diff
@@ -73,9 +73,9 @@ A new market proposal sets parent market Id to a market that has settled. The pa
...
 
-Successor markets can be enacted if the parent market is still in the "proposed" state. There is no virtual stake to copy over and no insurance pool balance to transfer  (<a name="0081-SUCM-008" href="#0081-SUCM-008">0081-SUCM-008</a>)
+Successor markets cannot be enacted if the parent market is still in the "proposed" state. Successor market proposals can be submitted when the parent market is still in proposed state, but the when the voting period for the successor market ends (the successor market moves from "proposed" in to opening auction/"pending" state) when the parent market is still in "proposed" state, the successor market is rejected. (<a name="0081-SUCM-008" href="#0081-SUCM-008">0081-SUCM-008</a>)
```
davidsiska-vega:
```suggestion
Successor markets cannot be enacted if the parent market is still in the "proposed" state. Successor market proposals can be submitted when the parent market is still in proposed state. When the voting period for the successor market ends then either: a) the parent market is already enacted in which case the successor market moves from "proposed" in to opening auction/"pending" state. Or the parent market is still in "proposed" state in which case successor market is rejected. (<a name="0081-SUCM-008" href="#0081-SUCM-008">0081-SUCM-008</a>)
```

davidsiska-vega (approved):

PR #1795, opened 2023-06-27
Update 0064-VALP-validator_performance_based_rewards.md

### LINKED ISSUE #1993: [SPEC ONLY?]: Validator Ethereum + L2s heartbeart and performance metric
- [Per spec PR here](https://github.com/vegaprotocol/specs/pull/2027)

Vegaklaus:
Added Ethereum event forwarding reward/punishment.

close https://github.com/vegaprotocol/specs/issues/1993


protocol/0064-VALP-validator_performance_based_rewards.md
```diff
@@ -34,7 +34,39 @@ This function is primarily for testing purposes to allow for very short epochs w
...
 
-Then `validator_performance = max(0.05, min((p'/expected, 1))`
+The calculation only based on this would be `validator_performance = max(0.05, min((p'/expected_p, 1))`; however
+another factor is added:
+
+Goal 2: Detect unforwarded Ethereum events and punish the validator that does not forward them
+Detection: Events forwarded by some validators are not forwarded by others.
+
+#### Ethereum Heartbeat
+For the Ethereum Heartbeat, we use the system parameter <ethereum_heartbeat_period>. This parameter should be either 0 or a value bigger than the number of validators; the recommended initial value is 128, which would create a hearbeat per validator about every 20 minutes (i.e., about 120 heartbeats per validator per epoch). Legal values are all integers larger or equal to 0.
```
davidsiska-vega: should this be a network parameter?

protocol/0064-VALP-validator_performance_based_rewards.md
```diff
@@ -34,7 +34,39 @@ This function is primarily for testing purposes to allow for very short epochs w
...
+The multiplication with 1.1 is adding preventing a validator to get penalised if they got slightly less than others (which always can happen due to the random distribution). Thus, a validator that forwards 95% of the events that others do is not penalised. We could get more precision here (differentiating between the heartbeat and the expected real events etc), but that'd be overthinking. 
+In the end, we make sure no score is bigger than 1 (which might happen due to the multiplicative bonus).
+
+### Total Performance
+As we have several performance measurements, they need to be combined to a total score. To this end, we have a system variable performance weights, 
+which has n+1 parameters (weight_0,.. weight_n) for n measurements (currently 2, the tendermint-performance and the ethereum-performance.Weights are normalised, so the sum of all weights needs to be 1. Also, all individual performance measurements are normalised to be between 0 and 1.
+
+The total performance then is
```
davidsiska-vega, line 79: Unless we go to the multiplicative version... that would remove the need for weights.

Vegaklaus: Indeed; the weights are now also used to choose between multiplicative and additive (or both). If we make a decision that we only need one, they can go.

protocol/0064-VALP-validator_performance_based_rewards.md
```diff
@@ -34,7 +44,39 @@ This function is primarily for testing purposes to allow for very short epochs w
...
+For the Ethereum Heartbeat, we use the network parameter `ethereum_heartbeat_period`. This parameter should be either 0 or a value bigger than the number of validators; the recommended initial value is 128, which would create a hearbeat per validator about every 20 minutes (i.e., about 120 heartbeats per validator per epoch). Legal values are all integers larger or equal to 0.
+
+For every Ethereum block, if the hash of this block mod `ethereum_heartbeat_period` equals the identity of the a validator (taken mod ethereum_heartbeat_period)+1, then this validator has to forward this as an Ethereum event. This event is confirmed by other validators just like any other Ethereum event, but then ignored. If that block also contains a valid Vega event that requires an action, this is forwarded independently by the normal event forwarding mechanisms.
+If the parameter is set to 0, the heartbeats are effectively turned off.
+
+
+#### Performance Measurements
+At the end of each epoch, it is counted how many Ethereum events have been forwarded by each validator; this is (number_of_ethereum_blocks_per_epoch)/`ethereum_heartbeat_period`)+number_of_ethereum_events_per_validator
```
wwestgarth, line 64: Just thinking about how we're going to know what `number_of_ethereum_blocks_per_epoch` actually is: 

If we want it exactly its something that each node will have to work out by pinging ETH blocks until it finds the first one over the Vega epoch time (which given ethereum block times of ~12 would be a a few Vega blocks _later_ that Vega's epoch end block) and then send their answer through consensus etc etc. Which is a bit faffy.

Is there an acceptable estimate we could use instead i.e using the difference in block-heights and the eth-times between heartbeat's to get a guess at the average eth-block time over the period. And then work out how many of those would fit in Vega's epoch length?

Vegaklaus: The performance formulas don't need to be exact - there's some random element in there anyhow, and the reward/punishment formula takes this into account by giving some tolerance; the only precision requirement is agreement on what that value actually is.
What I had in mind when writing is to take the 12/second block time for Ethereum as reliable; otherwise I would measure block height at the beginning and end of a Vega epoch and then run that through agreement. The heartbeats  could be used as well, but then I'd do it directly - everyone posting their own number of heartbeats that had that epoch, and then agreeing on the median.

Vegaklaus: In any case, the plan was also  (if we implement the full reward formula) to initially set the weight parameter for the ETH-event forwarding to 0, so we can get some real life values to verify that it all works before it has serious consequences.

gordsport:
closing as replaced by: 
https://github.com/vegaprotocol/specs/pull/2027

PR #1796, opened 2023-06-27
fix: correction on amending commitment

davidsiska-vega:


gordsport (approved):
Approved in protocol design meeting

PR #1801, opened 2023-06-30
Update ACs for proposals via governance

vega-paul:
0028-GOVE-121 dupe - removed
0028-GOVE-080 / 0028-GOVE-090 removed - Source/Destination type cannot be black for Insurance Pool transfer - has to contain market Id
0028-GOVE-096 - adjusted - transfers can occur for Pending markets


witgaw (approved):

protocol/0028-GOVE-governance.md
```diff
@@ -587,9 +584,7 @@ Below `*` stands for any of `asset, market, updateMarket, updateNetParam, freeFo
 
 - For enacted proposals a token transfer will occur at the time of enactment between the source and destination account if sufficient tokens are held in the source account. A transaction result event will show the successful transfer between two accounts  (<a name="0028-GOVE-094" href="#0028-GOVE-094">0028-GOVE-094</a>)
 - A governance approved recurring transfer will continue even if the source account balance is `0`. In such case the amount transferred will be seen to be `0`. (<a name="0028-GOVE-095" href="#0028-GOVE-095">0028-GOVE-095</a>)
-- Transfers can occur for terminated markets  (<a name="0028-GOVE-096" href="#0028-GOVE-096">0028-GOVE-096</a>)
-- Transfers cannot occur for settled markets and a transaction result event will show the transfer failing with an appropriate message  (<a name="0028-GOVE-097" href="#0028-GOVE-097">0028-GOVE-097</a>)
-- Transfers cannot occur for pending markets unless they become active on or before the enactment time of the transfer (<a name="0028-GOVE-098" href="#0028-GOVE-098">0028-GOVE-098</a>)
+- Transfers can occur for pending, terminated markets, settled markets (<a name="0028-GOVE-096" href="#0028-GOVE-096">0028-GOVE-096</a>)
```
witgaw, line 587: I don’t see a problem with pending, not sure about the settled market though. As I understand that’d require a significant rework from core (I think we still throw away the entire market state once the market is settled) for no obvious benefit (there’s the mechanism to move leftover funds to per-asset fund or treasury from what I remember).

vega-paul: Ive just reflected the behaviour the core exhibits when transfers from/to settled market - tokens are available and are transferred. If this is not expected behaviour can you update this spec to reflect that and I will raise the core issues to make the changes, thanks @witgaw

witgaw: OK, I suppose that mechanism must already be in place to support successor markets.

witgaw (approved):

PR #1804, opened 2023-07-03
feat: use ghcr image to run approbation

mkjmdski:
The approbation run on the specs repo CI is failing:

- https://github.com/vegaprotocol/specs/actions/workflows/quality_check.yml

All failing with the same error:
```npm WARN exec The following package was not found and will be installed: github:vegaprotocol/approbation
Error: Process completed with exit code 1.```

### DIFF
--- a/.github/workflows/quality_check.yml
+++ b/.github/workflows/quality_check.yml
@@ -68,8 +68,8 @@ jobs:
 
       - name: "Run Check AC codes"
         run: |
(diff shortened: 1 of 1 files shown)


edd (approved):
These are fine. Other runs may need to use the new `--output="/workspace/results"` param. This is only required if `--output-jenkins` or `--output-csv` are used (stupidly I defaulted it to `./results` which fails inside docker)

PR #1805, opened 2023-07-03
Update 0044 sla spec1

Sohill-Patel:


witgaw (approved):
The bonus account will get deleted, but let's merge this and I'll deal with it in a separate PR.

PR #1806, opened 2023-07-03
Include stop orders in gas calculations

witgaw:
- Include stop orders in gas calculations.
- Remove LP shapes.

### DIFF
--- a/protocol/0079-TGAP-transaction_gas_and_priority.md
+++ b/protocol/0079-TGAP-transaction_gas_and_priority.md
@@ -60,7 +60,7 @@ gas = min((maxGas/minBlockCapacity)-1,gasOrder)
 
 ```go
 gasCancel = network.transaction.defaultgas + peg cost factor x pegs
-                                        + stop cost factor x stop orders
+                                        + stop cost factor x stops
                                         + level factor x levels
 gas = min((maxGas/minBlockCapacity)-1,gasCancel)
 ```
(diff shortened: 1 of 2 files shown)


cdummett (dismissed):

barnabee (changes requested):
Need to add stop orders to the calculation of gas for a batch. They are a separate "section" of the batch submission.

witgaw:
> Need to add stop orders to the calculation of gas for a batch. They are a separate "section" of the batch submission.

@barnabee is it not enough to just count them along with other cancellations / amendments / submissions? 
https://github.com/vegaprotocol/specs/pull/1806/commits/6f53b4c4a64fa8905e344dc8318ca57ef901b9b6

protocol/0079-TGAP-transaction_gas_and_priority.md
```diff
@@ -87,7 +87,7 @@ gas = min((maxGas/minBlockCapacity)-1,batchGas)
 
 ```go
 gasOliq = network.transaction.defaultgas + peg cost factor x pegs
```
karlem: The LP submission, amendment and cancelation should not produce any orders anymore so why do we still need this?

witgaw: Good point, the gas is the same as for the the regular order submission so doesn't really matter, but I can remove that section.

barnabee (approved):

PR #1808, opened 2023-07-04
Small changes to various specs

candida-d:
Clear up some outdated info, confusing terminology in several specs.

### DIFF
--- a/protocol/0028-GOVE-governance.md
+++ b/protocol/0028-GOVE-governance.md
@@ -275,7 +275,7 @@ Enactment of an asset modification proposal is:
 
 The below table shows the allowable combinations of source and destination account types for a transfer that's initiated by a governance proposal.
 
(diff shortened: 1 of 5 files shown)


protocol/0028-GOVE-governance.md
```diff
@@ -275,7 +275,7 @@ Enactment of an asset modification proposal is:
 
 The below table shows the allowable combinations of source and destination account types for a transfer that's initiated by a governance proposal.
 
-| Source type | Destination type | Governance initiated transfer permitted |
+| Source type | Destination type | Transfer permitted |
```
candida-d, line 278: I had this changed before the follow-on PR came through. However as this is a table heading, the shorter version is still better as it should be clear contextually what the table's about

gordsport (approved):
LGTM

PR #1820, opened 2023-07-10
WIP: isolated margin

### LINKED ISSUE #1627: Margin Isolation
One of the processes that creates lots of data is the searching for margin. This issue is to start discussions around how margin may be able to be Segregated/separated in order to reduce the number of times the protocol needs to search for margin thus reducing the amount of data being created and stored.

In addition to the data creation point above there has been user feedback:
- users want to be able to segregate margin to guarantee maximum leverage / margin at risk relative to position notional (not currently possible in the presence of margin-bearing active orders)
- users should not have to pay fees to manage segregated margin and control leverage, however transfers require fees due to anti spam

This spec update should account for these too.

barnabee:
Add feature for isolated margin.


protocol/0019-MCAL-margin_calculator.md
```diff
@@ -40,23 +40,98 @@
...
 
 Margin levels are used by the protocol to ascertain whether a trader has sufficient collateral to maintain a margined trade. When the trader enters an open position, this required amount is equal to the *initial margin*. Subsequently, throughout the life of this open position, the minimum required amount is the *maintenance margin*. As a trader's collateral level dips below the *collateral search level* the protocol will automatically search for more collateral to be assigned to support this open position from the trader's general collateral accounts. In the event that a trader has collateral that is above the *collateral release level* the protocol will automatically release collateral to a trader's general collateral account for the relevant asset.
 
 **Whitepaper reference:** 6.1, section "Margin Calculation"
 
 In future there can be multiple margin calculator implementations that would be configurable in the market framework. This spec describes one implementation.
 
+
+
+## Isolated margin mode
```
witgaw, line 95: I'm perhaps missing the point, but could we not just implement margin segregation as user-initiated transfers between their general and margin account?

So when margin segregation is flipped on for a market we never search or release collateral. User can always transfer into margin account. They can only transfer from margin into general if the amount left after transfer is at or above the margin release level at that time (this could be a bit tricky as levels change all the time, but maybe we can just allow users to transfer out max allowed at the time of transfer - actual amount is never specified). Transfer is carried out after next MTM so that any inflows/outflows a user may have on their open position are account for first. 

So if they want to say significantly increase their position they'd first need to transfer enough funds into the margin account for that market, and then open then place order as usual.

Those transfers could be free or very cheap (time-throttled?) if we worry about spam.

davidsiska-vega: @barnabee / @TomMcL that sounds reasonable, what do you think?

barnabee: @witgaw I don't think that actually meets the requirement…

The requirement is that the margin allocated to a position is a fixed % of notional. This means the transfer needs to happen at the point that an order trades, not when you place the order.

This drives the key reason for the feature (control leverage) and also the UX for how you'd set that control.

davidsiska-vega: It feels like we should fairly carefully establish what’s needed exactly. It sounds like we want to protocol-level functionality to enable implementation of a “leverage slider” in Console, is that a fair summary?

Jiajia-Cui:
superseded by #1850

PR #1822, opened 2023-07-11
Perps: Consider external data timestamps separately

witgaw:
We should allow to sample timestamps from the oracle data relating to the price of the underlying asset in a flexible way that's not necessarily tied to the time on vega blockchain.

### DIFF
--- a/protocol/0053-PERP-product_builtin_perpetual_future.md
+++ b/protocol/0053-PERP-product_builtin_perpetual_future.md
@@ -13,7 +13,7 @@ Unlike traditional futures contracts, the perpetual futures never expire. Withou
 1. `settlement_asset (Settlement Asset)`: this is used to specify the single asset that an instrument using this product settles in.
 1. `settlement_schedule (Data Source: datetime)`: this data is used to indicate when the next periodic settlement should be carried out.
 1. `settlement_data (Data Source: number)`: this data is used by the product to calculate periodic settlement cashflows.
-1. `margin_funding_factor`: a parameter in the range $[0, 1]$ controlling how much the upcoming funding rate liability contributes to party's margin.
+1. `margin_funding_factor`: a parameter in the range $[0, 1]$ controlling how much the upcoming funding payment liability contributes to party's margin.
 
 Validation: none required as these are validated by the asset and data source frameworks.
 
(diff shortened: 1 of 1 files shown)


davidsiska-vega (approved):

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -53,33 +56,102 @@ cash_settled_perpetual_future.value(quote) {
...
+    }
+    previous_point = p
+}
+
+sum_product += previous_point.s*(funding_period_end-max(funding_period_start,previous_point.t))
+s_twap = sum_product / (funding_period_end - max(funding_period_start, oracle_data_points[0].t))
+```
 
-The next step is to calculate the periodic settlement funding payment. If there are no periodic settlement data points then the periodic settlement is skipped. Otherwise, consider all the periodic settlement data points and calculate the time-weighted average price difference as:
+Only the oracle data point with largest timestamp that's less than or equal to `funding_period_end` (and any data points with larger timestamps) need to be kept from that point on.
```
wwestgarth, line 107: An edge case that has been sitting in my head for a few days:
Given that the timestamp we are associating with the oracle data point may come from a different chain _is it at all possible_ that we might have an oracle data-point that comes in which _for whatever reason_ is a little bit in the future from vega's point-of-view of "now". Would we want want keep those for when they do become within the funding period?

Also another thing:
When we trigger a the periodic settlement from the incoming oracle data, I'm assuming we'll do it at the end of the block? I'm thinking we would also want to processing any oracle data that might also be waiting in to be processed in the same block but just happened to be after the oracle tx was processed. Also MTM happens at the end of a block and we'd want to include that point.

wwestgarth (approved):

protocol/0053-PERP-product_builtin_perpetual_future.md
davidsiska-vega: I guess if the external timestamp says it's ahead of vega time (ahead of a funding payment) then simply don't include data associated to that timepoint in the funding calculation but it will play a role in the next one.

witgaw: Yes, the pseudocode for spot TWAP should imply that behaviour.

So it's always the vega time we care about when choosing which bits of spot data to include in the calculation relevant for a particular funding payment (we come the oracle spot data timestamp to the vega times when deciding which data points fall in the interval). The spot data itself can be ignored at data ingestion time via filters.

It can be at the end of block but it doesn't really matter as we assume the observed price prevails until the end of the funding payment window or until the next observation arrives, so data arriving at the funding period end will have no effect on the funding payment being calculated (only on the next one).

PR #1824, opened 2023-07-11
Fix/sla sampling update

### LINKED ISSUE #1823: Update to SLA spec: sampling freq
Introduce a new netparam which will set after how many transactions, intra-block, we measure whether the SLA is met or not.

davidsiska-vega:


protocol/0044-LIME-lp_mechanics.md
```diff
@@ -23,6 +23,7 @@ Valid values: any decimal number `>= 0` with a default value of `0.1`.
 - `market.liquidity.earlyExitPenalty` (decimal ≥0), sets how much LP forfeits of their bond in case the market is below target stake and they wish to reduce their commitment. If set to `0` there is no penalty for early exit, if set to `1` their entire bond is forfeited if they exit their entire commitment, if set >1, their entire bond will be forfeited for exiting `1/earlyExitPenalty` of their commitment amount.
 - `market.liquidity.probabilityOfTrading.tau.scaling` sets how the probability of trading is calculated from the risk model; this is used to [measure the relative competitiveness of LPs supplied volume](0042-LIQF-setting_fees_and_rewarding_lps.md).
 - `market.liquidity.minimum.probabilityOfTrading.lpOrders` sets a lower bound on the result of the probability of trading calculation.
+- `market.liquidity.sla.evalEveryNumOfTransactions` sets how many transactions in a block are processed before each party that's committed to providing liquidty is evaluated for meeting (or not) their commitment.
```
TomMcL:
```suggestion
- `market.liquidity.sla.evalEveryNumOfTransactions` sets how many transactions in a block are processed before each party that's committed to providing liquidity is evaluated for meeting (or not) their commitment.  
```

ze97286: it's not "how many transactions in a block" right? it's a global interval, i.e. every how many transactions regardless of blocks isn't it?

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -223,10 +223,7 @@ During the epoch, the amount of time in nanoseconds (of Vega time) that each LP
 
   - If the LP is meeting their commitment, store the Vega time of the start of the epoch as the time the LP began meeting their commitment, otherwise store `nothing`.
 
-- At the start of each block generate a pseudorandom integer `k` between `1..N` (inclusive of `1` and `N`) where `N` is the number of transactions in the block (note: transactions not orders, a batch is one transaction for this purpose).
-Use a suitable deterministic seed to minimise the probability of an LP gaming `k` or being able to target transactions around (directly before or after) the point `k`.
-For example, the seed might combine the hash of all transactions in the block itself with the number of transactions `N`.
-Using only information from the prior block as the seed may allow exploits based on pre-generation of `k` and must be avoided.
+- At start of block, set `k` to `0` and then after every `market.liquidity.sla.evalEveryNumOfTransactions` transactions have been processed set `k` to the transaction number of the just processed transaction.
```
barnabee: Is it me or is this wording not quite correct?

protocol/0042-LIQF-setting_fees_and_rewarding_lps.md
```diff
@@ -245,6 +242,7 @@ Using only information from the prior block as the seed may allow exploits based
 
 - At the end of the epoch, calculate the actual observed epoch length `observed_epoch_length` = the difference in nanoseconds between the Vega time at the start of the epoch and the Vega time at the end of the epoch.
 
+Note that because vega time won't be progressing intra-block the above mechanism should ensure that `s_i` gets incremented only if the LP was meeting their commitment at every point this was checked within the block.
```
barnabee: What happens if during the block there are three checks for an LP who was meeting their commitment previously:

check 1: still meeting commitment
check 2: not meeting 
check 3: meeting again 

So they weren’t meeting it in every check but were only not meeting for zero time. I guess that’s ok despite not being true per the above?

barnabee: What about the same as above but where only checks 2 and 3 happen, so the end of last block they were meeting, first check in the block they are not, they are again by the end.

Is _that_ ok also?

ze97286: the more I think about it the more I convince myself (very possibly wrongly) that it's actually better to check once at the end of the block and take their lowest commitment for the block as the commitment to check against.

barnabee: If we just did the checks once per block but using their lowest supplied liquidity at any point during the block, that would be perfect.

barnabee (dismissed):
Approved but need to resolve [this comment thread](https://github.com/vegaprotocol/specs/pull/1824/files#r1262328733) before moving ahead.

barnabee (dismissed):

TomMcL (approved):

PR #1829, opened 2023-07-13
feat: on-chain referral program

### LINKED ISSUE #1669: on-chain referral program
https://docs.google.com/document/d/1k8Tf7ZzLekcV2VdXJQWJ3Zrmf6StGsSOj3h4y_MC-aM/edit

### Referral Program requirements

The aim of the referral program is to allow users and community members to refer new traders to Vega Protocol and provide rewards for both the referrer and the referee. The referral program is to be built as a decentralised framework on the Vega Protocol with initial parameters set to zero.

- The rewards will be distributed from the taker fees paid by the referred users. 
- Referrers will earn a return based on the trading fees paid by the referred users.
- The referred users will enjoy discounted trading fees
- Reward Distribution Rewards earned through the referral program will be paid out at the end of each epoch in the settlement currency of the respective markets.

The program will be activated and implemented as a governance proposal, with the following network parameters:

- Referrers reward: The percentage of trading fees paid by the referred users that will be distributed with the referrers as reward for each tier
- Referred user reward discount: The percentage of trading fees that referred users will receive as a discount when paid for each tier
- Duration of the Program: The specific period for which the referral program will run.
- Minimum VEGA Staked: The minimum quantity of VEGA tokens that need to be staked in order for a user to generate a referral code.

In order to reduce the risk of self-referrals, the program should include tiered rewards based on the referrer's performance. Tier structure will be defined based on the following criteria:

- Tier 1: anyone can create a code.
- Tier 2: volume above $x in the last 7 days
- Tier 3: volume above $x in the last 7 days

    - Referrer needs to maintain certain trading volume to enjoy the higher rewards in Tier 2 and 3
    - The percentage distributed to the referrer at the end of each epoch will be based on the referrers’ tier at that point in time
    - Referred users will automatically receive the applicable discount on their taker fees based on their referrer's tier at the start of the epoch

Referral Code Generation: 

- A separate URL hosted on IPFS, or a section on Console or the governance app needs to be created where users can connect their Vega wallet and generate a unique referral code. This code will be embedded in the referral URL and shared with potential referees.
- Referred users need to connect their wallet and sign an on-chain transaction that links their address to the referrer's address. - 

The referred user will now be able to enjoy discounted taker fees based on the tier of the referrer.
Referral Program Disclaimer: A disclaimer regarding the referral program and its reward parameters should be displayed prominently on the trading front end to inform users about network parameter changes

Dashboard:

- A dashboard or section within Console will be provided to referrers, displaying relevant referral statistics including the number of referred users, their trading volume, and earned rewards.
- The credited earnings will be visible in the referrer's account dashboard, providing transparency on their referral earnings.
- Example of referrer dashboard that should be used - https://app.hyperliquid.xyz/referrals
- The referred user should be able to see the tier that the referrer is on on

Discounted Fees Display:

- The trading front end will clearly display the discounted fees for referred users based on the tier of the referrer.
- This information will be visible during the trading process to ensure transparency.

Internal Reporting and Analytic:

- A reporting and analytics dashboard should be developed to provide referrers with insights into their referral performance.
- The dashboard will display information such as the number of successful referrals, total taker fees generated by referred users, and earned referral rewards.

cdummett:
## Summary

PR adds specs for the following benefit schemes which are funded through taker fees:

- a referral program
- a volume discount program

## Referral Program

The aim of the on-chain referral program is to allow users of the protocol to incentivise users and community members to refer new traders by voting to provide benefits for referrers and/or referees.

A party will be able to create a referral code and share this code with referees. Referees who apply the code will be added to the referrers "referral set". 

Whilst a referral program is active, the following benefits may be available to members of a referral set:

- a **referrer** may receive a proportion of all referee taker fees as a **reward**.
- a **referee** may be eligible for a **discount** on any taker fees they incur.

Providing a party has been associated with a referral set for long enough, they will become eligible for greater benefits as their referral sets running trading volume increases.

To create an emphasis on community, collaboration, and competition. Referrers will be able to designate their referral set as a team. Teams will have additional fields which allow them to be visible on leaderboards and later to compete for team based rewards.

0083-RFPR-on_chain_referral_program_referral_set_hierarchy

## Closes

Closes #1669


protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+The size of the reward or discount will be dependent on the benefit tier the referees [team](#team-mechanics) is currently placed in. A team becomes eligible for greater benefits by reaching a minimum running volume over a number of epochs. Both referrers and referees contribute to their teams running volume.
+
+On-chain referral programs can be created and updated through [governance proposals](#governance-proposals).
+
+## Network Parameters
+
+- `referralProgram.maxBenefitTiers` - limits the maximum number of [benefit tiers](#benefit-tiers) which can be specified as part of a referral program
+- `referralProgram.maxReferralRewardFactor` - limits the maximum reward factor which can be specified as part of a referral program
+- `referralProgram.maxReferralDiscountFactor` - limits the maximum discount factor which can be specified as part of a referral program governance proposal
+- `referralProgram.maxPartyVolumePerEpoch` - limits the volume which is eligible each epoch for referral program mechanisms
```
davidsiska-vega: Presumably this would need to be scaled by the asset's quantum?

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+- `referralProgram.minStakedVegaTokens` - limits team creation to parties staking at least this number of tokens
+
+## Governance Proposals
+
+An on-chain referral program can be proposed via governance. As part of the proposal, the proposer specifies the following fields:
+
+- `name`: name of the referral program
+- `benefit_tiers`: a list of dictionaries with the following fields
+  - `minimum_running_volume`: the required `running_team_volume` for a team to access this tier
```
davidsiska-vega: Another to scale by `quantum` I presume, to make it comparable across markets (I am assuming this is a cross-market thing).

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+        window_length: 7,
+    }
+}
+```
+
+When submitting a referral program proposal through governance the following conditions apply:
+- a proposer cannot set a `start_epoch` or `end_epoch` less than or equal to the current epoch.
+- a proposer cannot set an `end_epoch` less than or equal to the `start_epoch`.
+- the number of tiers in `benefit_tiers` must be less than or equal to the network parameter `referralProgram.maxBenefitTiers`.
+- all `referral_reward_factor` values must be less than or equal to the network parameter `referralProgram.maxReferralRewardFactor`.
```
davidsiska-vega: greater than or equal to `0` and ...

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+- all `referral_discount_factor` values must be less than or equal to the network parameter `referralProgram.maxReferralDiscountFactor`.
+- `window_length` must be an integer greater than zero.
```
davidsiska-vega: strictly greater than '0`?

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+
+To create a team and generate a referral code, a party must fulfil the following criteria:
+
+- party must not currently be a **referrer**
+- party must not currently be a **referee**
+- party must be staking at least `referralProgram.minStakedVegaTokens` tokens.
+
+This staking requirement is constant. If a referrer un-stakes enough tokens to fall below the requirement, they and their referees will no long be eligible for program benefits. If the referrer re-stakes enough tokens to fulfil the staking requirement, the team will become eligible for referral program benefits.
+
+A separate URL hosted on IPFS should allow a party to connect their Vega wallet and fill in the following optional team information:
```
davidsiska-vega: Is this a front end requirement? In that case that can be left as an aside but we should formulate what the node has to provide in terms of transactions it accepts to facilitate this. 

And perhaps it's best not to talk about the front-end part in the protocol part of the spec.

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+
+To become a referee, a referee must create a key (programmatically or through a dApp) and then send a transaction to associate their key with an existing team.
+
+message JoinTeam{
+    id: "abcd3fgh1jklmn0pqf5tuvwzyz"
+}
+
+A user should also be able to do this process by approving a transaction from a dApp.
+
+### Team epoch and running volumes
```
davidsiska-vega: More `quantum` scaling ...

davidsiska-vega: Ah, I see the `quantum` scaling now; so maybe just a clarification near the top of the spec that when we talk of volume we mean it's scaled by quantum already.

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+            "referral_discount_factor": 0.010,
+        },
+    ]
+
+And:
+    team_running_volume=22353
+
+Then:
+    referral_reward_factor=0.005
+    referral_reward_factor=0.005
```
davidsiska-vega:
```suggestion
    referral_discount_factor=0.005
```

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+- Calculate the rewards due to the referrer.
+    ```pseudo
+    infrastructure_fee_reward = floor(infrastructure_fee * referral_reward_factor) 
+    liquidity_fee_reward = floor(liquidity_fee * referral_reward_factor) 
+    maker_fee_reward = floor(maker_fee * referral_reward_factor) 
+    ```
+- Calculate the discounts due to the referee.
+    ```pseudo
+    infrastructure_fee_discount = floor(infrastructure_fee * referral_discount_factor)
+    liquidity_fee_discount = floor(liquidity_fee * referral_discount_factor)
```
davidsiska-vega: I just wonder whether the above are used consistently with the sprit of the numbers in the above example. Because with increasing volume, I see the reward factors increasing (good) but then I'd expect 

```
floor(liquidity_fee * (1-referral_discount_factor))
```

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
+# On Chain Referral Program Specification
+
+## Summary
+
+The aim of the on-chain referral program is to incentivise users and community members to refer new traders by providing benefits for both referrers and referees.
```
barnabee:
```suggestion
The aim of the on-chain referral program is to allow users of the protocol to incentivise users and community members to refer new traders by voting to provide benefits for referrers and/or referees. 
```

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
+# On Chain Referral Program Specification
+
+## Summary
+
+The aim of the on-chain referral program is to incentivise users and community members to refer new traders by providing benefits for both referrers and referees. 
+
+Whilst a referral program is active, the following benefits will be available to eligible parties:
```
barnabee:
```suggestion
Whilst a referral program is active, the following benefits may be available to eligible parties:
```

Either benefit could be set to zero

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+
+- a **referrer** will receive a proportion of all referee taker fees as a **reward**.
+- a **referee** will be eligible for a **discount** on any taker fees they incur.
+
+The size of the reward or discount will be dependent on the benefit tier the referees [team](#team-mechanics) is currently placed in. A team becomes eligible for greater benefits by reaching a minimum running volume over a number of epochs. Both referrers and referees contribute to their teams running volume.
```
barnabee:
```suggestion
The size of the reward or discount will be dependent on the benefit tier the referee's [team](#team-mechanics) is currently placed in. A team becomes eligible for greater benefits by reaching a minimum running volume over a number of epochs. Both referrers and referees contribute to their team's running volume.
```

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+
+An on-chain referral program can be proposed via governance. As part of the proposal, the proposer specifies the following fields:
+
+- `name`: name of the referral program
+- `benefit_tiers`: a list of dictionaries with the following fields
+  - `minimum_running_volume`: the required `running_team_volume` for a team to access this tier
+  - `referral_reward_factor`: the proportion of the referees taker fees to be rewarded to the referrer
+  - `referral_discount_factor`: the proportion of the referees taker fees to be discounted
+- `start_epoch`: the first epoch the program will become `STATE_ACTIVE` and benefits will be enabled
+- `end_epoch`: the first epoch the program will become `STATE_CLOSED` and benefits will be disabled
```
TomMcL: Intuitively to me this would be the last active epoch rather than the first inactive

protocol/0082-RFPR-on_chain_referral_program.md
TomMcL: Yeah I vote kill anything non-protocol related (even datanodey things make me sad)

protocol/0082-RFPR-on_chain_referral_program.md
barnabee: Yeah I'd expect this to be specified in terms of quantum

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+
+- `referralProgram.maxBenefitTiers` - limits the maximum number of [benefit tiers](#benefit-tiers) which can be specified as part of a referral program
+- `referralProgram.maxReferralRewardFactor` - limits the maximum reward factor which can be specified as part of a referral program
+- `referralProgram.maxReferralDiscountFactor` - limits the maximum discount factor which can be specified as part of a referral program governance proposal
+- `referralProgram.maxPartyVolumePerEpoch` - limits the volume which is eligible each epoch for referral program mechanisms
+- `referralProgram.minStakedVegaTokens` - limits team creation to parties staking at least this number of tokens
+
+## Governance Proposals
+
+An on-chain referral program can be proposed via governance. As part of the proposal, the proposer specifies the following fields:
```
barnabee:
```suggestion
Enabling or changing the terms of the on-chain referral program can be proposed via governance. As part of the proposal, the proposer specifies the following fields:
```

protocol/0082-RFPR-on_chain_referral_program.md
cdummett: This doesn't get scaled, rather the volume of a parties trades are scaled into quantum units to compare to this value.

Will make this clearer though.

protocol/0082-RFPR-on_chain_referral_program.md
cdummett: Same as this [comment](https://github.com/vegaprotocol/specs/pull/1829/files#r1262678183).

Will make it clearer again though 👍

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+
+- `name`: name of the referral program
+- `benefit_tiers`: a list of dictionaries with the following fields
+  - `minimum_running_volume`: the required `running_team_volume` for a team to access this tier
+  - `referral_reward_factor`: the proportion of the referees taker fees to be rewarded to the referrer
+  - `referral_discount_factor`: the proportion of the referees taker fees to be discounted
+- `start_epoch`: the first epoch the program will become `STATE_ACTIVE` and benefits will be enabled
```
barnabee: I think the interaction between this and the proposal's enactment date might be annoying.

Perhaps it would be better to say the programme becomes active on the next epoch change after the enactment date/time of the proposal is reached?

protocol/0082-RFPR-on_chain_referral_program.md
TomMcL: This is the discount no? It would be weird if `discount = x * (1-discount_factor)`

protocol/0082-RFPR-on_chain_referral_program.md
barnabee: I think this might be better off as a date/time, where the programme ends at the first epoch end after the date/time is reached.

Also, we should be clear whether other programmes remain "stacked" until they end or if the most recently enacted programme should replace all others, e.g. do we have this:

1. Propose programme A starting 1st Jan and ending 31st Dec
2. Proposal accepted and programme A begins on 1st Jan
3. Propose programm B starting 1st June ending 31st Aug
4. Proposal accepted and programme B begins on 1st June
5. All programmes stop on 31st Aug

OR

1. Propose programme A starting 1st Jan and ending 31st Dec
2. Proposal accepted and programme A begins on 1st Jan
3. Propose programm B starting 1st June ending 31st Aug
4. Proposal accepted and programme B begins on 1st June
5. Programme B stops and programme A resumes on 31st Aug
6. Programme A stops on 31st Dec

In my opinion it should probably be the first option, which seems simpler to reason about and implement. A series of programmes can still be created by using different enactment dates, and when we have multiple actions in a single governance proposal these could all be combined to one series.

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+- And then update the fees.
+    ```pseudo
+    infrastructure_fee = infrastructure_fee - infrastructure_fee_reward - infrastructure_fee_discount
+    liquidity_fee = liquidity_fee - liquidity_fee_reward - liquidity_fee_discount
+    maker_fee = maker_fee - maker_fee_reward - maker_fee_discount
+    ```
+
+Note the rewards and discounts are floored rather than raised to ensure the final infrastructure fee cannot be negative.
+
+The network can then carry out the normal fee transfers using the updated fee amounts followed by additional transfers from the referees general account to the referrers general account. These transfers will use the following new transfer types.
```
TomMcL, line 224: Should there instead be a new fee type which goes directly to the referrer rather than just a transfer? And maybe should be clear here that it is the `infrastructure_fee_reward` listed above (and maybe rename that to `_referrer_reward` or something. It threw me at first to see that as a discount to the fee the referee was paying.

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+
+- `TRANSFER_TYPE_MAKER_FEE_REWARD_PAY`
+- `TRANSFER_TYPE_MAKER_FEE_REWARD_DISTRIBUTE`
+- `TRANSFER_TYPE_LIQUIDITY_FEE_REWARD_PAY`
+- `TRANSFER_TYPE_LIQUIDITY_FEE_REWARD_DISTRIBUTE`
+- `TRANSFER_TYPE_INFRASTRUCTURE_FEE_REWARD_PAY`
+- `TRANSFER_TYPE_INFRASTRUCTURE_FEE_REWARD_DISTRIBUTE`
```
TomMcL: Do we need the pay and distribute side of the types if it's done in one transfer rather than to some holding bucket?

protocol/0082-RFPR-on_chain_referral_program.md
barnabee: Yeah I'd remove front end requirements, we just need the transactions we define and data node APIs to facilitate the front end, in this spec.

There should be a front end spec/ticket (check with FE team what they need) as well for this though.

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,235 @@
...
+| `STATUS_REJECTED`  | No               | Governance vote fails                                     | na                                                   |
+| `STATUS_PENDING`   | No               | Governance vote passes                                    | Current epoch greater than or equal to `start_epoch` |
+| `STATUS_STOPPED`   | No               | There is already an existing program with `STATUS_ACTIVE` | na                                                   |
+| `STATUS_ACTIVE`    | Yes              | Previously `STATUS_PENDING`                               | Current epoch less than or equal to `end_epoch`      |
+| `STATUS_CLOSED`    | No               | Previously `STATUS_ACTIVE`                                | na                                                   |              
+
+## Team Mechanics
+A team is comprised of a referrer and all their referees. There can only ever be one referrer per team but the number of referees is unlimited.
+
+### Creating a team
```
barnabee: We should make this "Creating or updating a team" and note that sending the transaction a second or subsequent time must not change the referal code/team ID but should cause the metadata (name, URL, image) to be updated.

barnabee (changes requested):

protocol/0082-RFPR-on_chain_referral_program.md
cdummett: Not sure I follow, your equation would give a discounted fee rather the discount to be applied no?

We could do it that way (and then not have to update the fees) but the logic for calculating the discount and then applying it to the fee is so we could then expose the discount value through an API.

(Which I now realise isn't actually in the spec - will add that).

protocol/0082-RFPR-on_chain_referral_program.md
davidsiska-vega: 💅

protocol/0082-RFPR-on_chain_referral_program.md
davidsiska-vega: Sorry my bad: I was reading it wrong - I was reading new fee (which isn't what it said, sorry), rather than discount. Yes setting discount to this is correct.

protocol/0082-RFPR-on_chain_referral_program.md
cdummett: Yep, good point can remove the distributes thanks 👍

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,238 @@
...
+
+- party must not currently be a **referrer**
+- party must not currently be a **referee**
+- party must be staking at least `referralProgram.minStakedVegaTokens` tokens.
+- party must not have a liquidity provision in any of the following states:
+    - `STATUS_ACTIVE`
+    - `STATUS_PENDING`
+    - `STATUS_UNDEPLOYED`
+
+This staking requirement is constant. If a referrer un-stakes enough tokens to fall below the requirement, they and their referees will no long be eligible for program benefits. If the referrer re-stakes enough tokens to fulfil the staking requirement, the team will become eligible for referral program benefits.
```
cdummett, line 97: Do we want a referrer un-staking to punish referees also?

Or should we do something more along the lines of, set `referral_reward_factor` of all referees to `0` until the staking requirement is met (i.e. they receive no rewards).

cdummett, line 97: Do we want a referrer un-staking to punish referees also?

Or should we do something more along the lines of, set referral_reward_factor of all referees to 0 until the staking requirement is met (i.e. they receive no rewards).

cdummett: Do we want a referrer un-staking to punish referees also?

Or should we do something more along the lines of, set `referral_reward_factor` of all referees to `0` until the staking requirement is met (i.e. they receive no rewards).

davidsiska-vega: I would just say that if referrer hasn't staked enough you don't get any of the benefits (both referrer and referred).

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,238 @@
...
+
+On-chain referral programs can be created and updated through [governance proposals](#governance-proposals).
+
+## Network Parameters
+
+- `referralProgram.maxBenefitTiers` - limits the maximum number of [benefit tiers](#benefit-tiers) which can be specified as part of a referral program
+- `referralProgram.maxReferralRewardFactor` - limits the maximum reward factor which can be specified as part of a referral program
+- `referralProgram.maxReferralDiscountFactor` - limits the maximum discount factor which can be specified as part of a referral program governance proposal
+- `referralProgram.maxPartyVolumePerEpoch` - limits the volume in quantum units which is eligible each epoch for referral program mechanisms
+- `referralProgram.minStakedVegaTokens` - limits team creation to parties staking at least this number of tokens
```
davidsiska-vega: @barnabee / @cdummett / @TomMcL do you see any value to have, on top of this minimum, a requirement for extra vega tokens for each team member on the team?

cdummett: Can definitely see the value, adds a disadvantage to teams merging to access higher tiers which I think is good

I think though if we add this we will need to add supporting features for the referrer like being able to remove "dead-weight" team members and possibly "approve" new team members which is added complexity for what I feel should be a fairly simple referral program (at least at first).

davidsiska-vega: @barnabee says - not needed.

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,282 @@
...
+    - Proposal for program B accepted and overrides program A the first epoch after 1st June.
+    - Program is closed first epoch after 31st Aug, there should be no active proposals.
+
+### Team Mechanics
+
+#### Creating / updating teams
+
+1. If a party **is not** currently a referrer, the party can **create** a team, by submitting a signed `CreateParty` transaction.
+1. If a party **is** currently a referrer, the party can **update** a team, by submitting a signed `CreateParty` transaction.
+1. If one or more of the following conditions are not met, any `CreateParty` transaction should be rejected.
```
jeremyletang:
```suggestion
1. If a party **is not** currently a referrer, the party can **create** a team, by submitting a signed `CreateTeam` transaction.
1. If a party **is** currently a referrer, the party can **update** a team, by submitting a signed `CreateTeam` transaction.
1. If one or more of the following conditions are not met, any `CreateTeam` transaction should be rejected.
```

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,282 @@
...
+1. If the referrer of a team currently not eligible for benefits re-stakes enough tokens, their team will become eligible for benefits from the next epoch.
+
+1. If a party has created a team (i.e. is a referrer) any future liquidity provision transactions from the party should be rejected.
+
+#### Joining / moving teams
+
+1. If a party **is not** currently a **referee**, the party can join a team by submitting a signed `JoinParty` transaction.
+1. If a party **is** currently a **referee**, the party can move team (at the start of the next epoch) by submitting a signed `JoinParty` transaction.
+1. If a party **is** currently a **referee** and submits multiple `JoinParty` transactions in an epoch, the latest valid `JoinParty` transaction will be applied.
+1. If one or more of the following conditions are not met,  any `JoinParty` transaction should be rejected.
```
jeremyletang:
```suggestion
1. If a party **is not** currently a **referee**, the party can join a team by submitting a signed `JoinTeam` transaction.
1. If a party **is** currently a **referee**, the party can move team (at the start of the next epoch) by submitting a signed `JoinTeam` transaction.
1. If a party **is** currently a **referee** and submits multiple `JoinTeam` transactions in an epoch, the latest valid `JoinTeam` transaction will be applied.
1. If one or more of the following conditions are not met,  any `JoinTeam` transaction should be rejected.
```

protocol/0082-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,298 @@
...
+The Trades API should now also expose the following additional information for every trade:
+
+- Referral program rewards
+  - `infrastructure_fee_reward`
+  - `liquidity_fee_reward`
+  - `maker_fee_reward`
+- Referral program discounts
+  - `infrastructure_fee_discount`
+  - `liquidity_fee_discount`
+  - `maker_fee_discount`
```
cdummett: Think this hopefully covers (1) from the [thread](https://vegaprotocol.slack.com/archives/CAHA5EX0F/p1690293264541239).

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,337 @@
...
+    - Propose program B with `enactment_timestamp` 1st June and `closing_timestamp` 31st Aug.
+    - Proposal for program B accepted and overrides program A the first epoch after 1st June.
+    - Program is closed first epoch after 31st Aug, there should be no active proposals.
+
+### Team Mechanics
+
+#### Creating / updating teams
+
+1. If a party **is not** currently a referrer, the party can **create** a team, by submitting a signed `CreateTeam` transaction.
+1. If a party **is** currently a referrer, the party can **update** a team, by submitting a signed `CreateTeam` transaction.
```
ValentinTrinque: I would rather have a dedicated payload for that `UpdateTeams` that can have the `teamID` a property (which also mean the team ID / referral code is immutable). Is there anything preventing that ? Is it possible to change the referral code ?

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,337 @@
...
+                "referral_reward_factor": 0.005,
+                "referral_discount_factor": 0.005,
+            },
+            {
+                "minimum_running_volume": 30000,
+                "minimum_epochs_in_team": 31,
+                "referral_reward_factor": 0.010,
+                "referral_discount_factor": 0.010,
+            },
+        ],
```
ValentinTrinque: So the Tier number (Tier **1**, Tier **2**, ...) is based on this list being sorted by `minimum_running_volume` ? Is that it ?

About validation, should we ensure the factor keeps getting higher as we climb the Tier ? Or is it possible to Tier 2 to have less return than Tier 1 ?

protocol/0083-RFPR-on_chain_referral_program.md
cdummett: If that is preferable then I can update the spec with that separate `UpdateTeam` transaction.

The `teamID` / referral code for a team should never change.

protocol/0083-RFPR-on_chain_referral_program.md
cdummett: Yeah exactly, should be sorted by `minimum_running_volume`.

I think re validating the factors, I would leave it up to governance, gives a bit more flexibility and I can't see it being exploited since the proposal has to be voted on.

There is a small concern referees will jump between teams (hence the epochs in team restriction) so having no validation opens up the option of setting a ceiling to the running volume as well (where rewards drop off if your team is creating too much volume). @barnabee thoughts?

protocol/0083-RFPR-on_chain_referral_program.md
ValentinTrinque: Ok, thanks. Let's have a separate payload then

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,337 @@
...
+
+### Setting benefit factors
+
+Whilst a referral program is `STATUS_ACTIVE`, at the start of an epoch (after pending `JoinTeam` transactions have been processed) the network must set the `referral_reward_factor` and `referral_discount_factor` for each referee.
+
+#### Setting the referral reward factor
+
+The `referral_reward_factor` should be set by identifying the "highest" benefit tier where the following conditions are fulfilled.
+
+- `team_running_volume` is greater than the tiers `minimum_running_volume`.
```
bwallacee: should this be greater than or equal to?

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,366 @@
...
+To generate a team id / referral code, the party must submit a signed `CreateTeam` transaction with the following optional team information.
+
+- `enable_rewards`: a boolean value which specifies whether the team should be eligible for team based rewards (spec not started).
+- `name`: an optional team name to be added to the referral banner.
+- `teamUrl`: an optional link to a team forum, discord, etc.
+- `avatarUrl`: an optional url to an image to be used as the teams avatar
+
+```protobuf
+message CreateTeam
+    enable_rewards: True,
```
cdummett: @ValentinTrinque How much of a pain would it be if the `enable_rewards` field changed name and datatype?

UX are designing the different journeys and raised the good point that having to sign a transaction with `enable_rewards=False` isn't the clearest when this isn't to do with enabling referral rewards...

Changes would be:
- `enable_rewards` renamed to `referral_type`
- datatype changed from `bool` to an `enum` with the following names
  - `REFERRAL_TYPE_UNSPECIFIED`:
  - `REFERRAL_TYPE_CODE`: Create a code to refer new users
  - `REFERRAL_TYPE_TEAM`: Create a team to compete on leaderboards

(cc: @ckpringle @mattrussell36  )

ValentinTrinque: It can be done. I will have to change the core implementation and the proto.

protocol/0029-FEES-fees.md
```diff
@@ -21,6 +21,44 @@ Fees are calculated and collected in the settlement currency of the market, coll
 
 Note that maker_fee = 0 if there is no maker, taker relationship between the trading parties (in particular auctions).
 
+## Applying benefit factors
+
+Before fees are transferred, if there is an [active referral program](./0083-RFPR-on_chain_referral_program.md) or [volume discount program](./0085-VDPR-volume_discount_program.md), each parties fee components must be modified as follows.
+
+1. Calculate any referral rewards due to the parties referrer.
```
barnabee: We should explicitly make sure this is applied *after* all fee discounts (from this or any other programme) so referral rewards are a % of the amount actually collected by the protocol?

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,402 @@
...
+The referral program will start the epoch after the `enactment_timestamp` is reached.
+
+## Referral program lifecycle
+
+After a referral program [proposal](#governance-proposals) is validated and accepted by the network, the network referral program is created / updated and can be one of the following states. The current state of the network referral program should be exposed via an API.
+
+| Status               | Benefits Enabled | Condition for entry                                       | Condition for exit                                                |
+| -------------------- | ---------------- | --------------------------------------------------------- | ----------------------------------------------------------------- |
+| `STATUS_PROPOSED`    | No               | Governance proposal valid and accepted                    | Governance proposal voting period ends (or proposal is invalid)   |
+| `STATUS_REJECTED`    | No               | Governance vote fails (or is invalid)                     | New governance proposal submitted to the network                 |
```
barnabee: If there is an active programme and a new proposal is rehected or fails, what is the status of the programme? Should strill be active but not sure if that is clear here.

I'd suggest maybe rejected is not a valid state for the programme, and you only have:

**INACTIVE**: there is no current programme and no proposal either still in voting or passed voting for there to be one
**PROPOSED**: there is no current programme and no proposal that has passed voting but at least one proposal is still being voted on
**PENDING**: there is no current programme but a proposal has passed voting (voting has closed and criteria were met) but not yet been enacted
**ACTIVE**: there is currently an active referral programme

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,402 @@
...
+
+### Creating a referral set
+
+To create a new referral set and become a referrer, a party must fulfil the following criteria:
+
+- party must not currently be a **referrer**
+- party must not currently be a **referee**
+- party must be staking at least `referralProgram.minStakedVegaTokens` tokens
+
+The staking requirement is constant. If a referrer un-stakes enough tokens to fall below the requirement, they and their referees will immediately no longer be eligible for referral benefits. If the referrer re-stakes enough tokens to fulfil the staking requirement, they and their referees will become eligible for referral benefits **at the start of the next epoch**. Note, for the case where a party does not re-stake, the protocol will allow referees will to "move" referral sets by [applying](#applying-a-referral-code) a new referral code.
```
barnabee:
```suggestion
The staking requirement is constant. If a referrer un-stakes enough tokens to fall below the requirement, they and their referees will immediately no longer be eligible for referral benefits. If the referrer re-stakes enough tokens to fulfil the staking requirement, they and their referees will become eligible for referral benefits **at the start of the next epoch**. Note, for the case where a party does not re-stake, the protocol will still allow referees to "move" referral sets by [applying](#applying-a-referral-code) a new referral code as normal.
```

barnabee (changes requested):

barnabee (approved):

PR #1832, opened 2023-07-13
Perps: include optional parameters in the funding payment calculation

### LINKED ISSUE #868: Review/Update spec for new product 'Perpetuals'

witgaw:
Include optional parameters in the funding payment calculation to allow modelling payment types observed in some of the existing markets.

close #868

### DIFF
--- a/protocol/0053-PERP-product_builtin_perpetual_future.md
+++ b/protocol/0053-PERP-product_builtin_perpetual_future.md
@@ -6,7 +6,7 @@ Background reading: [1](https://www.paradigm.xyz/2021/05/everlasting-options/#Pe
 
 Perpetual futures are a simple "delta one" product. Mark-to-market settlement occurs with a predefined frequency as per [0003-MTMK-mark_to_market_settlement](0003-MTMK-mark_to_market_settlement.md). Additionally, a settlement using external data is carried out whenever `settlement_schedule` is triggered. Data obtained from the `settlement_data` oracle between to consecutive `settlement_schedule` events is used to calculate the funding payment and exchange cashflows between parties with open positions in the market.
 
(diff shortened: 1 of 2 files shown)


protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -133,10 +136,11 @@ Only the internal data point with largest timestamp needs to be kept from that p
 
 #### Funding payment calculation
 
-The next step is to calculate the periodic settlement funding payment
+The next step is to calculate the periodic settlement funding payment. We allow the optional interest rate and clamp component, where $\text{clamp}(a,b;x)=min(b,max(a, x))$. The funding payment then takes the form:
 
 ```go
-funding_payment = f_twap - s_twap
+delta_t = funding_period_end - max(funding_period_start, internal_data_points[0].t)
+funding_payment = f_twap - s_twap + min(clamp_upper_bound*s_twap,max(clamp_lower_bound*s_twap, exp(interest_rate*delta_t)*s_twap-f_twap))
```
wwestgarth: I feel like I'm missing something here. In the case where the interest-rate calculation lands between the clamping bounds we'll have:
```
funding_payment = f_twap - s_twap + e^(r*d_t)*s_twap - f_twap
```
and the `f_twap` cancel out meaning the funding rate is independent of the internal mark-prices, which doesn't seem right?

davidsiska-vega: Yes, that's in fact (seemingly) the desired effect. 

image

wwestgarth: Ok interesting, the diagrams are helpful it seeing the effect it has. So we're basically saying, if the average-index price stays within these bounds payment it based on interest calc of the index price, if it moves outside we start to take into account the difference with the mark price and offset by the bounds. 

And the choice of `a,b` in figure 2 are down as percentages, but I guess in reality they will be chosen has +/- some percent of the expected interest calculation based on the _current_ index price? i.e I know `r` and the time-delta because I choose them when I setup the market, and can look at the current index price and go "right so i'll set `a = S*e^{blah} * 1.05, b=S*e^{blah} * 0.95`. And if the price move alot during the lifetime of the perp and is _always_ outside of those bounds they'd have to do a MarketUpdate proposal to bring it back in line.

wwestgarth (dismissed):

protocol/0053-PERP-product_builtin_perpetual_future.md
witgaw: Yes, that's all correct.

davidsiska-vega (approved):

wwestgarth (approved):

PR #1835, opened 2023-07-14
chore: update spot ACs after 1st QA review

gordsport:
This PR updates the spot ACs in both the specs and the feature.json file


Jiajia-Cui (approved):
LGTM

PR #1841, opened 2023-07-18
fix: remove spot ac 0026-AUCT-030

Sohill-Patel:
ac [0026-AUCT-030](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0026-AUCT-auctions.md#0026-AUCT-030) contradicts spot spec https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0080-SPOT-product_builtin_spot.md#8-auctions


Jiajia-Cui (approved):
LGTM

PR #1850, opened 2023-07-21
feat: isolated margin

### LINKED ISSUE #1627: Margin Isolation

TomMcL:
close #1627


protocol/0019-MCAL-margin_calculator.md
```diff
@@ -42,21 +42,115 @@
...
+## Isolated margin mode
+
+When in isolated margin mode, the position on the market has an associated margin factor.
+The margin factor must be greater than 0 and less than or equal to 1.
+
+Isolated margin mode can be enabled by placing an _update margin mode_ transaction.
+
+The default when placing an order with no change to margin mode specified must be to retain the current margin mode of the position.
+
+The margin mode cannot be changed whilst an account has open positions or orders on a market.
```
TomMcL: Remove this line

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -42,21 +42,115 @@
...
 In future there can be multiple margin calculator implementations that would be configurable in the market framework. This spec describes one implementation.
 
+
+
+## Isolated margin mode
+
+When in isolated margin mode, the position on the market has an associated margin factor.
+The margin factor must be greater than 0 and less than or equal to 1.
+
+Isolated margin mode can be enabled by placing an _update margin mode_ transaction.
```
TomMcL: Add detail on batch txs

TomMcL: - Also require collateral on mkt
- Order: Cancel -> Margin update -> Amend -> Create

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -42,21 +42,125 @@
...
+
+When switching to isolated margin mode, the following steps will be taken:
+  1. For any active position, calculate `average entry price * abs(position) * margin factor`. Calculate the amount of funds which will be added to, or subtracted from, the general account in order to do this. If additional funds must be added which are not available, reject the transaction immediately.
+  2. For any active orders, calculate the quantity `limit price * remaining size * margin factor` which needs to be placed in the order margin account. Add this amount to the difference calculated in step 1. If this amount is less than or equal to the amount in the general account, perform the transfers (first move funds into/out of margin account, then move funds into the order margin account). If there are insufficient funds, reject the transaction.
+  3. Move account to isolated margin mode on this market
+
+When switching from isolated margin mode to cross margin mode, the following steps will be taken:
+   1. Any funds in the order margin account will be moved to the cross margin account.
```
davidsiska-vega: Does this imply that we're adding two accounts per party per market i.e. 
- isolated margin account and
- order margin account? 

And do I understand correctly that the current margin account (cross margin account stays)?

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -42,21 +42,125 @@
...
+When in isolated margin mode, the position on the market has an associated margin factor.
+The margin factor must be greater than 0 and less than or equal to 1.
+
+Isolated margin mode can be enabled by placing an _update margin mode_ transaction.
+
+The default when placing an order with no change to margin mode specified must be to retain the current margin mode of the position.
+
+### Placing an order
+
+When submitting, amending, or deleting an order in isolated margin mode and continuous trading, a two step process will be followed:
```
davidsiska-vega, line 141: Maybe this is somehow implicitly implied somewhere, but: where exactly do we check the margin balances against the "maintenance x initial" level?

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -42,21 +42,125 @@
...
 **Whitepaper reference:** 6.1, section "Margin Calculation"
 
 In future there can be multiple margin calculator implementations that would be configurable in the market framework. This spec describes one implementation.
 
+
+
+## Isolated margin mode
+
+When in isolated margin mode, the position on the market has an associated margin factor.
+The margin factor must be greater than 0 and less than or equal to 1.
```
cdummett: How is the `margin_factor` set, is it a market parameter set in the market configuration?

Also does their need to be any validation to stop someone setting the margin factor < the smallest risk factor? I guess it wouldn't break anything but parties wouldn't be able to place any orders in isolated margin mode.

protocol/0019-MCAL-margin_calculator.md
TomMcL: I guess removing cross would be clearer here, it's just meaning "the margin account"

protocol/0019-MCAL-margin_calculator.md
TomMcL: That's what the user sets to determine their isolated leverage. Ultimately there should be a check that they can't open a position which *immediately* is closed out but that's about it

protocol/0010-MARG-margin_orchestration.md
```diff
@@ -37,6 +37,7 @@ This specification outlines:
 The [margin calculator](./0019-MCAL-margin_calculator.md) will calculate the margin levels when instructed to do so. It will return four margin levels for each trader:
```
witgaw:
```suggestion
The [margin calculator](./0019-MCAL-margin_calculator.md) will calculate the margin levels when instructed to do so. It will return five margin levels for each trader:
```

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,131 @@
...
+         2. If they do not have sufficient, the order will be rejected in it's entirety for not meeting margin requirements.
+      3. If the trade would decrease the party's position, that portion will trade and margin will be released as in the Decreasing Position section
+   2. If the order is not persistent this is the end, if it is persistent any portion of the order which has not traded in step 1 will move to being placed on the order book. 
+      1. At this point, the party's general account will be checked for sufficient margin to cover `new margin = limit price * remaining size * margin factor`, as this is the worst-case trade price of the remaining component. If there is insufficient, the remaining portion of the order will be `stopped`. If there is sufficient, the order margin will be calculated for the current position plus the party's orders including this potential new order. If this order margin is greater than the sum of `new margin + margin balance + order margin balance` then the order will be rejected. Otherwise, an amount `new margin` will be moved into the party's Order Margin account and the order will be placed on the book.
```
witgaw: I'd be up for using limit price rather than mark price also in cross-margining mode too. That should allow for performance optimisation (both within cross-margining mode and when switching modes).

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,131 @@
...
+
+### Potential position margin level:
+
+1. **Order margin**: the amount of additional margin on top of the amount in the margin account that is required for the party's current active orders.
+Note that this may be zero if the active orders can only decrease the position size. 
 
 1. ***Maintenance margin***
 1. ***Collateral search level***
 1. ***Initial margin***
 1. ***Collateral release level***
```
witgaw:
```suggestion
```

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,131 @@
...
+
+When submitting, amending, or deleting an order in isolated margin mode and an auction is active there is no concept of an order trading immediately on entry, however the case of someone putting in a sell order for a very low price must be handled (as it is likely to execute at a much higher price). To handle this, when in an auction the amount taken into the order margin account should be the larger of either `limit price * size * margin factor` or `max(mark price, indicative uncrossing price) * size * margin factor`. All other steps are as above.
+
+### Increasing Position
+
+When an order trades which increases the position (increasing the absolute value of the trader's position), the target amount to be transferred is calculated as:
+
+```math
+margin to add = margin factor * VWAP of new trades * total size of new trades
+```
```
witgaw: Might be worth using the notion of a sum-product so that this doesn't really get implemented as VWAP/total volume.

```suggestion
$$
\text{margin to add} = \text{margin factor} * \text{VWAP of new trades} * \abs{\text{total size of new trades}}
$$
```

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,131 @@
...
+
+NB: In implementation, for any volume that trades immediately on entry, the additional margin may be transferred directly from the general account to the margin account. 
+
+### Reducing Position
+
+When an order trades which reduces the trader's current position the amount to be withdrawn from the margin account is calculated as:
+
+```math
+margin to remove = current margin * abs(total size of new trades) / abs(position prior to trade)
+```
```
witgaw: ~~Unless I'm missing something I think same position notional should always result in the same margin:~~

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,131 @@
...
+
+### Changing Sides
+
+A single order could move a trader from a short to a long position (or vice versa) which either increases or decreases the required margin. When this occurs, the trade should be seen as two steps. One in which the position is reduced to zero (and so margin can be released) followed immediately by one which increases to the new position. (Note that these should *not* be two separate trades, merely modelled as such. The two account movements should occur at once with no other changes allowed between them).
+
+
+### Position Resolution
+
+Isolated margin acts slightly differently to cross margining mode at times when the position becomes distressed. In the case of an open position dropping below maintenance margin levels active orders will remain active as these are margined separately and will not be cancelled.
+
```
witgaw, line 196: Probably worth spelling out that general account should never be searched for funds in this mode.

witgaw (commented):
Looks sensible to me, just some comments (in addition to in-code comments below):
* `0010` might need a bit more work, unless we just want to say in `0019` that isolated margin mode sets `search level` to 0 and `release level` to infinity.
* In `0019` moving (or just repeating) this:
  ```
  The protocol will attempt to set the funds within the margin account equal to average entry price * current position * new margin factor. This value must be above the initial margin for the current position or the transaction will be rejected.
  ```
  towards the beginning of isolated section would make the whole concept much easier to follow imo.

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,131 @@
...
 **Whitepaper reference:** 6.1, section "Margin Calculation"
 
 In future there can be multiple margin calculator implementations that would be configurable in the market framework. This spec describes one implementation.
 
+
+
+## Isolated margin mode
+
+When in isolated margin mode, the position on the market has an associated margin factor.
+The margin factor must be greater than 0 and less than or equal to 1.
```
witgaw: Should we not require `margin factor >= initial_margin_scaling_factor*(linear slippage factor + risk_factor)` (can do max of risk factors or consider each side separately)? To assure position can always be opened irrespective of that the slippage currently is? 

There's also the quadratic risk factor, but that would need to consider volume, so we can either omit it or include it in the sum to assure a new position of size up to 1 can always be opened.

protocol/0019-MCAL-margin_calculator.md
barnabee: > same position notional should always result in the same margin

Agree, that's also my understanding where position_notional means `average_entry_price * position_size`.

However, I can't get my head around how the price of the closing trades could have an impact on the margin scaling for the reduction except insofar as it determines the [settlement](https://github.com/vegaprotocol/specs/blob/master/protocol/0002-STTL-settlement.md) cash flows for the closed portion. Perhaps your formula @witgaw incorporates both but if so it isn't obvious to me (nor obvious if that'd be a good idea), or perhaps I am just missing something basic about this 🙈 

As a starting point:
* we need to make the right settlement cashflows as for any reduction in size
* we need to adjust the margin correctly for the remaining position size
* we know the margin account may not contain `margin_factor * position_notional` any more as MTM settlement can (likely will) have changed it (and also that MTM settlement runs intermittently and not in "real time")

So I _think_ any formula that multiplies the current margin by something to calculate what to remove is probably on the wrong track, because this is not guaranteed to give the same result dependent on the price moves that have happened and when MTM settlement last ran, but would need to write it out and test it to be sure.

I've ended up back at the margin reduction being `change_in_size * average_entry_price`  (where average entry price is unchanged by definition as no position is entered in a reduction). Plus then the settlement cash flow is `change_in_size * (vwap_close_price - last_mark_price)`.

But I'm not even as sure as I was on Monday 🙃 

It'd be good to get a few scenarios tested on this.

witgaw: Here's the small example I mentioned on Monday:
https://docs.google.com/spreadsheets/d/1x32LzfS6WEofnFIU-7I7oAKKz2jojUQvA0sAtqudyIo/edit?usp=sharing

TomMcL: `current margin` here would be the balance in the margin account, which should be purely dependent on the price moves since average_entry_price, then you close a proportion of it based on your proportion closed. 

Checking the spreadsheet

TomMcL: First thought is that there's no PnL in here, I think the first trader would have had their position closed :D.

However I'm not immediately sure that "total margin taken after a series of trades should be equal at equal VWAP" is necessarily true. It feels like removing margin based on VWAP of the trades might make it tricky mentally to think about what the remaining amount means (given that isolated margin is meant to be 'simpler', "I close half my position and half my margin is released" Is easy to reason at least. "It depends how your VWAP is affected by the closing price" less so)

TomMcL: (I'm also probably being dense but I'm not sure calculating average entry price including sales like that works does it?)

protocol/0019-MCAL-margin_calculator.md
TomMcL: Will separate this into another PR into this one given it looks like it's going to touch a few areas

protocol/0019-MCAL-margin_calculator.md
TomMcL: Adding in an above section to clarify 👍

protocol/0019-MCAL-margin_calculator.md
TomMcL: I think it was decided that isolated margin would skip the initial margin requirement too and allow people to just get themselves in danger of being closed out immediately if they wanted. But `> risk_factor` certainly feels like it would make sense.

protocol/0019-MCAL-margin_calculator.md
witgaw: If if it doesn't in general it did for that particular example, you end up with same effective position in terms of volume and prices in both cases.

witgaw: Btw. I'm on board with going for ratio of sizes as originally suggested, removed my suggestion now and closing the conversation.

protocol/0019-MCAL-margin_calculator.md
TomMcL: Resolving as agreed it's probably not worth the time sink and risk to align this now

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,141 @@
...
+
+When submitting, amending, or deleting an order in isolated margin mode and an auction is active there is no concept of an order trading immediately on entry, however the case of someone putting in a sell order for a very low price must be handled (as it is likely to execute at a much higher price). To handle this, when in an auction the amount taken into the order margin account should be the larger of either `limit price * size * margin factor` or `max(mark price, indicative uncrossing price) * size * margin factor`. All other steps are as above.
+
+### Increasing Position
+
+When an order trades which increases the position (increasing the absolute value of the trader's position), the target amount to be transferred is calculated as:
+
+```math
+margin to add = margin factor * sum across executed trades(abs(trade size) * trade price)
+```
```
witgaw:
```math
\text{margin to add} = \text{margin factor} \cdot \text{sum across executed trades}(|\text{trade size}| \cdot \text{trade price})
```

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -84,21 +84,141 @@
...
+
+NB: In implementation, for any volume that trades immediately on entry, the additional margin may be transferred directly from the general account to the margin account.
+
+### Reducing Position
+
+When an order trades which reduces the trader's current position the amount to be withdrawn from the margin account is determined by the fraction of the position which is being closed. However, this fraction should also take into account that the entire position's margin may be due to change since the current trading price may have diverged from the last mark price update. As such the margin released should be calculated as:
+
+```math
+margin to remove = margin required for entire position at VWAP trade price * abs(total size of new trades) / abs(entire position prior to trade)
+```
```
witgaw:
```math
\text{margin to remove} = \text{margin required for entire position at VWAP trade price} \cdot \frac{|\text{total size of new trades}|}{|\text{entire position prior to trade}|}
```

witgaw (dismissed):
LGTM, just some minor latex formatting issues:

witgaw (approved):
LGTM

PR #1851, opened 2023-07-24
Govt mkt closure AC clarification

witgaw:


protocol/0028-GOVE-governance.md
```diff
@@ -533,6 +530,7 @@ APIs should also exist for clients to:
 - Enacting a market closure governance proposal on a market that is in a settled state has no effect. (<a name="0028-GOVE-137" href="#0028-GOVE-137">0028-GOVE-137</a>)
 - Enacting a market closure governance proposal on a market that is not in a settled state always uses the final price supplied with the proposal, even when the oracle settlement price is available at that time. (<a name="0028-GOVE-138" href="#0028-GOVE-138">0028-GOVE-138</a>)
```
vega-paul: This one is confusing, there is no closing proposal, only terminate, resume, suspend. Terminating (close state) proposals will reject unless active market. What does final price man here - there is only one price supplied on the proposal.

vega-paul (changes requested):

protocol/0028-GOVE-governance.md
witgaw: Closure and termination proposal is the same, specs refer to it as "move market to a closed state" in the spec. 
There is only one price supplied on the proposal and there's only one price that will be final from market's point of view, it's meant to be the same price. Rephrased now.

vega-paul (approved):

PR #1854, opened 2023-07-24
tests: Add more successor market ACs.

jgsbennett:


protocol/0081-SUCM-successor_markets.md
```diff
@@ -104,6 +106,14 @@ When fetching a market that is part of a "parent / successor chain", we should s
...
 
+A market which has been settled, but is still inside successor expiry window, is retained in a checkpoint, and can be used by a successor market after restart(<a name="0081-SUCM-029" href="#0081-SUCM-029">0081-SUCM-029</a>)
+
+A market which has been settled, and beyond the successor expiry window, is not retained in a checkpoint, and cannot be used by a successor market after restart(<a name="0081-SUCM-030" href="#0081-SUCM-030">0081-SUCM-030</a>)
+
+A market which has been settled, and already has a child which has succeeded it, is retained in a checkpoint, and can be found after restart. Both child and parent retain child/parent links in the API(<a name="0081-SUCM-031" href="#0081-SUCM-031">0081-SUCM-031</a>)
+
+For a parent/child (explicitly: the child has left opening auction), after a checkpoint restart, parent/child enter opening auction again. It is not possible to propose a new market which attempts to succeed that parent.(<a name="0081-SUCM-032" href="#0081-SUCM-032">0081-SUCM-032</a>)
```
davidsiska-vega: Both parent and child enter opening auction again?

protocol/0081-SUCM-successor_markets.md
```diff
@@ -104,6 +106,14 @@ When fetching a market that is part of a "parent / successor chain", we should s
 
 After a LNL checkpoint restart the successor (child) / parent market state is preserved where applicable inc. the LPs ELS	(<a name="0081-SUCM-016" href="#0081-SUCM-016">0081-SUCM-016</a>)
 
+A market which has been settled, but is still inside successor expiry window, is retained in a checkpoint, and can be used by a successor market after restart(<a name="0081-SUCM-029" href="#0081-SUCM-029">0081-SUCM-029</a>)
+
+A market which has been settled, and beyond the successor expiry window, is not retained in a checkpoint, and cannot be used by a successor market after restart(<a name="0081-SUCM-030" href="#0081-SUCM-030">0081-SUCM-030</a>)
+
+A market which has been settled, and already has a child which has succeeded it, is retained in a checkpoint, and can be found after restart. Both child and parent retain child/parent links in the API(<a name="0081-SUCM-031" href="#0081-SUCM-031">0081-SUCM-031</a>)
```
davidsiska-vega: > A market ... can be found after restart 

Too vague. Do we mean the marketId? Or more?

davidsiska-vega: Aha, I see that case 2. is in 030; so just clarify 031.

davidsiska-vega (changes requested):
See other comments.

protocol/0081-SUCM-successor_markets.md
EVODelavega: Yes, this is technically possible (more so than something that would happen in real life), but given that it's possible for a successor to leave opening auction before the parent is settled, this is a possibility. We could have a checkpoint taken at a point in time where the parent market wasn't settled yet, and the successor has already left opening auction. If that checkpoint is then used to restore state from while the parent is still valid, both markets will be restored in opening auction, but because we've already copied over ELS data from the parent to the successor, it ought to not be possible for a second successor market for that parent

gordsport (dismissed):
Can you please make sure that the:

https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/features.json

Is updated with new ACs being added before its merged

davidsiska-vega (approved):

protocol/0081-SUCM-successor_markets.md
```diff
@@ -104,6 +106,14 @@ When fetching a market that is part of a "parent / successor chain", we should s
 
 After a LNL checkpoint restart the successor (child) / parent market state is preserved where applicable inc. the LPs ELS	(<a name="0081-SUCM-016" href="#0081-SUCM-016">0081-SUCM-016</a>)
 
+A market which has been settled, but is still inside successor expiry window, is retained in a checkpoint, and can be used by a successor market after restart(<a name="0081-SUCM-029" href="#0081-SUCM-029">0081-SUCM-029</a>)
+
+A market which has been settled, and beyond the successor expiry window, is not retained in a checkpoint, and cannot be used by a successor market after restart(<a name="0081-SUCM-030" href="#0081-SUCM-030">0081-SUCM-030</a>)
+
+A market which has been settled, and already has a child which has succeeded it, is retained in a checkpoint. Market can be queried via APIs and settled market state can be retrieved. Both child and parent retain parent/child links in market state, and are listed in "successor chain" API request(<a name="0081-SUCM-031" href="#0081-SUCM-031">0081-SUCM-031</a>)
```
EVODelavega, line 113: I don't think this is right. The market itself isn't going to be retained in the checkpoint. I think it's more accurate to say that the successor will be retained and will reference its parent market, regardless of that market existing or not. If we were to implement this, checkpoints would have to retain the parent market data indefinitely, and therefore checkpoints would just continuously grow over time (ie if the parent were in turn a child market, one could argue that we'd need to keep its parent in a checkpoint and so on).

This, I think, relates mostly to the issue we found recently where a CP restore would erase the reference to the parent market if the parent market wasn't restored from the CP (because the parent had already been settled). We don't need to retain the parent state to address this, we just have to ensure that a successor market that had left opening auction when the checkpoint was taken still references its parent when we restore the checkpoint.

protocol/0081-SUCM-successor_markets.md
```diff
@@ -104,6 +106,14 @@ When fetching a market that is part of a "parent / successor chain", we should s
...
+
+For a parent and child (explicitly: the child has left opening auction), after a checkpoint restart, parent and child both enter opening auction again. It is not possible to propose a new market which attempts to succeed that parent.(<a name="0081-SUCM-032" href="#0081-SUCM-032">0081-SUCM-032</a>)
```
EVODelavega, line 115: Being even more explicit: this would be a case where a successor has left opening auction _before_ the parent was settled, and the checkpoint is restored while the parent can still be created (e.g. A parent that settled on 01/08/2023 can be included in the checkpoint, but if the checkpoint is restored on 02/08/2023, that market is no longer valid, and won't enter opening auction)

PR #1870, opened 2023-07-28
feat: adding perp spec copies

### LINKED ISSUE #1842: Review the ACs we have for futures and see which apply to the perps product
The cosmic elevator branch has a [feature.json](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/features.json) file that maps ACs to a feature.

We need to review:

- [ ] which ACs should be added for perps (use the futures for a starter/guidance)
- [ ] which of these are feature tests and which are system tests

TomMcL:
close #1842 .


protocol/0002-STTL-settlement.md
```diff
@@ -107,7 +107,7 @@ The [market lifecycle spec](./0043-MKTL-market_lifecycle.md) provides detail on
 1. If margin and general account of trader are insufficient to cover collateral transfers, then collateral is attempted to be taken from market's insurance pool. (<a name="0002-STTL-008" href="#0002-STTL-008">0002-STTL-008</a>)
 1. If the full required amount for collateral cannot be collected from individual or combination of these accounts, then as much as possible in the above sequence of accounts is collected and loss socialisation occurs. (<a name="0002-STTL-009" href="#0002-STTL-009">0002-STTL-009</a>)
 
-### Example 3 - Settlement data to cash settled future is submitted before trading is terminated (<a name="0002-STTL-010" href="#0002-STTL-010">0002-STTL-010</a>)
+### Example 3 - Settlement data to cash settled future is submitted before trading is terminated (<a name="0002-STTL-010" href="#0002-STTL-010">0002-STTL-010</a>)(Perpetual Future: <a name="0002-STTL-015" href="#0002-STTL-015">0002-STTL-015</a>)
```
wwestgarth: I don't think this is relevant for perps. As I understand the perps spec there is no oracle for termination or settlement like fixed expiry futures, and the idea is that when a perps market wants to be closed it'll happen via a governance state change proposal. So there is no way really to send in settlement data before its terminated.

(Probably the same for example 2 a well)

protocol/0002-STTL-settlement.md
```diff
@@ -64,11 +64,11 @@ The [market lifecycle spec](./0043-MKTL-market_lifecycle.md) provides detail on
 
 ## Acceptance Criteria
 
-### The typical "Happy Path" case (<a name="0002-STTL-001" href="#0002-STTL-001">0002-STTL-001</a>)
+### The typical "Happy Path" case (Expiring Future: <a name="0002-STTL-001" href="#0002-STTL-001">0002-STTL-001</a>, Perpetual Future: <a name="0002-STTL-011" href="#0002-STTL-011">0002-STTL-011</a>)
```
wwestgarth: I think the typical happy-path for a perps is too different that the desciption here. Really it needs to be:
- can get into continous trading
- with external oracle data submitting, cash flows happen when the time-triggered settlement cue fires
- perps market can be closed by market-state-change governance proposal.

Edit: basically what you have for `0043-MKTL-009`!

protocol/0043-MKTL-market_lifecycle.md
```diff
@@ -301,6 +301,40 @@ Market state is `suspended`.
...
+The market enters [liquidity monitoring auction](0035-LIQM-liquidity_monitoring.md).
+The market state is `suspended`.
+1. A party place bid; this becomes a best static bid.
+After the specified time the liquidity auction ends.
+The market state is `active`.
+1. Make sure that trades happen so that at least two parties have open positions.
+
+1. An oracle event arrives which triggers the perpetual market's interim settlement logic, causing cashflow transfers but the market remains open.
+1. Further trades happen, with parties still having different positions to previously. The mark price is `p`.
+1. The time specified at market proposal by the internal time oracle is reached.
```
wwestgarth: All of the below will be with a goverance state-change proposal and not oracles.

protocol/0002-STTL-settlement.md
TomMcL: Aha, was thinking there was still the ability to have a termination/settlement oracle, will update these accordingly!

TomMcL:
@wwestgarth Updated (Removed the STTL ones entirely, tests for the market change proposal to close the market should cover all those cases once we have the MKTL one for perps themselves)

davidsiska-vega (approved):

PR #1876, opened 2023-08-03
feat: rewards overhaul

cdummett:
PR updates the rewards specification with updates defined in the requirements [document](https://docs.google.com/document/d/1bmTqkp_X1QvCVmD6G6db9ZfiLXAobI1bIh2wWVSR5YM/edit).

### DIFF
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -248,17 +248,12 @@ Let:
 
 - $d_{i}$ be the payout factor for entity $i$
 - $s_{i}$ be the share of the rewards for entity $i$
-- $k$ be the decay factor specified in the recurring transfer funding the reward. This must be a float in the range $0\leq k \leq 2$. 
-- $R_{i}$ be the normalised rank of entity $i$
+- $r_{i}$ be the share ratio of entity $i$ as determined from the rank table
 - $M_{i}$ be the sum of all reward payout multipliers for entity $i$ (reward payout multipliers include the [activity streak multiplier](./0086-ASPR-activity_streak_program.md#applying-the-activity-reward-multiplier) and [bonus rewards multiplier](./0085-RVST-rewards_vesting.md#determining-the-rewards-bonus-multiplier)).
 
 $$d_{i}=M_{i} * r_{i}$$
 
 Note if the entity is a team, $M_{i}$ is set to 1 as reward payout multipliers are considered later when distributing rewards [amongst the team members](#distributing-rewards-amongst-team-members).
-
-To avoid exponential formulas in the core implementation, the above exponential equation can be approximated using the following 7th order Taylor expansion. An odd number of terms has intentionally been chosen so the expansion diverges to $-\infty$ rather than $+\infty$.
-
-$$d_{i} = M_{i} e^{-k R_i} \approx M_{i} \min(\sum_{j=0}^{7} \frac{(-k R_i)^j}{j!}, 0)$$
 
 Calculate each entities share of the rewards, $s_{i}$ pro-rata based on $d_{i}$, i.e.
 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -120,18 +120,17 @@ $$R = \{r_i \cdot \log(1 +\bar{OI_{i}}) \mid i = 1, 2, \ldots, N\}$$
 
 The reward metric $m_{rv}$ is the standard deviation of the set $R$.
 
-### Validator power metric
+### Validator ranking metric
 
-The validator power metric, $m_{vp}$, measures the voting power of validators.
+The validator ranking metric, $m_{vr}$, measures the ranking score of consensus and standby validators.
 
-At the end of each epoch, for each party who is a validator set their reward metric as follows.
+At the end of each epoch, for each party who **is** a consensus or standby validator set their reward metric as follows.
 
-$$m_{vp} = validatorScore$$
+$$m_{vr} = rankingScore$$
 
-Note the `validatorScore` is the score **after** any [penalties](./0061-REWP-pos_rewards.md#for-each-validator-we-then-do) have been applied.
-If a party is not a validator, their reward metric is simply:
+If a party **is not** a consensus or standby validator, their reward metric is simply:
 
-$$m_{vp} = 0$$
+$$m_{vr} = 0$$
 
 ### Market creation reward metrics
 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -52,13 +52,70 @@ If the reward account balance is 0 at the end of the epoch for a given market, a
 
 Fee-based reward metrics (the total fees paid/received by each party as defined above) are stored in [LNL checkpoints](./0073-LIMN-limited_network_life.md) and are restored after a checkpoint restart to ensure rewards are not lost.
 
-### Open interest metrics
+### Open interest metric
+
+The open interest metric, ($m_{oi}$, measures each parties time-weighted average open interest over a number of epochs.
+
+At the start of each epoch, the network must reset each parties time weighted average open interest for the epoch ($\bar{OI}$) to `0`. Whenever a parties position changes during an epoch, **and** at the end of the epoch, this value should be updated as follows.
+
+Let:
+
+- $\bar{OI}$ be the parties time weighted average open interest in the epoch so far
+- $OI_{n}$ be the parties open interest before their open interest changed
+- $t_{n}$ be the time the party held the previous position in seconds
+- $t$ be the amount of time elapsed in the current epoch so far
 
 
-### Pnl metrics
+$$\bar{OI} = \bar{OI} \cdot \left(1 - \frac{t_{n}}{t}\right) + \frac{OI_{n} \cdot t_{n}}{t}$$
 
+At the end of the epoch, the network must store the parties time weighted average open interest and then calculate their open interest reward metric as follows.
 
-### Volatility metrics
+Let:
+
+- $m_{oi}$ be the parties open interest reward metric
+- $\bar{OI_{i}}$ be the parties time weighted average open interest in the ith epoch
+- $N$ be the network parameter `rewards.metrics.openInterestWindow`
+
+$$m_{oi} = \frac{\sum_{i}^{n}\bar{OI_{i}}}{N}$$
+
+### Relative return metric
+
+The relative return metric, $m_{rr}$, measures each parties average relative return, weighted by their [time-weighted average open-interest](#open-interest-metric), over a number of epochs.
+
+At the start of the epoch, the network must store the parties current pnl.
+
+$$p_{0} = p_{realised} + p_{unrealised}$$
+
+At the end of each epoch, the network must calculate and store the parties change in pnl:
+
+Let:
+- $\Delta{p}$ be the parties change in pnl
+- $p_{realised}$ be the parties realised pnl at the end of the epoch
+- $p_{unrealised}$ be the parties unrealised pnl at the end of the epoch
+
+$$\Delta{p} = p_{realised} + p_{unrealised} - p_{0}$$
+
+And calculate their average relative return, weighted by the log of their [time weighted average open interest](#open-interest-metrics), over the last $N$ epochs as follows.
+
+Let:
+- $m_{rr}$ be the parties relative return reward metric
+- $\bar{OI_{i}}$ be the parties time weighted average open interest in the ith epoch
+- $\Delta{p}$ be the parties change in pnl in the ith epoch
+- $N$ be the network parameter `rewards.metrics.relativeReturnsWindow`
+
+$$m_{rr} = \frac{\sum_{i}^{n}{p_{i}\cdot\log(1 + \bar{OI_{i}})}}{N}$$
+
+### Returns volatility metric
+
+The return volatility metric, $m_{rv}$, measures the volatility of a parties returns across a number of epochs.
+
+At the end of an epoch, if a party has had net returns less than or equal to `0` over the last $N$ epochs (where $N$ is the network parameter `rewards.metrics.returnsVolatilityWindow`), their reward metric $m_{rv}$ is set to `0`. Otherwise, the network should calculate the standard deviation of the set of each parties returns weighted by the log of their [time weighted average open interest](#open-interest-metrics) over the last $N$ epochs.
+
+Given the set:
+
+$$R = \{r_i \cdot \log(1 +\bar{OI_{i}}) \mid i = 1, 2, \ldots, N\}$$
+
+The reward metric $m_{rv}$ is the standard deviation of the set $R$.
 
 ### Market creation reward metrics
 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -185,7 +185,7 @@ Reward accounts and balances must be saved in [LNL checkpoints](./0073-LIMN-limi
 
 ## Reward distribution
 
-All rewards are distributed to [vesting accounts](./0085-RVST-rewards_vesting.md) at the end of each epoch *after* [recurring transfers](0057-TRAN-transfers.md) have been executed.
+All rewards are distributed to [vesting accounts](./0085-RVST-rewards_vesting.md) at the end of each epoch *after* [recurring transfers](0057-TRAN-transfers.md) have been executed. Funds distributed to the vesting account will not start vesting until the [`lock period`](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts) defined in the recurring transfer has expired.
 
 The entire reward account balance is paid out every epoch unless the total value of the metric over all entities is zero, in which case the balance will also be zero anyway (there are no fractional payouts).
 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -41,6 +41,8 @@ This is the original precision for the metric source data.
 
 For reward metrics relating to trading, an individual must meet the [staking requirement](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts) **AND** [notional time-weighted average position requirement](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts)) set in the recurring transfer. If they do not then their reward metric is set to `0`. Note, these requirements do not apply to the [validator ranking metric](#returns-volatility-metric) or the [market creation reward metric](#market-creation-reward-metrics).
 
+For reward transfers where the [scope](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts) is set to teams, each party must meet the minimum time in team requirement. That is, given a party has been in a team for $N$ epochs, if $N$ is strictly less than the network parameter `rewards.minimumEpochsInTeam` (an integer defaulting to `0`) their reward metric is set to `0`.
+
 ### Fee-based reward metrics
 
 There will be three reward metrics calculated based on fees.
@@ -51,7 +53,7 @@ There will be three reward metrics calculated based on fees.
 
 These metrics apply only to the sum of fees for the epoch in question.
 That is, the metrics are reset to zero for all parties at the end of the epoch.
-If the reward account balance is 0 at the end of the epoch for a given market, any parties with non-zero metrics will not be rewarded for that epoch and their metric scores do not roll over (they are still zeroed).
+If the reward account balance is `0` at the end of the epoch for a given recurring transfer, any parties with non-zero metrics will not be rewarded for that epoch and their metric scores do not roll over (they are still zeroed).
 
 Fee-based reward metrics (the total fees paid/received by each party as defined above) are stored in [LNL checkpoints](./0073-LIMN-limited_network_life.md) and are restored after a checkpoint restart to ensure rewards are not lost.
 
@@ -170,15 +172,14 @@ A team’s reward metric is the weighted average metric score of the top perform
 
 ## Reward accounts
 
-Trading reward accounts are defined by the reward asset (the asset in which the reward is paid out), the market, and the reward type (metric).
-That is, there can be multiple rewards with the same type paid in different assets for the same market.
+Trading reward accounts are defined by a hash of the fields specified in the recurring transfer funding the reward account (see the [transfers](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts) spec for relevant details about each field). This allows multiple recurring transfers to fund the same reward pool.
 
-Note that the market settlement asset has nothing to do in particular with the asset used to pay out a reward for a market.
+Note as part of the recurring transfer a user specifies a settlement asset. The market settlement asset has nothing to do in in particular with the asset used to pay out a reward.
+
 That is, a participant might receive rewards in the settlement asset of the market, in VEGA governance tokens, and in any number of other unrelated tokens (perhaps governance of "loyalty"/reward tokens issued by LPs or market creators, or stablecoins like DAI).
 
-Reward accounts are funded by setting up recurring transfers, which may be set to occur only once for a one off reward.
-These allow a reward type to be automatically funded on an ongoing basis from a pool of assets.
-Recurring transfers can target groups of markets, or all markets for a settlement asset, in which case the amount paid to each market is determined pro-rata by the markets' relative total reward metrics for the given reward type. See [transfers](./0057-TRAN-transfers.md) for more detail.
+Reward accounts are funded by setting up recurring transfers, which may be set to occur only once for a one off reward. These allow a reward type to be automatically funded on an ongoing basis from a pool of assets.
+Recurring transfers can target groups of markets, or all markets for a settlement asset. See [transfers](./0057-TRAN-transfers.md) for more detail.
 
 Reward accounts and balances must be saved in [LNL checkpoints](./0073-LIMN-limited_network_life.md) to ensure all funds remain accounted for across a restart.
 
--- a/protocol/0056-REWA-rewards_overview.md
+++ b/protocol/0056-REWA-rewards_overview.md
@@ -1,6 +1,6 @@
 # Reward framework
 
-The reward framework provides a mechanism for measuring and rewarding a number of key activities on the Vega network.
+The reward framework provides a mechanism for measuring and rewarding individuals or [teams](./0083-RFPR-on_chain_referral_program.md#glossary) (collectively referred to within this spec as entities) for a number of key activities on the Vega network.
 These rewards operate in addition to the main protocol economic incentives which come from
 [fees](0029-FEES-fees.md) on every trade.
 These fees are the fundamental income stream for [liquidity providers LPs](0042-LIQF-setting_fees_and_rewarding_lps.md) and [validators](./0061-REWP-pos_rewards.md).
@@ -9,7 +9,6 @@ The additional rewards described here can be funded arbitrarily by users of the 
 Note that transfers via governance, including to fund rewards, is a post-Oregon Trail feature.
 
 Note that validator rewards (and the reward account for those) is covered in [validator rewards](./0061-REWP-pos_rewards.md) and is separate from the trading reward framework described here.
-
 ## New network parameter for market creation threshold
 
 The parameter `rewards.marketCreationQuantumMultiple` will be used together with [quantum](0040-ASSF-asset_framework.md) to asses market size when deciding whether a market qualifies for the payment of market creation rewards.
@@ -20,17 +19,18 @@ Therefore, for example, to reward futures markets when they reach a lifetime tra
 
 At a high level, rewards work as follows:
 
-- Reward metrics are calculated for each combination of [reward type, party, market].
-  - The calculation used for the reward metric is specific to each reward type.
+- Individual reward metrics are calculated for each combination of [reward type, party, market] and team reward metrics for each combination of [reward type, team, market].
 
 At the end of the epoch:
 
 1. Recurring reward transfers (set up by the parties funding the rewards) are made to the reward account(s) for a specific reward type, for one or more markets in scope where the total reward metric is `>0`. See [transfers](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts).
-1. Then the entire balance of each reward account is distributed to the parties with a non-zero reward metric for that reward type and market, pro-rata by their reward metric.
+1. Then the entire balance of each reward account is distributed amongst entities with a non-zero reward metric for that reward type and market using the mechanism specified in the recurring transfer.
+1. Distributed rewards are transferred to a [vesting account](./0085-RVST-rewards_vesting.md).
 
-## Reward metrics
+## Individual reward metrics
 
-Fee-based reward metrics are scoped by [`reward type`, `market`, `party`] (this triplet can be thought of as a primary key for fee-based reward metrics).
+Individual reward metrics are scoped by [`reward type`, `market`, `party`] (this triplet can be thought of as a primary key for fee-based reward metrics).
+
 Therefore a party may be in scope for the same reward type multiple times but no more than once per market per epoch.
 Metrics will be calculated at the end of every epoch, for every eligible party, in each market for each reward type.
 Metrics only need to be calculated where the [market, reward type] reward account has a non-zero balance of at least one asset.
@@ -38,10 +38,9 @@ Metrics only need to be calculated where the [market, reward type] reward accoun
 Reward metrics will be calculated once for each party/market combination in the reward metric asset which is the [settlement asset](0070-MKTD-market-decimal-places.md) of the market.
 This is the original precision for the metric source data.
 
-### Market activity (fee based) reward metrics
+### Fee-based reward metrics
 
-There will be three market activity reward metrics calculated based on fees (as a proxy for activity).
-Each of these represents a reward type with its own segregated reward accounts for each market.
+There will be three reward metrics calculated based on fees.
 
 1. Sum of maker fees paid by the party on the market this epoch
 1. Sum of maker fees received by the party on the market this epoch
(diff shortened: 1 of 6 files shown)


protocol/0085-RVST-rewards_vesting.md
```diff
@@ -0,0 +1,50 @@
+# Rewards Vesting Specification
+
+## Summary
+
+The aim of the rewards vesting mechanics is to prevent farming rewards by delaying the payout of rewards through vesting. To encourage longer term behaviour parties can accelerate their rewards vesting rate through the [activity streak program](./0086-ASPR-activity_streak_program.md).
+
+## Network Parameters
+
+- `rewards.vesting.baseRate`: the proportion of rewards in a vesting account which are vested each epoch
+- `rewards.vesting.minimumTransfer`: the minimum amount (expressed in quantum) which can be vested each epoch
```
jeremyletang: Could we specify the default values for these in the specs?

protocol/0085-RVST-rewards_vesting.md
```diff
@@ -0,0 +1,50 @@
...
+## Vesting mechanics
+
+As detailed in [distributing rewards](), each party has their rewards paid into vesting rewards accounts (one for each asset).
+
+At the end of each epoch, a proportion of the rewards accumulated in each "vesting" account should be released and transferred to the respective "vested" account. The percentage released can be scaled by the account owner increasing their [activity streak](./0086-ASPR-activity_streak_program.md) and a minimum transfer amount will be applied to ensure the account is eventually emptied. The proportion released and minimum applied are controlled for parameters for the asset.
+
+Now, let:
+
+- $T$ be the amount to be "vested" (transferred from the vesting account to the vested account)
+- $B_{vested}$ be the total quantum amount in the vesting account
```
jeremyletang, line 22: Should this be that?
```suggestion
- $B_{vesting}$ be the total quantum amount in the vesting account
```

protocol/0085-RVST-rewards_vesting.md
```diff
@@ -0,0 +1,50 @@
...
+- $r_{bvr}$ be the network parameter `rewards.vesting.baseRate`
+- $a$ be the account owners current [`activity_streak_vesting_multiplier`](./0086-ASPR-activity_streak_program.md#setting-activity-benefits)
+- $m_{mva}$ be the network parameter `rewards.vesting.minimumTransfer`
+
+The quantum amount to be transferred from each "vesting" account to the relevant "vested" account is defined as:
+$$
+T = min(B_{vested} * r * a, m)
+$$
```
jeremyletang: This bit doesn't render well.

protocol/0085-RVST-rewards_vesting.md
```diff
@@ -0,0 +1,50 @@
...
+When transferring funds from the vesting account to the vested account, a new transfer type should be used, `TRANSFER_TYPE_REWARDS_VESTED`.
+
+## APIs
+
+### Accounts API
+
+Must expose the following:
+
+- every account with `ACCOUNT_TYPE_VESTING_REWARDS` for each party
+- every account with `ACCOUNT_TYPE_VESTED_REWARDS` for each party
```
jeremyletang, line 71: The initial document mentioned:
```
Rewards are paid into a (new) locked (unvested) rewards account N epochs after they are earned (this is set on reward distribution, and can be different for each reward. 
There’s a network param for the maximum value of N.
```

That seems to translate better into 3 accounts.
- A first locked one where the funds are transfered from the reward pool onto, and locked for the given party
- A second one where the funds are vesting.
- An a last one were they are transfered once vested.

jeremyletang: hum actually 2 accounts would work, we can just keep track internally what amount is available for being distributed, and what amount is still locked before starting vesting.

protocol/0085-RVST-rewards_vesting.md
```diff
@@ -0,0 +1,50 @@
...
+
+- $T$ be the amount to be "vested" (transferred from the vesting account to the vested account)
+- $B_{vested}$ be the total quantum amount in the vesting account
+- $r_{bvr}$ be the network parameter `rewards.vesting.baseRate`
+- $a$ be the account owners current [`activity_streak_vesting_multiplier`](./0086-ASPR-activity_streak_program.md#setting-activity-benefits)
+- $m_{mva}$ be the network parameter `rewards.vesting.minimumTransfer`
+
+The quantum amount to be transferred from each "vesting" account to the relevant "vested" account is defined as:
+$$
+T = min(B_{vested} * r * a, m)
```
jeremyletang: Shouldn't this be: T = max(B_{vested} * r * a, m)

protocol/0085-RVST-rewards_vesting.md
cdummett: Sure @barnabee to confirm these are sensible but for now have set 
- `rewards.vesting.baseRate=0.1` (means ~50% of rewards are unlocked after 7 epochs and ~95% after 31 epochs)
- `rewards.vesting.minimumTransfer=100` (same as the minimum transfer network parameter).

protocol/0085-RVST-rewards_vesting.md
cdummett: Think I agree 3 accounts would actually be clearer to explain in docs and expose through APIs. If it is not a big deal implementation wise could we go with that?

If you want to stick with two, I would definitely rename the `ACCOUNT_TYPE_VESTING_REWARDS` to `ACCOUNT_TYPE_UNVESTED_REWARDS` to make it clear some funds in it are locked before they start to vest.

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -0,0 +1,403 @@
...
+
+```protobuf
+message CreateReferralSet{
+    is_team: True
+    team_details: {
+        name: "VegaRocks",
+        team_url: "https://discord.com/channels/vegarocks",
+        avatar_url: "https://vega-rocks/logo-360x360.jpg",
+        closed: False,
+}
```
jeremyletang:
```suggestion
message CreateReferralSet{
    is_team: True
    team_details: {
        name: "VegaRocks",
        team_url: "https://discord.com/channels/vegarocks",
        avatar_url: "https://vega-rocks/logo-360x360.jpg",
        closed: False,
     }
}
```

protocol/0085-RVST-rewards_vesting.md
jeremyletang:
```
If you want to stick with two, I would definitely rename the ACCOUNT_TYPE_VESTING_REWARDS to ACCOUNT_TYPE_UNVESTED_REWARDS to make it clear some funds in it are locked before they start to vest.
```
Yea can do that.

protocol/0085-RVST-rewards_vesting.md
barnabee: The minimum transfer means that someone with 1000 of rewards will be fully unlocked after 10 epochs regardless of the baserate. That seems too high.

Additionally I think the base rate ought to suggest unlocking over months rather than days for anyone with non-trivial rewards.

I think the minimum transfer amount should be something like 5 or 10, and the base rate probably in the range 1–2.5%.

**However**, I think this one we need to discuss with @MM0819 and Steven so lets hold off on confirming anything until that is done.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -51,7 +51,87 @@ These metrics apply only to the sum of fees for the epoch in question.
...
+
+The reward metric $m_{rv}$ is the standard deviation of the set $R$.
+
+### Validator power metric
+
+The validator power metric, $m_{vp}$, measures the voting power of validators.
+
+At the end of each epoch, for each party who is a validator set their reward metric as follows.
+
+$$m_{vp} = validatorScore$$
```
ze97286: what validator score? why not just the voting power directly?

ze97286: I generally don't understand how this is supposed to work - if I specify this metrics with no parties in scope, does anyone get it? the way it's written above, does it go only to consensus validators and not ersatz validators? pending validators? those would have non negative score (by some definition of score) but would have no voting power so not sure what the intention is. If I don't qualify parties, who gets the reward? no one? all validators ranked by their score?

protocol/0086-ASPR-activity_streak_program.md
```diff
@@ -0,0 +1,126 @@
...
+- a greater share of rewards schemes
+- an accelerated vesting rate on locked rewards
+
+Parties access higher tiers and greater benefits by maintaining an activity streak. The length of a streak is measured in epochs and a party is considered active if they made a trade or held an open position during the epoch. If a party is inactive for more than a specified number of epochs they lose their streak.
+
+## Network parameters
+
+- `rewards.activityStreak.benefitTiers`: is an ordered list of dictionaries defining the requirements and benefits for each tier.
+- `rewards.activityStreak.inactivityLimit`: the maximum number of epochs a trader can be inactive before loosing their streak.
+- `rewards.activityStreak.minQuantumOpenVolume`: the minimum open notional volume (expressed in quantum) for a trader to be considered active in an epoch
```
jeremyletang:
```suggestion
- `rewards.activityStreak.minQuantumOpenNotionalVolume`: the minimum open notional volume (expressed in quantum) for a trader to be considered active in an epoch
```

protocol/0086-ASPR-activity_streak_program.md
```diff
@@ -0,0 +1,126 @@
...
+
+- a greater share of rewards schemes
+- an accelerated vesting rate on locked rewards
+
+Parties access higher tiers and greater benefits by maintaining an activity streak. The length of a streak is measured in epochs and a party is considered active if they made a trade or held an open position during the epoch. If a party is inactive for more than a specified number of epochs they lose their streak.
+
+## Network parameters
+
+- `rewards.activityStreak.benefitTiers`: is an ordered list of dictionaries defining the requirements and benefits for each tier.
+- `rewards.activityStreak.inactivityLimit`: the maximum number of epochs a trader can be inactive before loosing their streak.
```
jeremyletang, line 13: this doesn't seems to be used anywhere in the spec?

cdummett: Yeah good catch, it used in the section about setting the streak length but under a different name.. updated it so it uses `inactivityLimit`.

protocol/0085-RVST-rewards_vesting.md
```diff
@@ -29,6 +30,35 @@ $$T = max(B_{vesting} * r * a, m)$$
 
 When transferring funds from the vesting account to the vested account, a new transfer type should be used, `TRANSFER_TYPE_REWARDS_VESTED`.
 
+## Vested bonus
+
+Once vested rewards are transferred to the vested account, the party will be able to transfer funds to their general account using a normal transfer.
+
+Alternatively, they can leave their rewards in the vested account to receive a multiplier on their reward payout share. The size of this multiplier is dependent on the balance (expressed in quantum) in the vested account before rewards are distributed.
```
barnabee: This should be based on the total in the vested account, the unvested account, and the earned rewards not yet transferred into the unvested account.

barnabee (changes requested):

protocol/0056-REWA-rewards_overview.md
```diff
@@ -51,7 +53,82 @@ These metrics apply only to the sum of fees for the epoch in question.
...
+
+Let:
+
+- $\bar{P}$ be the parties time weighted average position in the epoch so far
+- $P_{n}$ be the parties position before their position changed
+- $t_{n}$ be the time the party held the previous position in seconds
+- $t$ be the amount of time elapsed in the current epoch so far
+
+
+$$\bar{P} = \bar{P} \cdot \left(1 - \frac{t_{n}}{t}\right) + \frac{|P_{n}| \cdot t_{n}}{t}$$
```
```suggestion
$$\bar{P}  \leftarrow  \bar{P} \cdot \left(1 - \frac{t_{n}}{t}\right) + \frac{|P_{n}| \cdot t_{n}}{t}$$
```

davidsiska-vega: I suggest using $\leftarrow$ to indicate assignment - since \bar P is on both sides...

protocol/0056-REWA-rewards_overview.md
```diff
@@ -51,7 +53,82 @@ These metrics apply only to the sum of fees for the epoch in question.
...
+
+At the end of the epoch, the network must store the parties time weighted average position and then calculate their average position reward metric as follows.
+
+Let:
+
+- $m_{ap}$ be the parties average position reward metric
+- $\bar{P_{i}}$ be the parties time weighted average position in the i th epoch
```
```suggestion
- $\bar{P_{i}}$ be the parties time weighted average position in the $i$-th epoch
```

protocol/0056-REWA-rewards_overview.md
```diff
@@ -51,7 +53,82 @@ These metrics apply only to the sum of fees for the epoch in question.
...
+The relative return metric, $m_{rr}$, measures each parties average relative return, weighted by their [time-weighted average position](#average-position-metric), over a number of epochs.
+
+At the end of each epoch, the network must calculate and store the parties relative returns as follows.
+
+Let:
+
+- $r_i$ be the parties relative returns in the epoch
+- $m2m_{wins}$ be the sum of all mark-to-market win transfers in the epoch
+- $m2m_{losses}$ be the sum of all mark-to-market loss transfers in the epoch
+- $TWAP$ be the parties time-weighted average position in the epoch.
```
davidsiska-vega: Any reason not to use $\bar P$ introduced above (instead of new notation)?

protocol/0056-REWA-rewards_overview.md
davidsiska-vega: $\ot$

protocol/0056-REWA-rewards_overview.md
```diff
@@ -51,7 +53,82 @@ These metrics apply only to the sum of fees for the epoch in question.
...
+
+$$r = \frac{m2m_{wins} + m2m_{losses}}{TWAP}$$
```
```suggestion
$$r = \frac{m2m_{wins} - m2m_{losses}}{TWAP}$$
```

davidsiska-vega: In fact we should probable clarify whether we expect $\text{m2m}_{\text{losses}}$ to come with a negative sign above...

protocol/0056-REWA-rewards_overview.md
```diff
@@ -51,7 +53,82 @@ These metrics apply only to the sum of fees for the epoch in question.
...
+- $r_i$ be the parties change in pnl in the i th epoch
+- $N$ be the window length specified in the recurring transfer.
+
+$$m_{rr} = \max(\frac{\sum_{i}^{n}{r_{i}}}{N}, 0)$$
+
+### Returns volatility metric
+
+The return volatility metric, $m_{rv}$, measures the volatility of a parties returns across a number of epochs.
+
+At the end of an epoch, if a party has had net returns less than or equal to `0` over the last $N$ epochs (where $N$ is the window length specified in the recurring transfer), their reward metric $m_{rv}$ is set to `0`. Otherwise, the network should calculate the standard deviation of the set of each parties returns over the last $N$ epochs.
```
davidsiska-vega, line 113: We should connect negative net returns to the above calc of $r_i$ to make sure there is no ambiguity.

protocol/0056-REWA-rewards_overview.md
cdummett:
```suggestion
- $\bar{P}$ be the parties time-weighted average position in the epoch.
```

protocol/0056-REWA-rewards_overview.md
cdummett: Good point, I think since it will be the sum of the "loss" transfers it should be a positive value. For extra clarity, though I will add abs signs to both.
```suggestion
$$r = \frac{|m2m_{wins}| - |m2m_{losses}|}{\bar{P}}$$
```

protocol/0056-REWA-rewards_overview.md
```diff
@@ -98,25 +184,88 @@ Reward accounts and balances must be saved in [LNL checkpoints](./0073-LIMN-limi
...
+1. Calculate each entities reward metric
+2. Order each entity in a descending list by their reward metric value and determine their "rank" in the list
+3. Normalise the rank of each party within a range of 0 to 1, using the highest rank as the upper limit for normalisation.
+4. Calculate each entities share of the rewards using the below formula.
+
+Let:
+
+- $d_{i}$ be the payout factor for entity $i$
+- $s_{i}$ be the share of the rewards for entity $i$
+- $k$ be the decay factor specified in the recurring transfer funding the reward. This must be a float in the range $0\leq k \leq 2$.
```
```suggestion
- $k$ be the decay factor specified in the recurring transfer funding the reward. This must be a decimal in the range $0\leq k \leq 2$. 
```

protocol/0056-REWA-rewards_overview.md
```diff
@@ -98,25 +184,88 @@ Reward accounts and balances must be saved in [LNL checkpoints](./0073-LIMN-limi
...
+2. Order each entity in a descending list by their reward metric value and determine their "rank" in the list
+3. Normalise the rank of each party within a range of 0 to 1, using the highest rank as the upper limit for normalisation.
+4. Calculate each entities share of the rewards using the below formula.
+
+Let:
+
+- $d_{i}$ be the payout factor for entity $i$
+- $s_{i}$ be the share of the rewards for entity $i$
+- $k$ be the decay factor specified in the recurring transfer funding the reward. This must be a float in the range $0\leq k \leq 2$. 
+- $R_{i}$ be the normalised rank of entity $i$
```
davidsiska-vega: Did we say how it's normalised somehwere?

protocol/0056-REWA-rewards_overview.md
```diff
@@ -98,25 +184,88 @@ Reward accounts and balances must be saved in [LNL checkpoints](./0073-LIMN-limi
...
+- $M_{i}$ be the sum of all reward payout multipliers for entity $i$ (reward payout multipliers include the [activity streak multiplier](./0086-ASPR-activity_streak_program.md#applying-the-activity-reward-multiplier) and [bonus rewards multiplier](./0085-RVST-rewards_vesting.md#determining-the-rewards-bonus-multiplier)).
+
+$$d_{i}=M_{i} e^{-k R_{i}}$$
+
+Note if the entity is a team, $M_{i}$ is set to 1 as reward payout multipliers are considered later when distributing rewards [amongst the team members](#distributing-rewards-amongst-team-members).
+
+To avoid exponential formulas in the core implementation, the above exponential equation can be approximated using the following 7th order Taylor expansion. An odd number of terms has intentionally been chosen so the expansion diverges to $-\infty$ rather than $+\infty$.
```
```suggestion
To avoid exponential formulas in the core implementation, the above exponential function can be approximated using the following 7th order Taylor expansion. An odd number of terms has intentionally been chosen so the expansion diverges to $-\infty$ rather than $+\infty$.
```

davidsiska-vega: We should probably floor at $0$ for sanity (odd degree poly).

protocol/0056-REWA-rewards_overview.md
cdummett: Step 3 on line 227
> 3. Normalise the rank of each party within a range of 0 to 1, using the highest rank as the upper limit for normalisation.

davidsiska-vega (changes requested):
Looks basically fine, but I've left a couple of tiny comments. 

It needs review from Lewis and Steven as they're the ones who have the best picture of how this functionality will be used.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -218,28 +218,45 @@ Calculate each entities share of the rewards, $s_{i}$ pro-rata based on $d_{i}$,
...
 
-$$d_{i}=M_{i} e^{-k R_{i}}$$
+$$d_{i}=M_{i} * r_{i}$$
 
 Note if the entity is a team, $M_{i}$ is set to 1 as reward payout multipliers are considered later when distributing rewards [amongst the team members](#distributing-rewards-amongst-team-members).
 
-To avoid exponential formulas in the core implementation, the above exponential function can be approximated using the following 7th order Taylor expansion. An odd number of terms has intentionally been chosen so the expansion diverges to $-\infty$ rather than $+\infty$.
+To avoid exponential formulas in the core implementation, the above exponential equation can be approximated using the following 7th order Taylor expansion. An odd number of terms has intentionally been chosen so the expansion diverges to $-\infty$ rather than $+\infty$.
 
 $$d_{i} = M_{i} e^{-k R_i} \approx M_{i} \min(\sum_{j=0}^{7} \frac{(-k R_i)^j}{j!}, 0)$$
```
barnabee: Should this be gone too?

protocol/0056-REWA-rewards_overview.md
```diff
@@ -218,28 +218,45 @@ Calculate each entities share of the rewards, $s_{i}$ pro-rata based on $d_{i}$,
...
 - $R_{i}$ be the normalised rank of entity $i$
 - $M_{i}$ be the sum of all reward payout multipliers for entity $i$ (reward payout multipliers include the [activity streak multiplier](./0086-ASPR-activity_streak_program.md#applying-the-activity-reward-multiplier) and [bonus rewards multiplier](./0085-RVST-rewards_vesting.md#determining-the-rewards-bonus-multiplier)).
 
-$$d_{i}=M_{i} e^{-k R_{i}}$$
+$$d_{i}=M_{i} * r_{i}$$
 
 Note if the entity is a team, $M_{i}$ is set to 1 as reward payout multipliers are considered later when distributing rewards [amongst the team members](#distributing-rewards-amongst-team-members).
 
-To avoid exponential formulas in the core implementation, the above exponential function can be approximated using the following 7th order Taylor expansion. An odd number of terms has intentionally been chosen so the expansion diverges to $-\infty$ rather than $+\infty$.
+To avoid exponential formulas in the core implementation, the above exponential equation can be approximated using the following 7th order Taylor expansion. An odd number of terms has intentionally been chosen so the expansion diverges to $-\infty$ rather than $+\infty$.
```
barnabee: Should this be gone?

barnabee (commented):
Two comments about things which maybe aren't still needed?

protocol/0057-TRAN-transfers.md
```diff
@@ -87,6 +87,41 @@ To be able to dispatch rewards to reward pools of multiple markets pro rata to t
...
+  - `ENTITY_SCOPE_INDIVIDUALS` - the rewards must be distributed directly amongst eligible parties
+  - `ENTITY_SCOPE_TEAMS` - the rewards must be distributed amongst directly eligible teams (and then amongst team members)
+- `individual scope` - optional enum if the entity scope is `ENTITY_SCOPE_INDIVIDUALS` which defines the subset of individuals which are eligible to be rewarded.
+  - `INDIVIDUAL_SCOPE_ALL` - all parties on the network are within the scope of this reward
+  - `INDIVIDUAL_SCOPE_IN_TEAM` - all parties which are part of a team are within the scope of this reward
+  - `INDIVIDUAL_SCOPE_NOT_IN_TEAM` - all parties which are not part of a team are within the scope of this reward
+- `team scope` - optional list if the reward type is `ENTITY_SCOPE_TEAMS`, field allows the funder to define a list of team ids which are eligible to be rewarded from this transfer
+- `staking_requirement` - the required minimum number of tokens staked for a party to be considered eligible. Defaults to `0`.
+- `notional_time_weighted_average_position_requirement` - the required minimum notional time-weighted averaged position required for a party to be considered eligible. Defaults to `0`.
```
barnabee, line 104: this is in terms of quantum? assume that's covered elsewhere but if not, should be covered somewhere

protocol/0057-TRAN-transfers.md
```diff
@@ -87,6 +87,41 @@ To be able to dispatch rewards to reward pools of multiple markets pro rata to t
...
+
+- `entity scope` - mandatory enum which defines the entities within scope.
+  - `ENTITY_SCOPE_INDIVIDUALS` - the rewards must be distributed directly amongst eligible parties
+  - `ENTITY_SCOPE_TEAMS` - the rewards must be distributed amongst directly eligible teams (and then amongst team members)
+- `individual scope` - optional enum if the entity scope is `ENTITY_SCOPE_INDIVIDUALS` which defines the subset of individuals which are eligible to be rewarded.
+  - `INDIVIDUAL_SCOPE_ALL` - all parties on the network are within the scope of this reward
+  - `INDIVIDUAL_SCOPE_IN_TEAM` - all parties which are part of a team are within the scope of this reward
+  - `INDIVIDUAL_SCOPE_NOT_IN_TEAM` - all parties which are not part of a team are within the scope of this reward
+- `team scope` - optional list if the reward type is `ENTITY_SCOPE_TEAMS`, field allows the funder to define a list of team ids which are eligible to be rewarded from this transfer
+- `staking_requirement` - the required minimum number of tokens staked for a party to be considered eligible. Defaults to `0`.
```
barnabee:
```suggestion
- `staking_requirement` - the required minimum number of governance (e.g. VEGA) tokens staked for a party to be considered eligible. Defaults to `0`.
```

protocol/0056-REWA-rewards_overview.md
```diff
@@ -98,25 +184,100 @@ Reward accounts and balances must be saved in [LNL checkpoints](./0073-LIMN-limi
...
+
+$$s_{i} = \frac{d_{i}}{\sum_{i=1}^{n}d_{i}}$$
+
+#### Distributing based on rank
+
+Rewards funded using the rank-distribution strategy should be distributed as follows.
+
+1. Calculate each entity's reward metric.
+2. Arrange all entities in a list in descending order based on their reward metric values and determine their rank. If multiple entities share the same reward metric value, they should be assigned the same rank. The next entity's rank should be adjusted to account for the shared rank among the previous entities. For instance, if two entities share rank 2, the next entity should be assigned rank 4 (since there are two entities with rank 2).
+3. Set the entities `share_ratio` based on their position in the `rank_table` specified in the recurring transfer.
```
barnabee, line 228: Is there a need for a param for maximum rank table size? If not we should specify the max size that needs to be supported here. I guess it could be pretty big like a few hundred or low thousands entries without issue… can't imagine anyone want tons of entries really.

protocol/0057-TRAN-transfers.md
```diff
@@ -87,6 +87,41 @@ To be able to dispatch rewards to reward pools of multiple markets pro rata to t
...
+- `notional_time_weighted_average_position_requirement` - the required minimum notional time-weighted averaged position required for a party to be considered eligible. Defaults to `0`.
+
+A party should be able to configure the distribution of rewards by specifying the following fields:
+
+- `window_length` - the number of epochs over which to evaluate the reward metric.
```
ze97286: there must be an upper bound for this or else we have to keep the data forever

protocol/0057-TRAN-transfers.md
```diff
@@ -87,6 +87,41 @@ To be able to dispatch rewards to reward pools of multiple markets pro rata to t
 - `reward metric asset` — (the settlement asset of all markets that will be in scope for the transfer)
 - `market scope` — a subset of markets in which parties are eligible to be rewarded from this transfer.
   - If the market scope is not defined / an empty list, it is taken as all the markets that settle in the reward metric asset.
+
+A party can further control their recurring transfer funding the reward pool by defining the entities which are within scope. Entities within scope can be either individual parties or [teams](./0083-RFPR-on_chain_referral_program.md#glossary). When scoping individuals, a subset of keys can be detailed, and when scoping teams a specific set of team ids can be detailed.
```
ze97286, line 91: > When scoping individuals, a subset of keys can be detailed

I don't see how this is possible. Seems like with individuals you can only choose between the 3 options of everybody, everybody that's part of a team, everybody that's not part of a team.

protocol/0057-TRAN-transfers.md
ze97286: I would imagine that  `staking_requirement` and `notional_time_weighted_average_position_requirement` are not applicable for the validator ranking metric. same comment about scope `ENTITY_SCOPE_TEAMS`

protocol/0057-TRAN-transfers.md
cdummett: Do "Quarterly" games sound sensible as an upper-limit for the length of games @MM0819 @barnabee ?

So an upper bound of `93` for `window_length`.

protocol/0057-TRAN-transfers.md
cdummett: Hm good point, the original intention was it would be an additional filter but this requirement is overkill and will probably cause headaches when they conflict. I'll remove it.

protocol/0056-REWA-rewards_overview.md
cdummett: Good point, I think a hard limit of 500 rows should give more than enough room.

protocol/0057-TRAN-transfers.md
cdummett: I don't think it necessarily needs to be in terms of quantum?

As part of the recurring transfer at the moment you have to specify an `asset_for_metric` in the recurring transfer so all the markets in the scope of your reward transfer will be using the same settlement asset anyway.

Notional still needed though because you could have BTC / ETH / DOGE markets all settled in USDC within scope.

protocol/0057-TRAN-transfers.md
cdummett: Yeah for the validator ranking metric good point, I'll make that clear in the spec.

For `ENTITY_SCOPE_TEAMS` they should still apply. We calculate each team members individual reward metric first and this is where these requirements are applied (if they don't meet them their reward-metric is `0`). The teams reward metric is then calculated from the individual team members metrics as per the spec.

ze97286: what I meant is that I don't see how ENTITY_SCOPE_TEAMS is relevant for the validator ranking metric? this could only be relevant if the validator public keys are members in teams which I don't know if we want.

protocol/0057-TRAN-transfers.md
```diff
@@ -87,6 +87,41 @@ To be able to dispatch rewards to reward pools of multiple markets pro rata to t
...
+  - `INDIVIDUAL_SCOPE_NOT_IN_TEAM` - all parties which are not part of a team are within the scope of this reward
+- `team scope` - optional list if the reward type is `ENTITY_SCOPE_TEAMS`, field allows the funder to define a list of team ids which are eligible to be rewarded from this transfer
+- `staking_requirement` - the required minimum number of tokens staked for a party to be considered eligible. Defaults to `0`.
+- `notional_time_weighted_average_position_requirement` - the required minimum notional time-weighted averaged position required for a party to be considered eligible. Defaults to `0`.
+
+A party should be able to configure the distribution of rewards by specifying the following fields:
+
+- `window_length` - the number of epochs over which to evaluate the reward metric.
+- `lock_period` - the number of epochs after distribution to delay [vesting of rewards](./0085-RVST-rewards_vesting.md#vesting-mechanics) by.
+- `distribution_strategy` - enum defining which [distribution strategy](./0056-REWA-rewards_overview.md#distributing-rewards-between-entities) to use.
```
ze97286, line 110: I suppose this does not apply to market proposer metric? same comment on window_length

protocol/0057-TRAN-transfers.md
cdummett: Ah I'm with you. 

There is nothing in the referral spec stopping validators creating/joining teams so I would lean towards supporting this combination of ENTITY_SCOPE_TEAMS and validator reward metric. Also helps with consistency, ideally all metrics would work for both individual and team scopes.

If some complexity in the implementation comes up though we can re think it because using the individual scope is the more likely use case for this metric.

protocol/0057-TRAN-transfers.md
cdummett: Yeah neither would be relevant.

protocol/0057-TRAN-transfers.md
ze97286: same question on market proposer, in this case the reward is targeted to individuals, I'm not sure what it would mean to specify distribution strategy and scope teams for this metric.

cdummett: I think it harmless to support but since we now have two metrics where it might make sense to disable the `ENTITY_SCOPE_TEAMS` and `DISTRIBUTION_STRATEGY_RANK` it might be best to prioritise clarity over consistency.

I'll raise this in the protocol-design session now and let you know.

protocol/0085-RVST-rewards_vesting.md
gordsport: @cdummett - can we add some sensible values to both this and the spreadsheet and these will be used. A core isue to be raised to update once the real values have ben decided

protocol/0057-TRAN-transfers.md
gordsport: agreed this is sensible and we should use a round figure (say 100), this may need to be considered if the epoch length should change in the future. 

Core to consider upper/lower bounds based on memory / perf concerns and the spec updated

protocol/0057-TRAN-transfers.md
gordsport: @ze97286 - if the restrictions are preferred to be added then this should be done, no hard requirements from research

protocol/0057-TRAN-transfers.md
barnabee: But this is the average position requirement for the reward though? I didn't think this was specific to any asset?

protocol/0085-RVST-rewards_vesting.md
```diff
@@ -0,0 +1,81 @@
+# Rewards Vesting Specification
```
barnabee, line 1: @cdummett can you point me to where the initial delay to vesting start is in the spec?

wordlist.txt
```diff
@@ -315,3 +315,7 @@ dApp
 leaderboards
 ReferralSet
 CreateReferralSet
+enum
```
edd, line 318: **thought (non-blocking)**: The spoell checker already ignores things in backtick (`) codeblocks. It looks like these are required because $this style$ is... latex or something? Anyway. I wonder if we can configure the spellchecker to ignore lines with those in... shrug 

Non-blocking because this works fine too.

cdummett: Yeah exactly, anything in $$ is latex and doesn't get caught by the spellchecked. Definitely agree we should look into a more elegant solution then dumping everything in the wordlist when we have some time.

protocol/0085-RVST-rewards_vesting.md
cdummett: @barnabe the rewards spec has been updated to emphasise the `lock_period` currently described in the transfers spec.

barnabee (approved):

davidsiska-vega (approved):

PR #1881, opened 2023-08-07
feat: referral rewards are paid to the original referrer

cdummett:
## Summary
Currently, referral rewards are generated by each referee and paid to the creator of their current referral set. Since, parties are free to move between referral sets, their is no guarantee rewards generated by a referee will be paid to the original party who actually referred the user to the network.

## Changes
PR updates the spec so when a referral code is applied for the first time, the referee becomes permanently associated with the referrer. Referral rewards will now always be paid out to this referrer regardless of whether the party changes referral set.

Note, a party will contribute to their current referral sets volume and their benefit factors will be set using their current referral sets running volume.


protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -191,7 +193,9 @@ message ApplyReferralCode{
...
+If a party is not currently a referee, they must immediately be added to the referral set and [benefit factors updated](#setting-benefit-factors) accordingly. Their key must then become associated with the referrers key. All referral rewards will be transferred to this referrers key, regardless of whether the party reapplies a new referral code.
+
+If a party is already a referee, and submits another `ApplyReferralCode` transaction, they will be transferred to the new referral set at the start of the next epoch. Note, if the referee has submitted multiple transactions in an epoch, the referee will be associated with the set specified in the latest valid transaction. When a referee is transferred, the referrer for which they generate referral rewards for **is not** updated.
```
barnabee, line 198:
```suggestion
If a party is already a referee, and submits another `ApplyReferralCode` transaction, they will be transferred to the new referral set at the start of the next epoch. Note, if the referee has submitted multiple transactions in an epoch, the referee will be associated with the set specified in the latest valid transaction. When a referee is transferred, the referrer for which they generate referral rewards for **is not** updated.
```

Having looked at the rest of the spec I think this should be changed so that a new refferer CAN BE set _if and only if_ the existing referrer is not staking the required tokens at the time.

Thoughts @cdummett @MM0819?

barnabee (commented):
Is this superseded by https://github.com/vegaprotocol/specs/issues/1912 ?

cdummett:
> Is this superseded by #1912 ?

Yes this PR can be closed.

PR #1899, opened 2023-08-11
Add check-features check

edd:
Add a new approbation check - verifies that a features file doesn't have duplicate ACs and that all the ACs exist.

Cheeky additional commit [bd92a65](https://github.com/vegaprotocol/specs/pull/1899/commits/bd92a65f385b5614ae1ef3ac12f88186a188b965):
- Cleans up some spot spec grammar that was causing confusion when testing.


gordsport (approved):
@edd 

Works great - added ipynb files and added to the make file too.

I snuck in a bit of a clean up that came out of a review with the QA guys earlier.

PR #1906, opened 2023-08-21
feat: added Acceptance Criteria for Liquidity Provisioning

VanithaVega:


protocol/0044-LIME-lp_mechanics.md
```diff
@@ -341,11 +341,19 @@ In the case of spot markets it will be transferred into the network treasury for
   - LP can cancel their commitment without incurring penalties (<a name="0044-LIME-053" href="#0044-LIME-053">0044-LIME-053</a>)
 - For a market that is in continuous trading and a single LP has committed liquidity:
   - The LP can cancel their commitment at any time (though this may involve incurring a penalty) (<a name="0044-LIME-060" href="#0044-LIME-060">0044-LIME-060</a>) for spot (<a name="0044-LIME-056" href="#0044-LIME-056">0044-LIME-056</a>)
+  - The LP can increase their commitment at any time without incurring fee (<a name="0044-LIME-061" href="#0044-LIME-061">0044-LIME-061</a>) for spot (<a name="0044-LIME-062" href="#0044-LIME-062">0044-LIME-062</a>)
+  - When LP increases their commitment higher than their account balance are rejected and old provision persists (<a name="0044-LIME-063" href="#0044-LIME-063">0044-LIME-063</a>) for spot (<a name="0044-LIME-064" href="#0044-LIME-064">0044-LIME-064</a>)
+  - When LP decreases their commitment are not charged (<a name="0044-LIME-065" href="#0044-LIME-065">0044-LIME-065</a>) for spot (<a name="0044-LIME-066" href="#0044-LIME-066">0044-LIME-066</a>)
```
davidsiska-vega: This isn't correct.

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -341,11 +341,19 @@ In the case of spot markets it will be transferred into the network treasury for
...
+  - When LP decreases their commitment to zero are rejected (<a name="0044-LIME-067" href="#0044-LIME-067">0044-LIME-067</a>) for spot (<a name="0044-LIME-068" href="#0044-LIME-068">0044-LIME-068</a>)
```
davidsiska-vega: I don't believe this is correct either: setting commitment to `0` should work the same as cancelling the commitment - in fact it's probably the only way to do so.

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -341,11 +341,19 @@ In the case of spot markets it will be transferred into the network treasury for
...
+  - When LP amends the Fee Factor greater than `market.liquidityV2.maximumLiquidityFeeFactorLevel` are rejected (<a name="0044-LIME-070" href="#0044-LIME-070">0044-LIME-070</a>) for spot (<a name="0044-LIME-071" href="#0044-LIME-071">0044-LIME-071</a>)
```
davidsiska-vega: This should only be temporary core naming for the parameter. Once everything is on SLA the `V2` should go away. Specs shouldn't use `V2`.

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -341,11 +341,19 @@ In the case of spot markets it will be transferred into the network treasury for
...
+  - Distressed LP is cancelled automatically or closeout of LP order which doesn't meet minimum margin are cancelled (<a name="0044-LIME-071" href="#0044-LIME-071">0044-LIME-071</a>) for spot (<a name="0044-LIME-072" href="#0044-LIME-072">0044-LIME-072</a>)
```
davidsiska-vega: I don't believe this is correct. If an LP cannot meet it's maintenance margin (general account balance `0`) then 
1. we remove all their orders from the book.
2. we check the margin again, 
2a) either they're ok now in which case that's it 
2b) they are still not meeting maintenance margin requirement. In this case their bond balance is (partially) transferred to their margin account and now
2bi) they meet maintenance margin requirement (their stake has been reduced because their bond isn't what it used to be) or 
2bii) their bond account is empty (balance `0`) and their position is closed out. They are no longer an LP (because bond account balance is `0`).

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -341,11 +341,19 @@ In the case of spot markets it will be transferred into the network treasury for
...
+  - Party submits LP in multiple markets and multiple bond accounts are created and managed by vega (<a name="0044-LIME-072" href="#0044-LIME-072">0044-LIME-072</a>) for spot (<a name="0044-LIME-073" href="#0044-LIME-073">0044-LIME-073</a>)
```
```suggestion
  - Party submits LP in multiple markets and multiple bond accounts are created and managed by vega (<a name="0044-LIME-072" href="#0044-LIME-072">0044-LIME-072</a>) for spot (<a name="0044-LIME-073" href="#0044-LIME-073">0044-LIME-073</a>)
```
```suggestion
  - If a party submits LP provisions in multiple markets then multiple bond accounts are created and managed by Vega. (<a name="0044-LIME-072" href="#0044-LIME-072">0044-LIME-072</a>) for spot (<a name="0044-LIME-073" href="#0044-LIME-073">0044-LIME-073</a>)
```

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -341,11 +341,19 @@ In the case of spot markets it will be transferred into the network treasury for
...
 
 ### Qualifying Order Types
 
 - Once liquidity is committed LPs can meet their commitment by placing limit orders, pegged limit orders and iceberg orders. For iceberg orders only the visible peak counts towards the commitment. (<a name="0044-LIME-028" href="#0044-LIME-028">0044-LIME-028</a>)
 - Parked pegged limit orders and stop-loss orders do not count towards an LPs liquidity commitment. (<a name="0044-LIME-029" href="#0044-LIME-029">0044-LIME-029</a>)
+- LPs can submit GFA orders and do not count towards an LPs liquidity commitment (<a name="0044-LIME-069" href="#0044-LIME-069">0044-LIME-069</a>)
```
davidsiska-vega: I don't follow.

davidsiska-vega (changes requested):
Left comments.

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -341,11 +341,19 @@ In the case of spot markets it will be transferred into the network treasury for
   - LP can cancel their commitment without incurring penalties (<a name="0044-LIME-053" href="#0044-LIME-053">0044-LIME-053</a>)
 - For a market that is in continuous trading and a single LP has committed liquidity:
   - The LP can cancel their commitment at any time (though this may involve incurring a penalty) (<a name="0044-LIME-060" href="#0044-LIME-060">0044-LIME-060</a>) for spot (<a name="0044-LIME-056" href="#0044-LIME-056">0044-LIME-056</a>)
+  - The LP can increase their commitment at any time without incurring fee (<a name="0044-LIME-061" href="#0044-LIME-061">0044-LIME-061</a>) for spot (<a name="0044-LIME-062" href="#0044-LIME-062">0044-LIME-062</a>)
+  - When LP increases their commitment higher than their account balance are rejected and old provision persists (<a name="0044-LIME-063" href="#0044-LIME-063">0044-LIME-063</a>) for spot (<a name="0044-LIME-064" href="#0044-LIME-064">0044-LIME-064</a>)
```
davidsiska-vega: This isn't clear about what's being compared to what.

protocol/0044-LIME-lp_mechanics.md
VanithaVega: @davidsiska-vega , current sla vega system throws error message on amend operation "commitment amount is too low".

protocol/0044-LIME-lp_mechanics.md
VanithaVega: @davidsiska-vega , Normally GFA orders get cancelled when market moves from **Opening Auction** to **Continuous Trading** .  When already provisioned LP trader submits GFA order at Opening Auction and when market mode changes, it get cancelled.  No blond slashing happens on GFA orders.

gordsport (dismissed):
@VanithaVega - once the content of this is agreed please can you review the AC code numbering. 

new ACs should have totally new numbers

ACs that are changing in a way thats different to the test either need to have a new number or an issue created in the relevant repo to update the test.

protocol/0044-LIME-lp_mechanics.md
karlem: It isn't. When they decrease they can be penalised as it is mentioned in the spec

protocol/0044-LIME-lp_mechanics.md
karlem: @davidsiska-vega we do not allow it. But they can cancel which is as you said similar like reducing to 0.

protocol/0044-LIME-lp_mechanics.md
karlem: We only close out lp when their bond account gets to 0.

VanithaVega: changed spec to keep the LP in active state

protocol/0044-LIME-lp_mechanics.md
VanithaVega: yes, changed the spec as When LP decreases their commitment are charged as per market.liquidity.earlyExitPenalty

protocol/0044-LIME-lp_mechanics.md
VanithaVega: LP which decreases the commitment are charged as factor of market.liquidity.earlyExitPenalty and when the commitment is increased from x to x+1000, they are not charged any fee.

VanithaVega: However, when LP tries to increase the commitment greater than their available fund are rejected. Lets say LP got 1000 units in General and 1000 units in Margin and 100 unit in Bond accoutn. When they try to increase their commitment from 100 units to 3000 units , it fails .

davidsiska-vega: What you say may be fine but the AC needs to stand on its own (at least for most people) - the explanatory comments left here won't be visible to someone just reading the spec.

protocol/0044-LIME-lp_mechanics.md
davidsiska-vega: @karlem any reason why we shouldn't treat setting commitment to `0` same as cancelling (instead of throwing the error)?

protocol/0044-LIME-lp_mechanics.md
davidsiska-vega: > Normally GFA orders get cancelled when market moves from Opening Auction to Continuous Trading
True, in fact true for any auction not just the opening auction. 

But during an auction we evaluate the LPs commitment and they can choose to meet it with GFA orders. So the AC, as stated, sounds wrong.

protocol/0044-LIME-lp_mechanics.md
davidsiska-vega: This isn't correct either @VanithaVega - it depends on where the market target stake is and what other commitments there are.

protocol/0044-LIME-lp_mechanics.md
davidsiska-vega: @karlem can you specifically check whether the sequence I describe is correct?

protocol/0044-LIME-lp_mechanics.md
Jiajia-Cui: answer from @karlem the reason is we are not sure whether they have actually set the commitment to 0 or just omit it

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -340,11 +340,20 @@ In the case of spot markets it will be transferred into the network treasury for
...
+  - When an LP creates a new provision with zero commitment, it should be rejected with an error message stating that the commitment amount is zero. (<a name="0044-LIME-069" href="#0044-LIME-069">0044-LIME-069</a>) for spot (<a name="0044-LIME-070" href="#0044-LIME-070">0044-LIME-070</a>)
+  - When an LP amends the Fee Factor to a value greater than `market.liquidity.maximumLiquidityFeeFactorLevel`, the amendments are rejected (<a name="0044-LIME-071" href="#0044-LIME-071">0044-LIME-071</a>) for spot (<a name="0044-LIME-072" href="#0044-LIME-072">0044-LIME-072</a>)
+  - A distressed LP (when $ bond account == 0 && general account == 0 && margin account < maintenance margin$) will have their orders cancelled. If the LP remains distressed, the network takes over any positions the LP may hold and zeroes them out (closes them out) (<a name="0044-LIME-073" href="#0044-LIME-073">0044-LIME-073</a>) for spot (<a name="0044-LIME-074" href="#0044-LIME-074">0044-LIME-074</a>)
```
davidsiska-vega: Not sure about the $ there, does the formula render correct?

protocol/0044-LIME-lp_mechanics.md
```diff
@@ -340,11 +340,20 @@ In the case of spot markets it will be transferred into the network treasury for
...
-- Parked pegged limit orders and stop-loss orders do not count towards an LPs liquidity commitment. (<a name="0044-LIME-029" href="#0044-LIME-029">0044-LIME-029</a>)
+- Once liquidity is committed LPs can meet their commitment by placing limit orders, pegged limit orders and iceberg orders. For iceberg orders only the visible peak counts towards the commitment. (<a name="0044-LIME-028" href="#0044-LIME-028">0044-LIME-028</a>) for spot (<a name="0044-LIME-029" href="#0044-LIME-029">0044-LIME-029</a>) 
+- Parked pegged limit orders and stop-loss orders do not count towards an LPs liquidity commitment. (<a name="0044-LIME-077" href="#0044-LIME-077">0044-LIME-077</a>) for spot (<a name="0044-LIME-078" href="#0044-LIME-078">0044-LIME-078</a>)
+- GFA orders during auction (apart form opening auction) from LP count towards LPs liquidity commitment (<a name="0044-LIME-079" href="#0044-LIME-079">0044-LIME-079</a>) for spot (<a name="0044-LIME-080" href="#0044-LIME-080">0044-LIME-080</a>)
```
```suggestion
- GFA orders during auction from LP count towards LPs liquidity commitment (<a name="0044-LIME-079" href="#0044-LIME-079">0044-LIME-079</a>) for spot (<a name="0044-LIME-080" href="#0044-LIME-080">0044-LIME-080</a>)
```

davidsiska-vega (dismissed):
One question, one suggestion, otherwise it's good!

Jiajia-Cui (approved):
LGTM

PR #1913, opened 2023-08-23
feat: add transfer interval field

cdummett:
To support requirements for "daily / weekly / monthly games" a `transfer_interval` field should been added to the recurring transfer transaction.

The field will limit recurring transactions to transferring funds every n epochs. When a new recurring transfer is created, the first transfer will happen after n epochs.

**NOTE:** Since this PR was first created, reward caps based on total fees paid have been added which has implications for rewards which are not distributed every epoch.

- rewards should now be capped by some multiple of total fees paid since rewards were last payed out.
- this would require core to keep a record of the total fees paid by each party in each market. As such the transfer interval is limited at 100 epochs (this is the same limitation to the window length field in the recurring transfer).


protocol/0057-TRAN-transfers.md
```diff
@@ -65,6 +65,9 @@ A recurring transfers needs to contain this specific information:
 - start epoch: at the end of this epoch the first recurring transfer will be made between
 - end epoch (optional): at the end of this epoch the last recurring transfer will be made between, optional. If not specified the transfer run until cancelled (by its creator or by the network as described below).
 - factor, decimal > 0.0 (a factor used with the amount specified for the transfer).
+- transfer interval: number of epochs between transfers, i.e. when 4, funds will be transferred every 4 epochs with the first transfer occurring 4 epochs after the transfer is processed. Must be an integer strictly greater than `0`.
+
+Note, to avoid parties creating [reward pools](#recurring-transfers-to-reward-accounts) which are never funded, if a transfer interval strictly greater than `1` is specified. The funds for the next scheduled transfer should be taken straight away from the origin account, locked by the network, and distributed to the destination account at the end of the correct epoch. This way, if a user cancels a recurring transfer or withdraws all their funds, the next scheduled transfer will still take place.
```
barnabee: Why not apply this for all recurring transfers even with interval one?

cdummett: I have no strong arguments against it other than I didn't think we would want to lose the existing functionality (and break tests).

It would probably be better to have consistent behaviour tbh so I'll change it 👍

barnabee:
@gordsport is there a specific change I'm reviewing + is this all in Cosmic Elevator still?

gordsport:
> @gordsport is there a specific change I'm reviewing + is this all in Cosmic Elevator still?

@barnabee - its to re-review since the update from the comment above.

@jeremyletang has mentioned this could be included, however, IMO it would be a stretch and likely to be in the deployment after cosmic elevator

barnabee (approved):

barnabee (dismissed):

protocol/0057-TRAN-transfers.md
```diff
@@ -67,6 +67,9 @@ A recurring transfers needs to contain this specific information:
 - start epoch: at the end of this epoch the first recurring transfer will be made between
 - end epoch (optional): at the end of this epoch the last recurring transfer will be made between, optional. If not specified the transfer run until cancelled (by its creator or by the network as described below).
 - factor, decimal > 0.0 (a factor used with the amount specified for the transfer).
+- transfer interval: number of epochs between transfers, i.e. when 4, funds will be transferred every 4 epochs with the first transfer occurring 4 epochs after the transaction is processed. Must be an integer strictly greater than `0` and less than `100` (this ceiling is for performance reasons and matches the limit on the `window_length` field).
```
ze97286: I think it would be better to make this optional for backward compatibility and no value is interpreted as 1.

ze97286: it would be a lot easier (or at all feasible) if transfer_interval is part of the dispatch strategy. why? the reward has no access (and cannot have access) to which transfer funded the reward account, because there could be multiple transfers with the same dispatch strategy funding the account, so say we have transfer A and transfer B funding a reward account with the exact same dispatch strategy but different interval, the reward engine cannot enforce the cap as it doesn't have information about which look-back period it should use. So while it's not super clean, putting it in the dispatch strategy would work better I think.

ze97286: also, I'm assuming this goes for both recurring and governance recurring transfers but should probably be made explicit here or on the ACs.

protocol/0057-TRAN-transfers.md
```diff
@@ -67,6 +67,9 @@ A recurring transfers needs to contain this specific information:
...
+
+Note, to avoid parties intentionally creating transfers which never occur, either between parties or to fund a [reward pool](#recurring-transfers-to-reward-accounts). The funds for the next scheduled transfer should be taken straight away from the origin account, locked by the network, and distributed to the destination account at the end of the correct epoch. This way, if a user cancels a recurring transfer or withdraws all their funds, the next scheduled transfer will still take place.
```
ze97286: this is not straightforward. reward transfers do not go through unless there's eligible parties, so how is this gonna work? we first transfer the transfer amount to some temp account and then at the relevant epoch we either return it (and park the amount for the next round)? Also this also means that a transfer cannot be cancelled and is guaranteed to go at least once. I don't think this is desired. If this is spam protection maybe we need to think of an alternative.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -27,6 +27,8 @@ At the end of the epoch:
 1. Then the entire balance of each reward account is distributed amongst entities with a non-zero reward metric for that reward type and market using the mechanism specified in the recurring transfer.
 1. Distributed rewards are transferred to a [vesting account](./0085-RVST-rewards_vesting.md).
 
+Note, reward metrics should be calculated and exposed through an API even if the the reward pool is not to be funded this epoch (i.e. when the `transfer_interval` set in the [recurring transfer](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts) is strictly greater than `1`).
```
ze97286: what exactly does this mean? what are you referring to by _reward metrics_ that you're expecting to be calculated and published every epoch?

protocol/0057-TRAN-transfers.md
ze97286: how does transfer_interval play with window size, i.e. unless the window size is greater than or equal to the transfer interval then that means there are "dead epochs" which don't count towards the reward. E.g. window size is 1 and transfer_interval is 7, then we're only looking at the last epoch and the previous 6 are not counting towards anything (but the cap).

cdummett: Agreed on making the field optional and yes it should go for governance transfers as well. Will make updates for both of those.

For how it works with respect to window size, yes in the example you give I there will be dead epochs which don't count towards each parties score. Most likely users are going to setup rewards where `transfer_interval=window_size` (i.e. pay out every n epochs assessing performance over the last n epochs). 

I can't think of a reason why setting `transfer_interval!=window_size` would cause issues so my 2-cents is allow it, but in the docs make it clear how these fields relate to each other (can make some spec clarifications as well if needed).

ze97286: I don't think we shouldn't allow it but I think transfer_interval>window_size makes no sense as a configuration :)

cdummett: Yep sounds sensible, agree it will be clearer in the dispatch strategy regrading multiple transfers funding the same reward account.

protocol/0056-REWA-rewards_overview.md
cdummett: I will remove this entirely. 

It's a bit of legacy text which is somewhat redundant now we are considering publishing interim reward information.

It was supposed to ensure we were at least publishing scores each epoch even if the rewards were only distributed every n epochs.

protocol/0057-TRAN-transfers.md
ze97286: maybe makes-no-sense is too harsh though, anyhow shouldn't be a problem.

protocol/0057-TRAN-transfers.md
cdummett: Less spam protection (although we did discuss how it could help with that) and more a prevention of users creating "fake" competitions. 

e.g. I create a new market, I want to attract users so I create a large reward with a first payout far in the future (say  one month). Users then trade expecting rewards but at the end of the month I cancel the transfer or the transfer fails for whatever reason. Users left without rewards.

In the case of no eligible parties I think the flow you described is fine, "return funds then park funds for the next round."  In the case a user wants to cancel a transfer because a market has no activity, they could cancel the transfer and after the next relevant payout epoch, if there are still no eligible parties they would have their funds returned.

ze97286: Anyone can make a mistake in setting up transfers as we know well, and such a mistake can be very expensive as you're guaranteeing that at least one round of payout will be made and you can't do anything to prevent it once the transfer has been included in a block.

protocol/0057-TRAN-transfers.md
cdummett: > I don't think we shouldn't allow it but I think transfer_interval>window_size makes no sense as a configuration :)

Maybe a recurring weekend trading competition? (transfer_interval=7, window_size=2). I'm playing devils advocate though, I agree it's not the most sensible setup.

protocol/0057-TRAN-transfers.md
cdummett: Resolving with the decision to remove the requirement to lock funds for now as per slack discussions.

protocol/0057-TRAN-transfers.md
```diff
@@ -108,9 +108,10 @@ To support entity scoping, the transaction include the following fields:
...
+- `cap_reward_fee_multiple` [optional] - if set, the actual amount of reward transferred to each public key during distribution for this transfer will be `min(calculated_reward_in_quantum, cap_reward_fee_multiple × feed_paid_since_last_payout)` (fees paid since last payout is akin to checking the total fees paid over the last `transfer_interval` epochs). When calculating how much of the reward each one is getting, if some is left from the applied cap, we recalculate on the remaining balance only for parties that have not reached their cap until the leftover is less than 1 reward asset unit or the maximum rounds of recalculation is 10. If all keys are capped (i.e. the total amount of the transfer cannot be be sent to eligible keys without breaching the cap) then the remaining balance must be left in the reward pool and included in the distribution in future epochs. If this occurs, and the total transferred in a given epoch, this does not affect the size of the next iteration, which proceeds as normal (including decay factors etc.) as if the full transfer has been made.
```
Jiajia-Cui, line 113:
```suggestion
- `cap_reward_fee_multiple` [optional] - if set, the actual amount of reward transferred to each public key during distribution for this transfer will be `min(calculated_reward_in_quantum, cap_reward_fee_multiple × feed_paid_since_last_payout)` (fees paid since last payout is akin to checking the total fees paid over the last `transfer_interval` epochs). When calculating how much of the reward each one is getting, if some is left from the applied cap, we recalculate on the remaining balance only for parties that have not reached their cap until the leftover is less than 1 reward asset unit or the maximum rounds of recalculation which is set to 10. If all keys are capped (i.e. the total amount of the transfer cannot be be sent to eligible keys without breaching the cap) then the remaining balance must be left in the reward pool and included in the distribution in future epochs. If this occurs, and the total transferred in a given epoch, this does not affect the size of the next iteration, which proceeds as normal (including decay factors etc.) as if the full transfer has been made.
```

Jiajia-Cui (approved):

Issue #1912, opened 2023-08-23
Referral programme feature updates


barnabee:
We've received internal stakeholder feedback requesting the following.

### Referral programme

- Additional table of tiers mapping number of staked VEGA (staked for at least N epochs, where N is a network parameter) to additional referrer bonus multiple:
    - Referrer receives base share of fees earned by their refereee × staking multiple
    - I believe there will need to be a new net param setting the maximum total portion of fees that can be diverted to a referrer
- Change calculations so that the tier used for calculating the referrers bonus and the referee's discount come from the referee's referral set/team, not the trader's current team.
- [no spec or code change] The staked token requirement to create a referral code should be set to zero at launch. This could be increased later if necessary to mitigate issues.

### Other discounts (prev. volume discount)

- In addition to volume discounts for all users
- Add a table of tiers mapping number of staked VEGA (staked for at least N epochs, where N is a network parameter) to additional fee discount rate (%)

PR #1914, opened 2023-08-24
feat: update referral program for new requirements

cdummett:
PR updates specs touched by the referral program as per the additional requirements specified in #1912.


protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -13,8 +13,12 @@ Whilst a referral program is active, the following benefits may be available to
 
 Providing a party has been associated with a referral set for long enough, they will become eligible for greater benefits as their referral sets running taker volume increases.
 
+Referrers will also the ability to increase the proportion of referee taker fees they receive by staking governance tokens ($VEGA).
```
ValentinTrinque: I think this is missing the been "have"

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -40,11 +45,14 @@ If the community wish to update the referral program limits **and** apply these
 
 Enabling or changing the terms of the on-chain referral program can be proposed via governance. As part of the proposal, the proposer specifies the following fields:
 
-- `benefit_tiers`: a list of dictionaries with the following fields
+- `benefit_tiers`: a list of dictionaries (with the below fields) defining the reward and discount factors from referrals
   - `minimum_running_notional_taker_volume`: the required [`referral_set_running_notional_taker_volume`](#referral-set-volumes) in quantum units for parties to access this tier
   - `minimum_epochs`: the required number of epochs a party must have been in a referral set to access this tier
```
ValentinTrinque: Should we replace party by referee since it only affect the referees? 
> the required number of epochs a referee must have been in a referral set to access this tier

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -13,8 +13,12 @@ Whilst a referral program is active, the following benefits may be available to
 
 Providing a party has been associated with a referral set for long enough, they will become eligible for greater benefits as their referral sets running taker volume increases.
 
+Referrers will also have the ability to increase the proportion of referee taker fees they receive by staking governance tokens ($VEGA).
+
 To create an emphasis on community, collaboration, and competition. Referrers will be able to designate their referral set as a team. Teams will have additional fields which allow them to be visible on leaderboards and later to compete for team based rewards.
 
+Note, if a party wants to contribute towards a different referral set or compete as a member of a different team, a party is able to move between referral sets by "reapplying" a new referral code. However, if they do, their referral rewards will still be paid out to the referrer of the original referral set.
```
barnabee: Shouldn't this also be that their referral rewards will also still be calculated based on the volume of their original referral set?

I am wondering here if this does mean people can need to be in two sets: their original refferer one for referrals and their current team for incentives, else they stop contributing to their original referrer's volume and maybe that's not ideal? 🤔

ValentinTrinque:
This computation is wrong:
```diff
party_epoch_notional_taker_volume = 
   party_epoch_notional_taker_volume + 
-   (trade_price * trade_size * settlement_asset_quantum)
+   (trade_price * trade_size / settlement_asset_quantum)
```

protocol/0083-RFPR-on_chain_referral_program.md
cdummett: > Shouldn't this also be that their referral rewards will also still be calculated based on the volume of their original referral set?

Yes this is true and is described in the "Setting benefit factors" section. This paragraph should be making that clear though so will edit.

```suggestion
Note, if a party wants to compete as a member of a different team, a party is able to move between teams by "reapplying" a new referral code. However, if they do, their referral rewards will still be paid out to the referrer of the original referral set and they will still contribute to the original referral sets volume.
```

ValentinTrinque: We lack clarity on the naming here. I believe we should separate the concept of team and referral set in here for the sake of clarity.

Basically, you can change team, but not referral set. So this paragraph would be:
```suggestion
Note, if a referee wants to compete as a member of a different team, he is able to move between teams by "reapplying" a referral code. However, their referral rewards will still be computed based their original referral set performance and, thus, paid out to the their original referrer.
```

barnabee: Agree that makes sense with this change

barnabee (approved):

ValentinTrinque (approved):

PR #1933, opened 2023-09-07
fix: Minor spelling corrections, plus two new ACs for testing perp da…

jgsbennett:
…ta is visible in REST and GRPC.


gordsport (approved):
LGTM

PR #1953, opened 2023-09-13
feat: add clarification when no indicative price

Jiajia-Cui:


protocol/0044-LIME-lp_mechanics.md
```diff
@@ -211,6 +211,12 @@ We calculate the volume of notional that is in the range
 (1.0-market.liquidity.priceRange) x min(last trade price, indicative uncrossing price) <=  price levels <= (1.0+market.liquidity.priceRange) x max(last trade price, indicative uncrossing price).
 ```
 
+If there is no 'indicative uncrossing price' then any price should count, and the price range should be
```
```suggestion
If there is no 'indicative uncrossing price' then volume placed at any price should count towards the LP's commitment i.e the price range is interpreted as
```

davidsiska-vega (approved):

PR #1955, opened 2023-09-13
Fix/0081

### LINKED ISSUE #1847: Would suggest removing "(i.e. leaves the opening auction)" from successor market spec's as it is a bit confusing.
Reading the following spec's makes me believe that `enacted = leaves opening auction`. 

But a market can be enacted and still be in opening auction until it has orders and liquidity. 

I would suggest removing _"i.e leaves the opening auction"_ since I believe that it makes the description more clear and does not change the specs. 

https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0081-SUCM-successor_markets.md#0081-SUCM-011

davidsiska-vega:
Closes #1847


daunatv (approved):
The amendment has removed potential confusion

PR #1958, opened 2023-09-13
feat: add spec for staking discounts

### LINKED ISSUE #1912: Referral programme feature updates

cdummett:
Spec proposal for adding staking discounts to cover points in #1912.

### DIFF
--- a/protocol/0029-FEES-fees.md
+++ b/protocol/0029-FEES-fees.md
@@ -23,7 +23,7 @@ Note that maker_fee = 0 if there is no maker, taker relationship between the tra
 
 ## Applying benefit factors
 
-Before fees are transferred, if there is an [active referral program](./0083-RFPR-on_chain_referral_program.md) or [volume discount program](./0085-VDPR-volume_discount_program.md), each parties fee components must be modified as follows.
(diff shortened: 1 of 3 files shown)


barnabee (dismissed):

barnabee:
We should avoid merging this until we know if it's going to the Cosmic Elevator branch or the next milestone.

The strong likelihood at this point is for the next milestone.

PR #1962, opened 2023-09-18
feat: add additional referral program test ACs

cdummett:
PR adds additional and aligns existing ACs to cover core implementation of the referral program.

ACs are already tested and validated.


protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -465,10 +465,13 @@ The Estimate Fees API should now calculate the following additional information:
 
 #### Epoch and running volumes
 
-1. Each trade should increment the taker parties `party_epoch_notional_taker_volume` by the volume of the trade (expressed in quantum units) (<a name="0083-RFPR-031" href="#0083-RFPR-031">0083-RFPR-031</a>).
+1. Each trade in which a party is the "maker" **should not** increment the taker parties `party_epoch_notional_taker_volume` by the volume of the trade (expressed in quantum units) (<a name="0083-RFPR-048" href="#0083-RFPR-048">0083-RFPR-048</a>).
+1. Each trade in which a party is the "taker" **should** increment the taker parties `party_epoch_notional_taker_volume` by the volume of the trade (expressed in quantum units) (<a name="0083-RFPR-031" href="#0083-RFPR-031">0083-RFPR-031</a>).
 1. A trade generated during auction uncrossing should not contribute to either parties `party_epoch_notional_taker_volume` (<a name="0083-RFPR-032" href="#0083-RFPR-032">0083-RFPR-032</a>).
+1. Two trades with an identical price and size, but in separate markets using assets with different quantum values, should increment a parties `party_epoch_notional_taker_volume` by the same amount (<a name="0083-RFPR-049" href="#0083-RFPR-049">0083-RFPR-049</a>).
```
davidsiska-vega: Does the volume ever get compared between different assets?

davidsiska-vega (commented):
Still confused about the volume / notional / quantum stuff.

protocol/0083-RFPR-on_chain_referral_program.md
cdummett: Yeah a set only has one running volume value, so if we have trading in various markets using different settlement assets we we need to be able to aggregate them into a single value using quantum.

This AC is to basically check that we are scaling volume correctly into notional volume using settlement decimal places and quantum.

davidsiska-vega (approved):

PR #1978, opened 2023-09-27
feat: add spam protection rules for referral program transactions

cdummett:
PR adds the following rules (and ACs) for referral program transactions:

- `CreateReferralSet` and `UpdateReferralSet` transactions will be pre-block rejected if the party is not staking enough tokens.
- `ApplyReferralCode` transactions ill be pre-block rejected if the party has not deposited enough funds.
- All referral transaction types will be pre-block rejected if more than n (of each individual type) have been submitted in an epoch and post-block rejected if submitted in a block. Parties banned if they break the 50% rule.


protocol/0062-SPAM-spam_protection.md
```diff
@@ -117,5 +131,11 @@ More than 360 delegation changes in one epoch (or, respectively, the value of `s
...
+- A party who has submitted strictly more than `spam.protection.max.CreateReferralSet` `CreateReferralSet` transactions in an epoch should have any future `CreateReferralSet` transactions in that epoch pre-block rejected (<a name="0062-SPAM-029" href="#0062-SPAM-029">0062-SPAM-029</a>).
+- A party who has submitted strictly more than `spam.protection.max.UpdateReferralSet` `UpdateReferralSet` transactions in an epoch should have any future `UpdateReferralSet` transactions in that epoch pre-block rejected (<a name="0062-SPAM-030" href="#0062-SPAM-030">0062-SPAM-030</a>).
+- A party who has submitted strictly more than `spam.protection.max.ApplyReferralCode` `ApplyReferralCode` transactions in an epoch should have any future `ApplyReferralCode` transactions in that epoch pre-block rejected (<a name="0062-SPAM-031" href="#0062-SPAM-031">0062-SPAM-021</a>).
```
```suggestion
- A party who has submitted strictly more than `spam.protection.max.ApplyReferralCode` `ApplyReferralCode` transactions in an epoch should have any future `ApplyReferralCode` transactions in that epoch pre-block rejected (<a name="0062-SPAM-031" href="#0062-SPAM-031">0062-SPAM-031</a>).
```

vega-paul:
We need something here to describe the party banning if more blocks than allowed contain the referral trxs aka proposal party banning

protocol/0062-SPAM-spam_protection.md
```diff
@@ -117,5 +131,11 @@ More than 360 delegation changes in one epoch (or, respectively, the value of `s
 - Increase `spam.protection.minimumWithdrawalQuantumMultiple` and verify that a withdrawal transaction that would have been valid according to the former parameter value is rejected with the new one. (<a name="0062-SPAM-023" href="#0062-SPAM-023">0062-SPAM-023</a>)
 - Decrease `spam.protection.minimumWithdrawalQuantumMultiple` and verify that a withdrawal transaction that would have been invalid with the old parameter and is valid with the new value and is accepted.(<a name="0062-SPAM-024" href="#0062-SPAM-024">0062-SPAM-024</a>)
 - Issue a valid withdrawal bundle. Increase `spam.protection.minimumWithdrawalQuantumMultiple` to a value that would no longer allow the creation of the bundle. Ask for the bundle to be re-issued and verify that it's not rejected. (<a name="0062-PALAZZO-001" href="#0062-PALAZZO-001">0062-PALAZZO-001</a>)
+- A party staking less than `referralProgram.minStakedTokens` should have any `CreateReferralSet` transactions pre-block rejected (<a name="0062-SPAM-026" href="#0062-SPAM-026">0062-SPAM-026</a>).
```
vega-paul, line 136: We need something here to describe the party banning if more blocks than allowed contain the referral trxs aka proposal party banning. Also suggest a chat with @ze97286 on what he has delivered and what does/does not match this spec.

vega-paul (changes requested):

protocol/0062-SPAM-spam_protection.md
```diff
@@ -80,6 +80,22 @@ The minimum allowed withdrawal amount is `spam.protection.minimumWithdrawalQuant
...
+
+The [on-chain referral program](./0083-RFPR-on_chain_referral_program.md) adds three transaction types which can be submitted with no cost/risk to the party:
+
+- `CreateReferralSet`
+- `UpdateReferralSet`
+- `ApplyReferralCode`
+
+To avoid spamming of `CreateReferralSet` and `UpdateReferralSet` transactions, a party must meet the staked governance tokens ($VEGA) threshold set by the network parameter `referralProgram.minStakedVegaTokens`. A party who does not meet this requirement should have any transactions of the aforementioned types pre-block rejected.
+
+To avoid spamming of `ApplyReferralCode`, a party must meet the deposited funds threshold set by the network parameter `spam.protection.applyReferral.min.funds`.  All non-VEGA assets count towards this threshold and balances should be scaled appropriately by the assets quantum. A party who does not meet this requirement should have any transactions of the aforementioned type pre-block rejected. This requirement will be checked against snapshots of account balances taken at a frequency determined by the network parameter `spam.protection.balanceSnapshotFrequency`. This network parameter is a duration (e.g. `5s`, `1m5s`).
```
ze97286: question: why exclude vega? i.e. suppose I have a 1000 VEGA in my general account that I received in rewards, why not let it count towards the min funds? (as opposed to staking which is not included in this)

protocol/0062-SPAM-spam_protection.md
```diff
@@ -80,6 +80,22 @@ The minimum allowed withdrawal amount is `spam.protection.minimumWithdrawalQuant
...
+
+Further, each referral transactions is limited to `n` submissions per epoch per party where `n` is controlled by the respective network parameter for the transaction (`spam.protection.max.CreateReferralSet`, `spam.protection.max.UpdateReferralSet`, `spam.protection.max.ApplyReferralCode`).
+
+Any party who submits more than `n` transactions of a single referral transaction type in a single epoch will have all future for should have their transactions pre-block rejected.Any party who manages to fit more then `n` of a single referral transaction type into a single block will be banned for 1/48th of an epoch, or un till the end of the current epoch, whichever comes first.
```
ze97286: I'm assuming I know what you're tying to say but I think the first sentence could be rephrased for clarity. 
You're allowed to submit up to `n` transactions for a given transaction type in a given epoch. Once `n` transactions have been included in a block, all further transactions for this type in the same epoch will be pre-rejected and not included in a block.

protocol/0062-SPAM-spam_protection.md
cdummett: The thinking was this threshold is specified as a quantum and $VEGA is fairly volatile in comparison to the current settlement assets on mainnet (so the assets quantum might not be accurate). The reason it is singled out specifically is because it is given out as a reward so will likely be in a lot of parties wallets.

I'm not attached to this restriction though, and thinking about it more it will just seem weird if more volatile assets do get proposed later down the line.

vega-paul (approved):

PR #1989, opened 2023-10-06
Modify liquidation mechanics

### LINKED ISSUE #1641: [Liquidation improvements] Completely decouple MTM, margin checks and closeout trades
The current design around margins means that:
- closeout trades happen (or are attempted) when a party is in breach of margin requirement and when there is enough volume and when we're not in an auction. 
- we cannot handle (future) frequent batch auctions
- we may reward a party that has a big position (MTM move against them, we want to close out but can't, MTM move for them they're OK again - the network was at risk but party kept profit). 
- the logic in core is tightly coupled in places making moves towards WASM based smart products harder. 

The aim here is:
1. Separate margin checks, mark-to-market and closeout trades into three separate streams that can run at arbitrary frequencies, 
1. Accommodate closeout trades during auctions,
1. Avoid allowing parties with huge position breaching margin level hanging around. 
1. Do this without changing any existing margin calculations. 
1. Do this without changing how loss socialisation works. 

Roughly how: 
- The insurance pool will become "margin account" for a "network party" (think of the network party as a "bad bank"). 
- We will regularly check whether a party is distressed (but doesn't need to be tied to orders / trades etc. only at desired frequency e.g. every block or every 5s). If a party is in breach of margin level their margin is transferred to insurance pool and their position to the "network party". (*)
- We do mark-to-market as usual. If the network party has a position the insurance pool is used for MTM flows. If insurance pool is empty we loss socialise as before. 
- The network party (insurance pool party) will employ a simple heuristic (which later can be WASM but for now we hard code) according to which it will attempt, at a desired intensity, to reduce it's position. 

Insurance party position reduction heuristic:
- if we're in continuous trading place market orders once every desired number of seconds with size the smaller of "half available book volume", "half the position to close out" (unless the position is exactly `1` in which case close out all. 
- if we're in an auction place a limit order - more thinking needed about the size / price; e.g. we don't want to trigger price monitoring or move price too much. 

(*) An added benefit of this is that we may avoid a bunch of close-out trades - imagine network party is long `10` but now it is assigned a position of a party that was short `8`; its new position is `2` ... and a bunch of network trades didn't need to happen.

witgaw:
* Liquidate distressed party's position immediately and move it to the network party.
* Market's insurance pool is market's party margin account.
* Introduce a simple inventory management strategy for the network party.

close https://github.com/vegaprotocol/specs/issues/1641


protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,48 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
-`Trader3 open orders:   0`
-
-#### STEP 1
+#### Liquidating distressed party's position in a market
 
-No traders are removed from the distressed trader list.
+- There is a market parameter `market.positionResolutionFrequency` controlling the maximum (the observed frequency may be slightly lower depending on block time) frequency of position resolution evaluations which needs to be specified when proposing a new market and can be updated any time as long as the market is not in a closed, terminated or settled [state](./0043-MKTL-market_lifecycle.md#market-lifecycle-statuses). The update takes effect immediately (if the paramer is decreased then the next position resolution evaluation time may happen sooner than it would've had the old parameter value been kept).
 
-#### STEP 2
+- Position evolution evaluation is never carried out during an auction. Leaving auction of any type (including opening auction) triggers position resolution evaluation and restarts the timer. Subsequent evaluations are triggered once the elapsed time since last evaluation is greater than `market.positionResolutionFrequency` or when market exits some other auction (whichever comes first).
```
davidsiska-vega: resolution?

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,48 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
 
-LiquiditySourcingTrade2: {
-  buyer: Trader5,
-  seller: Network,
-  size: 1,
-  price: 100,
-  type: 'liquidity-sourcing'
+  - their open volume get's added to network's open volume for that market,
+  - party's open volume in that market gets set to 0,
+  - the full amount of party's margin account balance gets forcibly transferred to market's insurance pool.
```
davidsiska-vega:
```suggestion
  - the full amount of party's margin account balance gets transferred to market's insurance pool.
```

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,48 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
 
-}
+This concludes the position resolution from party's perspective.
 
-```
+#### Managing network's position
 
-#### STEP 4
+Whilst network has a non-zero position in a given market it's treated as any other party with market's insurance in that market acting as its margin account and with and exception that margin search, margin release or liquidation are never attempted on a network party.
```
davidsiska-vega, line 47: We could calculate whether the network party is "distressed" or not and provide that on API... as a bit of info about market health.

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,48 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
 
-Close out trades are generated with the distressed traders
+Whenever party has a non-zero position it attempts to unload it using an [immediate or cancel](./0014-ORDT-order_types.md) limit order. If the the network has a `long` position it will submit a `sell` order. If it has a short position it will submit a buy order. The size of that order is chosen according to the liquidation strategy which forms a part of the marke's configuration. The strategy can be updated at any point whilst market is active with a market change [governance vote](./0028-GOVE-governance.md#2-change-market-parameters).
```
davidsiska-vega:
```suggestion
Whenever the network party has a non-zero position it attempts to unload it using an [immediate or cancel](./0014-ORDT-order_types.md) limit order. If the the network has a `long` position it will submit a `sell` order. If it has a short position it will submit a buy order. The size of that order is chosen according to the liquidation strategy which forms a part of the marke's configuration. The strategy can be updated at any point whilst market is active with a market change [governance vote](./0028-GOVE-governance.md#2-change-market-parameters).
```

davidsiska-vega, line 59: Limit order or market order? Or is that left up to the strategy? If it's a limit order it will need a price. I am still reading so I may come back and delete this comment.

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,48 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
-
-```json
-TransferRequest4 {
-  from: [MarketInsuranceAccount],
-  to:  Trader4_MarginAccount,
-  amount: (120 - PreviousMarkPrice) * -2, // this is the movement since the last settlement multiplied by the volume of the closed out amount
-}
-
-```
+Assume the price range implied by the `market.liquidity.priceRange` is `[a, b]`. Once the network has worked out a size of its immediate or cancel limit order it sets it's price to `a` if it's a sell order or `b` if it's a buy order, and it submits the order.
```
davidsiska-vega, line 58: Why not a market order?

witgaw: Just thought that would reduce the scope for manipulation of a network party or for it to make non-sensical trades.

protocol/0012-POSR-position_resolution.md
witgaw: Could do, but it should hopefully be trivial to get the margin level of the network party so could leave that up to end users to implement.

protocol/0012-POSR-position_resolution.md
witgaw: Fixed now, thanks.

davidsiska-vega (dismissed):

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,45 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
+Whilst network has a non-zero position in a given market it's treated as any other party with market's insurance in that market acting as its margin account and with and exception that margin search, margin release or liquidation are never attempted on a network party.
 
-LiquiditySourcingTrade1: {
-  buyer: Trader4,
-  seller: Network,
-  size: 2,
-  price: 120,
-  type: 'liquidity-sourcing'
-}
+Whenever the network party has a non-zero position it attempts to unload it using an [immediate or cancel](./0014-ORDT-order_types.md) limit order. If the the network has a `long` position it will submit a `sell` order. If it has a short position it will submit a buy order. The size of that order is chosen according to the liquidation strategy which forms a part of the marke's configuration. The strategy can be updated at any point whilst market is active with a market change [governance vote](./0028-GOVE-governance.md#2-change-market-parameters).
```
EVODelavega, line 49: I suspect this should be an IOC _market_ order, rather than a limit order

protocol/0012-POSR-position_resolution.md
EVODelavega: So liquidation strategy will be part of the market framework. Where should these paramters be specified, what validation should we have in place, etc...? Esentially: we need to update the 0001-MKTF-market_framework spec file for this I think

gordsport (commented):
@witgaw 

Are there any new ACs to be added to this PR for the changes?

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,45 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
-LiquiditySourcingTrade2: {
-  buyer: Trader5,
-  seller: Network,
-  size: 1,
-  price: 100,
-  type: 'liquidity-sourcing'
+The liquidation strategy consists of:
 
-}
+- `disposal time step` (mandatory, e.g. `n=10s`): network attempts to unload its position in a given market every time it goes out of auction and then every `n` seconds as long as market is not in auction mode and while the network's position is not equal to `0`,
```
davidsiska-vega: Minimum valid value is `1s` maximum valid value is `1h`. Default value is `10s`.

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,45 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
+- `disposal fraction` (e.g. `0.1`): fraction of network's current open volume that it will try to reduce in a single disposal attempt,
```
davidsiska-vega: minimum valid value is `0.01`, maximum valid value is `1`. Default value is `0.1`. Note that when determining the amount to be disposed of (i.e. `inventory x disposal fraction`) then we want to round up.

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,45 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
+- `full disposal size` (e.g. `20`): once net absolute value of network's open volume is at or below that value, the network will attempt to dispose the remaining amount in one go,
```
davidsiska-vega: minimum valid value is `0`, maximum valid value is biggest unsigned int. Default value `20`. Note that e.g. on a BTC/USD market with 4 position decimal places the value of `20` means `20 x 1e-4`.

protocol/0012-POSR-position_resolution.md
```diff
@@ -22,147 +22,45 @@ Any trader that has insufficient collateral to cover their margin liability is r
...
+- `max fraction of book side within liquidity bounds consumed` (e.g. `m=0.05`): once the network chooses the size of its order (`s_candidate`) the effective size will be calcualted as `s_effective=min(m*N, s_candidate)`, where `N` is the sum of volume (on the side of the book with which the network's order will be matching) that falls within the range implied by the `market.liquidity.priceRange` [parameter](./0044-LIME-lp_mechanics.md#market-parameters),
```
davidsiska-vega: minimum valid value is `0` - which means that the network party can never dispose of anything, maximum valid value is `1.0`, default value is `0.05`. Note that when doing `m x N` we should be rounding up.

protocol/0012-POSR-position_resolution.md
witgaw: The idea was that the limit order would be safer for the network as it'd be more difficult to manipulate things to that it trades at some extremely unfavourable price

witgaw:
Moved here: https://github.com/vegaprotocol/specs/pull/2040

PR #1991, opened 2023-10-11
Feat: remove liquidity monitoring auctions

### LINKED ISSUE #1996: Remove liquidity monitoring auctions
- Remove the existing functionality for liquidity monitoring auctions
- Ensure remaining LP mechanism works as it needs to
- Contingent on doing "Liquidation improvements"

Optional variation: 
- add market configuration that would disable liquidity monitoring auctions on a per market basis. The downside is that it doesn't simplify code, the upside is that we can experiment / see what there is demand for. 

Added bonus: 
- at the moment we want LP commitment changes to happen at epoch boundary but we have to treat increases in a special way so we don't stay stuck in a liquidity auction till the end of epoch. Ditching liquidity auctions would simplify this.

davidsiska-vega:
close https://github.com/vegaprotocol/specs/issues/1996


protocol/0026-AUCT-auctions.md
```diff
@@ -6,7 +6,8 @@ Auctions are a trading mode that 'collect' orders during an *auction call period
 
 In comparison to continuous trading, the auction mode for a market, is a state of the orderbook where each order placed is just sitting on the book, for a given period of time or until some requirements are met (called `call period`), then the matching orders are uncrossed.
 
-They are mostly useful in less liquid markets, or in specific scenarios where a price must be determined, i.e. at opening of a market, when a potentially excessively large price move might occur (price monitoring) or when liquidity needs to be sourced and aggregated (liquidity monitoring). In traditional markets (where markets open and close every day) we can run an open and closing auction for the price to stabilise at both ends.
+They are mostly useful in less liquid markets, or in specific scenarios where a price must be determined, i.e. at opening of a market, when a potentially excessively large price move might occur (price monitoring).
```
EVODelavega, line 9: Just a point of clarification: should we still have some form of liquidity monitoring during the opening auction period (ie extend the opening auction period if no liquidity for the market has been provided)?

EVODelavega:
Something that I don't see changed in this spec is the bond penalties. In the past, when a market was in liquidity auction, no bond penalties were applied, now that we don't have liquidity auctions anymore, do we want bond penalties to be applied during a price monitoring auction?

Jiajia-Cui:
> Something that I don't see changed in this spec is the bond penalties. In the past, when a market was in liquidity auction, no bond penalties were applied, now that we don't have liquidity auctions anymore, do we want bond penalties to be applied during a price monitoring auction?

we evaluate whether LPs meet the SLA during auctions (apart from opening auctions), so we should keep the bond penalties to be applied during price monitoring auctions.

EVODelavega:
> > Something that I don't see changed in this spec is the bond penalties. In the past, when a market was in liquidity auction, no bond penalties were applied, now that we don't have liquidity auctions anymore, do we want bond penalties to be applied during a price monitoring auction?
> 
> we evaluate whether LPs meet the SLA during auctions (apart from opening auctions), so we should keep the bond penalties to be applied during price monitoring auctions.

That's what is currently happening, so no worries there. The main question remaining here is whether it's OK for a market to leave opening auction without any provided liquidity

protocol/0026-AUCT-auctions.md
davidsiska-vega: Nope, because any LP who commits during the opening auction can then remove the liquidity (in the SLA world they'll just get some small penalty). So let's not bother at all.

EVODelavega:
While going through the list of system tests that had to be fixed/reworked for this change, here's a couple of AC's that we should remove from the spec, too:

* 0068-MATC-017 (market should enter auction if either side of the book is empty)
* 0043-MKTL-002 (6 -> market in liquidity monitoring auction no longer applies)

EVODelavega (dismissed):
LGTM - this has since been implemented (or un-implemented)

gordsport (approved):
Reapproving as last commit was to remove trailing whitespace for MD linting

PR #2001, opened 2023-10-16
Perps: modify funding payment calculation

witgaw:
* Modify TWAP calculations to exclude time spent in auctions


protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -83,62 +83,16 @@ Every time a [mark to market settlement](./0003-MTMK-mark_to_market_settlement.m
 
 When the `settlement_schedule` event is received we need to calculate the funding payment. Store the current vega time as `funding_period_end`.
 
-If there are no oracle data points with a timestamp less than `funding_period_end` available then funding payment is skipped and `funding_period_start` gets overwritten with `funding_period_end`.
+Skip the funding payment calculation (set payment to `0`) if any of the following conditions is met:
 
-If such points are available then the calculations discussed in the following subsections get executed and funding payments get exchanged.
+- no spot (external) data has been ingested since market was created,
+- market has been in auction throughout the entire funding period.
```
EVODelavega, line 89: Does this imply we should discard external data points if the market is still in opening auction, or simply skip the funding payment if the market is in opening auction? Do we still factor in the data for margin calculation?

witgaw: Let me know if the latest commit clarifies it.

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -83,62 +83,16 @@ Every time a [mark to market settlement](./0003-MTMK-mark_to_market_settlement.m
...
 
-#### TWAP spot price calculation
+Please refer to the following subsections for the details of calculation of the funding payment if none of the above conditions are met.
 
-Traverse all the available oracle data point tuples `(s,t)` and calculate the time-weighted average spot price (`s_twap`) as:
+#### TWAP calculation
 
-```go
```
davidsiska-vega, line 94: are we really deleting the calculation formula / code?

witgaw: That was the intention as it will be fairly convoluted to explain in pseduocode now and was hoping the description + detailed AC example would be enough, but can add it if you think otherwise.

witgaw: @davidsiska-vega @EVODelavega I've [reinstated the pseudocode](https://github.com/vegaprotocol/specs/pull/2001/commits/68155efe0faa89df76906bf558e9cb212cac4121). If you could give it a once over when you have a moment and approve if you're happy that'd be amazing :pray:

davidsiska-vega (dismissed):

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -83,63 +83,62 @@ Every time a [mark to market settlement](./0003-MTMK-mark_to_market_settlement.m
...
+#### TWAP calculation
+
+Same methodology applies to spot (external) and perps (internal). The available prices (spot and perps considered separately of course) should be used to calculate the time weighted average price within the funding period. If no observations at or prior to the funding period start exist then the start of the period used for calculation should be moved forward to the time of the first observation. An observation is assumed to be applying until a subsequent observation is available. Periods spent in auction should be excluded from the calculation. This implies that spot datapoints received during the auction except for the latest one should be disregarded. Please refer to the acceptance criteria for a detailed example.
+
+Calculation of the TWAP is carried out by maintaining the following variables: `numerator`, `denominator`, `previous_price` and `previous_time`. It's also assumed that a function `current_time()` is available which returns the current Vega time, and a function `in_auction()` exists which returns `true` if the market being considered is currently in auction and `false` otherwise.
```
wwestgarth, line 97: So there is one thing that adds a layer of complication to all of this, and thats being able to support spot prices coming in out-of-order (which could happen in the general case where someone might be submitting this data a transactions by hand, and not eth-oracles). 

The current core code can handle this by working out where the new-old-point should be, unwinding the TWAP back, adding the new point and and winding forward again.

To be able to do this with these auction changes as well, it gets just that little bit more complicated. So, is there any possibility of just rejecting any new points that arrive which are before a point we already have, just to simplify it all? Or it is important to support it?

witgaw: @wwestgarth that's a very good point, we still want to maintain that property, so I've got an [absolute cop-out](https://github.com/vegaprotocol/specs/pull/2001/commits/486b5a987104ced08c44a1d14fbebad1f2aaabe1) of a solution as far as specs go:

> Note that depending on what type of oracle is used for the spot price it may be that the oracle points only become known shortly before or at the funding payment cue time, so the above pseudocode is just an illustration of how these quantities should be calculated and the implementation will need to be able to apply such calculation retrospectively.

What do you think? Just not sure if complicating this further in the spec itself adds much value. Happy to help figure it out in the actual implementation by pairing or taking on a `core` task focused on that if that'd help at all.

wwestgarth: The cop out is fine 😄 I've got an idea in my head for how to do it, but it felt fiddly enough to ask whether it was even in scope to support, and maybe save me some effort!

wwestgarth (approved):

davidsiska-vega (approved):

PR #2005, opened 2023-10-16
chore: removed descoped features from cosmic branch

gordsport:
This PR removed the batch proposal and spot feature details in readiness to merge the cosmic elevator branch into master with the validator deployment of 0.73.x

All the changes removed here have been left in the `palazzo` branch (however spot may need to be bumped from that again)

I have left reference to spot in the perps spec as I was not 100% sure how to change this if indeed it was needed at all


Jiajia-Cui (approved):
LGTM

spot reference can stay in Perps as we need spot price for arbitrage the price difference between futures price in perps market and spot price from oracle

PR #2008, opened 2023-10-17
Make referrals spec line up with implementation

### LINKED ISSUE #2007: Fix referral spec in line with implementation
Minimum variables were not implemented as described

candida-d:
Closes #2007


candida-d:
wrong branch!

PR #2018, opened 2023-10-26
feat: align discount application with core implementation

cdummett:
Spec currently states all fee discounts are to be calculated using the original fee. Core implementation instead calculates and applies the fees one after the other.

PR aligns the spec with the core implementation and adds ACs to cover the cases where there are multiple discount sources active.

Core behaviour confirmed in https://github.com/vegaprotocol/vega/pull/9907.


protocol/0029-FEES-fees.md
```diff
@@ -25,46 +25,54 @@ Note that maker_fee = 0 if there is no maker, taker relationship between the tra
 
 Before fees are transferred, if there is an [active referral program](./0083-RFPR-on_chain_referral_program.md) or [volume discount program](./0085-VDPR-volume_discount_program.md), each parties fee components must be modified as follows.
 
-Note, it is important discounts are calculated and applied **before** rewards are calculated and applied.
+Note, discounts are calculated and applied one after the after and **before** rewards are calculated.
```
ValentinTrinque: Should it be ?

```suggestion
Note, discounts are calculated and applied one after the other and **before** rewards are calculated.
```

cdummett: Yes thank you, read my mind 😆

ze97286 (approved):

PR #2019, opened 2023-10-30
feat: automated market maker

### LINKED ISSUE #1990: vAMMs
# High level aims: 
- We want a party to be able to deposit capital, choose some parameters and open a perp (or future) market
- We don't want them to go bankrupt quickly every time

# High level "problems":
- there is only the settlement asset so we can't just pick up an off-the-shelf CFM curve (except we can more on that later, but it won't be the same)
- there is margin for the "order volume" that has to be calculated
- short positions have unlimited downside so the party that set this up can go bust very quickly if there is no position management 
- we have to make it work with SLA

# Use CFM curves: possible approach 1:
- The curve is parametric, in the simplest case, which we won't use is $xy = k$ where initially the party will choose what $x$ and $y$ they want to go with (but don't support it with assets, they just need to provide margin). If a trade occurs we'd update $x$ and $y$ appropriately
- The "parametric curve" code / module then has to provide the answers to following questions: 1) for a trade of size $S$ (integer multiple of minimum position size), what's the best buy price / sell price? 2) what is the volume at each price level between last trade price until the bound implied by the SLA parameter `lpPriceRange`. Clearly answers to 1) will be used for trade execution, answers to 2) will be used for margin calculation (order margin) and SLA evaluation. And arguably we could use 2) with VWAP calc to answer 1) as well... 
- An incoming trade (market order / limit order) should be evaluated against book and ever volume, if it's a limit order crossing any volume then it will be treated as market order. There will be routing logic which executes against the best liquidity source (without splitting or by splitting into various chunks, details TBD). 
- The "curve" should help with some position management which is the reason why we won't use $xy=k$. MVP would be to do $xy = k_b$ for incoming buy orders and $xy = k_s$ for incoming sell orders with $k_b$ and $k_s$ adjusted based on position size, so that "if I am already short buy orders get a bad price from me but sell orders get a good price from me". 
- We may want to allow Uniswap v3 type CFM curves. 
- There should be API which will allow UI to visualise the volume implied by all the passive liquidity together with the order book volume; the logic for generating the data should sit on the data node. 

# Don't use CMF curves, just automatically place volume
This uses the observation that up to discretization CFM curves and LOB volumes are the same thing and we already *have* the LOB. 
- Setup automated orders that will allow "less active" liquidity. *However* don't do it using pegged orders but instead just place limit orders on the book based on a simple parametric formula; place the volume relative to last traded price and min / max price implied by the SLA parameter `lpPriceRange`. 

The good thing about the 2nd approach is that it requires on routing or special API work.

Something to consider: allow market to leave opening auction by setting the "mark price" to the underlying oracle input if there were no crossing orders.

TomMcL:
close https://github.com/vegaprotocol/specs/issues/1990 .


protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,49 @@
...
+- **Volume at Upper Limit**: The volume the market maker will hit at the upper limit (this will be a short volume). Note that as the market maker is operating a constant function market curve there is an inherent link between traded price and position which allows this assertion. The combination of this volume and the range between `base price` and `upper price` will determine the volume placed at each price level inbetween.
+- **Volume at Lower Limit**: The volume the market maker will hit at the lower limit (this will be a long volume). Note that as the market maker is operating a constant function market curve there is an inherent link between traded price and position which allows this assertion. The combination of this volume and the range between `base price` and `lower price` will determine the volume placed at each price level inbetween.
+
+_To Determine_: Should tick spacing be customisable or enforced?
```
davidsiska-vega: As we're providing the volume as LOB volume it may make sense to have a market or network wide parameter specifying the maximum number of ticks the volume would be deployed on (as an upper bound on complexity). 

So that would potentially enforce tick spacing (up to some rounding weirdness).

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,49 @@
...
+
+Note that the independent long and short ranges mean that at `base price` the market maker will be flat with respect to the market with a `0` position. This means that a potential market maker with some inherent exposure elsewhere (likely long in many cases as a token holder) can generate a position which is always either opposite to their position elsewhere (with a capped size), thus offsetting pre-existing exposure, or zero.
```
witgaw: I don't understand this, there's nothing in the design guaranteeing that if the market is at this level when AMM commits, then the market diverges from it and returns the position will be `0`, right? Or am I missing something?

In other words, more than one AMM and not necessarily just AMMs can operate within a given market, right?

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,49 @@
...
+  - Maximum position notional/absolute value
+  - Price bounds
+  - Granularity of price levels to post
+- Once accepted, the network will begin to monitor the configured strategy on that key.
+- At each block, immediately prior to the evaluation of liquidity provision for SLA purposes, for each configured AMM:
+  - The party's available balance (including margin and general accounts) for trading on the market will be checked. If the total balance is `0` the AMM configuration will be cancelled. 
+  - Each running AMM will be queried for it's provided orders. All orders from this party on this market will then be cancelled, followed by the placement of these new orders on the book. There are a couple of things to note here:
+  - All orders are placed as-if arriving from the party's key externally.
+  - No bond is taken for the market maker, the party is responsible for ensuring the key remains sufficiently collateralised.
+  - No lock is placed on the key, the party may continue to sign and submit other trading transactions and orders. However, any orders resting on the book will be cancelled alongside others each time the AMM updates.
```
davidsiska-vega: Is this necessary? It may be for efficiency but would be nice to avoid.

TomMcL: To be honest I quite like the idea of pushing "this key is currently in automated trading mode, don't mess with it" as much as possible :D. But yes I'm mainly thinking from an efficiency point of view such that the core doesn't have to track which orders came from the market maker constantly.

protocol/0087-VAMM-automated_market_maker.md
TomMcL: So with the current formulation I *believe* you always should because the market maker is updating their mid-price based exclusively on their position (which is perhaps not entirely clear in the spec as-here so far), rather than anything regarding the market's mid price. So it should act as an AMM would, and more AMMs would just mean more stickiness at a given price before moving upwards/downwards.

witgaw: Ah fair enough, thanks

davidsiska-vega: >  because the market maker is updating their mid-price based exclusively on their position 
I wonder whether that's what we'd want to go with. 

Of course the orders the AMM will be placing could be "post-only" so even if it's price is different from midprice it won't place anything that trades aggresively with other volume. 

But still, as a user I may want to update my AMM "mid price" based on the mark price or market mid price. 

Could we use position instead of the base price?

TomMcL: Yeah my initial agent (just pushed to the branch in market sim) used last trade price as the "mid" rather than inferring it from position. The challenge then is that if the AMM *is* the only one on the market it feels pretty exploitable (and you also lose the position guarantees, which makes your risk much less bounded).

protocol/0087-VAMM-automated_market_maker.md
davidsiska-vega: We can ask core when it comes to this detail.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Process Overview
+
+The configuration and resultant lifecycle of an automated market maker is as follows:
+
+- Party funds a key which will be used by the strategy with desired token amounts.
+- Party submits a transaction containing configuration for the strategy on a given market. This will contain:
+  - Amount of funds to commit
+  - Price bounds (upper, lower, base)
+  - Granularity of price levels to post
```
davidsiska-vega: This is only relevant if we put volume directly on book, no? 

If we go via the route @barnabee was describing we'd register each source range with the order book and then when a price level is empty on the order book ask the AMM: how much volume at this price?

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,94 @@
...
+    - All orders are placed as-if arriving from the party's key externally.
+- If the party submits a `CancelAMM` transaction the AMM configuration for that party, on that market, will be cancelled. All active orders from the AMM will be cancelled and all funds and positions associated with the sub-account will be transferred back to the main account.
+
+## Sub-Account Configuration
+
+Each main Vega key will have one associated sub account for a given market, on which an AMM may be set up. The account key should be generated through a hash of the main account key plus the ID of the market to generate a valid Vega address in a predictable manner. Outside of the AMM framework the sub-accounts are treated identically to any other account, they will have the standard associated margin/general/bond accounts and be able to place orders if required as with any other account. The key differentiator is that no external party will have the private key to control these accounts directly. The maintenance of such an account will be performed through a few actions:
```
davidsiska-vega: It may be nice if I could control the account at least for purpose of posting market orders(*) to reduce position - but that's probably an unnecessary level of complexity. 

(*) of course this can only help if there is other volume on the book not just my amm volume

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,94 @@
...
+- Separate off-book liquidity
+  - Benefits
+    - Only requires updating or interacting with on the arrival of orders which may immediately trade, improving performance
+    - Acts as a more 'pure' AMM structure, immediately updating on each trade before interacting with the next
+  - Risks
+    - The AMMs would have to be each checked for every tick on the market, adding a performance impact to the choice of decimal places, or another parameter which needed to be set for tick size of AMMs on a market. It is possible this performance impact is significant.
+      - This also requires that core matching code outside the new AMM logic is aware of these changes and can handle them, potentially adding uncertainty to scope of change work.
+    - A "complete" order book picture would now require understanding the AMM presence at a given timepoint too. To build a full order book, calculations would have to be performed in downstream systems to expand the AMM into virtual orders, combining those with real orders resting on the book. Failure to do this correctly would result in an inaccurate picture of available liquidity.
```
davidsiska-vega: In particular the data node needs to accommodate this.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,94 @@
...
+    - Additional margin considerations for the AMM. When expanding out the AMM to virtual orders for a book, how does one ensure liquidity is not shown which would actually not be tradable with the funds available to that AMM.
```
davidsiska-vega: I think something like once per block or once per mark price update we'd have to calculate the virtual order volume at each price level and calculate the order margin based on that. Or we look if there is some easy approximation.

protocol/0087-VAMM-automated_market_maker.md
TomMcL: Would have to change the way the bot works entirely for that to make sense though. A uniswap-esque liquidity pool's price is ultimately determined by it's position, so reducing position would just change it's idea of what the price should be and it would be back where it was seconds later :D. One proviso though is what should happen when someone changes the bounds, as then they probably do want to change position more aggressively (at the least with not post-only orders)

protocol/0087-VAMM-automated_market_maker.md
TomMcL: Yes, some more of this spec needs to be killed/reworded if we end up in the off-book AMM world

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,165 @@
...
+
+  1. Request from the AMM it's current `fair price` by requesting a quote of volume `0`
+  1. Request from the AMM the total trade volume required to move between `fair price` and `mark price`. Note that if `mark price` > `base price` this should be negative (i.e. the AMM will be selling) and it `mark price` < `base price` it should be positive (i.e. the AMM will be buying).
+     1. If no upper(/lower) price is specified then if the price is above(/below) the mid the AMM should return `0` volume
+  1. Create a limit order with `FOK` execution at a price equal to `mark price` shifted by the AMM's `max slippage %` for this volume. If the limit order would not trade, reject the entire creation/amendment as `CANNOT_REBASE_SLIPPAGE_BEYOND_LIMITS`.
```
davidsiska-vega: Is `max slippage %` the same as `slippage_tolerance`?

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,165 @@
...
+
+  1. First, identify the current position, `p`. If it is `0` then the current fair price is the base price, and we are done.
+  1. If `P > 0`:
+     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{L}{\sqrt{p_l}}$, where $L$ is the value for the lower range, $P$ is the market position and $p_l$ is the `base price`.
+     1. The virtual `y` of the position can be calculated as $y_v = \frac{c_c}{ (f_s + f_l) \cdotp f_i} + L \cdotp \sqrt{p_l}$ where $c_c$ is the current total dollar balance of the AMM across margin and general accounts and `p_l` is the `lower price`. Other variables are as defined above.
+  1. If `P < 0`:
+     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{c}{ p_u \cdotp (f_s + f_l) \cdotp f_i} + \frac{L}{\sqrt{p_l}}$ where $p_l$ is the `base price` and `p_u` is the `upper price`.
+     1. The virtual `y` can be calculated as $v_y = abs(P) * p_e + L * p_l$ where $p_e$ is the average entry price of the position and $p_l$ is the `base price`
```
davidsiska-vega: let's stick to $\cdot$ or simply nothing for multiplication in formulae

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,165 @@
...
+
+  1. First, identify the current position, `p`. If it is `0` then the current fair price is the base price, and we are done.
+  1. If `P > 0`:
+     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{L}{\sqrt{p_l}}$, where $L$ is the value for the lower range, $P$ is the market position and $p_l$ is the `base price`.
+     1. The virtual `y` of the position can be calculated as $y_v = \frac{c_c}{ (f_s + f_l) \cdotp f_i} + L \cdotp \sqrt{p_l}$ where $c_c$ is the current total dollar balance of the AMM across margin and general accounts and `p_l` is the `lower price`. Other variables are as defined above.
+  1. If `P < 0`:
+     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{c}{ p_u \cdotp (f_s + f_l) \cdotp f_i} + \frac{L}{\sqrt{p_l}}$ where $p_l$ is the `base price` and `p_u` is the `upper price`.
+     1. The virtual `y` can be calculated as $v_y = abs(P) * p_e + L * p_l$ where $p_e$ is the average entry price of the position and $p_l$ is the `base price` 
+  1. Now the `fair` price is simply `v_y / v_x`
```
davidsiska-vega: $v_y$ or $y_v$? Above we calculate $y_v$... in the formula it's $v_y$.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,165 @@
...
+  1. If `P < 0`:
+     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{c}{ p_u \cdotp (f_s + f_l) \cdotp f_i} + \frac{L}{\sqrt{p_l}}$ where $p_l$ is the `base price` and `p_u` is the `upper price`.
+     1. The virtual `y` can be calculated as $v_y = abs(P) * p_e + L * p_l$ where $p_e$ is the average entry price of the position and $p_l$ is the `base price` 
+  1. Now the `fair` price is simply `v_y / v_x` 
+
+#### Volume between two prices
+
+For the second interface we need to calculate the volume which would be posted to the book between two price levels. In order to calculate this for an AMM we are ultimately asking the question "what volume of swap would cause the fair price to move from price A to price B?"
+
+To calculate this, the interface will need the `starting price` $p_s$, `ending price` $p_e$, `upper price of the current range` $p_u$ (`upper price` if `P < 0` else `lower price`) and the `L` value for the current range. At `P = 0` use the values for the range which the volume change will cause the position to move into.
```
davidsiska-vega: else `lower_price` or `base_price`?

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,165 @@
...
+     1. Check all active AMMs, querying their quote price API with the smallest trade unit on the market in the direction of trading (if the incoming order is a `buy`, query the AMM's `ask`, or vice versa). Retain those where this price < `outer price`
+     1. Within these, select either the minimum `upper price` (if the incoming order is a buy) or the maximum `lower price` (if the incoming order is a sell), call this `amm bound price`. This is the range where all of these AMMs are active. Finally, select either the minimum (for a buy) or maximum (for a sell) between `amm bound price` and `outer price`. From this form an interval `current price, outer price`. 
+     1. Now, for each AMM within this range, calculate the volume of trading required to move each from the `current price` to the `outer price`. Call the sum of this volume `total volume`.
+     1. If `remaining volume <= total volume` split trades between the AMMs according to their proportional contribution to `total volume` (e.g. larger liquidity receives a higher proportion of the trade). This ensures their mid prices will move equally (TODO: Is trade splitting more involved than this?).
```
davidsiska-vega: we should probably mention the `maker_fee` going to the respective AMMs for the volume they contributed.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,165 @@
...
+The configuration and resultant lifecycle of an automated market maker is as follows:
+
+- Party funds a key which will be used by the strategy with desired token amounts.
+- Party submits a transaction containing configuration for the strategy on a given market. This will contain:
+  - Amount of funds to commit
+  - Price bounds (upper, lower, base)
+- Additionally, the transaction should contain data related to the setup of the position but which does not need to be stored:
+  - Maximum slippage (%), used for rebasing position when creating/updating AMM
+- Once accepted, the network will transfer funds to a sub-account and use the other parameters for maintaining the position.
+- At each block, the party's available balance (including margin and general accounts) for trading on the market will be checked. If the total balance is `0` the AMM configuration will be cancelled.
```
jeremyletang: Not a big thing, but usually when the network initiate cancelling something, we call it stopped, when the users does we call it cancellation, and map it to appropriate status (e.g: active, rejected (at submission), cancelled (by user), stopped (by the network).

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,165 @@
...
+
+- **Base Price**: The base price is the central price for the market maker. When trading at this level the market maker will have a position of `0`. Volumes for prices above this level will be taken from the `upper` curve and volumes for prices below will be taken from the `lower` curve.
+- **Upper Price**: The maximum price bound for market making. Prices between the `base price` and this price will have volume placed, with no orders above this price. This is optional and if not supplied no volume will be placed above `base price`. At these prices the market maker will always be short
+- **Lower Price**: The minimum price bound for market making. Prices between the `base price` and this will have volume placed, with no orders below this price. This is optional and if not supplied no volume will be placed below `base price`. At these prices the market maker will always be long
+- **Commitment**: This is the initial volume of funds to transfer into the sub account for use in market making. If this amount is not currently available in the main account's general account the transaction will fail.
+
```
davidsiska-vega: Allow specification of "leverage at bounds" which will be reflected in the "liquidity" $L$ you're implying.

protocol/0087-VAMM-automated_market_maker.md
TomMcL: :+1:

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,177 @@
...
+- Creation: A sub-account will be funded when a user configures an AMM strategy with a set of criteria and a commitment amount. At this point in time the commitment amount will be transferred to the sub-account's general account and the AMM strategy will commence
+- Cancellation: When the AMM is cancelled all active orders are first cancelled. Following this cancellation, all funds in the sub-account's margin account should be transferred to the associated main account's margin account, with the same then happening for funds in the general account. Finally, any associated non-zero position on the market should be reassigned from the sub-account to the main account. At this point processing can continue, allowing the standard margining cycle to perform any required transfers from margin to general account.
```
jeremyletang: `When the AMM is cancelled all active orders are first cancelled.` I think this is now outdated and should be removed.

jeremyletang: `Finally, any associated non-zero position on the market should be reassigned from the sub-account to the main account.` I'm not sure this is right, we should probably just try to close the position, if it's possible then the pool can be cancelled, if not we close as much of the position as we can, and leave the pool open.

TomMcL: The challenge I see with that would be that a large LP could end up with a pretty substantial position compared to the market, which they'd have to either close all at once and eat a load of slippage or manually bring down their base position towards the current price over time, which is a bit of a pain.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,177 @@
...
+- At each block, the party's available balance (including margin and general accounts) for trading on the market will be checked. If the total balance is `0` the AMM configuration will be cancelled. 
+- If the party submits a `CancelAMM` transaction the AMM configuration for that party, on that market, will be cancelled. All active orders from the AMM will be cancelled and all funds and positions associated with the sub-account will be transferred back to the main account.
+
+## Sub-Account Configuration
+
+Each main Vega key will have one associated sub account for a given market, on which an AMM may be set up. The account key should be generated through a hash of the main account key plus the ID of the market to generate a valid Vega address in a predictable manner. Outside of the AMM framework the sub-accounts are treated identically to any other account, they will have the standard associated margin/general/bond accounts and be able to place orders if required as with any other account. The key differentiator is that no external party will have the private key to control these accounts directly. The maintenance of such an account will be performed through a few actions:
```
jeremyletang: This bit mention the bond account here, do we still need those?

protocol/0087-VAMM-automated_market_maker.md
davidsiska-vega: I certainly don't love the sound of "if the position can't be closed we leave the pool open and you can't exit".  

So I am leaning towards: if the user wishes to cancel their AMM then their position is dumped on their main key and the AMM is removed. 

The UI can offer to submit a market order for them to close the position (if there is other volume available).

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,177 @@
...
+- **Lower Price**: The minimum price bound for market making. Prices between the `base price` and this will have volume placed, with no orders below this price. This is optional and if not supplied no volume will be placed below `base price`. At these prices the market maker will always be long
+- **Commitment**: This is the initial volume of funds to transfer into the sub account for use in market making. If this amount is not currently available in the main account's general account the transaction will fail.
+- **Margin Ratio at Bounds**: The exact volume scaling is defined by the position at the upper and lower prices. To determine this we must compare the commitment with what leverage that might allow at the price bounds. One way to do this is to assume we will use the value at which `commitment == initial margin` for the position at that price, however users may wish to take a more conservative approach. Using this parameter allows them to set a value such that `position = commitment / margin ratio`, however with the restriction that commitment must still be `>= initial margin`. This parameter should be optional.
```
davidsiska-vega: Should we have different ratios at the upper and lower bounds?

davidsiska-vega: For consistency let's also write `margin_ratio_at_bounds` here.

TomMcL: Done

protocol/0087-VAMM-automated_market_maker.md
TomMcL: After discussing with @jeremyletang it sounds like this transfer of positions would be pretty painful to implement, so it may be easiest to say if you don't want to eat the slippage you just have size down bit-by-bit

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,197 @@
...
+
+where $v_{worst}$ is as above, $p_u$ is the price at the top of the range (`upper price` for the upper range and `base price` for the lower range) and $p_l$ is the price at the bottom of the range (`base price` for the lower range and `lower price` for the lower range). This gives the two `L` values for the two ranges.
+
+#### Fair price
+
+From here the first step is calculating a `fair` price, which can be done by utilising the `L` value for the respective range to calculate `virtual` values for the pool balances. From here on `y` will be the cash balance of the pool and `x` the position.
+
+  1. First, identify the current position, `p`. If it is `0` then the current fair price is the base price.
+  2. If `P > 0`:
+     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{L}{\sqrt{p_l}}$, where $L$ is the value for the lower range, $P$ is the market position and $p_l$ is the `base price`.
```
wwestgarth: Minor: `p_l` is defined as the base price, here, and then as the lower price just below in the next line, so can we just change it to `p_u` here?
```suggestion
     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{L}{\sqrt{p_u}}$, where $L$ is the value for the lower range, $P$ is the market position and $p_u$ is the `base price`.
```

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,197 @@
...
+     2. The virtual `y` of the position can be calculated as $y_v = c_c \cdotp r_f + L \cdotp \sqrt{p_l}$ where $c_c$ is the current total dollar balance of the AMM across margin and general accounts and `p_l` is the `lower price`. Other variables are as defined above.
+  3. If `P < 0`:
+     1. The virtual `x` of the position can be calculated as $x_v = P + \frac{c}{p_u} \cdotp r_f + \frac{L}{\sqrt{p_l}}$ where $p_l$ is the `base price` and `p_u` is the `upper price`.
```
wwestgarth: Just confirming whether this `c` here is supposed to be `c_c` as above (current total balance), or if it is actually referring to the `c` for the initial committment all the way at the top of the file?

protocol/0087-VAMM-automated_market_maker.ipynb
davidsiska-vega, line 1: I think this
```
def initialise(
        self,
        vega,
    ):
        risk_factors = vega.get_risk_factors(self.market_id)
        self.short_factor, self.long_factor = risk_factors.short, risk_factors.long
``` 

can be removed.

davidsiska-vega, line 1: And maybe this too?
```
def _generate_shape(
        self, bid_price_depth: float, ask_price_depth: float
    ) -> Tuple[List[MMOrder], List[MMOrder]]:
    ... 
    ...
```

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,268 @@
...
+
+
+- If a vAMM is cancelled and set in `Reduce-Only` mode when it is currently short, then: (<a name="0087-VAMM-021" href="#0087-VAMM-021">0087-VAMM-021</a>)
+  - It creates no further sell orders even if the current price is below the configured upper price.
+  - When one of it's buy orders is executed it still does not produce sell orders, and correctly quotes buy orders from a lower price.
+  - When the position reaches `0` the vAMM is closed and all funds are released to the user after the next mark to market.
+
+- If a vAMM is cancelled and set in `Reduce-Only` mode when it currently has no position then all funds are released after the next mark to market. (<a name="0087-VAMM-022" href="#0087-VAMM-022">0087-VAMM-022</a>)
+
+- If a vAMM is cancelled and set into `Reduce-Only` mode, then an amend is sent by the user who created it, the vAMM is amended according to those instructions and is moved out of `Reduce-Only` mode back into normal operation. (<a name="0087-VAMM-023" href="#0087-VAMM-023">0087-VAMM-023</a>)
```
wwestgarth: A couple of things that have popped into mind that we will probably want AC's for:
- Amending a vAMM such that when it sends in an order to "pull" it into position, check that it doesn't try to trade with itself i.e the matching process excludes the vAMM being amended.
- If someone uses a key to create a vAMM then uses the same key to submits a normal limit order like a normal person, should it be a considered a wash trade if it trades with the vAMM?

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,288 @@
...
+        res += i;
+        if(((1LL << res) - 1) & num)
+            res -= i;
+    }
+    return res;
+}
+```
+
+
+## Acceptance Criteria
```
wwestgarth: Can we have a section/AC for what should happen when a market is settled/closed. I'm assuming we will settle like normal and then transfer all assets from the subaccount back to the main account when close the market.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,288 @@
...
+As the computation of this virtual order shape may be heavy when run across a large number of passive AMMs the number of AMMs updated per block should be throttled to a fixed maximum number, updating on a rolling frequency, or when updated/first created.
+
+A given AMM's average liquidity score across the epoch should also be tracked, giving a time-weighted average at the end of each epoch (including `0` values for any time when the AMM either did not exist or was not providing liquidity on one side of the book). From this, a virtual stake amount can be calculated by dividing through by the `market.liquidity.stakeToCcyVolume` value and the AMM key'
+s ELS updated as normal.
+
+## Setting Fees
+
+The `proposed_fee` provided as part of the AMM construction contributes to the fee determination logic on the market, if a setup where LPs decide on the market fee is in use. In the case where it is the AMM's current assigned ELS, or the running average liquidity provided so far if the commitment was made in the current epoch, is used for weighting the AMM's vote for the fee.
+
+## Calculating Square Roots
```
davidsiska-vega: I think this can go, no? We're not using it AFAIK.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,288 @@
...
+  - If other traders trade to move the market mid price to `90` and then in one trade move the mid price to `110` then trade to move the mid price to `120` the vAMM will have a larger (more negative) but comparable position to if they had been moved straight from `100` to `120`. (<a name="0087-VAMM-014" href="#0087-VAMM-014">0087-VAMM-014</a>)
+  
+- A vAMM which has been created and is active contributes with it's proposed fee level to the active fee setting mechanism. (<a name="0087-VAMM-015" href="#0087-VAMM-015">0087-VAMM-015</a>)
+- A vAMM's virtual ELS should be equal to the ELS of a regular LP with the same committed volume on the book (i.e. if a vAMM has an average volume on each side of the book across the epoch of 10k USDT, their ELS should be equal to that of a regular LP who has a commitment which requires supplying 10k USDT who joined at the same time as them). (<a name="0087-VAMM-016" href="#0087-VAMM-016">0087-VAMM-016</a>)
+  - A vAMM's virtual ELS should grow at the same rate as a full LP's ELS who joined at the same time. (<a name="0087-VAMM-017" href="#0087-VAMM-017">0087-VAMM-017</a>)
```
EVODelavega: @TomMcL @wwestgarth after we discussed the scoring and ELS for vAMMs, we agreed that a vAMM would not have an ELS until the next epoch. These two ACs  (16 and 17) either need to be updated, or we need to re-evaluate the current implementation.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,306 @@
...
+The configuration and resultant lifecycle of an automated market maker is as follows:
+
+- Party funds a key which will be used by the strategy with desired token amounts.
+- Party submits a transaction containing configuration for the strategy on a given market. This will contain:
+  - Amount of funds to commit
+  - Price bounds (upper, lower, base)
+  - Margin ratios at upper and lower bound (ratio for leverage at bounds. Reciprocal of leverage multiplier e.g. 0.1 = 10x leverage)
+- Once accepted, the network will transfer funds to a sub-account and use the other parameters for maintaining the position.
+- At each block, the party's available balance (including margin and general accounts) for trading on the market will be checked. If the total balance is `0` the AMM configuration will be stopped.
+- If the party submits a `CancelAMM` transaction the AMM configuration for that party, on that market, will be cancelled. All active orders from the AMM will be cancelled and all funds and positions associated with the sub-account will be transferred back to the main account.
```
candida-d: > all funds **and positions** associated with the sub-account will be transferred back to the main account.

This isn't how it's being implemented, as I understand it - does this need an update?

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,306 @@
...
+
+- Creation: A sub-account will be funded when a user configures an AMM strategy with a set of criteria and a commitment amount. At this point in time the commitment amount will be transferred to the sub-account's general account and the AMM strategy will commence
+- Cancellation: When the AMM is cancelled all funds in the sub-account's margin account should be transferred to the associated main account's margin account, with the same then happening for funds in the general account. Finally, any associated non-zero position on the market should be reassigned from the sub-account to the main account. At this point processing can continue, allowing the standard margining cycle to perform any required transfers from margin to general account.
```
candida-d: > Finally, any associated non-zero position on the market should be reassigned from the sub-account to the main account.

Same comment as above about implementation of cancelling AMMs.

protocol/0087-VAMM-automated_market_maker.md
```diff
@@ -0,0 +1,306 @@
...
+   - If the AMM is currently short, the `lower bound` is removed
+   - If the AMM is currently long, the `upper bound` is removed
+   - The `upper`/`lower` bound (if the AMM is currently short/long) is then set to the AMM's current `fair price`. In this mode the AMM should only ever quote on the side which will reduce it's position (it's `upper`/`lower` bound should always be equal to the current `fair price` belief).
+   - Once the position reaches `0` the AMM can be cancelled and all funds in the general account can be returned to the creating party
+   - This acts similarly to the mode when an AMM is synchronising, except that the position will be closed in pieces as the price moves towards the `base price` rather than all at once at the nearest price.
+
+Note that, whilst an `Abandon Position` transaction immediately closes the AMM a `Reduce-Only` transaction will keep it active for an uncertain amount of time into the future. During this time, any Amendment transactions received should move the AMM out of `Reduce-Only` mode and back into standard continuous operation.
+
```
candida-d: The implementation allows going from `Abandon Position` to `Reduce-Only`, which would be nice to reference in the spec.

protocol/0087-VAMM-automated_market_maker.md
TomMcL: 👍

protocol/0087-VAMM-automated_market_maker.md
TomMcL: I'm not sure how that would be possible given `Abandon Position` should totally close the position?

candida-d: Sorry @TomMcL, I meant the other way around, from reduce to abandon.

davidsiska-vega (approved):

PR #2020, opened 2023-10-30
feat: fixed size order amendments

### LINKED ISSUE #1998: Order amendment improvements
- See feedback below
- Also requested an "amend ID"

Image

cdummett:
Currently order amendments amending the size of an order require a size delta to be specified. This causes issues when a user wants to amend an order to a specific amount but the remaining part of their order changes in between the amendment being submitted and committed in a block. 

Spec PR changes order amendments to specify a new size. Both the size and remaining part of the order should then be updated accordingly. Note, time priority behaviour is unchanged, i.e. an order reducing the size maintains time priority and an order increasing its size loses time priority. 

closes https://github.com/vegaprotocol/specs/issues/1998 .


Jiajia-Cui (approved):
LGTM

Issue #1999, opened 2023-10-16
Teams improvements


gordsport:
- Fully separate teams and referral sets
- Allow setting a key name / metadata (to show in a team)
- Invitations to closed teams (owner maintains a whitelist who can join)

davidsiska-vega:
Closed with PR #2058

PR #2030, opened 2023-11-02
chore: tidy up the features.json for palazzo milestone

gordsport:
This archives off the cosmic features.json file and tidies up current one ready for the addition of more palazzo ACs

- The tidied features.json file will be used with this [PR](https://github.com/vegaprotocol/jenkins-shared-library/pull/703) to get metrics on coverage
- the cosmic-features.json is purely a record of what was there at this time. It could be removed when we are sure that we do not need the information


protocol/cosmic-features.json
```diff
@@ -0,0 +1,927 @@
...
+      "0056-REWA-101",
+      "0056-REWA-102",
+      "0056-REWA-103",
+      "0056-REWA-104",
+      "0056-REWA-105",
+      "0056-REWA-106",
+      "0056-REWA-107",
+      "0056-REWA-108"
+    ]
+  },
```
cdummett, line 534: Possibly these should be removed as they are disabled for cosmic?

Or do we include as they are implemented?

protocol/cosmic-features.json
```diff
@@ -0,0 +1,927 @@
...
+    "milestone": "deployment-3",
+    "acs": [
+      "0028-GOVE-145",
+      "0028-GOVE-146",
+      "0028-GOVE-147",
+      "0028-GOVE-148",
+      "0028-GOVE-149"
+    ]
+  },
+  "Spot": {
```
cdummett, line 571: Not sure this Spot section should be here.

protocol/cosmic-features.json
```diff
@@ -0,0 +1,927 @@
...
+      "0074-BTCH-015",
+      "0074-BTCH-016",
+      "0074-BTCH-019",
+      "0079-TGAP-006",
+      "0079-TGAP-007",
+      "0081-SUCM-033",
+      "0081-SUCM-034"
+    ]
+  },
+  "Fixed Size Amendments": {
```
cdummett, line 915: Not a cosmic feature, should probably be removed.

cdummett (commented):
Few small comments re features which were deferred / disabled / not implemented for cosmic.

protocol/cosmic-features.json
gordsport: I thiink we should keep this in as we may add more Teams ACs under: https://github.com/vegaprotocol/specs/issues/1999

The feature will be enabled for palazzo milestone

protocol/cosmic-features.json
gordsport: I left this in so we had a record of all the spot ACs for when spot gets prioritised.

protocol/cosmic-features.json
gordsport: Happy to remove, I just archived the old features.json as it was so we have a record of how it looked.

cdummett (approved):

PR #2031, opened 2023-11-02
feat: support both size amend methods

cdummett:
#2020 added a method for amending the size of an order but was ambiguous re breaking changes / deprecation.

PR tidies up the amend spec so it is clear an order can be amended by either:

- specifying a `sizeDelta`
- specifying a new desired `size`


protocol/0004-AMND-amends.md
```diff
@@ -42,7 +45,7 @@ The amend order can alter the quantity, price and expiry time/`TIF` type. For pe
 
 ## Guide-level explanation
 
-The amend order message is a custom message containing the `orderID` of the original order and optional fields that can be altered. Prices and sizes can be changed with new absolute values. Expiry time can be set to a new value and the `TIF` type can be toggled between `GTC` and `GTT`. Changing the `TIF` field will impact the value in the `ExpiryTime` field as it will either be blanked or set to a new valid value.
+The amend order message is a custom message containing the `orderID` of the original order and optional fields that can be altered. Prices can be changed with a new absolute value, sizes can be changed by either specifying a size delta or a new absolute value. Expiry time can be set to a new value and the `TIF` type can be toggled between `GTC` and `GTT`. Changing the `TIF` field will impact the value in the `ExpiryTime` field as it will either be blanked or set to a new valid value.
```
davidsiska-vega: Should we be more explicit that if both are specified then amend gets rejected?

davidsiska-vega (dismissed):

protocol/0004-AMND-amends.md
cdummett: Yep, good point thanks, needs an AC as well.

davidsiska-vega (approved):

PR #2036, opened 2023-11-03
Remove banning ACs from the spam spec

### LINKED ISSUE #2035: Removal of spam ACs detailing banning as ComentBFT 0.38-> will not support this.
With the introduction of cometBFT 0.38 into core (see below issue) there will no longer be banning. Instead TXs that would have previously been post-rejected will not be included in the blocks.

- https://github.com/vegaprotocol/vega/issues/8051

In light of this we need to:

- Remove ACs referencing banning
- See if new ACs covering the new functionality can be crafted (if indeed this can be tested)

gordsport:
close https://github.com/vegaprotocol/specs/issues/2035 .

DO NOT MERGE - until https://github.com/vegaprotocol/vega/issues/8051 is confirmed to be in palazzo milestone version.


gordsport:
@Vegaklaus 

Please can you take a look at the paragraphs on lines:

- 51
- 54
- 97

As these give details about spamming. If the core side upgrade to CometBFT 0.38 goes into the palazzo milestone we will need to update this spec.

PR #2037, opened 2023-11-03
Update 0062-SPAM-spam_protection.md

### LINKED ISSUE #2035: Removal of spam ACs detailing banning as ComentBFT 0.38-> will not support this.

Vegaklaus:
Removed banning from anti-spam


gordsport (changes requested):
@Vegaklaus I believe this AC will need to be removed as well [0072-SPPW-006](https://github.com/vegaprotocol/specs/blob/Vegaklaus-patch-7/protocol/0072-SPPW-spam-protection-PoW.md#0072-SPPW-006)

Please can you take a look at that spec too and make sure not content needs changing too?

gordsport (approved):

PR #2040, opened 2023-11-06
Modify liquidation mechanics

### LINKED ISSUE #1641: [Liquidation improvements] Completely decouple MTM, margin checks and closeout trades

witgaw:
- Liquidate distressed party's position immediately and move it to the network party.
- Market's insurance pool is market's party margin account.
- Introduce a simple inventory management strategy for the network party.

close https://github.com/vegaprotocol/specs/issues/1641

### DIFF
--- a/protocol/0012-POSR-position_resolution.md
+++ b/protocol/0012-POSR-position_resolution.md
@@ -67,14 +67,23 @@ Whenever the network party has a non-zero position it attempts to unload it usin
 
 Currently only one liquidation strategy is supported and its defined by the following parameters:
 
-- `disposal time step` (min: `1s`, max: `1h`, default: `n=10s`): network attempts to unload its position in a given market every time it goes out of auction and then every `n` seconds as long as market is not in auction mode and while the network's position is not equal to `0`,
-- `disposal fraction` (min: `0.01`, max: `1`, default: `0.1`): fraction of network's current open volume that it will try to reduce in a single disposal attempt,
-- `full disposal size` (min: `0`, max: `max int`, default: `20`): once net absolute value of network's open volume is at or below that value, the network will attempt to dispose the remaining amount in one go,
-- `max fraction of book side within liquidity bounds consumed` (min: `0`, max: `1`, default: `m=0.05`): once the network chooses the size of its order (`s_candidate`) the effective size will be calculated as `s_effective=min(m*N, s_candidate)`, where `N` is the sum of volume (on the side of the book with which the network's order will be matching) that falls within the range implied by the `market.liquidity.priceRange` [parameter](./0044-LIME-lp_mechanics.md#market-parameters),
+- `disposal time step` (min: `1s`, max: `1h`): network attempts to unload its position in a given market every time it goes out of auction and then every `disposal time step` seconds as long as market is not in auction mode and while the network's position is not equal to `0`,
+- `disposal fraction` (min: `0.01`, max: `1`): fraction of network's current open volume that it will try to reduce in a single disposal attempt,
+- `full disposal size` (min: `0`, max: `max int`): once net absolute value of network's open volume is at or below that value, the network will attempt to dispose the remaining amount in one go,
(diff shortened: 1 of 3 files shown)


davidsiska-vega (dismissed):

gordsport:
The core implementation issue is:

- https://github.com/vegaprotocol/vega/issues/9945

protocol/0012-POSR-position_resolution.md
```diff
@@ -2,167 +2,82 @@
...
+Whenever the network party has a non-zero position it attempts to unload it using an [immediate or cancel](./0014-ORDT-order_types.md) limit order. If the the network has a `long` position it will submit a `sell` order. If it has a short position it will submit a buy order. The size of that order is chosen according to the liquidation strategy which forms a part of the market's configuration. The strategy can be updated at any point whilst market is active with a market change [governance vote](./0028-GOVE-governance.md#2-change-market-parameters).
 
-`// OpenPosition of Network =  -3 +5 -4 +2 = 0`
-`// OpenPosition of Trader1 =  +5 -5 = 0`
-`// OpenPosition of Trader2 = -4 +4 =  0`
-`// OpenPosition of Trader3 =  +2 - 2 = 0`
+Currently only one liquidation strategy is supported and its defined by the following parameters:
 
-#### STEP 5
+- `disposal time step` (min: `1s`, max: `1h`, default: `n=10s`): network attempts to unload its position in a given market every time it goes out of auction and then every `n` seconds as long as market is not in auction mode and while the network's position is not equal to `0`,
```
EVODelavega: Seeing as these are parameters that can be specific to each market, but we want to have default values, should the defaults be hard-coded, or set as network parameters? Can those network parameters be changed via governance, or not?

witgaw: Let's hardcore them, the netparam default for a market parameter ends up just confusing everyone. I'll mention that in the spec.

EVODelavega: @barnabee wonderiing what you make of default values in proposals here

barnabee: I'd suggest no defaults. Make them specify it on the proposal.

jeremyletang: what about market already existing after a protocol upgrade?

EVODelavega: Defaults in protocol upgrade are fine, but that distinction should be made clearer probably -> proposals _must_ specify all 4 parameters, existing markets will default to these default params, which can be updated through governance after protocol upgrade

witgaw: OK, let me know how it's looking now.

EVODelavega (approved):

davidsiska-vega (approved):

PR #2038, opened 2023-11-05
Transfer fee

Jiajia-Cui:
closes [#1995 ]

PR #2041, opened 2023-11-07
chore: add ACs to features.json from transfer fee improvements

gordsport:
This adds the ACs from:

- https://github.com/vegaprotocol/specs/pull/2038


Jiajia-Cui (approved):
LGTM

PR #2043, opened 2023-11-07
feat: stop order position linked

### LINKED ISSUE #2025: Position/order linked stop orders
- Attach stop to order (volume “activates” as order trades, unactivated is cancelled if order no longer on book or parked)
- Attach stop to position (volume grows/shrinks with position, cancelled if position 0 or changes direction)

TomMcL:


protocol/0014-ORDT-order_types.md
```diff
@@ -71,10 +71,16 @@ If it has an expiry then it can be set either to cancel on expiry (i.e. it is de
 An OCO contains TWO stop order submissions, and must include one in each trigger direction.
 OCOs work exactly like two separate stop orders except that if one of the pair is triggered, cancelled, deleted, or rejected, the other one is automatically cancelled.
 An OCO submission allows a user to have a stop loss and take profit applied to the same amount of their position without the risk of both trading and reducing their position by more than intended.
-  - An OCO submission cannot be set to execute at expiry.
-
+  - An OCO submission can be set to trigger either one side or the other at expiry, or cancel.
```
gordsport: Can this be re-worded to ensure its clear.

davidsiska-vega:
Waiting for @JonRay15 to mock up UI flows, meeting booked for 7th Dec, should know more soon.

JonRay15 (dismissed):

cdummett (commented):
LGTM - just two small comments for ACs

- ACs need to be added to `features.json`
- maybe an AC for the behaviour if both sides are set to submit on expiry? I assume the submission should be rejected

TomMcL:
@cdummett Added specs in to features.json. I believe the OCO part should be a flag where you set which side executes so it should be impossible to set both to execute.

cdummett:
@TomMcL  Ah I might not be aware of a coming change but at the moment I thought to make an OCO you submit a [StopOrdersSubmission](https://docs.vega.xyz/testnet/api/grpc/vega/commands/v1/commands.proto#stoporderssubmission) specifying both a `rises_above` and `falls_below` [StopOrderSetup](https://docs.vega.xyz/testnet/api/grpc/vega/commands/v1/commands.proto#stopordersetup). 

I'm just wondering what happens if the expiry time is the same and the expiry strategy is submit for both of those setups?

TomMcL:
Ah you're right I wasn't thinking about how each had their own strategy already, will update

cdummett (approved):

PR #2050, opened 2023-11-09
Allow using network's trade price as mark price

witgaw:
Allow using network's trade price as mark price + ACs


davidsiska-vega (dismissed):

protocol/0012-POSR-position_resolution.md
```diff
@@ -85,6 +88,8 @@ is closest to reproducing the legacy setup where party would get liquidated imme
 
 Different liquidation strategies with different parameters might be proposed in the future, hence implementation should allow for easy substitution of strategies.
 
+Note that network's trade is treated as any other trade in a sense that its price can be used as the mark price.
```
EVODelavega, line 93: If these trades can define the mark price, then surely they must also factor in to price monitoring, so we probably will want to add an AC along those lines

witgaw: Factor in as in feed into the engine when it determines the reference price?

EVODelavega: Yes, if we didn't do that, we could see the mark price move up/down, but the price bounds would be based on the last non-network initiated trade. In practice, the impact should be fairly minimal, but still, where we would previously have a mark price of 1,000 with min/max bounds 950-1,050, we could now have a mark price of 1,010 with the same price bounds.

The issue here is that price monitoring expects us to pass in the prices of all trades, to check the bounds and then trigger price auction if they exceed the bounds. Once the price auction is triggered, trading would essentially end up getting disabled, so if the network order results in 5 trades, the second of which trips a price auction, should we still push the other trades through, or should we roll back the subsequent trades somehow - after all, that's what we should do if we decide to treat these trades exactly like any other trade.

Currently, after we reduce/zero out the network, we do perform a MTM settlement only applied to the parties that traded with the network, but if the trades can determine the mark price, I suspect that this no longer applies (we should just set the last traded price, and we'll MTM all the updated positions as part of the next MTM settlement)

witgaw: Summarising from Slack:
* network trade's are meant to be IOC, so no risk of triggering an auction,
* there's not meant to be any special flow for those trades, so no special MTM, they will update `last traded price` like any other trade, and if that happens to be the price used for next mark-to-market then that's fine too.

If you believe existing ACs are incorrect or more are needed then please suggest a code change or add your commits.

PR #2052, opened 2023-11-09
Perps: additional parameters

### LINKED ISSUE #2023: For consideration: scaling and caps for perps funding payments
Consider: 
1) add a linear scaling factor to perps payments - basically control how strongly the perp market should be pulled towards the spot price
2) add caps on perps funding payments

witgaw:
1. `scaling_factor`: optional scaling factor applied to funding payment.
1. `rate_lower_bound`: optional lower bound applied to funding payment such that the resulting funding rate will never be lower than the specified value.
1. `rate_upper_bound`: optional upper bound applied to funding payment such that the resulting funding rate will never be greater than than the specified value.


EVODelavega (dismissed):

protocol/0053-PERP-product_builtin_perpetual_future.md
```diff
@@ -152,14 +165,32 @@ funding_payment = f_twap - s_twap + min(clamp_upper_bound*s_twap,max(clamp_lower
...
+If `scaling_factor` is specified set:
+
+```go
+funding_payment = scaling_factor * funding_payment
+```
+
+If `rate_lower_bound` is specified set:
+
+```go
+funding_payment = max(rate_lower_bound*s_twap, funding_payment)
```
wwestgarth, line 177: I'm assuming if all 3 new options are set, we check these mins/maxes _after_ we've done the scaling? If so can we just add a line so thats clear.

witgaw: Clarified now

wwestgarth (dismissed):

wwestgarth (approved):

PR #2058, opened 2023-11-15
feat: team improvments

### LINKED ISSUE #1999: Teams improvements

cdummett:
closes https://github.com/vegaprotocol/specs/issues/1999


protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -202,27 +219,53 @@ message UpdateReferralSet{
...
+
+If a party is already a referee in a valid set, to join or move between teams, a party can submit a `JoinTeam` transaction with the following fields:
+
+- `id`: the id of the team they wish to join
+- `alias`: Optional string defining an alias which should be associated with the key and should appear on leaderboards in the keys place.
+
+```protobuf
+message JoinTeam{
+    id: "mYr3f3rra1c0d3"
+    alias: "myPuBlicA1ias"
```
ValentinTrinque: Why use the alias to join a team ?

cdummett: So users have a way to change their `alias`.

Unless you would prefer a separate set/update alias transaction?

ValentinTrinque: Oh I understood the alias the other way around.

I am wondering if that kind of mechanics is not better as a separate domain, such as a "party / game profile" where we would be free to extend later on and keep other API free from such aspect. 

That would create less coupling.

cdummett: I agree it may be worth it if we planned to add more fields but think it would probably be overcomplicating it otherwise.

@MM0819 do you think it is likely there might be other metadata a party might want to add to their key in the future with respect to teams?

cdummett: Also @ValentinTrinque joining a team and then updating your `alias` (name on leaderboards) would then be two transactions which whilst not horrible I think we want to avoid?

ValentinTrinque: It's also worth considering to *not* necessarily associated a 1-to-1 relation between a form and an endpoint. The front-end can expose a single form that calls 2 endpoints. That might not be the cleanest solution, but I think it's still worth considering about domain extensibility, flexibility and API evolution.

Basically, what happen if you want to change your alias while staying in the team ?
- Should you _re-join_ the team with a different alias ? 
- What kind of event should we emit for that, for the data-node ? Emit a `MemberJoinTeam` would be _wrong_. But, emitting an `MemberAliasUpdated`
 would clearly show that were are mixing 2 things.

These are the question that led me to consider 2 different needs:
1. Joining a team
2. Aliasing a party, network wide for team, referral set, games, whatever.

protocol/0083-RFPR-on_chain_referral_program.md
```diff
@@ -202,27 +219,53 @@ message UpdateReferralSet{
...
 
 To apply a referral code and become a referee, a party must fulfil the following criteria:
 
 - party must not currently be a **referrer**
+- party must not currently be a **referee** associated with a valid referral set (Note: for a set to be valid, the sets referrer must be meeting the staking requirement.)
 
 To become a referee, a referee must submit a signed `ApplyReferralCode` transaction with the following fields:
 
 - `referral_code`: the referral code they wish to apply
+- `alias`: Optional string defining an alias which should be associated with the key applying the referral code and should appear on leaderboards in the keys place.
```
ValentinTrinque: Should the alias be unique ? Is there anything to prevent every party to be named _wen-moon_ ? Do we have restrictions in size, characters, etc ?

protocol/features.json
```diff
@@ -66,7 +66,14 @@
       "0056-REWA-105",
       "0056-REWA-106",
       "0056-REWA-107",
-      "0056-REWA-108"
+      "0056-REWA-108",
+      "0083-RFPR-052",
+      "0056-REWA-053",
```
gordsport, line 60: @cdummett - should these be RFPR rather than REWA?

cdummett: Yes it should! Thanks for catchng

protocol/0083-RFPR-on_chain_referral_program.md
cdummett: Good points, for length a restriction of `20` characters works, for character restriction can we use the same as what we have for the team name?

For whether the alias has to be unique I am less sure as the benefit seems small to me as the alias is purely cosmetic for leaderboards. If the alias replaced keys in forums / governance views etc then yeah I think unique to prevent impersonation would be important.

Out of interest, how expensive would it to be check uniqueness? Is there any risk of DoS attacks?

ValentinTrinque: I agree, it all depends on the reach this alias will have.

It can be expensive on the long run, I guess. Because it means to keep track of all aliases in memory.

protocol/0083-RFPR-on_chain_referral_program.md
JonRay15: Surely the issue with having alias on the team is that if I ever swap teams i have to keep re-inputting my alias each time?

We have the concept in the sketches of a user profile, and a user profile "card" ..... plus you can choose to compete when not in a team and would want an alias there too I suspect.

I dont know right now if we definitely will have more metadata associated to a user profile, but i can certainly imagine it ... discord ID, twitter handle, maybe awards / ranks / achievements etc.

To me a seperate user metadata thing seems more appropriate.

cdummett: Resolving as "alias"/"profiles" delayed to future release.

protocol/0083-RFPR-on_chain_referral_program.md
cdummett: Resolving as "alias"/"profiles" delayed to future release.

ValentinTrinque (approved):

PR #2078, opened 2023-12-11
fix: update amend pegged order margin expected behaviour

Sohill-Patel:
updated expected behaviour of a pegged order amend
remove duplicate ac

### DIFF
--- a/protocol/0019-MCAL-margin_calculator.md
+++ b/protocol/0019-MCAL-margin_calculator.md
@@ -163,11 +163,11 @@ There should be an additional amount `limit price x size x margin factor = 15910
 
 - When the party increases the pegged sell order offset price, the order margin should be reduced (<a name="0019-MCAL-057" href="#0019-MCAL-057">0019-MCAL-057</a>)
 
-- When the party increases the pegged order size and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the amendment is rejected and the original order is not effected (<a name="0019-MCAL-052" href="#0019-MCAL-052">0019-MCAL-052</a>)
+- When the party increases the pegged order size and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the amendment is rejected and the original order is not affected (<a name="0019-MCAL-052" href="#0019-MCAL-052">0019-MCAL-052</a>)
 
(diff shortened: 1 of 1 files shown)


protocol/0019-MCAL-margin_calculator.md
```diff
@@ -163,13 +163,11 @@ There should be an additional amount `limit price x size x margin factor = 15910
 
 - When the party increases the pegged sell order offset price, the order margin should be reduced (<a name="0019-MCAL-057" href="#0019-MCAL-057">0019-MCAL-057</a>)
 
-- When the party increases the pegged order size and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the order should be stopped (<a name="0019-MCAL-052" href="#0019-MCAL-052">0019-MCAL-052</a>)
+- When the party increases the pegged order size and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the amendment is rejected and the original order is not effected (<a name="0019-MCAL-052" href="#0019-MCAL-052">0019-MCAL-052</a>)
```
TomMcL:
```suggestion
- When the party increases the pegged order size and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the amendment is rejected and the original order is not affected (<a name="0019-MCAL-052" href="#0019-MCAL-052">0019-MCAL-052</a>)
```

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -163,13 +163,11 @@ There should be an additional amount `limit price x size x margin factor = 15910
...
 
-- When the party increases the pegged order price and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the order should be stopped (<a name="0019-MCAL-075" href="#0019-MCAL-075">0019-MCAL-075</a>)
+- When the party increases the pegged order price and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the amendment is rejected and the original order is not effected (<a name="0019-MCAL-075" href="#0019-MCAL-075">0019-MCAL-075</a>)
```
TomMcL:
```suggestion
- When the party increases the pegged order price and the party's general account does not contain sufficient funds to cover any increases to the order margin account to be equal to side margin then the amendment is rejected and the original order is not affected (<a name="0019-MCAL-075" href="#0019-MCAL-075">0019-MCAL-075</a>)
```

TomMcL (changes requested):
I think the deleted AC will need removing from features.json too

Sohill-Patel:
> I think the deleted AC will need removing from features.json too

@TomMcL i did check and ac `0019-MCAL-077` is not in there

TomMcL (approved):

PR #2086, opened 2023-12-21
feat: update market state when market name/code is changed while governance proposal

Jiajia-Cui:


protocol/0028-GOVE-governance.md
```diff
@@ -546,8 +546,8 @@ APIs should also exist for clients to:
...
 - A market change that's to modify any parameters on a market in `pending` state (i.e. voting has successfully completed on the market creation and the market is in the opening auction) will run voting rules the same as market creation proposals i.e. LPs don't get a vote. (<a name="0028-GOVE-072" href="#0028-GOVE-072">0028-GOVE-072</a>)
-- A market change proposal that aims to modify the market name in any state will modify the market name at the time of vote enactment. (<a name="0028-GOVE-159" href="#0028-GOVE-159">0028-GOVE-159</a>)
-- A market change proposal that aims to modify the market code in any state will modify the market code at the time of vote enactment. (<a name="0028-GOVE-166" href="#0028-GOVE-166">0028-GOVE-166</a>)
+- A market change proposal that aims to modify the market name in any state (apart from closed/terminated) will modify the market name at the time of vote enactment. (<a name="0028-GOVE-159" href="#0028-GOVE-159">0028-GOVE-159</a>)
```
davidsiska-vega, line 549: Can you check with core whether the ` (apart from closed/terminated) ` makes life easier or harder? In a way I don't care - if someone renames a closed market - that's their job.

davidsiska-vega (changes requested):
I think there may be other places in the spec where it says that these two can't be modified apart from the ACs. Can you please look?

protocol/0028-GOVE-governance.md
Jiajia-Cui: Its already implemented by Zohar like this :)

davidsiska-vega (approved):

PR #2095, opened 2024-01-05
feat: few acs for stop orders in auction

Sohill-Patel:
split existing acs that are actually two cases, since the current tests only cover one case
Added a couple acs for stop orders in auction / opening auction

### DIFF
--- a/protocol/0014-ORDT-order_types.md
+++ b/protocol/0014-ORDT-order_types.md
@@ -377,6 +377,7 @@ In Spot market, for multiple iceberg orders submitted as a batch of orders with 
 
 - A stop order placed prior to an auction will not execute during an auction, nor will it participate in the uncrossing. (<a name="0014-ORDT-134" href="#0014-ORDT-134">0014-ORDT-134</a>)
 - A stop order placed prior to an auction, where the uncrossing price is within the triggering range, will immediately execute following uncrossing. (<a name="0014-ORDT-135" href="#0014-ORDT-135">0014-ORDT-135</a>)
+- An order with a stop is placed during continuous trading. The market goes into auction. The market exits auction, the condition for triggering the stop is not met. The stop order is still present. (<a name="0014-ORDT-136" href="#0014-ORDT-136">0014-ORDT-136</a>)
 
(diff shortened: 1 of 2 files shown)


protocol/0014-ORDT-order_types.md
```diff
@@ -370,8 +372,15 @@ In Spot market, for multiple iceberg orders submitted as a batch of orders with
...
-- A stop order placed either prior to or during an auction, where the uncrossing price is within the triggering range, will immediately execute following uncrossing. (<a name="0014-ORDT-066" href="#0014-ORDT-066">0014-ORDT-066</a>)
+- A stop order placed during an auction will not execute during an auction, nor will it participate in the uncrossing. (<a name="0014-ORDT-065" href="#0014-ORDT-065">0014-ORDT-065</a>)
+- A stop order placed during an auction, where the uncrossing price is within the triggering range, will immediately execute following uncrossing. (<a name="0014-ORDT-066" href="#0014-ORDT-066">0014-ORDT-066</a>)
+
+- A stop order placed prior to an auction will not execute during an auction, nor will it participate in the uncrossing. (<a name="0014-ORDT-134" href="#0014-ORDT-134">0014-ORDT-134</a>)
+- A stop order placed prior to an auction, where the uncrossing price is within the triggering range, will immediately execute following uncrossing. (<a name="0014-ORDT-135" href="#0014-ORDT-135">0014-ORDT-135</a>)
```
davidsiska-vega, line 379: Is this different to 0014-ORDT-066?

davidsiska-vega (changes requested):
One question there...

protocol/0014-ORDT-order_types.md
Sohill-Patel: looking at the feature tests 0014-ORDT-066 and 0014-ORDT-065 we only run the scenario when stop order is created during auction. 
I created these two acs to cover the scenario when stop order is created prior to an auction

davidsiska-vega: I see it - one is during one is prior!

davidsiska-vega (approved):

PR #2138, opened 2024-01-23
feat: party profile

cdummett:
Adds a specification to cover the `UpdatePartyProfile` command added in PR https://github.com/vegaprotocol/vega/pull/10340.

Also adds spam protection rules to cover the new command type.


protocol/0088-PPRF-party_profile.md
```diff
@@ -0,0 +1,44 @@
...
+- the alias must be unique (i.e. the alias must not already be associated with an existing party profile)
+
+**Note:**
+In order to validate uniqueness of an alias, core must keep a record of the aliases associated with each party. If a party is no longer active.
+
+### Alias Squatting
+
+To prevent alias squatting the network will track the number of consecutive epochs each party has funds totalling less than `spam.protection.updatePartyProfile.min.funds` (quantum) across all their accounts.
+
+If the number of consecutive epochs is greater than or equal to `spam.protection.updatePartyProfile.maxInactiveEpochs` the network will no longer associate the alias and metadata with the key.
```
jeremyletang: I don't think name squatting makes much sense, and should not be implement like so. I could have fund on my party, decide to move it away and come back after that time. If I come back after I exceeded the number of epoch without funds someone can have stollen my profile and usurpate the identity of someone in the network somehow (obviously you can check what alias is attached to what pubkey and all but..).

I think the spam mecanics needs to be good enough so it's difficult to squat a name and it should be enough.

protocol/0088-PPRF-party_profile.md
```diff
@@ -0,0 +1,44 @@
...
+An `UpdatePartyProfile` has the following fields.
+
+- `alias`: an optional string which can be as an alternative identifier for the public key in any dApp (e.g. in competition leaderboards).
+- `metadata`: an optional list of metadata key value pairs to further describe a public key.
+
+If metadata is specified, the metadata must adhere to the following rules:
+
+- no more than 10 key value pairs can be specified
+- a key must be no longer than 33 characters
+- a value must be no longer than 256 characters
```
jeremyletang: What is the reasonning for these requirements?

cdummett: Matching what was spoken about in the [slack thread](https://vegaprotocol.slack.com/archives/C02KVKMAE82/p1704729335940369?thread_ts=1704711479.332909&cid=C02KVKMAE82) and implemented in the [PR](https://github.com/vegaprotocol/vega/pull/10340/files#diff-b0254a8b84ccb6f3b38b2d955a45acbd13f1bacb208fc51a38cddbf9d805f549) 

Believe this was a memory requirement and to stop users spamming the database with enormous metadata.

protocol/0088-PPRF-party_profile.md
cdummett: This was also to manage storage used, a party could infinitely move funds between keys and update a profile which the network would then have to store.  There would be transfer fees for doing this but ideally `spam.protection.updatePartyProfile.min.funds` is as low as possible so the fees may not be enough of a deterrent to prevent a spam attack.

cdummett: I do agree though, having someone usurp an identity may actually be worse though than name squatting..

So if you don't think memory issues are a concern maybe we can drop this anti name squatting feature? Thoughts @barnabee ?

gordsport: @cdummett - for palazzo we plan to only add spam protection, can we make the name unique and add details regarding spam. 

We can pick up the conversation on name squatting for the nebula milestone

jeremyletang (approved):

PR #2141, opened 2024-01-25
chore: update ACs moved from palazzo to nebula

gordsport:
part 2 of:

- https://github.com/vegaprotocol/specs/pull/2140


Jiajia-Cui (dismissed):

davidsiska-vega (changes requested):
I think we need to see the values a new feature provides - otherwise it's pretty unusable. 

The overall aim is to give users more trust in our funding payments. Without seeing the numbers going in are we achieving this?

davidsiska-vega (approved):

PR #2148, opened 2024-02-01
feat: add taker rewards capp by trading fees reward

### LINKED ISSUE #2142: Taker rewards capped by trading fees reward scheme
Consider adding a taker reward which caps the reward by trading fees paid. 

At the moment the taker rewards are roughly calculated as follows: the scheme specifies the fee type which makes you eligible (taker / maker)
1) there is an amount to be distributed
2) there are keys that have paid the relevant fees; we calculate the fraction of `1` of relevant fees each eligible key paid/received. 
3) each key then gets fraction times the total. 

In particular it means that e.g. if the reward pot is 500 and there is only one trader who did one trade they get all 500. 

As a separate scheme we want to add the following
1) there is an amount to be distributed
2) there are keys that have paid the relevant fees; we calculate the fraction of `1` of relevant fees each eligible key paid/received. 
3) each key then gets max(fraction times total, taker fees paid in epoch times conversion factor from quantum to reward asset). 
4) whatever is left over stays in the reward pool for next epoch (if this is valid) or sent to the key that funded the pool or whatever makes most sense in the current design.

Jiajia-Cui:
closes [#2142 ](https://github.com/vegaprotocol/specs/issues/2142)


protocol/0056-REWA-rewards_overview.md
```diff
@@ -56,6 +56,14 @@ If the reward account balance is `0` at the end of the epoch for a given recurri
 
 Fee-based reward metrics (the total fees paid/received by each party as defined above) are stored in [LNL checkpoints](./0073-LIMN-limited_network_life.md) and are restored after a checkpoint restart to ensure rewards are not lost.
 
+The taker reward for each key is capped at `market.maxreward.taker * rewards_taker_paid_fee`. Whatever left in the reward pool will remain for the next epoch.
+
+The maker reward for each key is capped at `market.maxreward.maker * rewards_maker_fee`. Whatever left in the reward pool will remain for the next epoch.
+
+Market parameter `market.maxreward.taker` (decimal) —  maximum fraction of `rewards_taker_paid_fee` each key can receive each epoch. This is a decimal number in the interval $[0,1]$, when set to $0$ then each key will not receive any taker fee reward, when set to $1$ and if the key is the only price taker on the market, then the key will get all the `rewards_taker_paid_fee` in the epoch.
```
barnabee: It seems odd to me to make this a market parameter not part of the reward definition. What's the reasoning for this?

Jiajia-Cui: I think the taker fee reward is not clearly defined in the spec, so I added to the most relevant part, which part of the reward definition you have in mind?

protocol/0057-TRAN-transfers.md
```diff
@@ -109,6 +109,7 @@ A party should be able to configure the distribution of rewards by specifying th
 
 - `window_length` - the number of epochs over which to evaluate the reward metric. The value should be limited to 100 epochs.
 - `lock_period` - the number of epochs after distribution to delay [vesting of rewards](./0085-RVST-rewards_vesting.md#vesting-mechanics) by.
+- `cap_reward` -  maximum fraction of `reward` each key can receive each epoch from the reward pool. Once the cap is applied, whatever left in the reward pool will remain in the reward pool for the next epoch.
```
barnabee: I this should be as a multiple of fees paid, suggested change:

```suggestion
- `cap_reward_fee_multiple` [optional] - if set, the actual amount of reward transferred to each public key during distribution for this transfer will be `min(calculated_reward, cap_reward_fee_multiple × fees_paid_this_epoch)`. If all keys are capped (i.e. the total amount of the transfer cannot be be sent to eligible keys without breaching the cap) then the remaining balance must either not leave the source account for an iteration of the recurring transfer, or be transferred back to the source account with no fee. If this occurs, and the total transferred in a given epoch this does not affect the size of the next iteration, which proceeds as normal (including decay factors etc.) as if the full transfer has been made.
```

protocol/0057-TRAN-transfers.md
```diff
@@ -109,6 +109,7 @@ A party should be able to configure the distribution of rewards by specifying th
 
 - `window_length` - the number of epochs over which to evaluate the reward metric. The value should be limited to 100 epochs.
 - `lock_period` - the number of epochs after distribution to delay [vesting of rewards](./0085-RVST-rewards_vesting.md#vesting-mechanics) by.
+- `cap_reward_fee_multiple` [optional] - if set, the actual amount of reward transferred to each public key during distribution for this transfer will be `min(calculated_reward, cap_reward_fee_multiple × fees_paid_this_epoch)`. If all keys are capped (i.e. the total amount of the transfer cannot be be sent to eligible keys without breaching the cap) then the remaining balance must either not leave the source account for an iteration of the recurring transfer, or be transferred back to the source account with no fee. If this occurs, and the total transferred in a given epoch this does not affect the size of the next iteration, which proceeds as normal (including decay factors etc.) as if the full transfer has been made.
```
davidsiska-vega: You need to use the quantum values for both assets involved and create a conversion factor.

davidsiska-vega: > If all keys are capped (i.e. the total amount of the transfer cannot be be sent to eligible keys without breaching the cap) 

Isn't that "if at least one of the keys got capped"? Because that is the condition to have leftover reward balance.

davidsiska-vega: > then the remaining balance must either not leave the source account for an iteration of the recurring transfer, or be transferred back to the source account with no fee.

Decide. Maybe ask on the thread on Slack what people prefer.

barnabee: > Isn't that "if at least one of the keys got capped"? Because that is the condition to have leftover reward balance.

No, that balance could go to other keys that wouldn't otherwise have gotten all of their fees back in rewards. If any key remains uncapped you can use it all.

You don't want a situation where a few large traders who paid very little fees end up with 95% of the rewards assigned to them but only take 2% of them and other takers don't get all of their fees back.

barnabee:
```suggestion
- `cap_reward_fee_multiple` [optional] - if set, the actual amount (in quantum) of reward transferred to each public key during distribution for this transfer will be `min(calculated_reward_in_quantum, cap_reward_fee_multiple × fees_paid_this_epoch_in_quantum)`. If all keys are capped (i.e. the total amount of the transfer cannot be be sent to eligible keys without breaching the cap) then the remaining balance must be left in the reward pool and included in the distribution in future epochs. If this occurs, and the total transferred in a given epoch this does not affect the size of the next iteration, which proceeds as normal (including decay factors etc.) as if the full transfer has been made.
```

davidsiska-vega: Does that work though? In my mind 
1) I work out uncapped amount to send. 
2) I do the cap and I send. 

If you want to do what you suggest I'd need to do multiple rounds or have some more complex algo to work out the exact amounts so I try to use it all.

davidsiska-vega (changes requested):
- quantum conversion
- make it clear how the cap is actually implemented because it's not at the moment, see the thread above.

protocol/0057-TRAN-transfers.md
barnabee: Yes the algo needs to be more compelx than the one you outline above. I am not sure the spec needs to be the place to work out the algorithm if the requirement is clear?

protocol/0057-TRAN-transfers.md
```diff
@@ -109,6 +109,7 @@ A party should be able to configure the distribution of rewards by specifying th
...
 - `lock_period` - the number of epochs after distribution to delay [vesting of rewards](./0085-RVST-rewards_vesting.md#vesting-mechanics) by.
+- `cap_reward_fee_multiple` [optional] - if set, the actual amount of reward transferred to each public key during distribution for this transfer will be `min(calculated_reward, cap_reward_fee_multiple × fees_paid_this_epoch)`. When calculating how much of the reward each one is getting, if some is left from the applied cap, we recalculate on the remaining balance only for parties that have not reached their cap until the leftover is less than `transfer.fee.maxQuantumAmount x quantum`.If all keys are capped (i.e. the total amount of the transfer cannot be be sent to eligible keys without breaching the cap) then the remaining balance must be left in the reward pool and included in the distribution in future epochs. If this occurs, and the total transferred in a given epoch this does not affect the size of the next iteration, which proceeds as normal (including decay factors etc.) as if the full transfer has been made.
```
davidsiska-vega: Still missing the conversion to be done using assets' quantum, no?

davidsiska-vega (changes requested):
the quantum conversion?

davidsiska-vega (approved):

PR #2149, opened 2024-02-01
feat: changeable market tick size

### LINKED ISSUE #2122: Add changeable tick size to markets
... so that a coin that moons / tanks doesn't need us to close / re-open a perp market. 

A new market parameter, tick size arbitrary decimal, check that if this is divided by 10^-marketDPs we get an integer. E.g. 0.005 with 3 mdps is 0.005/0.001 = 5 is valid but 0.005 with 2 mdps is not.

Reject orders that are not integer multiples of tick size.

If a market proposal changes tick size:
0) check that if this is divided by 10^-mdps we get integer - this can be part of proposal validation since mdps won't change. 
2) go through the book at enactment and remove all volume that is not integer multiple of new tick size

cdummett:
closes #2122 

Adds a changeable tick size field to market definitions.

- Order submissions, amendments must quote prices which are an integer multiple of the tick size. 
- Pegged orders must specify offsets which are an integer multiple of the tick size.

NOTE:
- There is no requirement for stop orders to require a trigger price which is an integer multiple of the tick size.
- There is no requirement for vAMMs to require a base, lower and upper price which is an integer multiple of the tick size.


protocol/0001-MKTF-market_framework.md
```diff
@@ -32,6 +32,7 @@ Data:
 - **Position Decimal Places**: number of decimal places for orders and positions, i.e. if this is 2 then the smallest increment that can be traded is 0.01, for example 0.01 BTC in a `BTSUSD` market.
   - If this is negative e.g. -3 this means that the smallest order and position is of size 1000.
   - Accepted values are `-6,...,-1,0,1,2,...,6`.
+- **Tick size**: the minimum change in quote price for the market. Order prices and offsets for pegged orders must be given as an exact multiple of the tick size. For example if the tick size is 0.02 USD. then a price of 100.02 USD is acceptable and a price of 100.03 USD is not. The tick size of a market can be updated through governance.
```
TomMcL: Would it be clearer/cleaner if tick size were specified *in* market price decimals? e.g. for the above example it would just be 2. Would enforce that tick size is always compatible with the market decimals and possibly make the maths easier.

cdummett: That was definitely the intention when writing but reading back I realise that is not clear so thanks for flagging!

davidsiska-vega (changes requested):
Could we clarify what happens on enactment of new tick size and some ACs around that?

In the ticket we have this i.e. remove orders. You could also snap then to nearest valid tick (making sell orders more expensive and buy orders cheaper I suppose).
```
If a market proposal changes tick size:
0) check that if this is divided by 10^-mdps we get integer - this can be part of proposal validation since mdps won't change.
2) go through the book at enactment and remove all volume that is not integer multiple of new tick size
```

cdummett:
> Could we clarify what happens on enactment of new tick size and some ACs around that?
> 
> In the ticket we have this i.e. remove orders. You could also snap then to nearest valid tick (making sell orders more expensive and buy orders cheaper I suppose).
> 
> ```
> If a market proposal changes tick size:
> 0) check that if this is divided by 10^-mdps we get integer - this can be part of proposal validation since mdps won't change.
> 2) go through the book at enactment and remove all volume that is not integer multiple of new tick size
> ```

So in the protocol-design session last week (possibly whilst you were away) we spoke about reducing the complexity of this part of the feature.

The outcome was we were happy for existing orders at invalid price levels to stay on the book until they were filled/expired. The new tick sizing would only be applied to validation for new order types.

davidsiska-vega (approved):

PR #2169, opened 2024-02-08
Set order to stopped if cancelled by the core

peterbarrow:
If an order is cancelled by a user we mark it as cancelled, if the core is responsible for cancelling an order we set it to stopped. In this case as the core is changing the order from active to cancelled we mark it as stopped.


peterbarrow:
While this change makes sense it would make the spot markets act differently to the other market types. So I'm removing this update as it would create too much work to make everything consistent.

PR #2173, opened 2024-02-13
feat: update fees on auction exit AC

cdummett:
PR removes AC `0080-SPOT-019` and replaces it with `0080-SPOT-021`.

The change states that fees non longer need to be recalculated in the middle of an auction should the fee factors change and instead only at the end of the auction.


peterbarrow (approved):

gordsport:
@cdummett @peterbarrow FYI:, I have merged this branch into:
https://github.com/vegaprotocol/specs/pull/2163

Where I have been doing some tidying up on spot

PR #2175, opened 2024-02-15
refactor: amend a few perp ACs

witgaw:
Amend a few ACs that used incorrect margin expectation. These are only covered via integration tests and I've fixed them on a branch according to amended formulation so went for a change rather than deleting and incrementing number.


Jiajia-Cui (approved):
LGTM

PR #2180, opened 2024-02-19
feat: Add order spam

### LINKED ISSUE #2097: Order spam: margin required for an order is checked to be greater than a network parameter
Consider a spam protection where the margin required for an order is checked to be greater than a preset network parameter. Reject orders that are too cheap. 

This is a) for Nebula where you can expect mis-configured markets and b) for pre-launch perps / futures token where it will be genuinely hard to guess what the price ought to be.

gordsport:
Copies content from https://github.com/vegaprotocol/specs/pull/2179 with base branch as Colosseo

close https://github.com/vegaprotocol/specs/issues/2097


protocol/0062-SPAM-spam_protection.md
```diff
@@ -83,6 +83,17 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
 
 **Note** `spam.protection.max.updatePartyProfile` must be an integer greater than or equal to `0` (and default to `5`).
 
+### Transaction Spam
+
+Network parameter: `spam.order.MinimalMarginQuantumMultiple` (between 0 and infinite)
+
+If the maintenance margin for a given transaction is smaller than the parameter `spam.order.MinimalMarginSizeQuantumMultiple`, then the transaction is pre-block rejected.
+
+If the market doesn't exist or hasn't been enacted yet (and thus the maintenance margin cannot be computed), the transaction is rejected.
```
cdummett: Bit of a terminology nitpick but I'm not sure about "hasn't been enacted yet."

I think a "pending" market which is in an opening auction un till the enactment timestamp is reached still needs to be able to accept orders?

Vegaklaus: True, I was referring to undefined markets where the margin can't be calculated yet. Will clarify that

protocol/0062-SPAM-spam_protection.md
```diff
@@ -83,6 +83,17 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
 
 **Note** `spam.protection.max.updatePartyProfile` must be an integer greater than or equal to `0` (and default to `5`).
 
+### Transaction Spam
+
+Network parameter: `spam.order.MinimalMarginQuantumMultiple` (between 0 and infinite)
+
+If the maintenance margin for a given transaction is smaller than the parameter `spam.order.MinimalMarginSizeQuantumMultiple`, then the transaction is pre-block rejected.
```
jeremyletang, line 90: Is the minimal margin size the right thing to do here? I kinda remember us discussing instead using the quantum notional value of the order as it'd be much less computing heavy?

davidsiska-vega: Margin is what we want because it takes into account leverage. 

In terms of computation heavyness:  take `(rf + linear scaling param) x size x price` it's not much heavier than just `size x price` no?

protocol/0062-SPAM-spam_protection.md
```diff
@@ -83,6 +83,17 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
...
+
+If the market does not exist and thus the maintenance margin is not defined, the transaction is rejected.
```
davidsiska-vega, line 93:
```suggestion
If the market does not exist and thus the maintenance margin is not defined, the transaction is rejected.

If floating point consensus hasn't delivered a risk factor yet use `rf = 1` for the spam check. 
```

protocol/0062-SPAM-spam_protection.md
davidsiska-vega, line 90:
```suggestion
If the maintenance margin for a given transaction is smaller than the parameter `spam.order.MinimalMarginSizeQuantumMultiple`, then the transaction is pre-block rejected. 
I.e. if `(rf + linear slippage param) x size x price <  spam.order.MinimalMarginSizeQuantumMultiple x asset quantum amount` then the order is rejected. Here `rf` is the risk factor (and will be different for long and for short) `linear slippage param` is a market parameter and `size` and `price` are assumed to be correctly scaled by, PDPs and MDPs respectively.
```

protocol/0062-SPAM-spam_protection.md
Vegaklaus: That can lead to inconsistencies between validators (some seeing the transaction before the consensus finished and some after). This is not a problem anymore, right @ze97286 ?

davidsiska-vega: They will all have an agreement on whether they have the true `rf` from the floating point consensus or not. So either none of them have it or all of them have it. In case none of them have it (because rf hasn't gone through consensus yet) use the value of `1`.

ze97286: well, for as long as two nodes are on the same block, they will have the same value of rf, but if an order is submitted to a node that is behind then it may use an older value of rf.

ze97286 (commented):
what about spots?

protocol/0062-SPAM-spam_protection.md
Vegaklaus: But that won't cause any issues anymore, correct ?

ze97286: no

Vegaklaus:
> what about spots?

Spots needs discussion :)
I think we need to allow occasional low-amount spot (e.g., allow someone to transfer over staking rewards), but then only once per epoch/key ?

protocol/0062-SPAM-spam_protection.md
```diff
@@ -83,6 +83,18 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
 
 **Note** `spam.protection.max.updatePartyProfile` must be an integer greater than or equal to `0` (and default to `5`).
 
+### Transaction Spam
+
+Network parameter: `spam.order.MinimalMarginQuantumMultiple` (between 0 and infinite)
+
+If the maintenance margin for a given transaction is smaller than the parameter `spam.order.MinimalMarginSizeQuantumMultiple`, then the transaction is pre-block rejected.
+I.e. if `(rf + linear slippage param) x size x price <  spam.order.MinimalMarginSizeQuantumMultiple x asset quantum amount` then the order is rejected. Here `rf` is the risk factor (and will be different for long and for short) `linear slippage param` is a market parameter and `size` and `price` are assumed to be correctly scaled by, PDPs and MDPs respectively.
```
davidsiska-vega, line 91:
```suggestion
I.e. if `(rf + linear slippage param) x size x price <  spam.order.MinimalMarginSizeQuantumMultiple x asset quantum amount` then the order is rejected. Here `rf` is the risk factor (and will be different for long and for short) `linear slippage param` is a market parameter and `size` and `price` are assumed to be correctly scaled by, PDPs and MDPs respectively.
For spot markets, the calculation is the same except the value of `rf = 1` and `linear slippage factor = 0`.
```

ze97286 (commented):
this seems to focus on order submission, what about order amendment?

davidsiska-vega:
> > what about spots?


Take `rf = 1` and `linear slippage factor = 0`: 

https://github.com/vegaprotocol/specs/pull/2180/files#r1502490707

protocol/0062-SPAM-spam_protection.md
gordsport: @davidsiska-vega - should we also have an AC for spot markets on this?

davidsiska-vega: One for @Vegaklaus but yes please, A spot-specific AC will be good.

ze97286 (commented):
what about pegged orders?

ze97286 (commented):
maybe I've missed it but I don't see what should happen for a batch transaction?

protocol/0062-SPAM-spam_protection.md
```diff
@@ -83,6 +83,20 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
...
+
+If the market does not exist and thus the maintenance margin is not defined, the transaction is rejected.
+
+Non-persistent orders (e.g., spot orders) are not blocked by this mechanism as long as the market exists.
```
davidsiska-vega:
```suggestion
Non-persistent orders (on both spot and futures markets) are not blocked by this mechanism as long as the market exists.
```

davidsiska-vega (commented):
One more small edit.

davidsiska-vega (approved):

PR #2188, opened 2024-02-26
Order spam

### LINKED ISSUE #2097: Order spam: margin required for an order is checked to be greater than a network parameter

davidsiska-vega:
Aims to address comments from @ze97286 that got missed (by me) on the previous PR here https://github.com/vegaprotocol/specs/pull/2180:

1. "this seems to focus on order submission, what about order amendment?"
2. "what about pegged orders?"
3. "maybe I've missed it but I don't see what should happen for a batch transaction?" 

close https://github.com/vegaprotocol/specs/issues/2097


protocol/0062-SPAM-spam_protection.md
```diff
@@ -89,6 +89,9 @@ Network parameter: `spam.order.MinimalMarginQuantumMultiple` (between 0 and infi
 
 If the maintenance margin for a given transaction is smaller than the parameter `spam.order.MinimalMarginSizeQuantumMultiple`, then the transaction is pre-block rejected.
 I.e. if `(rf + linear slippage param) x size x price <  spam.order.MinimalMarginSizeQuantumMultiple x asset quantum amount` then the order is rejected. Here `rf` is the risk factor (and will be different for long and for short) `linear slippage param` is a market parameter and `size` and `price` are assumed to be correctly scaled by, PDPs and MDPs respectively.
+- For peg orders we check the `price` implied by the peg and the book at submission time.
```
```suggestion

- For peg orders we check the `price` implied by the peg and the book at submission time. 
```

ze97286: what if the peg doesn't exist?

ze97286: and during an auction.

protocol/0062-SPAM-spam_protection.md
```diff
@@ -90,14 +90,18 @@ Network parameter: `spam.order.MinimalMarginQuantumMultiple` (between 0 and infi
 If the maintenance margin for a given transaction is smaller than the parameter `spam.order.MinimalMarginSizeQuantumMultiple`, then the transaction is pre-block rejected.
 I.e. if `(rf + linear slippage param) x size x price <  spam.order.MinimalMarginSizeQuantumMultiple x asset quantum amount` then the order is rejected. Here `rf` is the risk factor (and will be different for long and for short) `linear slippage param` is a market parameter and `size` and `price` are assumed to be correctly scaled by, PDPs and MDPs respectively.
 
+- For peg orders use `mark price +/- offset` instead of price. The `mark price` value can be cached from end of last block (if that helps for performance, we need this to be cheap). In opening auctions mark price won't exist; in this case allow pegged order submission *but* upon opening auction uncrossing run a check on all the pegged orders in the market and cancel all those that don't meet the minimum margin condition.
```
davidsiska-vega: @davidsiska-vega change to not accepting in opening auctions.

davidsiska-vega (commented):
Check that `general + margin > 0' (after checking pow) but before any of these more fancy checks.

protocol/0062-SPAM-spam_protection.md
davidsiska-vega: done

protocol/0062-SPAM-spam_protection.md
```diff
@@ -85,19 +85,28 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
 
 ### Transaction Spam
 
+Before any order or liquidity commitment is accepted for a perpetual futures or expiring futures check that the party has `margin + general > 0` with `margin` being the balance in the margin account for the relevant market and `general` the balance in the general account for the relevant asset. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
+
+Before any order or liquidity commitment is for a spot market check that the party has `general > 0` with `general` the balance in the general account for the asset that the order is potentially disposing of / commiting in terms of liquidity. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
```
cdummett: So parties are able to submit an order amendment transaction reducing the size of an order, I think this check should include the holding account i.e.

```suggestion
Before any order or liquidity commitment is accepted for a spot market, check that the party has balances where `holding + general > 0` for the asset that the order is potentially disposing of / committing in terms of liquidity. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
```

protocol/0062-SPAM-spam_protection.md
```diff
@@ -145,5 +154,31 @@ More than 360 delegation changes in one epoch (or, respectively, the value of `s
...
+#### Balance checks 
+On perps and futures markets order are rejected `margin + general = 0` with margin being the margin account balance for the relevant market and general being the general account balance for the settlement asset for the market for 
+
+- market orders (<a name="0062-SPAM-053" href="#0062-SPAM-053">0062-SPAM-053</a>).
+- limit orders (<a name="0062-SPAM-054" href="#0062-SPAM-054">0062-SPAM-054</a>).
+- pegged orders (<a name="0062-SPAM-055" href="#0062-SPAM-055">0062-SPAM-055</a>).
+- liquidity commitment (<a name="0062-SPAM-056" href="#0062-SPAM-056">0062-SPAM-056</a>).
+- stop-loss / position-linked order (<a name="0062-SPAM-057" href="#0062-SPAM-057">0062-SPAM-057</a>).
+
+On spot markets orders are rejected if `general = 0` for the asset that the order is (or potentially is) disposing of for
```
cdummett: From the above suggestion
```suggestion
On spot markets orders are rejected if `holding + general = 0` for the asset that the order is (or potentially is) disposing of for
```

cdummett (changes requested):

protocol/0062-SPAM-spam_protection.md
TomMcL:
```suggestion
On spot markets orders are rejected if `general = 0` for the asset that the order is (or potentially is) disposing of
```

davidsiska-vega: fixed 👍

protocol/0062-SPAM-spam_protection.md
```diff
@@ -85,19 +85,28 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
 
 ### Transaction Spam
 
+Before any order or liquidity commitment is accepted for a perpetual futures or expiring futures check that the party has `margin + general > 0` with `margin` being the balance in the margin account for the relevant market and `general` the balance in the general account for the relevant asset. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
```
ze97286: we need to account for order margin I'd imagine too

davidsiska-vega: Yes, please have a look again after the recent edit.

ze97286 (dismissed):

cdummett (dismissed):

protocol/0062-SPAM-spam_protection.md
```diff
@@ -85,19 +85,28 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
 
 ### Transaction Spam
 
+Before any order or liquidity commitment is accepted for a perpetual futures or expiring futures check that the party has `margin + order margin + general > 0` with `margin` being the balance in the margin account for the relevant, `order margin` is the balance for the order margin if party is in isolated margin mode for the for the relevant market and `general` the balance in the general account for the relevant asset. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
+
+Before any order or liquidity commitment is accepted for a spot market, check that the party has balances where `holding + general > 0` for the asset that the order is potentially disposing of / committing in terms of liquidity. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
+
```
davidsiska-vega, line 93: This can be done first at mempool-level pre consensus.

protocol/0062-SPAM-spam_protection.md
```diff
@@ -85,19 +85,28 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
...
+
 Network parameter: `spam.order.MinimalMarginQuantumMultiple` (between 0 and infinite)
 
 If the maintenance margin for a given transaction is smaller than the parameter `spam.order.MinimalMarginSizeQuantumMultiple`, then the transaction is pre-block rejected.
```
davidsiska-vega: Spot?

protocol/0062-SPAM-spam_protection.md
```diff
@@ -85,19 +85,28 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
 
 ### Transaction Spam
 
+Before any order or liquidity commitment is accepted for a perpetual futures or expiring futures check that the party has `margin + order margin + general > 0` with `margin` being the balance in the margin account for the relevant, `order margin` is the balance for the order margin if party is in isolated margin mode for the for the relevant market and `general` the balance in the general account for the relevant asset. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
+
+Before any order or liquidity commitment is accepted for a spot market, check that the party has balances where `holding + general > 0` for the asset that the order is potentially disposing of / committing in terms of liquidity. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
+
+
 Network parameter: `spam.order.MinimalMarginQuantumMultiple` (between 0 and infinite)
```
davidsiska-vega: Maybe different param for spot - the name doesn't work for spot anyway.

protocol/0062-SPAM-spam_protection.md
```diff
@@ -83,6 +83,42 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
...
+If the holding requirement for a given transaction is smaller than the parameter `spam.order.minimalHoldingQuantumMultiple`, then the transaction is pre-block rejected.
+I.e. if `size x price <  spam.order.minimalHoldingQuantumMultiple x quote asset quantum amount` then the order is rejected. Here `size` and `price` are assumed to be correctly scaled by, PDPs and MDPs respectively. 
+
+For pegged orders `last traded price +/- offset` should be used in place of price. The `last traded price` value can be cached from end of last block (if that helps for performance). In opening auctions pegged orders should be rejected.
+
+#### All markets
+
+The following points apply generally to both derivative and spot markets:
+
+- Non-persistent orders are not blocked by the mechanisms described above as long as the market exists.
```
davidsiska-vega: What exactly do we mean by this? Is it that market orders are not subject to the minimum size checks? @cdummett

davidsiska-vega:
```suggestion
```

protocol/0062-SPAM-spam_protection.md
```diff
@@ -128,4 +163,44 @@ More than 360 delegation changes in one epoch (or, respectively, the value of `s
...
+- In a derivatives market, issue a set of orders for an existing, but not yet enacted market, starting with the minimum price, and doubling the order price with every order. Once the first order passes the spam filter, quadruple the parameter `spam.order.minimalMarginQuantumMultiple` and continue. Once the next order passes the filter, quadruple the quantum size for the underlying asset, and continue until an order passes the filter again. Verify that all rejected orders had a margin smaller than `spam.order.minimalMarginQuantumMultiple`, and all accepted ones one bigger or equal. (<a name="0062-SPAM-044" href="#0062-SPAM-044">0062-SPAM-044</a>).
+- Create an order for a non-existing derivatives market, and verify that it is rejected by the spam filter. (<a name="0062-SPAM-045" href="#0062-SPAM-045">0062-SPAM-045</a>).
+- In a derivatives market, create a non-persistent order for an existing market with a minimum order price, and verify that it is not rejected by the spam filter. (<a name="0062-SPAM-046" href="#0062-SPAM-046">0062-SPAM-046</a>).
```
davidsiska-vega:
```suggestion
```

protocol/0062-SPAM-spam_protection.md
```diff
@@ -128,4 +163,44 @@ More than 360 delegation changes in one epoch (or, respectively, the value of `s
...
+- In a derivatives market, order amends are rejected if the order with the new price / size meets *fails to meet* the `spam.order.minimalMarginQuantumMultiple` requirement (<a name="0062-SPAM-052" href="#0062-SPAM-052">0062-SPAM-052</a>).
+
+- In a spot market, issue a set of orders starting with the minimum price, and doubling the order price with every order. Once the first order passes the spam filter, quadruple the parameter `spam.order.minimalHoldingQuantumMultiple` and continue. Once the next order passes the filter, quadruple the quantum size for the underlying asset, and continue until an order passes the filter again. Verify that all rejected orders had a margin smaller than `spam.order.minimalHoldingQuantumMultiple`, and all accepted ones one bigger or equal. (<a name="0062-SPAM-064" href="#0062-SPAM-064">0062-SPAM-064</a>).
+- In a spot market, create a non-persistent order for an existing market with a minimum order price, and verify that it is not rejected by the spam filter. (<a name="0062-SPAM-065" href="#0062-SPAM-065">0062-SPAM-065</a>).
```
davidsiska-vega:
```suggestion
```

protocol/features.json
```diff
@@ -574,6 +574,41 @@
...
+      "0062-SPAM-056",
+      "0062-SPAM-057",
+      "0062-SPAM-058",
+      "0062-SPAM-059",
+      "0062-SPAM-060",
+      "0062-SPAM-061",
+      "0062-SPAM-062",
+      "0062-SPAM-063",
+      "0062-SPAM-064",
+      "0062-SPAM-065",
```
davidsiska-vega:
```suggestion
```

protocol/features.json
```diff
@@ -574,6 +574,41 @@
       "0028-GOVE-185"
     ]
   },
+  "Order spam": {
+    "milestone": "colosseo",
+    "acs": [
+      "0062-SPAM-043", 
+      "0062-SPAM-044",
+      "0062-SPAM-045",
+      "0062-SPAM-046",
```
davidsiska-vega:
```suggestion
```

ze97286 (dismissed):

protocol/0062-SPAM-spam_protection.md
```diff
@@ -83,6 +83,41 @@ Further, each party is allowed to submit up to `n` transactions per epoch where
...
 
+### Transaction Spam
+
+#### Derivative markets
+
+Before any order or liquidity commitment is accepted for a perpetual futures or expiring futures check that the party has `margin + order margin + general > 0` with `margin` being the balance in the margin account for the relevant, `order margin` is the balance for the order margin if party is in isolated margin mode for the for the relevant market and `general` the balance in the general account for the relevant asset. Orders from parties that don't meet this criteria are rejected. This is to be done after the PoW check.
+
+Further, given the network parameter: `spam.order.minimalMarginQuantumMultiple` (between 0 and infinite)
+
+If the maintenance margin for a given transaction is smaller than the parameter `spam.order.minimalMarginQuantumMultiple`, then the transaction is pre-block rejected.
```
EVODelavega, line 94: So there will be cases where `margin + order margin`  could be 0, but the transaction should still be accepted. I'm thinking, especially, about cases where a party has a general balance of 0, and margin balance < initial margin, but >= maintenance level (ie they're close to getting closed out). They might want to submit a transaction closing their current position. In that case the maintenance level for the given transaction == 0.

As per an exceptionally long thread on slack, there are also cases, specifically when dealing with capped futures, where the margin requirement for an order can be 0. This will, most likely cause such orders to be rejected as a rule.

To this end, I think it makes more sense for the markets themselves to receive a call to a function like `OrderAcceptanceLikelihood`, which the order spam component keeps track of over time. Should a given party submit N orders that are below 50% likely to get accepted over the course of N blocks, any order transaction that is less than 50% likely to get accepted should get rejected.

The benefits are that individual markets can determine whether or not, in the context of that market, an order makes more or less sense, and therefore should or should not be accepted, without relying on a single network parameter. Because we can keep a tally of these transaction types over time, we could also adjust the threshold for rejection/acceptance of any given transaction on a per-party basis, based on their behaviour. A party that submits, say, a zero-margin order to a capped market once isn't a worry as such. A party doing so every block, or several times per block, may see their order acceptance threshold increased to 75%, or even 100% (effectively timing them out for N blocks).

protocol/0062-SPAM-spam_protection.md
cdummett: Resolving as PR updated since them and comment outdated.

protocol/0062-SPAM-spam_protection.md
cdummett: Resolving as PR updated to use last traded price as an estimate instead of exact peg.

cdummett (approved):

PR #2191, opened 2024-02-28
feat: cancel pegged orders when updating tick size

cdummett:
## Current

Currently the spec says for pegged orders.

- Round pegged buy orders **UP** to the nearest market tick.
- Round pegged sell orders **DOWN** to the nearest market tick.

This ensured the smallest possible spread (1 tick) can be achieved with pegged orders.

However as orders are not cancelled if the markets tick size is updated through governance and orders are only validated on submission. If the tick size is increased, we can enter a situation where pegged orders now cross.

## Change

When the tick size is updated, all existing pegged orders are checked. If the offset of an order is no longer an integer multiple of the tick size, the order is canceled.


davidsiska-vega (changes requested):
Consider cancelling all pegged orders instead of this.

davidsiska-vega (approved):

Issue #12, opened 2020-03-27
Security review


ashleyvega:
In GitLab by @edd on Jun 4, 2018, 11:41

# General accounts
- [x] Require 2fa on Google accounts (13th June)
- [x] Require 2fa on Gitlab 

# Domains
We should have a registry locked domain on top of Registrar lock (verisign). Suggestion: maybe lets buy vegaprotocol.com, migrate it to safenames then lock it down:
- [ ] Buy vegaprotocol.com
- [ ] Transfer vegaprotocol.com to Safenames.net
- [ ] Turn registry lock on

ashleyvega:
In GitLab by @edd on Jun 4, 2018, 11:44

changed the description

ashleyvega:
In GitLab by @edd on Jun 17, 2019, 17:06

closed

PR #2218, opened 2024-03-21
Feat/liq strat range

### LINKED ISSUE #2211: Network liquidation strat to explicity spec disposal range instead of relying on SLA range
Currently, if the network has a position it will only place liquidation orders within the SLA range. 

This ticket is to de-couple this and have a liquidation range parameter as part of the liquidation strategy inside the market proposal and use this in the same way the SLA range is currently used.

davidsiska-vega:
Closes #2211


protocol/0012-POSR-position_resolution.md
```diff
@@ -110,15 +110,17 @@ Currently only one liquidation strategy is supported and its defined by the foll
...
 
-Assume the price range implied by the `market.liquidity.priceRange` is `[a, b]`. Once the network has worked out a size of its immediate or cancel limit order it sets it's price to `a` if it's a sell order or `b` if it's a buy order, and it submits the order.
+Assume the price range implied by the `disposal slippage range` is `[a, b]`. Once the network has worked out a size of its immediate or cancel limit order it sets its price to `a` if it's a sell order or `b` if it's a buy order, and it submits the order.
 
 Note that setting:
 
 - `disposal time step` = `0s`,
 - `disposal fraction` = `1`,
 - `full disposal size` = `max int`,
+- `disposal slippage range` = `10`,
```
EVODelavega, line 128: > `disposal slippage range` (decimal `>0` with default of `0.1` which is interpreted as `10%`,

If slippage range is 10, doesn't that mean 1000%? I'd imagine setting this to 1 should be enough, but I guess 10 or 100 means we should make sure the slippage doesn't result in a negative value, and setting it to a value > 1 should be fine, and this is just a value to ensure we can trade with the entire opposite side of the book if we have to

davidsiska-vega: Indeed.

davidsiska-vega: I am also ok with `1`; anyway it's just an example and for migration we explicitly say "take the SLA range".

protocol/0012-POSR-position_resolution.md
```diff
@@ -110,15 +111,19 @@ Currently only one liquidation strategy is supported and its defined by the foll
...
+- `max fraction of book side within bounds consumed` (min: `0`, max: `1`): once the network chooses the size of its order (`s_candidate`) the effective size will be calculated as `s_effective=min(m*N, s_candidate)`, where `N` is the sum of volume (on the side of the book with which the network's order will be matching) that falls within the range implied by the `disposal slippage range` and `m` is the `max fraction of book side within liquidity bounds consumed`.
+When vAMMs are implemented and provide liquidity then volume implied by vAMMs that lies within the relevant range must be included in the calculation.
+
+Assume the price range implied by the `disposal slippage range` is `[a, b]`. Once the network has worked out a size of its immediate or cancel limit order it sets its price to `a` if it's a sell order or `b` if it's a buy order, and it submits the order.
```
cdummett: Doest this contradict with the AC added in this [PR](https://github.com/vegaprotocol/specs/pull/2189/files) slightly?(https://github.com/vegaprotocol/specs/pull/2189/files).

> - A network disposal order will not cross with orders outside price monitoring bounds (internally has to behave like an IOC at one tick inside the price monitoring bound). Hence a network disposal cannot trade at a price outside the tightest price monitoring and it won't ever trigger a price monitoring auction. (<a name="0012-POSR-030" href="#0012-POSR-030">0012-POSR-030</a>)

If the price monitoring bounds are tighter should we use those prices so the order isn't stopped if it is large enough to trigger those bounds?

protocol/0012-POSR-position_resolution.md
```diff
@@ -110,15 +111,19 @@ Currently only one liquidation strategy is supported and its defined by the foll
...
 - `full disposal size` (min: `0`, max: `max int`): once net absolute value of network's open volume is at or below that value, the network will attempt to dispose the remaining amount in one go,
-- `max fraction of book side within liquidity bounds consumed` (min: `0`, max: `1`): once the network chooses the size of its order (`s_candidate`) the effective size will be calculated as `s_effective=min(m*N, s_candidate)`, where `N` is the sum of volume (on the side of the book with which the network's order will be matching) that falls within the range implied by the `market.liquidity.priceRange` [parameter](./0044-LIME-lp_mechanics.md#market-parameters) and `m` is the `max fraction of book side within liquidity bounds consumed`.
+- `disposal slippage range` (decimal `>0` with default of `0.1` which is interpreted as `10%`, existing markets to default to their current SLA range). Just like [SLA range](./0044-LIME-lp_mechanics.md) these are taken from `mid_price` during continuous trading or indicative uncrossing price during auctions so the lower bounds becomes `max(0,price x (1-range)` and upper bound `price x (1+range)`.
```
witgaw:
```suggestion
- `disposal slippage range` (decimal `>0` with default of `0.1` which is interpreted as `10%`, existing markets to default to their current SLA range). Just like [SLA range](./0044-LIME-lp_mechanics.md) these are taken from `mid_price` during continuous trading or indicative uncrossing price during auctions so the lower bound becomes `max(0,price x (1-range)` and upper bound `price x (1+range)`.
```

protocol/0012-POSR-position_resolution.md
witgaw: I think that's a good shout, if order violating PM bounds would be allowed to fill partially it'd be fine as is, but since that's not currently the case we should probably account for PM bounds pre-emptively from disposal strategy side.

EVODelavega: Orders that are placed to close out the network position don't pass through the price monitoring checks, and can't trigger price auctions. They also don't factor in to the mark price. They're treated as an always-valid, one off trade that doesn't impact the normal flow of the market.

protocol/0012-POSR-position_resolution.md
EVODelavega: I do think it's worth considering values > 1 here. We should handle cases where we end up with a price of `mid_price - 10 * mid_price` doesn't "overflow"/ignore the sign and results in an order of `9 * mid price`, but instead we should treat it as a de-facto market order

cdummett (approved):

protocol/0012-POSR-position_resolution.md
davidsiska-vega: I've implemented @cdummett 's comment, resolving.

PR #2221, opened 2024-03-22
chore: Add AC for vega issue #10973

gordsport:
Adds and AC for the vegaprotocol/vega#10973 issue seen on testnet


gordsport:
@Sohill-Patel @ze97286 - there are 2 ACs above this new one that have had their codes removed (I cant remember why this was done). 

Can you please review these and see if we should also give them a code?

EVODelavega (dismissed):

PR #2231, opened 2024-04-08
feat: tx trade ordering

### LINKED ISSUE #2232: Add spec for order transaction ordering within blocks
Add a spec for ordering transactions within a block for a given market such that post-only changes and cancellations come first, allowing market makers to offer tighter prices with less risk from toxic flow.

https://github.com/vegaprotocol/specs/pull/2231

TomMcL:
Closes #2232


protocol/0088-TRTO-trading_transaction_ordering.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Execution Ordering
+
+Trading transactions (those which create, update or remove orders of any type on any market) should be executed in a specific way once included in a block.
```
davidsiska-vega: We should probably add that the re-ordering only matters on a per-market basis and furthermore it should be possible to be enabled on a per-market basis.

TomMcL: 👍

protocol/0088-TRTO-trading_transaction_ordering.md
```diff
@@ -0,0 +1,21 @@
...
+
+Furthermore, transactions which can cause a trade by acting aggressively, such as non-post-only orders and amends, will be delayed by one block prior to execution. This results in the pattern where:
+
+ 1. Prior to block N, post only order A and market order B arrive to the chain, these are both included in block N.
+ 1. When block N is executed, order A takes effect.
+ 1. Prior to block N + 1, post only order C then market order D and finally a cancellation of order A arrive to the chain, these are both included in block N + 1.
+ 1. When block N + 1 is executed, order C first takes effect, then the cancellation of order A, then finally market order B.
+ 1. When block N + 2 is executed, market order D takes effect
+
```
Vegaklaus: We should probably add some implementation discussion here to make sure we're all on one page and this doesn't interfere poorly with other (planned) features; my understanding so far:
- delaying by a block is done post-consensus, i.e., a (block) delayed message is put in a special buffer and then added to the next block
- in-block ordering takes both the current block and the delayed transactions, and then re-sorts
- If we add other order policies (e.g., Wendy), this should then use the same code here - otherwise we may end up with conflicting order requirements at different places.
- For now, a transaction needs to be delayed only for one block, but that may change in the future
- Being able to take a message from consensus, put it run a buffer, and execute it later is also a good part of the work for asynchronous execution (i.e., execute the core code while CBft processes the next block)
- We need to make sure the post-consensus transaction validity check is still sufficient if transactions can be delayed more (it should be)

protocol/0088-TRTO-trading_transaction_ordering.md
barnabee: I don't think this is a requirement so we need to asses if this complicates the implementation before committing it.

protocol/0088-TRTO-trading_transaction_ordering.md
```diff
@@ -0,0 +1,41 @@
...
...es arriving and wishing to trade immediately, others have to be willing to place limit orders on the book which will remain there and wait for an incoming order to match with it. These are required for a limit order book to exist (as implied by the name) but expose the party placing the initial limit order to a range of risks, as often the reason these orders will eventually trade is because the price has become unfavourable to the party who placed the order, but they either have not realised or have not had a chance to cancel the order yet. This is often referred to as "toxic flow". Whilst another party obviously gains from this transaction, it is generally acknowledged that the higher a given venue's proportion of toxic flow to more balanced flow, the wider market makers will end up quoting to avoid being the victim of this. As such, exchange and transaction designs which allow for the reduction of this without unfairly impacting other parties using the network may allow for the exchange as a whole to provide traders with better prices than otherwise. This specification covers one such proposed methodology, comprising of a specified ordering of order executions on a given market.
```
barnabee: Should note that the problem is larger in an decentralised environment with both much higher latency than a centralised exchange and a transparent mempool/gossip protocol.

protocol/0088-TRTO-trading_transaction_ordering.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Execution Ordering
+
+Trading transactions (those which create, update or remove orders of any type on any market) should be executed in a specific way once included in a block. This ordering is per-market (inter-market ordering is unspecified) and should be enableable at a market configuration level (i.e. it can be enabled for some markets and not for others). For a market where this is not enabled, all transactions on the market enact in the order specified in the block itself.
```
barnabee: > (inter-market ordering is unspecified)

Inter market ordering would need to remain in the order seen in the block

> This ordering is per-market

TBD, only if doing so can be done in a way that is not significantly more complex than applying the logic network-wide.

protocol/0088-TRTO-trading_transaction_ordering.md
```diff
@@ -0,0 +1,41 @@
...
+
+ 1. Prior to block N, post only order A and market order B arrive to the chain, these are both included in block N.
+ 1. When block N is executed, order A takes effect.
+ 1. Prior to block N + 1, post only order C then market order D and finally a cancellation of order A arrive to the chain, these are both included in block N + 1.
+ 1. When block N + 1 is executed, order C first takes effect, then the cancellation of order A, then finally market order B.
+ 1. When block N + 2 is executed, market order D takes effect
+
+## Batch Transactions
+
+Batch transactions, as they contain different order types, must be split apart for the execution of their different components at the specified times. All cancellations, and any post-only order creations, should be separated from amendments and non-post-only orders. The first group will be executed all together within the first section of the inclusion block, whilst the second group will be executed all together within the second section of the block after the inclusion block.
```
barnabee: A possible simplification would be to assign the whole batch to the lowest priority set of any order in the batch (where P1 is cancellations, P2 post-only orders, and P3 anything else), and execute the whole thing then. This would post no issues that I can see.

protocol/0088-TRTO-trading_transaction_ordering.md
```diff
@@ -0,0 +1,41 @@
...
+
+Furthermore, transactions which can cause a trade by acting aggressively, such as non-post-only orders and amends, will be delayed by one block prior to execution. This results in the pattern where:
+
+ 1. Prior to block N, post only order A and market order B arrive to the chain, these are both included in block N.
+ 1. When block N is executed, order A takes effect.
+ 1. Prior to block N + 1, post only order C then market order D and finally a cancellation of order A arrive to the chain, these are both included in block N + 1.
+ 1. When block N + 1 is executed, order C first takes effect, then the cancellation of order A, then finally market order B.
```
barnabee: Would it be better to do cancellations before post only orders?

Two reasons come to mind:

1) performance, in case the post onlies are pegs and cancellations cause them to need repricing immediately after they are initially priced and placed on the book;l and
2) post only may be rejected because it should trade with something that's about to be cancelled, so you'd sometimes get better results doing cancels first.

barnabee (dismissed):

Vegaklaus (commented):
Some comments on the implementation (given that interacts with other planned features)
It would be nice if the delay function could also delay by several blocks, as that would be halt of the integration of order policies in Wendy. I.e., a flow like
0. Keep an extra delay-buffer with delayed transactions and some meta data (e.g., how long has this transaction been delayed yet; can be more in the future)
1. Increase the 'has-been-delayed' counter of all transactions by one
2. Take all transactions out of the tender mint buffer into the delay-buffer with 'has-been-dalayed'=0
3. [For future optimisations: Release CBFT to process the next block to parallelise block evaluation with the consensus. Assure that  CBFT is only ahead by one block to avoid too much asynchrony]
4. For all transactions in the delay buffer, run a function that verifies if they can leave the buffer now. This implements only the function from this spec now, but can be extended by additional policies that also may be market-dependent
5. Remove all functions that can be delivered from the delay buffer, and sort them by the second ordering function (which also might be extended in the future)
6. Run a transaction validity verification on every transaction again (as the state of the system may have changed in the meantime), and feed all valid ones into core.
7. If a transaction has been delayed too long (e.g., 50 blocks), it is dropped (or immediately delivered ,that's up to discussion)

davidsiska-vega:
Do we need a netparam to enable / disable this re-ordering? @barnabee / @TomMcL

barnabee:
> Do we need a netparam to enable / disable this re-ordering? @barnabee / @TomMcL

Ideally yes. I know an early version of the spec had this per market, which is most ideal but we were also willing to drop that if it added excessive complexity and just have a network wide switch.

I would assume "disabling" this (whether per-market or net wide) would basically mean something like treating every order/market instruction for which the feature is disabled as if it was given the high priority, or some equivalent.

Probably need @ze97286 input here as to what is built and what makes sense…

protocol/0028-GOVE-governance.md
```diff
@@ -602,6 +604,7 @@ APIs should also exist for clients to:
 - A market change proposal specifying a new tick size less than or equal to `0` gets rejected (<a name="0028-GOVE-184" href="#0028-GOVE-184">0028-GOVE-184</a>).
 - At enactment, a market change proposal updating the tick size cancels all pegged orders where their offset is no longer an exact integer multiple of the tick size (<a name="0028-GOVE-183" href="#0028-GOVE-183">0028-GOVE-183</a>).
 - A market LP with ELS > 0 can vote on a market change proposal even if the key doesn't meet the `governance.proposal.updateMarket.minVoterBalance` for governance token.  (<a name="0028-GOVE-185" href="#0028-GOVE-185">0028-GOVE-185</a>).
+- A market amended to enable or disable `Transaction Prioritisation` will have that take effect immediately upon enactment. (<a name="0028-GOVE-188" href="#0028-GOVE-188">0028-GOVE-188</a>)
```
davidsiska-vega:
```suggestion
- An amendment to a market to enable or disable `Transaction Prioritisation` will have that effect immediately upon enactment. (<a name="0028-GOVE-188" href="#0028-GOVE-188">0028-GOVE-188</a>)
```

davidsiska-vega (dismissed):

davidsiska-vega (approved):

PR #2241, opened 2024-04-15
feat: make relative return reward include losses

### LINKED ISSUE #2238: [Feature]: Make relative return reward include losses
# Feature Overview

**In order to** improve the current games functionality
**We will** look into these hopefully "quick fixes"
**So that** people find the games more engaging

## Tasks

- [ ] Adjust the `relativeReturn` reward type so that instead of disregarding any key with a loss we rank ALL eligible keys in terms of P&L and reward them.  This would allow the game to account for the situation where the market turns and everyone is losing, but some are losing "better" than others.
  - [ ] For a pro rata game this would mean rewarding all eligible keys some amount
  - [ ] For a ranked game this would mean that keys in loss can still be in the ranking table

## Feature test scenarios
Detailed scenarios that can be executed as feature tests to verify that the feature has been implemented as expected.

GIVEN (setup/context) 
WHEN (action) 
THEN (assertion) For example...
See [here](https://github.com/vegaprotocol/vega/tree/develop/core/integration) for more format information and examples.

### Additional Details (optional)
Any additional information that provides context or gives information that will help us develop the feature.

### LINKED ISSUE #2240: [Feature]: Make relative PNL reward only reward realised PnL
# Feature Overview

**In order to** improve the current games functionality
**We will** look into these hopefully "quick fixes"
**So that** people find the games more engaging

## Tasks

- [ ] Change the `relativePnL` game to be only realised PnL (or create a new reward type to cover this)

And alternative approach might be to 

- [ ] Add a `runningVolumeRequirement` so that in addition to the notional requirement reward creators can ensure that in order to qualify for a reward users must be actively trading and not just sitting on positions

But that will still be confusing because users will find it hard to track if they have done volume until vegaprotocol/specs#2239 is done ... and even then it is easy to game by just trading back and forth a few times on anotehr market while sitting on a position in another. So we favour optioin 1.

## Feature test scenarios
Detailed scenarios that can be executed as feature tests to verify that the feature has been implemented as expected.

GIVEN (setup/context) 
WHEN (action) 
THEN (assertion) For example...
See [here](https://github.com/vegaprotocol/vega/tree/develop/core/integration) for more format information and examples.

### Additional Details (optional)
Any additional information that provides context or gives information that will help us develop the feature.

cdummett:
closes #2238


Jiajia-Cui (dismissed):

protocol/0056-REWA-rewards_overview.md
```diff
@@ -205,6 +205,8 @@ Let:
 - $M_{i}$ be the sum of all reward payout multipliers for entity $i$ (reward payout multipliers include the [activity streak multiplier](./0086-ASPR-activity_streak_program.md#applying-the-activity-reward-multiplier) and [bonus rewards multiplier](./0085-RVST-rewards_vesting.md#determining-the-rewards-bonus-multiplier)).
 - $s_{i}$ be the share of the rewards for entity $i$
 
+NOTE: As reward metrics can be negative (e.g. a party has negative relative returns, all reward metrics must be offset by the lowest reward metric to ensure all metrics are positive before calculating each parties share of the rewards )
```
ze97286: a few comments on that @cdummett:
1. I think a more sensible way to do this would be to shift the range of values to [1,2] 1 + (x-min)/(max-min) - with what you suggested the most negative one become zero and gets nothing - however this is also skewing the relations between the scores. E.g. if we have 3 scores 1, 0.5, 0.2 - this translates to 2, 1.375,1 which is not equivalent. Another option is just say we adjust by the max negative, and this guy doesn't get a reward, unless they are alone and then they do get a reward. 
2. depending on what, when, and if we're publishing the value of this score it may become (even more) confusing. Now there will be 3 levels for this score - the raw, actual value which may be positive or negative, the min-max normalised, i.e. projected to the [1,2] range, and then the normalised by total which is the one actually used for the reward calculation. I don't know if/that we're publishing any of the above so not sure if it's a problem but may add a bit of confusion.

cdummett: For point 1.

> shift the range of values to [1,2]

I think this one is undesirable because of the point you mentioned, keeping the relations between the scores is quite important. If we normalise into that range the best performing party can at best receive twice the rewards of the worst performing party?

> Another option is just say we adjust by the max negative, and this guy doesn't get a reward, unless they are alone and then they do get a reward.

This is the approach I was trying to describe in the spec but wasn't clear enough. Agree if all the scores are positive we shouldn't shift all the scores down resulting in the lowest score being zero 👍 

Will update the spec to make it clearer if everyone is happy that offsetting scores by the lowest negative score is the right approach.

cdummett: For point 2.

If we do end up publishing scores, I think it should be the actual value.

e.g. for relative returns, this would be the average returns across the previous n epochs and would consider all the markets within the scope of the dispatch strategy (it could be -ve or +ve).

Any offsetting (depending on the approach we choose from point 1) and the calculation of each parties pro-rata share of the rewards would be done only at the end of the epoch and I don't think we would need to publish this "share".

ze97286: regarding point 1, the reason I think it's a bit unfair is that suppose you have two parties, one got a score of -10 the other got a score of -9.99999999 and there are only them qualifying for the reward - the one with -10 will get 0% of the reward, the one with -9.9999999 will get 100% of the reward. That seems a bit odd at least. But I'm ok with it.

Jiajia-Cui (approved):

PR #2242, opened 2024-04-15
feat: publish ongoing game score snapshots during epochs

### LINKED ISSUE #2239: [Feature]: Publish ongoing game score snapshots during epochs
# Feature Overview

**In order to** make games more fun
**We will** work out how to publish snapshots of running game scores at intervals in the epoch
**So that** users can see how they are doing rather than just waiting until the very end to see what they have won 

## Proposal 

Can we define some "calculation interval" or something on the game definition?  We already have a "window length" which sets how many epochs the game is ultimately calculated over.  We need some other parameter that allows us to get some interim snapshots of the data during the game.  Maybe this calculation interval could be enforced to be 0 < x < 1 so that it is always some fraction of an epoch? Then we could work out what the min value of this is to avoid any kind of performance concerns? 

## Acceptance Criteria
- [ ] I can get a view of my eligibilty across all metrics during the in-progress epoch
  - [ ] Open volume (whether we're meeting `notionalTimeWeightedAveragePositionRequirement`)
  - [ ] ~Traded volume (since soon we should be adding a `tradedVolumeRequirement`)~
  - [ ] Total fees paid this epoch (since there is discussion on whether the reward cap could be used here too)
- [ ] I can get a view of my my current place in the ranking for the reward metric

## Feature test scenarios
Detailed scenarios that can be executed as feature tests to verify that the feature has been implemented as expected.

GIVEN (setup/context) 
WHEN (action) 
THEN (assertion) For example...
See [here](https://github.com/vegaprotocol/vega/tree/develop/core/integration) for more format information and examples.

### Additional Details (optional)
Any additional information that provides context or gives information that will help us develop the feature.

cdummett:
closes #2239


protocol/0056-REWA-rewards_overview.md
```diff
@@ -42,6 +41,32 @@ For reward metrics relating to trading, an individual must meet the [staking req
 
 For reward transfers where the [scope](./0057-TRAN-transfers.md#recurring-transfers-to-reward-accounts) is set to teams, each party must meet the minimum time in team requirement. That is, given a party has been in a team for $N$ epochs, if $N$ is strictly less than the network parameter `rewards.minimumEpochsInTeam` (an integer defaulting to `0`) their reward metric is set to `0`.
 
+## Publishing Interim Reward Information
+
+Although rewards are distributed at the end of an epoch, to give users of the protocol an understanding of what rewards they are likely to receive at the next distribution event, the protocol must publish interim data during the epoch to populate data nodes.
+
+For each game, each entities (individual or team) **current reward metric** and rank in the game is to be published at the specified frequency. Note, the actual reward metric should be published rather than the underlying data in the current epoch, this is important for games where the window length is greater than `1`.
```
davidsiska-vega: @cdummett I am not seeing where we specify the frequency, probably should be a network parameter so we can fine-tune the load this creates on the data node.

davidsiska-vega (changes requested):

protocol/0056-REWA-rewards_overview.md
cdummett: The frequencies are in the table below as per what we decided in the research hour.

We discussed adding a network parameter for any reward metrics which were only updated at the end of an epoch but having double checked these (`Liquidity Fees`, `Market Creation`, `Validator Ranking`) the underlying events for these rewards only happen at the end of the epoch anyway so we can't calculate them at a faster frequency.

Jiajia-Cui (approved):

davidsiska-vega (approved):

Issue #2230, opened 2024-04-08
Prediction markets / capped payout future


barnabee:
This ticket is to write a spec. that would enable users of Vega to create markets like prediction markets, etc. in which the market settles between (e.g.) 100 and 0 depending on whether an outcome occurs or does not occur, respectively. The price of the market would therefore reflect the market view of the probability of the outcome, and the spread and volatility would reflect (among other things) the level of uncertainty.

This would be based on the existing derivatives functionality and would add parameters to:

1. Allow apecifying a maximum price/quote as well as the implicit minimum of zero. For example, 1 or 100. 
    - This is pretty much just validation and would mean that orders could only be submitted for prices 0 ≤ price ≤ MAX, and any settlement price > MAX would beset to = MAX. (Users wishing to filter values above MAX can do so in the oracle framework.)
    - We would need to ensure the test suite fully covers trading and settling at prices of both zero and MAX.
    - This feature should ideally be able to be used with or without full collateralisation, however if there is a significant benefit to "special casing" this than that could be done.
2. Add a new risk model _Fully Collateralised_.
    - This would be available to be used only when the max price/quote feature above is used.
    - When set, longs would need to provide `position size × average_entry_price` in collateral, and shorts would need to provide `position_size × (max_price - average_entry_price)`. 
    - Isolated margin and cross margin would be the same with this risk model and would both require the full amount of margin at all times, including being able to cover the full worst-case for open orders. 
    - Similarly, there would be no move in a traders' favour in which margin was released, or against them in which more margin was searched. The position would always be fully collateralised.

These should otherwise work like any other product. Price monitoring, LP SLA protocol, etc. would be possible and available. Some, like price monitoring may not need to be enabled where full collateralisation is used. Mark-to-market, etc. would happen using all the same methods that are available now. 

It would be preferable architecturally, and to enable maximum experimentaton and customisation for the max. price and fully collateralised risk model to work equally for futures and perpetuals, however the hard requirement is that they would need to at least be available for vash settled futures.

We will need to consider how features that are tied to the selected risk model operate where they are not compatible with fujll collateralisation (which does not include/involve a statistical risk model). They can use a separately speciifed log-normal model, or some other weighting function, or potentially be turned off.

We will need to think about pegged orders, such that they can never create a price outside the range 0–max.

[...]

PR #2251, opened 2024-04-22
Capped futures

witgaw:
Minimal spec changes required to allow creation of fully-collateralised binary options markets.

The main things left to decide:
1.  ~~If we want to allow capping of perps?~~  No.

2. ~~If we go with 1. do we want to allow fully-collateralised mode with perps?~~  No.

Closes https://github.com/vegaprotocol/specs/issues/2230

### DIFF
--- a/protocol/0019-MCAL-margin_calculator.md
+++ b/protocol/0019-MCAL-margin_calculator.md
@@ -425,6 +425,16 @@ There should be an additional amount `limit price x size x margin factor = 15910
 - All order margin balances are restored after a protocol upgrade (<a name="0019-MCAL-152" href="#0019-MCAL-152">0019-MCAL-152</a>).
 - The `margin mode` and `margin factor` of any given party must be preserved after a protocol upgrade (<a name="0019-MCAL-153" href="#0019-MCAL-153">0019-MCAL-153</a>).
 
+## Acceptance Criteria (Fully collateralised mode)
+
+Assume a [capped future](./0093-CFUT-product_builtin_capped_future.md) market with a `max price = 100` and mark-to-market cashflows being exchanged every block:
+
+- Party A posts an order to buy `10` contracts at a price of `30`, there's no other volume in that price range so the order lands on the book and the  maintenance and initial margin levels for the party and order margin account balance are all equal to `300`. (<a name="0019-MCAL-154" href="#0019-MCAL-154">0019-MCAL-154</a>)
+- Party B posts an order to sell `15` contracts at a price of `20`, a trade is generated for `10` contracts at price of `30` with party A. The maintenance and initial margin levels party A remains at `300`, order margin account balance is now `0` and margin account balance is `300`, the position is `10` and there are no open orders. The  maintenance and initial margin levels for party B are equal to `10 * (100 - 30) + 5 * (100 - 20) = 1100`, the margin account balance is `700`, order margin account balance is `400`, the position is `-10` and the remaining volume on the book from this party is `5` at a price of `20`. (<a name="0019-MCAL-155" href="#0019-MCAL-155">0019-MCAL-155</a>)
+- Party B posts an order to buy `10` contracts at a price of `18`, the orders get placed on the book and margin levels as well margin account balances and position remain unchanged. (<a name="0019-MCAL-156" href="#0019-MCAL-156">0019-MCAL-156</a>)
+- Party B posts an order to buy `30` contracts at a price of `16`, the orders get placed on the book, the maintenance and initial margin levels for party B grow to `1180`, and the margin account balance remains unchanged at `700` and the order margin account balance grows to `480 = max (5 * (100 - 20), 30 * 16)`. The position remains unchanged at `-10`. (<a name="0019-MCAL-157" href="#0019-MCAL-157">0019-MCAL-157</a>)
+- Party A posts an order to sell `20` contracts at a price of `17`. A trade is generated for `10` contracts at a price of `18` with party B. A sell order for `10` contracts at a price of `17` from party A gets added to the book. The maintenance and initial margin levels for party A is now `10 * (100 - 17) = 830`, the position is `0` and the remaining volume on the book from this party is `10` at a price of `18`. Party A lost `120` on its position, hence `830 - (300 - 120) = 410` additional funds get moved from the general account as part of the transaction which submitted the order to sell `20` at `17`. Party B now has a position of `0` and following orders open on the book: sell `5` at `20` and buy `30` at `16`. The maintenance and initial margin levels are `max(5 * (100 - 20), 30 * 16) = 480`. The margin account momentarily becomes `820` (`700` + `120` of gains from the now closed position of `-10`), order margin account balance is `480`, hence `820` gets released back into the general account and margin account becomes `0`. (<a name="0019-MCAL-158" href="#0019-MCAL-158">0019-MCAL-158</a>)
+
 ## Summary
 
 The *margin calculator* returns the set of margin levels for a given *actual position*, along with the amount of additional margin (if any) required to support the party's *potential position* (i.e. active orders including any that are parked/untriggered/undeployed).
@@ -432,7 +442,9 @@ The *margin calculator* returns the set of margin levels for a given *actual pos
 
 ### Margining modes
 
-The system can operate in one of two margining modes for each position.
+#### Partially-collateralised
+
+The system can operate in one of two partially-collateralised margining modes for each position.
 The current mode will be stored alongside of party's position record.
 
 1. **Cross-margin mode (default)**: this is the mode used by all newly created positions.
@@ -443,6 +455,23 @@ In this mode, the entire margin for any newly opened position volume is transfer
 This includes completely new positions and increases to position size. Other than at time of future trades, the general account will then
 *never* be searched for additional funds (a position will be allowed to be closed out instead), nor will profits be moved into the
 general account from the margin account.
+
+#### Fully-collateralised
+
+For certain derivatives markets it may be possible to collateralise the position in full so that there's no default risk for any party.
+
+If a product specifies an upper bound on price (`max price`) (e.g. [capped future](./0093-CFUT-product_builtin_capped_future.md)) then a fully-collateralised [wrapped risk model](./0018-RSKM-quant_risk_models.ipynb) can be specified for the market. If such a risk model is chosen then, it's mandatory for all parties (it's not possible to self-select any of the above partially-collateralised margining modes).
+
+In this mode long positions provide `position size * average entry price` in initial margin, whereas shorts provide `postion size * (max price - average entry price)`. The initial margin level is only re-evaluated when party changes their position. The [mark-to-market](./0003-MTMK-mark_to_market_settlement.md) is carried out as usual. Maintenance and initial margin levels should be set to the same value.  Margin search and release levels are set to `0` and never used.
+
+In this mode it is not possible for a party to be liquidated. Even if the price moves to the extremes of zero or the `max price` and parties may therefore have zero in their margin account, the parties must not be liquidated. Fully collateralised means that the posted collateral explicitly covers all eventualities and positions will only be closed at final settlement at maturity.
+
+Order margin is calculated as per [isolated margin mode](#placing-an-order) with:
+
+- `side margin = limit price * size` for buy orders
+- `side margin = (max price - limit price) * size` for sell orders.
+
+Same calculation should be applied during auction (unlike in isolated margin mode).
 
 ### Actual position margin levels
 
--- a/protocol/0016-PFUT-product_builtin_future.md
+++ b/protocol/0016-PFUT-product_builtin_future.md
@@ -14,6 +14,13 @@ Futures are a simple "delta one" product and the first product supported by Vega
 
 Validation: none required as these are validated by the asset and data source frameworks.
 
+Optional parameters:
+
+1. `max_price`: specifies the price cap for the market, an integer understood in the context of market decimal places,
+1. `binary_settlement`: if set to `true` settlement price other than `0` or `max_price` will be ignored.
+
+Validation: `max_price` > 0.
+
 ## 2. Settlement assets
 
 1. Returns `[cash_settled_future.settlement_asset]`.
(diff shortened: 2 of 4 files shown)


TomMcL (approved):

protocol/0047-DSRF-data_source_filter.md
```diff
@@ -60,6 +60,8 @@ To specify a filtered data source the following parameters can be specified:
 
 - `data`: (required) another data source definition defining the input data
 - `filters`: (required) a list of _at least one_ filter to apply to the data
+- `data_if_true`: (optional) any data source, if specified the latest value received from that downstream data source gets returned from the current data source whenever the result of applying `filters` to `data` is not null, if downstream data source has not been specified then the result of applying `filters` to `data` gets returned directly.
+- `data_if_false`: (optional) any data source, if specified the latest value received from that downstream data source gets returned from the current data source whenever the result of applying `filters` to `data` is null, if downstream data source has not been specified then nothing gets returned.
```
witgaw: @vegaprotocol/core please note that this is what seemed to be the easiest change from spec point of view to achieve the following functionality:
* the marke settles at `max price` if `settlement price` >= `max price`,
* the market settles at `0` if `settlement price < max price`.

But I don't think this spec accurately represents the implementation so definitely open to suggestions on how to achieve the functionality specified above with least amount of core work while maintaining a sensible UX.

wwestgarth: I don't think doing it this would would be much core work, maybe how we represent that this happened in the data-node would probably be the more fiddly bit. I might suggest changing the names to `override_if_XXX` instead.

I also can't help feel that this logic should be in the product code though? So the product gets the data-source as usual and then the code in the product just compares the settlement price in the data-source to `max price` and makes the decision itself. Because doing that would just be a couple of lines of Go and avoids adding complexity to the data-source definitions.

wwestgarth: So the idea is to have a data-source, with two other internal data-sources where the result of whether the first incoming data source filters or not causes core to trigger one of the internal data-sources?

Can this logic not be in the product code itself? So a data-source comes in as normal and then a few lines of Go compare the settlement price to `max_price` and make a decision itself? or is it that the data-source _won't_ be of the form of a settlement price and thats why we need to "override" it with something that is a settlemet price?

witgaw: It's more that we wanted to both:
-  allow experimentaion (have a capped future that behaves exactly as a regular fixed-expiry future, but all prices are capped)
-  as well as be able to reproduce binary options markets: the market settles either at `0` or `max_price` and nothing in between.

witgaw: Though I guess something like `allow_intermediate_settlement_price` flag in the product itself would do it.

barnabee: I am not super comfortable having this outside data sourcing because all views and APIs will report the data source value as the settlement price. Any tooling, fornt ends, etc. would have to special-case this. The settlement data source should match the price used to settle a future or perp IMO.

So I think it'd be good to dig into why we are proposing to build it this way instead of in the data source pipeline?

wwestgarth: For perpetual funding-rates we put all the "processed" data from the data-sources in the market-data (external/internal TWAP calc etc.) so I don't see why we can't have a `data-source -> [product does stuff with it] -> result in market-data` flow here too. So when a market settles you ask the market-data API what the settlement price was.

My reason for not being too keen on changing the data-source framework is because its already complicated, and having this indirection of "a data-source comes in that triggers another data-source that actually gets sent in" is going to add to that.

What are the expected data-sources for this binary markets, and is there a reason can't just emit the data in a form that it is expected to be in i.e `0` or `max-price`?

witgaw: Any external oracle. I think this needs to be covered within the system so parties don't have to take on a risk of a payout they aren't expecting (so they don't need to trust that the oracle emits either `0` or `max-price`, but it's guaranteed by the product).

barnabee: > What are the expected data-sources for this binary markets, 

Any existing DEX on Ethereum/Gnosis/Arbitrum, price oracles like Pyth, and truth oracles like UMA. We would expect to be able to settle directly using the data and also off the answers to questions about the current and historical data.  

> and is there a reason can't just emit the data in a form that it is expected to be in i.e 0 or max-price?

I think this is actually the right question. We have the Eth oracle framework so data sourcing logic that ***should*** be in Vega can be built into wrapper contracts until we fix data sourcing properly.

So I guess the `allow_intermediate_settlement_price` flag will do fine. 

I am not so sure about making any intermediate number 0 otherwise, not about capping the settlement price to MAX. I am trying to to come up with a reason not to rejected invalid settlement data that's bigger than MAX, or between 0 and MAX if the intermediate price flag is false rather than changing the value?

witgaw: My motivation was so that it's easy to allow markets "like will BTC price a week from now be >70k" without any extra oracle setup using fairly simple oracles that emit a price (something that we'd use for settling a vanilla fixed-expiry futures market).

Though we can go for a cleaner setup of either :
1. allow 0<=x<=max and ignore anything else
2. allow only 0 or max and ignore anything else

depending on the flag and defer that more complicated logic to oracle itself / WASM / specialised pre-WASM oracle component should such a need arise from community.

protocol/0093-CFUT-product_builtin_capped_future.md
```diff
@@ -0,0 +1,63 @@
...
+- mark prices,
+- final settlement price.
+
+The `max_price` is set at market creation time and it's immutable throughout the market's lifecycle.
+
+## 1. Product parameters
+
+Capped future should have all the product parameters and behaviour of a standard fixed-expiry cash settled [future](./0016-PFUT-product_builtin_future.md#1-product-parameters) as well as: 
+
+1. `max_price`: specifies the price cap for the market, specified formatted with market decimal places,
```
davidsiska-vega:
```suggestion
1. `max_price`: specifies the price cap for the market, an integer understood in the context of market decimal places,
```

protocol/0093-CFUT-product_builtin_capped_future.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Acceptance Criteria
+
+- Attempt to create a capped futures market with a price cap of `0` fails. (<a name="0093-CFUT-001" href="#0093-CFUT-001">0093-CFUT-001</a>)
+
+- Order with a `price > max_price` gets rejected. (<a name="0093-CFUT-002" href="#0093-CFUT-002">0093-CFUT-002</a>)
+
+- When the reference of a pegged sell order moves so that the implied order price is higher than `max_price` the implied order price gets capped at. `max_price` (<a name="0093-CFUT-003" href="#0093-CFUT-003">0093-CFUT-003</a>)
+
+- When market is setup to use oracle based mark price and the value received from oracle is less than `max_price` then it gets used as is and mark-to-market flows are calculated according to that price. (<a name="0093-CFUT-004" href="#0093-CFUT-004">0093-CFUT-004</a>)
```
Jiajia-Cui:
```suggestion
- When market is setup to use oracle based mark price and the value received from oracle is less than `max_price` then it gets used as it is and mark-to-market flows are calculated according to that price. (<a name="0093-CFUT-004" href="#0093-CFUT-004">0093-CFUT-004</a>)
```

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -444,6 +456,21 @@ This includes completely new positions and increases to position size. Other tha
...
 general account from the margin account.
 
+#### Fully-collateralised
+
+For certain derivatives markets it may be possible to collateralise the position in full so that there's no default risk for any party.
+
+If a product specifies an upper bound on price (`max price`) (e.g. [capped future](./0093-CFUT-product_builtin_capped_future.md)) then a fully-collateralised mode can be chosen for the market. This mode is to be selected at market creation time and, if chosen, it's mandatory for all parties (it's not possible to self-select any of the above partially-collateralised margining modes).
+
+In this mode long positions provide `position size * average entry price` in initial margin, whereas shorts provide `postion size * (max price - average entry price)`. The initial margin level is only re-evaluated when party changes their position. The [mark-to-market](./0003-MTMK-mark_to_market_settlement.md) is carried out as usual. Maintenance and initial margin levels should be set to the same value.  Margin search and release levels are set to `0` and never used.
+
```
barnabee, line 466:
```suggestion

In this mode it is not possible for a party to be liquidated. Even if the price moves to the extremes of zero or the `max price` and parties may therefore have zero in their margin account, the parties must not be liquidated. Fully collateralised means that the posted collateral explicitly covers all eventualities and positions will only be closed at final settlement at maturity.

```

protocol/0093-CFUT-product_builtin_capped_future.md
```diff
@@ -0,0 +1,63 @@
+# Built-in [Product](./0051-PROD-product.md): Capped future
```
barnabee: Note that this is expected to be implemented by adding two parameters two the standard Future product. These parameters will be such that leaving them out (i.e. setting them to default values) will result in a standard future.

There won't be a separate product. It would make sense to rework the content in this file to align with this approach.

protocol/0093-CFUT-product_builtin_capped_future.md
```diff
@@ -0,0 +1,63 @@
...
+
+The `max_price` is set at market creation time and it's immutable throughout the market's lifecycle.
+
+## 1. Product parameters
+
+Capped future should have all the product parameters and behaviour of a standard fixed-expiry cash settled [future](./0016-PFUT-product_builtin_future.md#1-product-parameters) as well as: 
+
+1. `max_price`: specifies the price cap for the market, an integer understood in the context of market decimal places,
+1. `binary_settlement`: if set to `true` settlement price other than `0` or `max_price` will be ignored,
+1. `fully_collateralised`: if set to `true` the [fully collateralised margin mode](./0019-MCAL-margin_calculator.md#fully-collateralised) is used.
```
barnabee: Ideally fully collateralised would not be a true/false flag but an alternative risk calculator / risk model that can be specified instead of log normal. (It won't need any parameters.)

**_Unless there's a technical implementation reason from @jeremyletang that we won't be doing that_** then we should describe it that way rather than as a separate parameter. This would avoid any ambiguity and any idea that a risk model could also be specified. If a log normal model is required for other aspects of the protocol it would be preferable for that to be done via separate parameters.

protocol/0093-CFUT-product_builtin_capped_future.md
```diff
@@ -0,0 +1,63 @@
...
+
+## 2. Additional considerations
+
+### 2.1 Order price validation
+
+In order to aid smooth operation of the market and safeguard against user errors or manipulation attempts order prices should be validated so that the maximum price of any order is `max_price`. Peg orders should be validated so that if the resulting price would be higher at any point it gets temporarily capped at `max_price`.
+
+### 2.2 Mark price validation
+
+Irrespective of how the mark price is obtained in the market, the final value to be applied used in the [mark-to-market](./0003-MTMK-mark_to_market_settlement.md) calculation must be capped so that `mark_price = min(max_price, mark_price_candidate)`.
```
barnabee: Capping is inconsistent with ignoring out of bounds values below. Given we changed those recently I wanted to check we definitely want to cap here?

protocol/0093-CFUT-product_builtin_capped_future.md
```diff
@@ -0,0 +1,63 @@
...
+
+- When market is setup with `fully_collateralised = true` option and it has parties with open positions settling it at a price of `max_price` works correctly and the sum of all final settlement cashflows equals 0 (loss socialisation does not happen). Assuming general account balances of all parties were `0` after opening the positions and all of their funds were in the margin accounts: long parties end up with balances equal to `position size * max_price` and short parties end up with `0` balances. (<a name="0093-CFUT-009" href="#0093-CFUT-009">0093-CFUT-009</a>)
+
+- When market is setup with `fully_collateralised = true` option and it has parties with open positions settling it at a price of `0` works correctly and the sum of all final settlement cashflows equals 0 (loss socialisation does not happen). Assuming general account balances of all parties were `0` after opening the positions and all of their funds were in the margin accounts: short parties end up with balances equal to `abs(position size) * max_price` and long parties end up with `0` balances. (<a name="0093-CFUT-010" href="#0093-CFUT-010">0093-CFUT-010</a>)
```
barnabee: TODO: add scenarious for the mark price being 0 and max price during trading and no close outs occurring

barnabee (changes requested):

protocol/0093-CFUT-product_builtin_capped_future.md
witgaw: Risk model is still required for price monitoring bounds and liquidity score calculation at the moment. 

How would that play out with keeping it all within standard futures product? We'd have to specify risk models separately depending on use (margin calculation, price monitoring bounds, liquidity score)?

protocol/0093-CFUT-product_builtin_capped_future.md
witgaw: I don't have a strong opinion, ignoring it sounds fine too, I'll change it to that.

protocol/0093-CFUT-product_builtin_capped_future.md
witgaw: Sure, I'll move it to a futures product spec.

protocol/0093-CFUT-product_builtin_capped_future.md
witgaw: Added these now

protocol/0093-CFUT-product_builtin_capped_future.md
barnabee: Yeah

So from an (idealised, conceptual…) architectural perspective, at least as imagined to date, "risk model" (or "risk calculator" or whatever we want to call it) attached to an instrument to make it tradable is the calculation logic that is used to determine how much margin is required for a [real or potential] position.

That means that it is conceptually something like this:

```
Market {
    product: Future {
        … future params …
    }
    risk_model: LogNormal {    // or FullyCollateralised { }
        … risk model params …
    }
    extensions: [
        PriceMonitoring {
            // optional: model: SomeModel { … and params … } to not use the market's default
            … price monitoring params …
        }
        LiquiditySla {
            // optional: model: SomeModel { … and params … } to not use the market's default
            … liquidity SLA params …
        }
    ]
}
```

…where the market can't exist without a way to calculate margin, but could theoretically:

* have no price monitoring or no liquidity protocol;
* have them but use different models/calculators for each; or even
* have price monitoring or liquidity protocols that use some other calculation (e.g. simple linear maths and specific parameters) rather than a risk model of the type that powers margin calcs and provides probabilities of price moves.

_NB:_ price monitoring and liquidity SLA live under `extensions` here because conceptually they are blobs of code that hook into product and market events (lifecycle events, trades, price moves, external price data, order book changes, etc.) and iteract with the market via an API to do things like change its trading mode (suspend/resume, enter/extend auction, end auction, block trade(s), cancel order(s), etc.). There could be many other such blobs in future providing extra protections, additional reward schemes, or whatever.

Currently it is **implicit by the naming** (_risk model_ rather than, say, _margin model_) that the market has a single risk model, and that is also **de facto the case in implementation**, because none of the modules that require such calculation allow specifying their own.

It would be possible, and is likely a good idea, to make this more explicit in future, with markets being more like:

```
Market {
    product: Future {
        … future params …
    }
    margin_calculator: RiskBasedMargin {    // where you could also have, e.g.: FixedMargin { ratio = 0.1 }
       model: LogNormal {    // or FullyCollateralised { }
            … risk model params …
        }
    }
    extensions: [
        RiskBasedPriceMonitoring {
            model: SomeModel { … and params … }    // no longer optional
            … other price monitoring params …
        }
        RiskBasedLiquiditySla {
            model: SomeModel { … and params … }    // no longer optional
            … liquidity SLA params …
        }
    ]
}
```


Of course, if we did that then the next step, for some markets would probably be to remove the risk model entirely, as for [at least higher liquidity/volume, and maybe all] delta one markets it's not really needed!

```
Market {
    product: Future {
        … future params …
    }
    margin_calculator: FixedMargin {
        ratio = 0.025    // 40x leverage
    }
    extensions: [
        PriceMonitoring {
            triggers: [
                … trigger definitions use simple relative moves in a horizon, or similar …
            ]
        }
        LiquiditySla {
            … liquidity SLA params specify requirements and fee distribution without ref. to a risk model …
        }
    ]
}
```

That's probably something we want to be able to do in the not too distant future.

As for what that exactly means for this specification, I am not entirely sure. But I do think we should try not to  end up keeping the market-wide risk model in a situation where it's used for everything _**except**_ margin calculation, which is special cased by a flag. That doesn't seem great.

barnabee (approved):

PR #2263, opened 2024-04-29
chore: fix typos and add in sqrt's to align with notebook

wwestgarth:


davidsiska-vega (approved):

protocol/0090-VAMM-automated_market_maker.md
```diff
@@ -173,7 +174,7 @@ where $P_v$ is the virtual position from the previous formula, $p_u$ and $p_l$ a
 From here the first step is calculating a `fair` price, which can be done by utilising the `L` value for the respective range to calculate `virtual` values for the pool balances. From here on `y` will be the cash balance of the pool and `x` the position.
 
   1. First, identify the current position, `P`. If it is `0` then the current fair price is the base price.
-  1. If `P != 0` then calculate the implied price from the current position using the virtual position $P_v$ which is equal to $P$ when $P > 0$ or $P + \frac{c}{p_u} \cdotp r_f$ where $P < 0$.
+  1. If `P != 0` then calculate the implied price from the current position using the virtual position $P_v$ which is equal to $P$ when $P > 0$ or $P + P_{v_u}$ where $P < 0$.
```
wwestgarth, line 177: Would be good to clarify (and I don't know the answer): Can this be `P_{v_u}` as calculated before when we used it to calculate `L`, or do we need to re-calculate it each time given potential balance changes from funding payments etc?

TomMcL (approved):

protocol/0090-VAMM-automated_market_maker.md
TomMcL: Should only need doing the once. Since this setup stores the "commitment" now you need to update that value if you want it to use any fees received in the market making

PR #2278, opened 2024-05-15
feat: add validation for fully collateralised mode

Jiajia-Cui:


Jiajia-Cui:
the two flags (fully-collateralised, and binary settlement) should be independent

PR #2280, opened 2024-05-15
refactor: modify specs in line with implementation

witgaw:
- disallow updates of `fully_collateralised` flag on running markets;
- remove the notion of a "wrapped risk model";


protocol/0019-MCAL-margin_calculator.md
```diff
@@ -436,6 +436,9 @@ Assume a [capped future](./0093-CFUT-product_builtin_capped_future.md) market wi
...
 - `fully-collateralised mode` is only allowed when there is a `max_price` specified (<a name="0019-MCAL-170" href="#0019-MCAL-170">0019-MCAL-170</a>)
 - `binary_settlement` is only allowed when there is a `max_price` specified (<a name="0019-MCAL-171" href="#0019-MCAL-171">0019-MCAL-171</a>)
+- When market gets updated so that `fully-collateralised=true` all parties should have their margin brought to the initial level implied by this mode. Any funds above that value should be released back to the general account. If party was in a cross-margin mode and current margin account balance is now insufficient under the fully-collateralised mode then additional funds should be searched for in the general account. No search should be carried out for parties which were in isolated mode. If parties can't meet the new margin level (irrespective of the mode they were in before) their orders should get cancelled and margin requirements should get recalculated. If the new initial margin level still can't be met then they should get closed-out. (<a name="0019-MCAL-177" href="#0019-MCAL-177">0019-MCAL-177</a>)
```
EVODelavega: So this point is perhaps a bit contentious (some would argue this is changing the nature of the market).

Regardless, some things to consider should we choose to allow this:

* It'd be possible that switching to fully collateralised mode closes out quite a few parties, potentially more on one side of the book than the other (depending on how close/far the mid price is from the max price)
* The network would be left holding the bag, and its insurance pool would be insufficient to cover its orders. That's fine by itself, but the parties still on the market may decide to close their positions ASAP - effecitively leaving the market dead in the water
* The parties who didn't close their position would incur loss socialisation, potentially fairly quickly depending on just how big the margin shortfall was for the closed-out positions (and therefore how shallow the insurance pool is). These parties probably won't be too happy.

protocol/0019-MCAL-margin_calculator.md
```diff
@@ -436,6 +436,9 @@ Assume a [capped future](./0093-CFUT-product_builtin_capped_future.md) market wi
...
+- When market gets updated so that `fully-collateralised=false` then all parties should be in cross-margining mode. All parties should have their margin brought to the initial level implied by this mode. Any funds above that value should be released back to the general account. f parties can't meet the new margin level their orders should get cancelled and margin requirements should get recalculated. If the new initial margin level still can't be met then they should get closed-out. (<a name="0019-MCAL-178" href="#0019-MCAL-178">0019-MCAL-178</a>)
```
Jiajia-Cui:
```suggestion
- When market gets updated so that `fully-collateralised=false` then all parties should be in cross-margining mode. All parties should have their margin brought to the initial level implied by this mode. Any funds above that value should be released back to the general account. If parties can't meet the new margin level their orders should get cancelled and margin requirements should get recalculated. If the new initial margin level still can't be met then they should get closed-out. (<a name="0019-MCAL-178" href="#0019-MCAL-178">0019-MCAL-178</a>)
```

protocol/0019-MCAL-margin_calculator.md
witgaw: Good point, modified the spec now to disallow updating `fully_collateralised` flag.

protocol/0019-MCAL-margin_calculator.md
witgaw: AC removed now

davidsiska-vega (approved):

PR #2281, opened 2024-05-15
feat: AMM Order Book Estimation

TomMcL:


cdummett (commented):
Sounds good to me 👍 - do we need some ACs though in a similar fashion to this non-protocol [spec](https://github.com/vegaprotocol/specs/blob/master/non-protocol-specs/0012-NP-LIPE-liquidation-price-estimate.md)?

cdummett (approved):

PR #2266, opened 2024-05-08
feat: LP performance reward metric

cdummett:
> [!WARNING]
> Changes in this PR will be superseded by changes from the liquidity mechanisms overhaul #2299 and #2309. This PR should only be merged and work ticketed if need for improved liquidity rewards cannot be delayed to `genbutemple`.

### Summary

PR adds a new reward metric for reward Liquidity Providers for providing liquidity within a specified range.

To make the rewards independent of the liquidity fees accumulated on a market, the transfer implements the concept of "virtual fees". That is, the fees that would have been distributed for a fixed amount of accumulated fees (1 quantum).

This reward metric will require core to compute SLA performance for a number of price ranges and as such a new network parameter is added to restrict the minimum transfer size of transfers to reward pools.

PR #2299, opened 2024-06-10
feat: liquidity mechanisms overhaul

### LINKED ISSUE #2253: Remove bond account and remove becoming an LP mechanism
*More radical* 

Remove bond account and remove becoming an LP mechanism.

To remove the transaction to become an LP, and to avoid having to evaluate SLA for all keys all the time, consider that a market param that sets what minimum % of maker fees you have to collect to count as an LP for the next epoch. 

If in epoch `n-1` you were given % of maker fees and in epoch `n` you meet the SLA you get included in the rewards. 

In general, it would be good for LPs if at the end of epoch `n-1` they *knew* what fraction of LP fees they're eligible at the end of epoch `n` (as long as they met the SLA that epoch).  

*Old suggestion; ignore for now*
Possibly two parts:
1) remove the need for an LP bond being provided upfront: 
- instead for a fixed "bond" amount provided with the LP transaction, just provide "bond amount intent", 
- the LP fees are moved into the "bond account" at the end of each epoch, until the desired "intent" amount is met. 
- Liquidity (volume on book) must be supplied in line with the "intent" amount otherwise penalties kick in. 
- At each epoch end update the ELS according to actual bond account balance (same logic as if the LP was increasing the commitment). 
- Commitment increase increases the "intent" (and you go back to accumulating fees into bond account). 
- Commitment decrease decreases the "intent" and if the amount in the bond account is higher than the intent amount then that's returned according to current rules.  

2) Providing liquidity on both sides of the book above the amount implied by the minimum bond amount netparam and meeting the minimum SLA will automatically register the party as an LP with the corresponding "intent" (backed out as the time-weighted average of the volume on the book divided by the scaling factor). From that point we create the bond account and start cumulating LP fees there (without the party having to send a transaction). At end of each epoch "intent" is recalculated increases, reductions are treated as in 1) above.

### LINKED ISSUE #2268: Replace ELS with internal LP volume asset
Rip out ELS. 

Instead each market will instantiate a new asset, uniquely tied to the market, called something like e.g. BTCUSD-Perp-LP (blobs / points/ shares / drops). Successor market doesn't create a new LP token asset and just uses the old one (needs some decimals worry). 

- at the end of each epoch one token of the LP asset is minted for each unit of volume traded during the epoch and distributed pro-rata amongst LPs based on their SLA scores for the epoch.
- during the epoch the relative ownership of the LP tokens for the market amongst LPs is used as a weight instead of the ELS.   
- have per-epoch inflation cap.

cdummett:
closes #2253
closes #2268 

> [!WARNING]
> This PR will make the new liquidity reward metric #2266 redundant as the existing liquidity fees metric will no longer be abused from wash trading. 
>
> Instead #2309 should be merged, as this introduces an alternative method for the new liquidity reward metric where rewards are distributed pro-rata by the volume of notional each LP provides.

## Summary

PR deprecates previous liquidity specs and replaces them with simplified liquidity mechanisms which:

- require no bond
- require no sign-up transaction
- yields more predictable liquidity fees

In short the mechanics work as follows...

- parties accrue ELS points over time proportional to their maker fees earned.
- parties are designated as LPs on an epoch by epoch basis if they receive above a specified proportion of the markets maker fees.
- parties receive a proportion of the liquidity fees accumulated in an epoch relative to their [accrued](#accruing-els-points) ELS points and volume of notional (but only when designated as an LP).

## Open questions:

1. Points accrued and LP designation driven by maker fees received - how open is this to abuse through wash trading?
    -  Resolved: with introduction of "volume of notional" weighting
2. Does this incentivise deep books? - advantageous for minimal slippage and quick disposal of network position.
    -  Resolved: with introduction of "volume of notional" weighting
3. Does this incentivise being on the book throughout the epoch enough? 
    -  Resolved: with introduction of "volume of notional" weighting


protocol/0091-LIQM-liquidity_mechanisms.md
```diff
@@ -0,0 +1,118 @@
...
+
+A party will only be able to receive liquidity fees or earn liquidity rewards providing they are designated as an LP for that epoch.
+
+A party will only be designated as an LP providing they received more than a specified proportion of the markets total maker fees in the previous epoch, let this requirement be $N$ (the network parameter `liquidityProvider.proportionRequirement`). Throughout the epoch the network will track each parties maker fees, $M$.  At the end of epoch $i$, a party will be designated as an LP for epoch $i+1$ providing:
+
+$$\frac{V_{i_j}}{\sum_{k}^{n}{M_{i_j}}} >= N$$
+
+Where:
+
+- $M_{i_j}$ is the maker fees of party ${j}$ in epoch ${i}$
```
witgaw: Are `M` and `V` not representing the same quantity?

protocol/0091-LIQM-liquidity_mechanisms.md
```diff
@@ -0,0 +1,118 @@
...
+- If there is an indicative uncrossing price calculate the volume of notional that is in the range.
+
+```text
+(1.0-market.liquidity.priceRange) x min(last trade price, indicative uncrossing price) <=  price levels <= (1.0+market.liquidity.priceRange) x max(last trade price, indicative uncrossing price).
+```
+
+- If there is no 'indicative uncrossing price' then volume placed at any price should count towards the LP's commitment i.e the price range is interpreted as
+
+```text
+-infinity <=  price levels <= infinity
```
witgaw: Perhaps we could narrow it down to the auction uncrossing range (used to calculate the uncrossing price) or some version of it stretched around the uncrossing price and calculate it retrospectively only once the auction uncrosses? Though I guess that's potentially more data to be stored in the interim.

protocol/0091-LIQM-liquidity_mechanisms.md
```diff
@@ -0,0 +1,118 @@
...
+
+Updates to this parameter will be used the next time LPs are designated, i.e. updating the network parameter will not result in LPs instantly being redesignated.
+
+## Accruing ELS Points
+
+Whenever a market proposal is enacted (passes opening auction), an internal Vega asset is created to track the ELS points of that market. For a successor market, a new asset should not be created and instead the ELS asset of the parent market should be used.
+
+At the end of each epoch, each party accrues ELS points for the relevant market as follows:
+
+$$ELS_{i_j} = V_{i_j} \cdot 10^{-Q}$$
```
witgaw: I was wondering if we wouldn't want to introduce some cut-off so that only some portion minimum proportion of maker fees in epochs qualifies for the ELS points, but the more I think about it the more I like the currently proposed design. It's in the spirit of user-ownership and could potentially incentivise smaller users to eventually become LPs or at least make some extra profit on the re-sale of their ELS points.

protocol/0091-LIQM-liquidity_mechanisms.md
```diff
@@ -0,0 +1,118 @@
...
+## Distributing Liquidity Fees
+
+At the end of epoch, after each party accrues ELS points for that epochs volume, the accumulated liquidity fees are distributed pro-rata amongst liquidity providers weighted by their accrued ELS points as follows.
+
+$$f_{i_j} = f_{i} \cdot \frac{ELS_j\cdot{L_j}}{\sum_{k}^{n}{ELS_k\cdot{L_k}}}$$
+
+Where:
+
+- $f_{i_j}$ is the liquidity fees distributed to party $j$ in epoch $i$
+- $f_{i_j}$ is the liquidity fees accumulated by the market in epoch $i$
```
witgaw:
```suggestion
- $f_{i}$ is the liquidity fees accumulated by the market in epoch $i$
```

protocol/0091-LIQM-liquidity_mechanisms.md
cdummett: Auction uncrossing range is a good idea, in hind sight not a massive fan of infinity as an LP can game it by placing small orders at large prices during auctions when there is no uncrossing price.

Do we have an uncrossing range when the book isn't crossed though?

witgaw: We don't, though I was thinking just `0` commitment in that case. So LPs only get commitment implied during auction if they successfully contributed (by placing volume in a sensible range) to auction having uncrossed with non-zero volume.

protocol/0091-LIQM-liquidity_mechanisms.md
cdummett: Definitely agree with this point
> incentivise smaller users to eventually become LPs

On this point...

> make some extra profit on the re-sale of their ELS points

I think it's okay but something I was wondering is whether a late-arriving LPs with more capital would be able to buy up all the ELS points accrued by early traders wanting to dump? I think the early LPs should still have a much larger supply of ELS points in this case so we would be fine but worth considering.

protocol/0091-LIQM-liquidity_mechanisms.md
cdummett: Yes thanks, silly typo 🙏

protocol/0091-LIQM-liquidity_mechanisms.md
witgaw: Not sure, though I think it's got more positive externalities. More fun to trade on markets and be earning something that's potentially worth something in the future and a late arrival of an LP commited enough to buy ELS off the smaller participants is probably a good thing for the market overall.

protocol/0091-LIQM-liquidity_mechanisms.md
```diff
@@ -0,0 +1,118 @@
...
+
+- "in-active" parties with a large number of ELS points will no longer be rewarded through liquidity mechanisms should they stop providing liquidity (or provide uncompetitive liquidity).
+- "late-arriving" parties with a small number of ELS points will be rewarded through liquidity mechanisms if they provide competitive liquidity and as such comprise a larger proportion of the markets volume.
+
+
+## Implied Commitment Amount
+
+Each LPs `implied commitment amount` is defined as, the maximum volume of notional liquidity that they supplied within a specified range for at least N % of the epoch (where the liquidity range and N are market configurable parameters).
+
+To calculate the implied commitment amount, throughout the epoch, the network must sample and store the volume of notional liquidity supplied by each LP at that point. For now this is done once a block but could be sampled randomly to reduce the amount of data stored.
```
davidsiska-vega: "implied commitment" -> "volume of notional" 

Consider whether we need weighting. (Remember simple means clever). We don't think weighting is needed because your maker fees collected registered you as an LP or not. So if you quote far you get diddly squat.

witgaw (dismissed):
LGTM

witgaw (approved):

PR #2309, opened 2024-06-24
feat: liquidity SLA reward metric

### LINKED ISSUE #2254: LP protocol reward schemes carry own SLA def
1) More flexible scheme
The LP protocol reward scheme can optionally additionally specify the who "SLA blob" with its own range and time on book etc and the rewards are distributed according to the SLA rules specified here. 

There can be multiple such reward schemes per one market. 

One reward scheme can still have multiple markets in scope.

cdummett:
closes #2254


witgaw (approved):
LGTM

PR #2319, opened 2024-07-17
feat: restricted ELS assets and changes to accrual

cdummett:
### Restricted internal assets (for ELS)

To enable internal built-in assets too represent ELS points for the new liquidity mechanisms, it is important certain uses are restricted to avoid having to build complicated spam protection for assets which are hard to value (at least initially).

PR updates the asset framework so internal asset can have restrictions set and introduces network parameters to default these values for assets representing ELS points.

PR also introduces a method of defaulting the quantum of these assets.

### Changes to ELS accrual

Only designated LPs should now accrue ELS points. To ensure parties properly accrue points for the volume they generate the order in which steps are executed in the liquidity mechanisms is modified such that.

- Liquidity fees are distributed only to parties who were delegated as LPs for the epoch that is ending.
- ELS points are accrued by parties who have been delegated for the next epoch based on the volume they created in the epoch that is ending.

### DIFF
--- a/protocol/0040-ASSF-asset_framework.md
+++ b/protocol/0040-ASSF-asset_framework.md
@@ -229,7 +229,7 @@ Built-in assets work much the same as any Vega asset except they can not be depo
 - as the quote asset in a spot market
 - as the settlement asset in a derivatives market (future or perpetual)
 
-These restrictions should be updatable at any point through governance. When updating an asset the following cases should be considered:
+These restrictions can be updated at any point through governance. When updating an asset the following cases should be considered:
 
 - given an asset allows transfers, if a delayed one-off or recurring transfer is created and then the asset is updated to not allow transfers, any outstanding transfers are not cancelled and are executed as normal.
 - given an asset allows use as a base, quote, or settlement asset, if a market is created using the asset and the asset is later updated to not allow the respective use, the market is not cancelled and should operate as normal.
--- a/protocol/0040-ASSF-asset_framework.md
+++ b/protocol/0040-ASSF-asset_framework.md
@@ -67,6 +67,7 @@ message DevAssets {
 }
 
 message AssetRestrictions {
+  bool faucetEnabled = 1;
   bool transferEnabled = 1;
   bool baseAssetEnabled = 1;
   bool quoteAssetEnabled = 1;
@@ -222,8 +223,11 @@ This section will be expanded if additional ethereum based token standards are s
 
 #### Built-In Assets
 
-Built-in assets work much the same as any Vega asset except they can not be deposited or withdrawn. They also optionally can be restricted from use in specific functions. These functions include:
+Built-in assets work much the same as any Vega asset except they can not be deposited or withdrawn. The creation of internal assets through governance proposals is restricted by the network parameter, `limits.assets.proposeBuiltinEnabled`.
 
+They also optionally can be restricted from use in specific functions. These functions include:
+
+- minting
 - transfers
 - as the base asset in a spot market
 - as the quote asset in a spot market
--- a/protocol/0040-ASSF-asset_framework.md
+++ b/protocol/0040-ASSF-asset_framework.md
@@ -53,6 +53,7 @@ message BuiltinAsset {
   string symbol = 3;
   string totalSupply = 4;
   uint64 decimals = 5;
+  AssetRestrictions assetRestrictions = 6;
 }
 
 message ERC20 {
@@ -63,6 +64,13 @@ message ERC20 {
 
 message DevAssets {
   repeated AssetSource sources = 1;
+}
+
+message AssetRestrictions {
+  bool transferEnabled = 1;
+  bool baseAssetEnabled = 1;
+  bool quoteAssetEnabled = 1;
+  bool settlementAssetEnabled = 1;
 }
 ```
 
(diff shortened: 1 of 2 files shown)


protocol/0095-LIQM-liquidity_mechanisms.md
```diff
@@ -21,48 +21,112 @@ The mechanisms also enable the following features:
...
 - "late-arriving" parties with a small number of ELS points will be rewarded through liquidity mechanisms if they provide competitive liquidity and as such comprise a larger proportion of the markets volume.
 
-## Volume of Notional
+### ELS Points
+
+#### Creating ELS points
+
+Whenever a market proposal is enacted (passes opening auction), an internal Vega asset is created to track the ELS points of that market. For a successor market, a new asset should not be created and instead the ELS asset of the parent market should be used.
+
+When ever an internal asset representing ELS points is created the [internal asset restrictions](./0040-ASSF-asset_framework.md#built-in-assets) are defaulted to the respective [network parameter](#network-parameters). It's quantum should be set using the following formula. Note, both the restrictions and quantum can later be updated through governance.
```
davidsiska-vega:
```suggestion
When ever an internal asset representing ELS points is created the [internal asset restrictions](./0040-ASSF-asset_framework.md#built-in-assets) are defaulted to the respective [network parameter](#network-parameters). Its quantum should be set using the following formula. Note, both the restrictions and quantum can later be updated through governance.
```

protocol/0095-LIQM-liquidity_mechanisms.md
```diff
@@ -21,48 +21,112 @@ The mechanisms also enable the following features:
...
 
 A number in the range $[0, 1]$ which defines the minimum proportion of a markets maker fees a party must receive in order to be [designated](#designating-liquidity-providers) as an LP for the next epoch. The parameter should default to `0.1`.
 
 Updates to this parameter will be used the next time LPs are designated, i.e. updating the network parameter will not result in LPs instantly being re-designated.
 
-## Accruing ELS Points
+### `liquidity.els.defaultQuantumFactor`
 
-Whenever a market proposal is enacted (passes opening auction), an internal Vega asset is created to track the ELS points of that market. For a successor market, a new asset should not be created and instead the ELS asset of the parent market should be used.
+A number in the range $[0, 100]$ which defines the amount to scale the quantum of the markets settlement or quote asset by (for derivative and spot markets respectively) when setting the initial quantum of the internal asset representing the ELS points for the market.
```
davidsiska-vega:
```suggestion
A number in the range $(0, 10000]$ which defines the amount to scale the quantum of the markets settlement or quote asset by (for derivative and spot markets respectively) when setting the initial quantum of the internal asset representing the ELS points for the market.
```

protocol/0095-LIQM-liquidity_mechanisms.md
```diff
@@ -21,48 +21,112 @@ The mechanisms also enable the following features:
...
+A boolean, defaulting to `false`, which sets whether internal assets representing ELS points are allowed to be used as the quote asset of a spot market when they are first created.
+
+Note, the restriction on the asset can always be updated later through governance.
+
+## Mechanisms
+
+There are three liquidity mechanisms that take place at the end of the epoch.
+
+1. Liquidity fees are distributed amongst liquidity providers
+2. New liquidity providers are delegated
```
witgaw: Perhaps "designated" rather than "delegated" to remove any possible association with delegation of tokens?

witgaw (dismissed):
LGTM on first read. I'll try think about it a bit more and let you know if I can think of any problems.

Also a small nit-pick:

protocol/0040-ASSF-asset_framework.md
```diff
@@ -212,6 +220,20 @@ NOTE 2: Before running this function, the user must run the ERC-20-standard `app
...
+- transfers
+- as the base asset in a spot market
+- as the quote asset in a spot market
+- as the settlement asset in a derivatives market (future or perpetual)
+
+These restrictions can be updated at any point through governance. When updating an asset the following cases should be considered:
+
+- given an asset allows transfers, if a delayed one-off or recurring transfer is created and then the asset is updated to not allow transfers, any outstanding transfers are not cancelled and are executed as normal.
+- given an asset allows use as a base, quote, or settlement asset, if a market is created using the asset and the asset is later updated to not allow the respective use, the market is not cancelled and should operate as normal.
+
```
davidsiska-vega, line 240: At the moment built-in assets allow anyone to mint - clearly we don't want that ;-) the ELS asset should only get created by the protocol in a controlled way. Let's mention that clearly. 

Alternatively - a new kind of asset @jeremyletang .

protocol/0040-ASSF-asset_framework.md
```diff
@@ -212,6 +220,20 @@ NOTE 2: Before running this function, the user must run the ERC-20-standard `app
 
 This section will be expanded if additional ethereum based token standards are supported by Vega. New bridges will be expected to implement `IVega_Bridge`.
 
+#### Built-In Assets
```
davidsiska-vega, line 224: Asset will either:
- get created by the market proposal (on enactment the ELS asset is created) in that case arbitrary minting is disabled; only the "liquidity mechanism" can mint at epoch end.
- XOR the builtin asset is created by governance in which case the anyone-can-mint function is enabled; "liquidity mechanism" cannot mint them. 

On top of that it must be possible to DISABLE creation of builtin assets by governance.

davidsiska-vega:
@cdummett Sneak in change in ELS accrual so that each ELS point creation is tied to maker volume you participated in a trade (not maker fee received) - it's just a scaling factor.

protocol/0040-ASSF-asset_framework.md
cdummett: Now addressed in commit [269534d](https://github.com/vegaprotocol/specs/pull/2319/commits/269534d34d4fd3f7412a56493655be32e0491d9)...

- New network parameter added to enable/disable creation of builtin assets via governance.
- New asset restriction added to restrict minting through a faucet.
- ELS assets default to restricting minting through a faucet.

protocol/0040-ASSF-asset_framework.md
cdummett: Now addressed in commit [269534d](https://github.com/vegaprotocol/specs/pull/2319/commits/269534d34d4fd3f7412a56493655be32e0491d9)...

- New asset restriction added to restrict minting through a faucet.
- ELS assets default to restricting minting through a faucet.

Can update if we go down the new asset route rather than using built-in assets.

PR #2332, opened 2024-08-03
feat: eligible entities reward metric and lottery dispatch strategy

cdummett:


protocol/0056-REWA-rewards_overview.md
```diff
@@ -215,6 +215,28 @@ This flag is used to prevent any given funder from funding a creation reward in
...
+- it specifies no metric asset, no markets within the market scope, no staking requirement, and no position requirement - in this case, all parties on the network are given a score of $1$.
+- it specifies no metric asset, no markets within the market scope, no position requirement, but does specify a staking requirement - in this case, all parties meeting the staking requirement are given a score of $1$.
+
+If however a position requirement is specified, an asset must be specified also and then parties must meet the position requirement to receive rewards.
+
+### Reward windows and transfer delays
+
+Note the following interactions with reward windows and transfer delays.
+
+- if a reward window greater than one is specified, an entity only needs to meet the eligibility requirements in the epoch the reward is distributed.
```
barnabee: Per the discussion on Slack, ideally this would require eligibility during all epochs in the window

protocol/0056-REWA-rewards_overview.md
```diff
@@ -215,6 +215,28 @@ This flag is used to prevent any given funder from funding a creation reward in
...
+- if a transfer delay greater than one is specified, an entity still **only** needs to meet the eligibility requirements in the epochs in the reward window (as per the point above).
```
barnabee: If the change to check all epochs in the reward window is made, this sentence reads correctly. However as currently specified they only need to meet the requirements in the distribution epoch, not the epochs in the reward window.

protocol/0056-REWA-rewards_overview.md
```diff
@@ -314,6 +336,22 @@ Calculate each entities share of the rewards, $s_{i}$ pro-rata based on $d_{i}$,
...
+
+1. At the end of the epoch, take the hash of the most recent block as the seed.
+2. Iterate over the rank table, randomly assigning an entity to that rank where an entities probability of selection is as follows.
+
+Let:
+
+- $p_i$ be the probability entity $i$ is selected
+- $r_i$ be the reward score of entity $i$
+
+$$p_{i} = \frac{r_i}{\sum_{i}^{n}{r_i}}$$
```
barnabee, line 351: Should be explicit that this is selecting only from the entities that don't already appear in the rank table, i.e. an entity can never appear in the rank table more than once.

barnabee (changes requested):

protocol/0056-REWA-rewards_overview.md
witgaw: Should be addressed now.

protocol/0056-REWA-rewards_overview.md
witgaw: Should be addressed now.

protocol/0056-REWA-rewards_overview.md
witgaw: Should be addressed now.

barnabee (approved):

PR #2343, opened 2024-08-14
tweaks to features configuration for approbation

edd:
Note that this change is being made on the main branch, so that the suzukacastle branch can then be rebased with it
which will bring all the changes in that @cdummett merged from colosseo_II. 

- move model free to historic_distillery from suzuka (@jonray15)
- move community tags from to historic_distillery from suzuka (@jonray15)
- remove volume discount changes (@cdummett)
- move some vamm acs to a new rewards improvements (@cdummett)

As I understand it, as of 14th August 2024 this is the correctest version. When it is merged in to `suzukacastle` and the approbation config is updated to point to that as the most recent branch, the final result will be:

Screenshot 2024-08-14 at 18 24 04

# Note
The `features.json` now has a new field, which... exists only because it'll be totally ignored. Ahem. The `acsToEnableInSuzukaBranch` all exist in that branch, but fail status checks on `master`, because they don't exist there.


edd:
Note: https://github.com/vegaprotocol/approbation/pull/142 will be merged after this

cdummett (approved):

PR #2361, opened 2024-09-16
Cover the case of the early termination of a capped market

### LINKED ISSUE #2359: Settlement for capped futures states that it has to be at the bounds...
... but if you need to settle a market early for any other reason this doesn't work. 

Update needed. 

For termination by oracle I guess we still want to adhere to the binary settlement.

witgaw:
Closes #2359


protocol/colosseo-features.json
cdummett, line 1: These changes should be going in to `features.json` instead I think.

witgaw: Thanks, moved now

cdummett (approved):

PR #2362, opened 2024-09-16
refactor: add flag to determine behaviour on failure of a nested instruction

### LINKED ISSUE #2355: Network parameter validation on proposals
We can validate parameter *ranges* at proposal submission (and I believe we do).

However, relation to anything dynamic, especially other parameter values must *only* happen at enactment time: 
- if, at enactment time validation against other dynamic values (other parameter values) passes, proposal goes ahead
- else proposal fails (and if this is part of a batch then only this part of the batch fails, the rest remains scheduled as when it passed voting). 

We need to:
- update spec to make this clear
- add an AC

witgaw:
Closes #2355


davidsiska-vega (commented):
This is what I had in mind; let's add some ACs and call this good.

protocol/0074-BTCH-batch-market-instructions.md
```diff
@@ -26,14 +26,16 @@ Overall, building the ability to handle batches of market instructions in a sing
...
-- All instructions within each list must be validated as normal **at the time that the instruction is processed**. That is, instructions cannot be pre-validated as a batch. If a prior instruction, would create a state that would cause a later instruction to fail validation, the later instruction must fail validation (and vice verse). If validation fails, that instruction must be skipped and the subsequent instructions must still be processed. Any validation or other errors should be returned, as well as a reference to the instruction to which they relate, in the response.
+- All instructions within each list must be validated as normal **at the time that the instruction is enacted**. That is, instructions cannot be pre-validated as a batch. If a prior instruction, would create a state that would cause a later instruction to fail validation, the later instruction must fail validation (and vice verse).
```
barnabee, line 37: I think this needs to be clear that this is only where the validation needs to be based on state that is not known at the time of submission, e.g. it validates against a network parameter or some app state (order/market/position/whatever).

Any validation that _can_ unambiguously and correctly be done and is currently done at submission time should continue to be done then.

barnabee (dismissed):
One clarification needed then this is good

protocol/0074-BTCH-batch-market-instructions.md
witgaw: Addressed now I believe

protocol/0074-BTCH-batch-market-instructions.md
```diff
@@ -26,14 +26,16 @@ Overall, building the ability to handle batches of market instructions in a sing
...
-- All instructions within each list must be validated as normal **at the time that the instruction is processed**. That is, instructions cannot be pre-validated as a batch. If a prior instruction, would create a state that would cause a later instruction to fail validation, the later instruction must fail validation (and vice verse). If validation fails, that instruction must be skipped and the subsequent instructions must still be processed. Any validation or other errors should be returned, as well as a reference to the instruction to which they relate, in the response.
+- All instructions within each list must be validated as normal **at the time that the instruction is enacted**. That is, instructions cannot be pre-validated as a batch. If a prior instruction, would create a state that would cause a later instruction to fail validation, the later instruction must fail validation (and vice verse). Any validation not dependent on other parameters or app state should be carried out as early as possible (e.g. check if type is numerical etc.).
```
davidsiska-vega:
```suggestion
- All instructions within each list must be validated as normal **at the time that the instruction is enacted**. That is, instructions cannot be pre-validated as a batch. If a prior instruction, would create a state that would cause a later instruction to fail validation, the later instruction must fail validation (and vice verse).
Any validation not dependent on other parameters or app state (i.e. static e.g. check if type is numerical etc.) should be carried on batch submission.
```

davidsiska-vega (approved):

Issue #2367, opened 2024-09-23
vAMM base price oracle


davidsiska-vega:
Within the existing oracle framework a vAMM can provide an oracle input (specifing the number of decimals) for setting and updating its base price.

davidsiska-vega:
Closed by #2375 .

PR #2375, opened 2024-09-26
Oracle-driven base price for vAMMs

witgaw:
Closes https://github.com/vegaprotocol/specs/issues/2367

### DIFF
--- a/protocol/0090-VAMM-automated_market_maker.md
+++ b/protocol/0090-VAMM-automated_market_maker.md
@@ -133,7 +133,7 @@ The base price can either be a fixed scalar specified by a user or an [oracle co
 
 When oracle-driven base price is selected during the vAMM creation then the system first awaits the receipt of the next data point from the oracle and then follows the process outlined in [creation/amendment process](#creationamendment-process) with `base price` set to that value.
 
-Similarly, when the vAMM configuration is amended to use the oracle-driven base price rather than a fixed user supplied value, the system awaits the receipt of the next data point from the oracle and then uses that value as the `base price` in the regular amendment transaction outlined above.
+Similarly, when the vAMM configuration is amended to use the oracle-driven base price rather than a fixed user supplied value, the system awaits the receipt of the next data point from the oracle and then uses that value as the `base price` in the regular amendment transaction outlined above. Please note that **configuration amendment** from a fixed to oracle-driven base price should only be rejected if the oracle specification is incorrect. If the `base price` received from the oracle is such that the **value amendment** is rejected then the value should remain unchanged, but the next **value amendment** should still be attempted the next time around.
 
 #### Maximum update frequency
 
--- a/protocol/0090-VAMM-automated_market_maker.md
+++ b/protocol/0090-VAMM-automated_market_maker.md
@@ -65,7 +65,7 @@ The `Concentrated Liquidity` AMM is a market maker utilising a Uniswap v3-style 
 
 The concentrated liquidity market maker consists of two liquidity curves of prices joined at a given `base price`, an `upper` consisting of the prices above this price point and a `lower` for prices below it. At prices below the `base price` the market maker will be in a long position, and at prices above this `base price` the market maker will be in a short position. This is configured through a number of parameters:
 
-- **Base Price**: The base price is the central price for the market maker. When trading at this level the market maker will have a position of `0`. Volumes for prices above this level will be taken from the `upper` curve and volumes for prices below will be taken from the `lower` curve.
+- **Base Price**: The base price is the central price for the market maker. When trading at this level the market maker will have a position of `0`. Volumes for prices above this level will be taken from the `upper` curve and volumes for prices below will be taken from the `lower` curve. This price can be either user supplied or set (and maintained with an oracle). Please see [oracle-driven base price](#oracle-driven-base-price) section for more detail on the later case.
 - **Upper Price**: The maximum price bound for market making. Prices between the `base price` and this price will have volume placed, with no orders above this price. This is optional and if not supplied no volume will be placed above `base price`. At these prices the market maker will always be short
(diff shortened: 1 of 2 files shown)


protocol/0090-VAMM-automated_market_maker.md
```diff
@@ -127,6 +127,18 @@ In addition to amending to reduce the size a user may also cancel their AMM enti
...
+
+#### Maximum update frequency
+
+An optional non-negative parameter controlling the maximum update frequency should be included in the oracle-driven base price configuration. When set to `0` the `base price` will update each time a new value is received from the oracle. When set to a positive value, the `base price` should be amended to the last value received from the oracle once the specified time elapses. If such value is the same as the currently specified `base price` then there should be no effect. The default value of this parameter should be `0`.
```
davidsiska-vega: This is a bit confusing, we have 'frequency' but then say the update happens after a time interval elapses. Should we call it "Maximum update time interval"?

davidsiska-vega (changes requested):
One really nit-picky comment, the real review will have to come from William @wwestgarth anyway.

protocol/0090-VAMM-automated_market_maker.md
wwestgarth: Maybe I'm missing something, but I don't think we even need this field since I think it can be handled solely by the existing data source framework we have. 

For example if we set it up an Ethereum contract call to drive the base price, there is already a field in the data-source definition for "how frequently to read the contract". And if you are sending in new prices by hand using a normal trusted-key oracle-submission-transaction then you'll just only send it in when you want it updated anyway.

protocol/0090-VAMM-automated_market_maker.md
```diff
@@ -127,6 +127,18 @@ In addition to amending to reduce the size a user may also cancel their AMM enti
 
 Note that, whilst an `Abandon Position` transaction immediately closes the AMM a `Reduce-Only` transaction will keep it active for an uncertain amount of time into the future. During this time, any Amendment transactions received should move the AMM out of `Reduce-Only` mode and back into standard continuous operation.
 
+### Oracle-driven base price
+
+The base price can either be a fixed scalar specified by a user or an [oracle configuration](./0045-DSRC-data_sourcing.md). It should be possible to switch between the two with a regular vAMM amendment transaction.
+
+When oracle-driven base price is selected during the vAMM creation then the system first awaits the receipt of the next data point from the oracle and then follows the process outlined in [creation/amendment process](#creationamendment-process) with `base price` set to that value.
```
wwestgarth, line 134: Maybe a word on what to do about slippage? When you submit an AMM the transaction comes with a slippage field that controls how far to allow the AMM's fair-price can move from its base-price when it tries to "rebase" the AMM so that it doesn't cross with whats already on the book. An amend transaction has one too which works the same but between the amended AMM's current fair price and where it would now need to be to not cross.

Do we want something similar here too? For example if the oracle-drive price suddently is wildly different from where the AMM is currently at should we have some sort of protection? Or should it always try to rebase itself even if it means eat up a lot of volume to get there?

protocol/0090-VAMM-automated_market_maker.md
```diff
@@ -127,6 +127,18 @@ In addition to amending to reduce the size a user may also cancel their AMM enti
 
 Note that, whilst an `Abandon Position` transaction immediately closes the AMM a `Reduce-Only` transaction will keep it active for an uncertain amount of time into the future. During this time, any Amendment transactions received should move the AMM out of `Reduce-Only` mode and back into standard continuous operation.
 
+### Oracle-driven base price
+
+The base price can either be a fixed scalar specified by a user or an [oracle configuration](./0045-DSRC-data_sourcing.md). It should be possible to switch between the two with a regular vAMM amendment transaction.
```
wwestgarth: I guess its implied, but I'm assuming when a new oracle price comes in to change the base price, the lower and upper boundary prices are also changed by the same delta?

protocol/0090-VAMM-automated_market_maker.md
davidsiska-vega: Would it be better to have % change of how far out the base price is from the oracle input?

That's to avoid paying taker fees when changes to base price would result in trades for the vAMM.

protocol/0090-VAMM-automated_market_maker.md
davidsiska-vega: First iteration: unless we hear strongly that someone needs the bunds to move as well, let's keep these fixed.

davidsiska-vega: @wwestgarth is one of them much cleaner than the other from implementation point of view?

protocol/0090-VAMM-automated_market_maker.md
davidsiska-vega: Should use the slippage specified when the vAMM was first set up, if this is an amendment that specifies slippage then the new slippage should be used.

protocol/0090-VAMM-automated_market_maker.md
wwestgarth: It doesn't make any difference to the implementation really, so its mostly what would be most useful to people.

witgaw: Went with fixed bounds for now, let's give people the option to either leave them fixed or updated them proportionally once they ask for it.

protocol/0090-VAMM-automated_market_maker.md
witgaw: Modified the spec to use the slippage

protocol/0090-VAMM-automated_market_maker.md
witgaw: Modified the spec to use the relative price change.

wwestgarth (dismissed):

davidsiska-vega (approved):

PR #2379, opened 2024-10-21
feat: amm spread

### LINKED ISSUE #2371: vAMM spread around fair price
Figure out how easy or hard, given current design and implementation would it be to allow some % spread around fair price.

cdummett:


wwestgarth:
Core work for this is done and seems to work nicely. One thing to note is that when a market enters auction the AMM's spread term is ignored and they just have the normal default spread. The reason being that the auction uncrossing code is already super complicated and prone to horrible edge cases. I don't really want to add to this complexity right now when there isn't enough resource to resolve any issues that may arise. An AMM's spread will return again once the book is uncrossed.

